Issue #34, opened 2025-03-05
Social Logins for Authentication


mrgrauel:
## As a user, I want to use social login to access Sokosumi without going through a complex registration process.

Please add the following social logins:

* Google
* Microsoft

### Technical Decisions:

We use [`react-social-login-buttons`](https://www.npmjs.com/package/react-social-login-buttons) as UI library for social buttons.

### Missing:

Secrets and IDs to login via socials

### Status (Update: 2025-03-18)

Postponed until further notice
**Reason**: Email & password login is sufficient for a soft launch.

### Status (Update: 2025-05-28)

@patrick organizes the required information for google and microsoft to enable social logins

mrgrauel:
@PatrickTobler will update us as soon as he can provide the necessary information to resolve the issue.

PR #38, opened 2025-03-05
Add social auth buttons and componentize auth pages

golddydev:
Add google, microsoft, linkedin and apple social provider buttons.

Related to #34 

Until we have social provider's secrets
This PR is only for auth page (sign in and sign up) UI

* Sign in and Sign up buttons from `react-social-login-buttons`
* Component sign in and sign up page


mrgrauel (commented):
The code looks fine, but I have a few questions:

1. How did you test it? Do you have the required ids and secrets for testing? If yes, could you send them to me on a privat channel.

3. Do we have the required ids and secrets for our deployments?

golddydev:
> The code looks fine, but I have a few questions:
> 
> 1. How did you test it? Do you have the required ids and secrets for testing? If yes, could you send them to me on a privat channel.
> 2. Do we have the required ids and secrets for our deployments?

We don't have those ids yet. (Soon we will get)
For now, I just added them to auth.ts and not using them when sign in or sign up.

web-app/.env.example
schaier-io, line 1: Probably we need the Client Id on the frontend too, so have it as NEXT_PUBLIC prefixed?

web-app/src/app/(auth)/signin/page.tsx
schaier-io, line 1: I would prefer we use a library which implements the social auth buttons, as the providers can be somewhat picky if their design guidelines are not 100% followed. What do you think?

mrgrauel: Example Library: 
* https://github.com/michaeljscript/react-social-login-buttons

web-app/.env.example
golddydev: Oh, yes.
For client id, we can make it NEXT public.

web-app/src/app/(auth)/signin/page.tsx
golddydev: I make social buttons using `react-social-login-buttons` library

web-app/.env.example
golddydev: I update .env.example file

But until we have all secrets, this PR won't do anything with those envs

mrgrauel (changes requested):
The figma has a slightly different design. The box is narrower due to the values and benefits on the left. I recommend stacking the four social login options vertically, making the box smaller (width), and adding dummy or placeholder values.

This would be much closer to the intended design.

image

golddydev:
Yeah, I think what I have implemented is just mobile design.
Let me add desktop design also.

golddydev:
What I suggest is that
This PR is only for add social auth buttons in auth pages.

If I implement those design in this PR, PR will get bigger, and hard to review

mrgrauel:
Fine just add the buttons, but let's stick to the design if it's not a problem.

The box should be smaller in width and social buttons should be a vertical stack.

mrgrauel (dismissed):
You are right, sorry. I'm confused about why the buttons are arranged in a 2x2 grid instead of a vertical stack. However, it's fine for now since we don't know the final design.

Please wait until @schaier-io also reviewed because of the open discussions.

mrgrauel (approved):

schaier-io:
Do not focus on the design to much, it is not final anyway. Make it somehow responsive and thats it. Not sure if the designers followed the social auth requirements anyway

golddydev:
> Do not focus on the design to much, it is not final anyway. Make it somehow responsive and thats it. Not sure if the designers followed the social auth requirements anyway

Okay. got that point.

Can you resolve your discussion?
So I can merge this PR?

PR #42, opened 2025-03-06
Adds a layout for the app route group

golddydev:
Make app layout (with responsiveness)

* main header
* sidebar


golddydev:
# Desktop

image

golddydev:
# Mobile

image

image

image

web-app/src/app/(app)/components/left-panel/mobile-left-panel.tsx
```diff
@@ -0,0 +1,43 @@
...
+  SheetHeader,
+  SheetTitle,
+  SheetTrigger,
+} from "@/components/ui/sheet";
+
+import AgentAddButton from "./components/agent-add-button";
+import AgentsList from "./components/agents-list";
+
+export default function MobileLeftPanel() {
+  const [open, setOpen] = React.useState(false);
```
mrgrauel, line 20: How can I open or close the panel? Do we need this function?
I would suggest to remove it for now, because we dunno how it should it's supposed to look closed, wo we?

image

web-app/src/app/(app)/components/main-nav.tsx
```diff
@@ -3,34 +3,30 @@
...
   const pathname = usePathname();
 
   return (
     <div className="mr-4 hidden w-full md:flex">
-      <Link href="/" className="mr-4 flex items-center gap-2 md:mr-6">
-        <span className="hidden font-bold md:inline-block">
-          {siteConfig.name}
-        </span>
-      </Link>
+      <SokosumiLogo />
```
mrgrauel, line 21: The SokosumiLogo is part of the Sidebar instead of the top navigation.

web-app/src/app/(app)/components/mobile-nav.tsx
```diff
@@ -14,83 +21,80 @@ import {
...
+                    },
+                  )}
+                >
+                  {nav.icon}
+                  {nav.label}
+                </Link>
+              </li>
+            );
+          })}
+        </ul>
```
mrgrauel, line 96: What is the purpose of the `icon` in the `navItem`?

I don't see it in your implementation and it's correct according to the design, but why did you implement it?

image

mrgrauel (changes requested):
I receive the following warnings in the console. I would highly recommend that we handle warnings as errors in the development process.

```
hook.js:608 Image with src "/logos/sokosumi-logo-black.svg" was detected as the Largest Contentful Paint (LCP). Please add the "priority" property if this image is above the fold.
Read more: https://nextjs.org/docs/api-reference/next/image#priority
```

```
dashboard:1 The resource http://localhost:3000/_next/static/css/app/layout.css?v=1741330991996 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
```

web-app/src/app/(app)/components/mobile-nav.tsx
golddydev: This is only for mobile nav.

web-app/src/app/(app)/components/left-panel/mobile-left-panel.tsx
golddydev: Nope, in the design, the sidebar is not possible to open or close.

golddydev: That is why I assume that that is may be depends on breakpoints

web-app/src/app/(app)/components/main-nav.tsx
golddydev: That is in top nav bar

image

golddydev: Well this is pretty normal, that the logo is one sidebar, and we can easily open and close sidebar. 
But it is not in our design.

This is how most website use side
https://shadcn-ui-sidebar.salimi.my/dashboard

golddydev:
> I receive the following warnings in the console. I would highly recommend that we handle warnings as errors in the development process.
> 
> ```
> hook.js:608 Image with src "/logos/sokosumi-logo-black.svg" was detected as the Largest Contentful Paint (LCP). Please add the "priority" property if this image is above the fold.
> Read more: https://nextjs.org/docs/api-reference/next/image#priority
> ```
> 
> ```
> dashboard:1 The resource http://localhost:3000/_next/static/css/app/layout.css?v=1741330991996 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
> ```

I am checking this warnings now

golddydev:
> I receive the following warnings in the console. I would highly recommend that we handle warnings as errors in the development process.
> 
> ```
> hook.js:608 Image with src "/logos/sokosumi-logo-black.svg" was detected as the Largest Contentful Paint (LCP). Please add the "priority" property if this image is above the fold.
> Read more: https://nextjs.org/docs/api-reference/next/image#priority
> ```
> 
> ```
> dashboard:1 The resource http://localhost:3000/_next/static/css/app/layout.css?v=1741330991996 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
> ```

About preload warning: I just fixed. (caused by image src being "#")

About image priority issue: I am going to fix that in another PR which update logo components

web-app/src/app/(app)/components/mobile-nav.tsx
golddydev: Without icons, It seems not good.

mrgrauel (changes requested):
The logo should be in the sidebar according to the design specs.

image

web-app/src/app/(app)/components/main-nav.tsx
golddydev: I just updated it

web-app/src/app/(app)/components/main-nav.tsx
schaier-io, line 1: we should componetize this

web-app/src/app/(app)/components/mobile-nav.tsx
schaier-io, line 1: lets also extract the component here

golddydev: I extracted `nav-menu` from `main-nav` and `mobile-nav`

web-app/src/app/(app)/components/main-nav.tsx
golddydev: Make this component simple using `nav-menu` and `nav-link`

web-app/src/app/(app)/components/nav-link.tsx
mrgrauel, line 1: Nice one! I'm going to reuse this for the landing page. I could have thought of it myself 🤦 With a small adjustment, it can become a component for both the app and the landing section.

web-app/.prettierignore
```diff
@@ -6,4 +6,6 @@
 src/lib/api/generated/**
 
 # Ignore UI components
-src/components/ui/**
\ No newline at end of file
+src/components/ui/**
+
+tailwind.config.ts
```
mrgrauel, line 11: Why did you add this file to the `.prettierignore`? As I understand it, we would reformat the entire file each time we make a change. 

It might even be deleted if #46 gets merged. I recommend removing this file.

```suggestion

```

mrgrauel (changes requested):
Should we implement a specific mobile design? As I understand it, we are focusing on a responsive design for desktop that should work reasonably well on mobile. However, there should not be an explicit mobile implementation.

mrgrauel:
Also the folder `app-sidebar` should just be named `sidebar`, doesn't it? Same idea as with the `app-header.tsx` renaming to `header.tsx`.

golddydev:
> Should we implement a specific mobile design? As I understand it, we are focusing on a responsive design for desktop that should work reasonably well on mobile. However, there should not be an explicit mobile implementation.

I don't understand what you are saying.
I don't implement any specific mobile design.

If you are talking about `mobile-nav`, that is for responsive.

Main nav is top sticky nav bar,
mobile nav is drawer from right side.

web-app/src/app/(app)/components/nav-link.tsx
golddydev: Then it doesn't should be in `(app)` folder.
It should be some where common (like `src/components`)

web-app/src/app/(app)/components/app-header.tsx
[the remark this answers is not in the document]
golddydev: The reason why I attached app before `header` and `sidebar`, is that 
we have `Sidebar` (from shadcn), which makes name conflicts

So that is why I named it as `app-sidebar`

Well do you prefer the say `import Sidebar as ShadcnSidebar`?

web-app/.prettierignore
golddydev: That is because everytime we add shadcn components, that is all formatted by itself (shadcn rule)
So why do we have to re format them in our way every time we add components?

That will make PR bigger.

golddydev: By the way, deletion of file doesn't have anything to do with `.prettierignore`, does it?

web-app/src/app/(app)/components/app-header.tsx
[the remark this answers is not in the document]
mrgrauel: I see the issue with the Sidebar, that makes sense to name the export it as `AppSidebar`, but I would rename the folder. 

Sorry but naming is for me important... I proposed a change how I think it makes it clear to understand even in 6 months.

Rename `AppHeader` to `Header` -> 77343864177e61d060a517a38ee1fb83668c7c9a
Rename `app-sidebar` folder to `sidebar` -> 57a270413d1996c981f1735faa94613ef2ec5c60

What do you think? Any opinion about these changes?

web-app/.prettierignore
mrgrauel: > By the way, deletion of file doesn't have anything to do with .prettierignore, does it?

If we agree to remove the file in #46, we shouldn't add it back since it no longer exists. However, I can remove it in #46 if it is merged after this one.

> That is because everytime we add shadcn components, that is all formatted by itself (shadcn rule)
So why do we have to re format them in our way every time we add components?

That's a great point! 👍

web-app/src/app/(app)/components/app-header.tsx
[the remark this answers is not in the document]
golddydev: Okay, then Let's stick to that convention.

`Never prefix components folder or exported name with parent folder`

Then we can use `import as` and default export `Sidebar` instead of `AppSidebar`

mrgrauel:
@golddydev I did some additional changes. After a `npm install` I head a dirty head on 8aebea0.

- I noticed that the `Sidebar` wasn't readded after the upgrade, which could use some issues.
- Removes `tailwind.config.ts` from `.prettierignore` because the file is removed
- Updates Packages because of dirty head after `npm install`

I hope you are fine with these changes.

mrgrauel:
> I don't implement any specific mobile design.
> 
> If you are talking about `mobile-nav`, that is for responsive.
> 
> Main nav is top sticky nav bar, mobile nav is drawer from right side.

Would you agree to rename the navigation components as follows:

`main-nav` => `sticky-navigation`  
`mobile-nav` => `drawer-navigation`  

The current names are confusing, and I initially thought they referred to specific implementations for mobile devices.

golddydev:
> > I don't implement any specific mobile design.
> > If you are talking about `mobile-nav`, that is for responsive.
> > Main nav is top sticky nav bar, mobile nav is drawer from right side.
> 
> Would you agree to rename the navigation components as follows:
> 
> `main-nav` => `sticky-navigation` `mobile-nav` => `drawer-navigation`
> 
> The current names are confusing, and I initially thought they referred to specific implementations for mobile devices.

Changed name.

`main-nav` => `sticky-navigation`

`mobile-nav` => `sheet-navigation` (because I use `sheet` not `drawer`)

golddydev:
> @golddydev I did some additional changes. After a `npm install` I head a dirty head on [8aebea0](https://github.com/masumi-network/sokosumi/commit/8aebea0d55d3a321788e4e64399b98a8e096ef48).
> 
> * I noticed that the `Sidebar` wasn't readded after the upgrade, which could use some issues.
> * Removes `tailwind.config.ts` from `.prettierignore` because the file is removed
> * Updates Packages because of dirty head after `npm install`
> 
> I hope you are fine with these changes.

This is all fine.

I also updated `globals.css` because it is dirty after updating. [e729e80](https://github.com/masumi-network/sokosumi/pull/42/commits/e729e806e8e92b60ceb8d15911978cc8e846a420)
And update `components.json` since we are using `tailwind.config.ts` anymore [d9431f5](https://github.com/masumi-network/sokosumi/pull/42/commits/d9431f58db99f61d1240496e46941445ddcd5b2b)

mrgrauel (changes requested):
The landing page is broken after your `global.css` changes. 

image

golddydev:
> The landing page is broken after your `global.css` changes.
> 
> image

Oh, I never know you add custom changes to `globals.css` 😅 

Could you tell me what custom changes you added in `globals.css`, expect 3 landing page colors?

golddydev:
```css
@utility container {
  margin-inline: auto;
  @media (width >= --theme(--breakpoint-sm)) {
    max-width: none;
  }
  @media (width >= 1400px) {
    max-width: 1400px;
  }
}
```

I see, this is what causes change on landing pages' hero disalign.
Can you make this as inline class name, rather than change `container` tag in `globals.css`?

mrgrauel (approved):

schaier-io (approved):

Issue #41, opened 2025-03-06
Agent Gallery


mrgrauel:
# As a user, I want to explore all available agents so I can use one or more of them for a job.

* Create a Landing Page Layout
* Add `/gallery` route
* Use a mock for the featured agent
* Use a mock for available agents

### Design
<img width="860" alt="Image" src="https://github.com/user-attachments/assets/f5b55511-d6fe-44dc-b710-03d1e3c7c48e" />

Source: https://www.figma.com/design/jLGG1K6RmRgDzKtaCZvdJz/Sokosumi-UX-%26-Wireframes?node-id=185-33611&t=Nrjy90pGpYTtwWDh-0

PR #44, opened 2025-03-07
Add Agent Gallery with Mocks

mrgrauel:
Related to #41

This pull requests adds the `/gallery` route with a simple implementation of the gallery with mock data. I reused the initial agent card design, but added tags below the description because of responsive issues if no tags are available for an agent.

image


web-app/src/app/(landing)/page.tsx
schaier-io, line 1: I like the split

schaier-io (approved):

PR #51, opened 2025-03-10
Refactor import paths to use @/shadcn/* for UI components

mrgrauel:
* Refactor import paths to use `@/shadcn/*` for UI components
* Extract cspell settings in `cspell.json`
* Disable ESLint rule for Next.js Jest configuration import


web-app/.vscode/settings.json
schaier-io, line 1: I think without this change, prettier formatted it wrong for me. Is it necessary to remove it?

schaier-io (commented):
Is there a reason why we differ from the default shadcn path?

web-app/.vscode/settings.json
```diff
@@ -14,10 +12,6 @@
   "files.associations": {
     "*.css": "tailwindcss"
   },
-  "prettier.tabWidth": 2,
-  "prettier.semi": true,
-  "prettier.singleQuote": false,
-  "prettier.trailingComma": "all",
```
mrgrauel, line 20: > I think without this change, prettier formatted it wrong for me. Is it necessary to remove it?

By @schaier-io

mrgrauel: We have two sources of truth. Prettier has a configuration `prettier.config.mjs`.
I would prefer to have one source of truth.

Could it be a configuration issue in your IDE?

mrgrauel:
> Is there a reason why we differ from the default shadcn path?

No, I changed it back. 👍 I thought it would be easier to identify when `shadcn` is used, but this seems to be related only to my knowledge of the framework.

web-app/.vscode/settings.json
schaier-io: ok lets try it for now, if I does not work at all, I will revert it

schaier-io (approved):

PR #72, opened 2025-03-12
I18n for auth pages

golddydev:
Add `i18n` to Sign In and Sign Up pages

There are 2 things which need `i18n`.

* Form fields (`label`, `description`, `placefolder`)
* Error messages
We use `zod` for form validation
But NOT using [`zod-i18n-map`](https://www.npmjs.com/package/zod-i18n-map) library
because it has several depedencies.

Rather make form `type` and use with `next-intl` itself


schaier-io (approved):
Good work

PR #74, opened 2025-03-12
fix(prisma): make agent rating required and update seed data

schaier-io:
# Pull Request Template

[] Documentation update  
[x] Bug fix  
[] New feature  
[] Other: 

## **Overview of Changes**
Aggregated rating added to seeding

## **Checklist**

### DIFF
--- a/web-app/package-lock.json
+++ b/web-app/package-lock.json
@@ -3440,13 +3440,6 @@
         "tailwindcss": "4.0.13"
       }
     },
-    "node_modules/@tailwindcss/node/node_modules/tailwindcss": {
-      "version": "4.0.12",
-      "resolved": "https://registry.npmjs.org/tailwindcss/-/tailwindcss-4.0.12.tgz",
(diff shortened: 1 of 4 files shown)


golddydev:
Is there any specific reason to have `Rating` as `required`?
Or do I miss something?

mrgrauel (changes requested):
After working with it I want to propose to keep the `AgentRating` nullable for the Agent.

Why?
`null` and `0` are two different pieces of information that we should handle in the frontend, not in the data.

schaier-io:
No it makes sense to have it always existing, this is just the aggregation of the rating. If there are none, it will show 0 ratings. The only think which would be ok, is to have the totalRating int nullable, but I feel like this is unnecessary

mrgrauel:
Now I finally understand the model 🤦 Sorry, I misinterpreted the model. It makes total sense that both are `0` when there are no ratings.

mrgrauel (approved):

PR #104, opened 2025-03-19
feat(db): changed to have a seperate tag table

schaier-io:
# Pull Request Template

[] Documentation update  
[] Bug fix  
[x] New feature  
[] Other: 

## **Overview of Changes**
Extract tags into seperate tables

Please understand that the Tag is unique and mapped. Meaning that the same Tag (name: A) will implicitly be mapped to possible multiple agent. Therefore do not delete tags neither directly nor via something like 
```
  const agent = await prisma.agent.findFirst({
    where: {
      Tags: {
        some: {
          name: "Marketing",
        },
      },
    },
  });
  await prisma.agent.update({
    where: { id: agent?.id },
    data: {
      Tags: { delete: { name: "Marketing" } },
    },
  });
```

As it would delete the tag for all agents referencing that tag use 
```
  await prisma.agent.update({
    where: {
      id: "test-123",
    },
    data: {
      Tags: { disconnect: { name: "Marketing" } },
    },
  });
```

## **Checklist**

### DIFF
--- a/web-app/prisma/migrations/20250319154402_initial/migration.sql
+++ b/web-app/prisma/migrations/20250319154402_initial/migration.sql
@@ -274,11 +274,11 @@ CREATE TABLE "Tag" (
 );
 
 -- CreateTable
-CREATE TABLE "_AgentTag" (
+CREATE TABLE "_AgentOnChainTag" (
     "A" TEXT NOT NULL,
     "B" TEXT NOT NULL,
 
-    CONSTRAINT "_AgentTag_AB_pkey" PRIMARY KEY ("A","B")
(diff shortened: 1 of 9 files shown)


mrgrauel (approved):
I just committed (d1518d76b51525eb8460d2510c04ae7b6ae5879c) to renaming `tags` to `onChainTags`. 

Please check and let me know if there was a specific reason for the original name. That was my only change request. You can either merge it or revert and explain why the previous name was used. I don't understand it, as everything else has the prefix.

I would understand if we removed the prefix `onChain` completely, but not just for one variable.

mrgrauel:
Oh damn a build error... I'm also looking into it.

mrgrauel (approved):

mrgrauel (approved):

PR #105, opened 2025-03-19
refactor(dto): migrate Tag and Status models to DTOs

mrgrauel:
This PR introduces Data Transfer Objects (DTOs) for Tag and Status models to decouple the database layer from the presentation layer.

Changes:
- Create `TagDTO` and `StatusDTO` interfaces
- Add conversion functions createTagDTO and createStatusDTO
- Replace direct Prisma Tag and Status usage with DTOs in components:
  - AgentDetails
  - FeaturedAgent
  - AgentCard
  - BadgeCloud
- Update AgentDTO to use new DTOs for consistency
- Rename related interfaces to follow DTO naming convention

This change improves the separation of concerns by preventing direct Prisma model usage in the UI layer and provides better type safety through explicit DTO interfaces.

### DIFF
--- a/web-app/src/app/(landing)/gallery/[agentId]/components/agent-details.tsx
+++ b/web-app/src/app/(landing)/gallery/[agentId]/components/agent-details.tsx
@@ -1,9 +1,9 @@
-import { Tag } from "@prisma/client";
 import Image from "next/image";
 import { useTranslations } from "next-intl";
 
 import BadgeCloud from "@/components/badge-cloud";
 import { Button } from "@/components/ui/button";
+import { TagDTO } from "@/lib/db/dto/TagDTO";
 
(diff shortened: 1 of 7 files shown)


schaier-io (changes requested):
Really see no benefit for that? Why don’twe keep the db types?

mrgrauel:
> Really see no benefit for that? Why don’twe keep the db types?

DTOs separate concerns between entities like the presentation layer and the domain model.

If we change the domain model, we only need to update the mapping to the presentation model rather than every component. 

See what you needed to do for the tags in #104. If you had simply mapped it to `string[]`, it would have worked out of the box without any changes to any components, as the presentation layer only requires a `string[]`.

schaier-io (approved):

PR #109, opened 2025-03-20
feat(auth): implement conditional header navigation based on auth state

mrgrauel:
This PR enhances the header navigation by implementing conditional rendering based on the user's authentication state. When a user is logged in, they will see a dashboard button instead of the auth buttons.

## Changes
- Created new `AppConnection` component that conditionally renders auth buttons or dashboard link
- Updated header component to use `AppConnection` instead of `AuthButtons`
- Added translation string for dashboard navigation
- Made header component asynchronous to support auth state checking

## Technical Details
- Uses `auth.api.getSession` to check authentication state
- Implements server-side rendering for optimal performance
- Maintains consistent UI with Shadcn UI button components


mrgrauel:
Sorry I forgot to add a [Suspense](https://react.dev/reference/react/Suspense). I added it with 7f782cbad4b4bc6f6e4f4d6ee4cb5f6fc09e87a0

web-app/src/lib/utils.ts
```diff
@@ -4,3 +4,6 @@ import { twMerge } from "tailwind-merge";
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
+
+export const simulateNetworkDelay = (ms: number) =>
```
golddydev, line 8: All is good, only what about naming this function `sleep` other than `simulateNetworkDelay`

This is typical sleep function

golddydev (dismissed):

golddydev (approved):

PR #119, opened 2025-03-21
refactor(auth): rename username to name and standardize password constraints

mrgrauel:
This PR improves the authentication system by standardizing field names and password constraints across the application.

Key Changes:
- Renamed "Username" field to "Name" throughout the application for better clarity
- Added PASSWORD_MAX_LENGTH constant (128 characters)
- Implemented consistent password length validation across all auth forms
- Configured better-auth to use the same password length constraints
- Removed debug console.log from signin action
- Added SOKOSUMI_URL constant to constants file
- Updated email templates to use "name" instead of "username"

### Technical Details
- Password constraints are now centralized in constants (min: 8, max: 128)
- Removed regex validation for name field (previously restricted to letters, numbers, and underscores)
- Updated translations to reflect the field name changes
- Applied consistent password validation across signup and reset password forms

### Testing
Please verify:
- Sign up flow with the new "Name" field
- Password validation on both signup and reset password forms
- Email templates render correctly with the user's name

### DIFF
--- a/web-app/src/constants/index.ts
+++ b/web-app/src/constants/index.ts
@@ -2,3 +2,7 @@ export const KEYBOARD_INPUT_DEBOUNCE_TIME = 300; // milliseconds
 
 export const MASUMI_URL = "https://masumi.network";
 export const KODOSUMI_URL = "https://kodosumi.com";
+export const SOKOSUMI_URL = "https://sokosumi.com";
+
+export const PASSWORD_MIN_LENGTH = 8;
+export const PASSWORD_MAX_LENGTH = 128;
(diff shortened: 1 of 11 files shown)


web-app/src/constants/index.ts
```diff
@@ -2,3 +2,7 @@ export const KEYBOARD_INPUT_DEBOUNCE_TIME = 300; // milliseconds
 
 export const MASUMI_URL = "https://masumi.network";
 export const KODOSUMI_URL = "https://kodosumi.com";
+export const SOKOSUMI_URL = "https://sokosumi.com";
+
+export const PASSWORD_MIN_LENGTH = 8;
+export const PASSWORD_MAX_LENGTH = 128;
```
mrgrauel, line 8: This is the default setting according to [better-auth](https://www.better-auth.com/docs/authentication/email-password#configuration), but I prefer to set it explicitly for zod-object validation.

PR #126, opened 2025-03-24
fix(eslint): enforce and implement nullish coalescing operator

mrgrauel:
### Changes
- Add [`@typescript-eslint/prefer-nullish-coalescing`](https://typescript-eslint.io/rules/prefer-nullish-coalescing) ESLint rule
- Configure TypeScript parser options for ESLint
- Replace logical OR (||) with nullish coalescing (??) operator in:
  - Authentication configuration
  - Forgot password form
  
### Why
The nullish coalescing operator (??) provides more precise handling of falsy values by only falling back to the default when the value is null or undefined. This prevents unexpected behavior with other falsy values like empty strings or zero.

### Technical Details
- Updated ESLint config to enforce nullish coalescing usage
- Fixed social provider configurations to use ?? for environment variables
- Updated form default values to use ?? for better type safety

### DIFF
--- a/web-app/eslint.config.mjs
+++ b/web-app/eslint.config.mjs
@@ -31,6 +31,7 @@ const eslintConfig = [
       "import/newline-after-import": "error",
       "import/no-duplicates": "error",
       "unused-imports/no-unused-imports": "error",
+      "@typescript-eslint/prefer-nullish-coalescing": "error",
       "@typescript-eslint/no-unused-vars": [
         "warn",
         {
@@ -44,6 +45,8 @@ const eslintConfig = [
     parserOptions: {
       sourceType: "module",
       ecmaVersion: "latest",
+      project: "./tsconfig.json",
+      tsconfigRootDir: __dirname,
     },
   }),
 ];
--- a/web-app/eslint.config.mjs
+++ b/web-app/eslint.config.mjs
@@ -1,12 +1,7 @@
 import { FlatCompat } from "@eslint/eslintrc";
-import { dirname } from "path";
-import { fileURLToPath } from "url";
-
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = dirname(__filename);
 
 const compat = new FlatCompat({
-  baseDirectory: __dirname,
+  baseDirectory: import.meta.dirname,
   recommendedConfig: {
     extends: ["eslint:recommended", "plugin:react/recommended"],
   },
@@ -46,7 +41,7 @@ const eslintConfig = [
       sourceType: "module",
       ecmaVersion: "latest",
       project: "./tsconfig.json",
-      tsconfigRootDir: __dirname,
+      tsconfigRootDir: import.meta.dirname,
     },
   }),
 ];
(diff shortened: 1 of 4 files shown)


web-app/eslint.config.mjs
```diff
@@ -31,6 +31,7 @@ const eslintConfig = [
       "import/newline-after-import": "error",
       "import/no-duplicates": "error",
       "unused-imports/no-unused-imports": "error",
+      "@typescript-eslint/prefer-nullish-coalescing": "error",
```
schaier-io, line 34: We have strictNullCheck enabled implicitly, right? As far as I see yes, but maybe you can double-check

schaier-io: Ok as far as I see it is by default enabled as long as we have https://www.typescriptlang.org/tsconfig/#strict maybe we should still add it in case this ever changes?

mrgrauel: Please Google this yourself in the future if you're unsure. It doesn't help if I search for it and then paste the response here. You can refer to this link: https://www.typescriptlang.org/tsconfig/#strict.

I checked it based on this reference:
> This rule will not work as expected if [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks) is not enabled.

If you had Googled it, you would have seen that it is enabled. Now, I need to write this comment and request a review again.

schaier-io: Ok I see the second comment with the context was not added directly, very strange. Should we add the strictNull checks regardless as mentinoed? In order to do be explicit

mrgrauel: Make a call. I don't care.

schaier-io: added in https://github.com/masumi-network/sokosumi/pull/126/commits/0a110609f7cb88be5882a472e348211db999964c

web-app/eslint.config.mjs
```diff
@@ -44,6 +45,8 @@ const eslintConfig = [
     parserOptions: {
       sourceType: "module",
       ecmaVersion: "latest",
+      project: "./tsconfig.json",
+      tsconfigRootDir: __dirname,
```
golddydev, line 49: We need to use `import.meta.dirname` since it is `.mjs`

web-app/tsconfig.json
golddydev, line 1: We need to include `eslint.config.mjs`, since that is not included, and IDE shows an error

image

golddydev (changes requested):

web-app/eslint.config.mjs
mrgrauel: Nice catch 👍 

64e3d3e213c9ef39a39bd49e57da378d97f7f971

web-app/tsconfig.json
mrgrauel: Nice catch 👍 

f02d6ec633a177a8c0b20e7cfef093ade1a52908

mrgrauel: [9027d61](https://github.com/masumi-network/sokosumi/pull/126/commits/9027d6115317e021df4f4e0366b09ca8517f281b)

golddydev (approved):

PR #127, opened 2025-03-24
feat: ensure env variables are valid and globally available

schaier-io:
# Pull Request Template

[] Documentation update  
[] Bug fix  
[x] New feature  
[] Other: 

## **Overview of Changes**
This PR shall ensure that all the .env variables are configured correctly.
No more use of process.env everywhere in server code.

For now there is only a need for server side code, but we can extend it later on for a client side code.
Naming explicitly has the `envServer`, that nobody sends this object to a frontend, as it contains all config variables. 
In the future, we might consider using a lib to have the env data encrypted for production

## **Checklist**


web-app/src/lib/better-auth/auth.ts
```diff
@@ -12,6 +12,7 @@ import { getTranslations } from "next-intl/server";
 
 import { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH } from "@/constants";
 
+import { envServer } from "../../../config/env.config";
```
golddydev, line 15: Everything is great.

Only can we add path alias for this?
Importing path is not looking good for me.

something like `config/env.config`?

Or even we can have this file under `src/configs` which is already there and
import `@/configs/env.config`

schaier-io: yes will resolve this for all in an upcoming new PR with an eslint config, as it is in many places in the code. Updated the location for now in https://github.com/masumi-network/sokosumi/pull/127/commits/c78937b21e28015b237785ff2cc56cad49f4ff0d

mrgrauel (changes requested):
Build fails

schaier-io:
I will wait until I have DO access to fix the deploy there

mrgrauel:
The DATABASE_URL isn't read using this approach because it's located within the `prisma.schema`. Instead, we could add the `validateEnv()` function to the `next.config.ts` before the app is created. This way, all environment variables are checked upfront rather than when we access the first variable.

This approach also prevents incorrect access to the environment variables. If we implement this, everyone must remember to access environment variables using `envServer`. It would also resolve the issue with CI/CD, as it is not needed at build time but rather at runtime when it is crucial.

schaier-io:
> The DATABASE_URL isn't read using this approach because it's located within the `prisma.schema`. Instead, we could add the `validateEnv()` function to the `next.config.ts` before the app is created. This way, all environment variables are checked upfront rather than when we access the first variable.
> 
> This approach also prevents incorrect access to the environment variables. If we implement this, everyone must remember to access environment variables using `envServer`. I would propose to remove `envServer` and only validate the existence of the variables at via a method call in `next.config.ts`.
> 
> This way we and everybody can use `process.env` as people are used to to access environment variables, but the variables are verified.

I am not sure I did not find any way to set this up in the `next.config.ts` or do you mean to set it up manually? Because if you use process.env.* it will not be correctly validated. You could for example access a `console.log("middleware", process.env.testInvalidVariable);` which builds fine. With the above aproach you would have a build time error.

The leaking of the secret is also more a theoretical problem, you just should be aware, that the envServer has the secrets. So just don't do something like `return {envServer}` in a server side function or similar. Which is also possible with the `process.env`

The only thing where we still might want to use the process.env is in the NEXT_PUBLIC variant, as they apparently get inlined during the build process. This however is not the case for serverside vars, just tested it.

mrgrauel:
@schaier-io my idea can be seen in #128. It was easier to implement as to describe fully.

schaier-io:
closed as of [this change](https://github.com/masumi-network/sokosumi/pull/128)

PR #216, opened 2025-04-01
feat(web-app): add job inputs

golddydev:
Added Inputs according to each job input schema

- [x] String type
- [x] Boolean type
- [x] Number type
- [x] Option type
- [x] None type

## How to test it?

`/app/jobs/[agentId]` will show Job inputs (from dummy data) on the right section.

[Dummy data](https://github.com/masumi-network/sokosumi/pull/216/files#diff-487a3f91d4a7b627ae8808cdfa58437601a73e1a027270e6e8c784b2eda5c428)

image

Will use real data from agent api in next PR


golddydev:
Just added max validation to multi select UI for test

mrgrauel:
@golddydev When I click `clear`, I see validation errors. We haven't discussed the requirements, but I don't expect this to happen when I click clear as a user. Did you implement it this way on purpose?

web-app/src/config/env.config.ts
```diff
@@ -47,23 +47,23 @@ const envSchemaSecrets = z.object({
     .default("https://payment.masumi.network/api/v1"),
   // BetterAuth Settings
   BETTER_AUTH_SESSION_EXPIRES_IN: z
-    .number()
+    .number({ coerce: true })
```
mrgrauel, line 50: Why did you add all these? 🤔

golddydev: because process.env.[key] is string by default.
So we need coerce to convert them to number

mrgrauel (approved):

PR #235, opened 2025-04-01
chore: remove requests dependency from job-cron-caller

mrgrauel:
This PR removes the `requests` package dependency from the project. The package was previously listed in `requirements.txt` but is no longer needed.

Changes:
- Deleted `requirements.txt` file containing `requests==2.31.0`
- Replace `npm ci` with `npm i` in workflow files


.github/workflows/deploy-main-job-function.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           token: ${{ secrets.DIGITALOCEAN_ACCESS_TOKEN }}
 
       - name: Install dependencies
-        run: npm ci
```
schaier-io, line 34: This was on purpose, as digital ocean does not want the package lock json file to be included and then you need to run the full install

schaier-io (approved):

PR #291, opened 2025-04-07
refactor(credits): improve credit cost handling and display

mrgrauel:
This PR refactors the credit system to improve consistency and correctness in handling credit calculations and display.

Key Changes:
- Rename `calculateAgentCreditCost` to `calculateAgentHumandReadableCreditCost` for clarity
- Fix credit transaction validation logic in `creditTransactionSpend`
- Update credit cost calculations to work with bigint throughout the system
- Adjust credit display formatting to be more consistent
- Update seed data with corrected credit cost values
- Fix credit cost comparison logic in job start validation

Technical Details:
- Credit calculations now maintain bigint precision until final display
- Credit transaction validation properly checks for negative values
- Credit costs in seed data updated to reflect correct pricing (0.50$ per ADA)
- User credit balance display moved to human readable format at the service level


web-app/prisma/seed.ts
```diff
@@ -72,11 +72,11 @@ const seedCreditCost = async () => {
       unit: "usdm",
     },
     update: {
-      creditCostPerUnit: convertCreditsToBaseUnits(1),
+      creditCostPerUnit: 1,
```
golddydev, line 75: what about we name this `creditCostInUnits`?

Sorry to bring this naming all time 😓 (really crucial to me)

web-app/src/lib/db/services/credit.service.ts
```diff
@@ -100,7 +98,7 @@ export async function calculateAgentCreditCost(
  */
 export async function calculateCreditCost(
   amounts: { unit: string; amount: number }[],
-): Promise<number> {
+): Promise<bigint> {
```
schaier-io, line 101: Can you check if this works across boundaries, as serilization works, as bigint is not supported on some function returns

schaier-io (dismissed):

web-app/src/lib/db/services/credit.service.ts
mrgrauel: It's used three times in the codebase:

1. In the API for credit requests, where it is converted to a string.
2. In the credit service - no issues
3. In the job service - no issues

As far as I know, the only boundary issue is with JSON.

schaier-io: @mrgrauel Do we use this function somewhere on the client side? I think there was a bug with serilization, when I tested it for the demo and using bigint. 

If the whole flow with agents works, then you can resolve and merge

schaier-io (approved):

web-app/src/lib/db/services/credit.service.ts
```diff
@@ -119,15 +117,14 @@ export async function calculateCreditCost(
     if (!creditCost) {
       throw new Error(`Credit cost not found for unit ${amount.unit}`);
     }
-    const cost = Number(creditCost.creditCostPerUnit) * amount.amount;
+    const cost = amount.amount * Number(creditCost.creditCostPerUnit);
```
golddydev, line 120: This formula seems weird to me.
`amount.amount` means the amount of unit (`lovelace` or `usdm`)
`creditCost.creditCostPerUnit` means "how many units for 1 credit"

And multiplying `amount of unit` and `how many units for 1 credit` seems werid

E.g.

If amount is `[ { amount: 1, unit: "usdm" }, { amount: 1, unit: "" } ]`, it is 1 usdm and 1 lovelace

`cost` is 1 * 1  + 1 * 500_000

Do I think something wrong?

web-app/prisma/seed.ts
mrgrauel: I updated the comment in code....

Does it help you:
```
1 usdm = 10^6 base unit usdms; 1usdm = 10^12 credits (1.0 credits) -> 1 base unit usdm == 10^6 credits
```

web-app/src/lib/db/services/credit.service.ts
schaier-io: No it is the opposite its not how many units for 1 credit it is -> how many credits it cost to spend one unit

mrgrauel: You have a misunderstanding of `creditCostPerUnit`. It doesn't mean "How many units per 1 credit". It means that 1 lovelace costs 500_000 credits.

schaier-io: So for example if one agent would cost 1 lovelace you would need to spend 500_000 credits

schaier-io: if it costs 1 ada it would cost 1_000_000 lovelace=1ada ->1_000_000*500_000 credits

golddydev (approved):
Thanks for explanations from both of you

PR #326, opened 2025-04-09
fix: update option input values as index number

golddydev:
## Changes

* Use index (number) for option values instead of option values
* Remove `transformFormData` function and just add `filterOutNullValues` to filter out `null` value
* Check option data `values` are unique or not

## Note

The select (both single and multi) will not work correctly if there is duplicate value in data `values` from schema
Because index is depend on `Array.indexOf` which only takes first index

### DIFF
--- a/web-app/messages/en.json
+++ b/web-app/messages/en.json
@@ -576,7 +576,8 @@
             "value": {
               "required": "Option is required"
             },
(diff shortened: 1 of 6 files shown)


mrgrauel (approved):
Awesome 👍

PR #347, opened 2025-04-11
feat(stripe): move customer creation to the checkout

mrgrauel:
# Description
This PR introduces a mechanism for creating Stripe customers during the checkout process. Although @better-auth/stripe tries to create customers upon signup, this may fail. Therefore, we need to implement a customer creation solution during checkout. This PR ensures that customer creation occurs reliably during the user's first credit top-up, if necessary.

We don't use any other features of @better-auth/stripe, which is why I removed it from the codebase. A user with a `stripeCustomerId` is someone who has purchased credits.

## Key Changes

- Add `customer_creation: "always"` parameter to checkout session creation when no `stripeCustomerId` exists
- Implement webhook handler for `customer.created` events to link newly created customers to users
- Refactor webhook handlers into separate functions for better maintainability

## Technical Details

- The checkout session will now attempt to create a Stripe customer if one doesn't exist
- New webhook handler listens for `customer.created` events and links them to users
- Added error handling for cases where customer email is missing or user not found

## Testing

- Test checkout flow for users without and without `stripeCustomerId`
- Validate `customer.created` webhook handling


mrgrauel:
I bypass the branch rules, because @golddydev and @bytegen-dev cannot test the webhook without a Stripe account and @schaier-io is focusing on other tasks.

PR #353, opened 2025-04-12
fix(agent): Add agent tags and refactor query state management

mrgrauel:
This PR introduces agent tags to the registry sync and refactors the filter state management using the `nuqs` library.

### Features

-   **Agent Tag Filtering:** Users can now filter agents in the gallery based on associated tags. The selected tags are reflected in the URL query parameters.
-   **Registry Sync Update:** The cron job responsible for syncing agent data from the registry (`/api/cron/registry-sync`) now includes processing and storing agent tags.

### Refactoring

-   **Query State Management:** Replaced the custom implementation in `use-gallery-filter.ts` with the `nuqs` library for managing URL query state (`query` and `tags`). This simplifies the hook and removes the need for manual debouncing and URL parsing/updating.
-   **Root Layout:** Integrated `NuqsAdapter` into the root layout (`src/app/layout.tsx`) as required by `nuqs`.

### Chores

-   **Dependencies:** Added `nuqs` to project dependencies.
-   **Cleanup:** Removed the unused `NEXT_PUBLIC_KEYBOARD_INPUT_DEBOUNCE_TIME` environment variable and its usage.

### DIFF
--- a/web-app/package.json
+++ b/web-app/package.json
@@ -59,6 +59,7 @@
     "next": "^15.3.0",
     "next-intl": "^4.0.2",
     "next-themes": "^0.4.6",
+    "nuqs": "^2.4.1",
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
     "react-error-boundary": "^5.0.0",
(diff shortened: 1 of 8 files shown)


golddydev:
Found a little issue which caused there is no debounce query input.

We still need debounce for input query.

golddydev (approved):
That is super simple 👍🏼

PR #489, opened 2025-04-24
feat: add with-authentication hooks and apply to hire agent button

golddydev:
## Changes

* Added `AuthenticationModal`  component
* Added `use-with-authentication` and `use-authentication-modal` hooks
* Apply use-with-authentication to `Hire Agent` button

## Benefit

* Re-usability of `use-with-authentication` hook

### DIFF
--- a/web-app/messages/en.json
+++ b/web-app/messages/en.json
@@ -93,6 +93,14 @@
       "jobs": "Jobs",
       "agents": "Gallery",
       "account": "Account"
+    },
+    "Modals": {
+      "AuthenticationModal": {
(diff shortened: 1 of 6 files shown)


mrgrauel (changes requested):
It's really nice that you implemented the feature, but the view is very width compared to the design.

image

image

golddydev:
Okay, let me fix width.

golddydev:
@mrgrauel  updated width to `w-sm`

mrgrauel (approved):

Issue #484, opened 2025-04-23
Update visuals for the authentication flow


mrgrauel:
### User Story:
As a user, I want to see a visual instead of a placeholder so that I have a positive emotion.

### Description:
* Update Visuals on Authentication Flow Pages:
   * `/login`
   * `/register`
   * `/forgot-password`
   * `/reset-password`

PR #520, opened 2025-04-28
feat: add auth background

golddydev:
## Changes

* Move auth layout out of landing layout
* Add auth background with Kanji
* Add auth layout's header (logo) and footer (legal links)

## Related

 #484


mrgrauel (changes requested):
Shouldn't the design be 50/50 in the width? 🤔 

What do you think, should we limit the height and/or width? At some point it's too much, isn't it?

golddydev:
> Shouldn't the design be 50/50 in the width? 🤔
> 
> What do you think, should we limit the height and/or width? At some point it's too much, isn't it?

Hmmm, well I thought I looks weird when screen is too big (so I just put the limit)
Well if that is what it should be, then I am fine

mrgrauel (approved):

PR #568, opened 2025-05-05
fix(agents): prevent crash when agent CreditCost is unavailable

mrgrauel:
### Summary  
This PR addresses an issue where the `/agents` landing page would crash if an agent's `CreditCost` could not be determined (e.g., when `getAgentCreditsPrice` throws). The solution ensures that only agents with a successfully resolved `CreditCost` are rendered, and gracefully handles cases where no agents are available.

### Details  
- Uses `Promise.allSettled` to fetch agent credit prices, filtering out agents where `getAgentCreditsPrice` fails.
- Only agents with a valid `CreditCost` are passed to the UI components.
- If no agents with a valid `CreditCost` are available, the `AgentsNotAvailable` component is rendered.
- Prevents runtime errors and improves user experience by handling missing or failing price data.

### Testing  
- Verified that the page no longer crashes when some agents have missing or failing `CreditCost` data.
- Confirmed that agents with valid prices are displayed as expected.
- Confirmed that the "not available" state is shown if all agents fail to provide a price.

### DIFF
--- a/web-app/src/app/(landing)/agents/page.tsx
+++ b/web-app/src/app/(landing)/agents/page.tsx
@@ -22,27 +22,48 @@ export async function generateMetadata(): Promise<Metadata> {
 export default async function GalleryPage() {
   const agents: AgentWithRelations[] = await getOnlineAgents();
 
-  if (!agents.length) {
-    return <AgentsNotAvailable />;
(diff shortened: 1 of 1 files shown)


mrgrauel:
@golddydev I merged because it fixes a crash on prod database.

PR #601, opened 2025-05-07
feat(auth): grant free credits to new users using the `FREE_CREDITS_ON_SIGNUP` environment variable

mrgrauel:
### 🆕 New Feature  
This PR introduces an automatic credit reward for new users upon signup:

- Every new user now receives 0 free credits when their account is created.
- The amount is configurable via the `FREE_CREDITS_ON_SIGNUP` environment variable (default: 50).
- Credits are granted using a new credit transaction, leveraging the existing credit system.
- Implementation uses a database hook in the authentication flow to ensure credits are assigned after user creation.

### 🛠️ Technical Details  
- Added `FREE_CREDITS_ON_SIGNUP` to the environment config schema.
- Updated the authentication logic to create a credit transaction for new users.
- Extended the credit transaction repository with a `createCreditTransaction` function.

### 🧪 Testing  
- Verified that new users receive the correct amount of credits on signup.
- Confirmed that the transaction is created and associated with the user.


golddydev (approved):
👍🏼

PR #616, opened 2025-05-08
fix(db): resolve race conditions in lock implementation for sync endpoints

mrgrauel:
## Summary

This PR refactors the distributed lock implementation used by the `/api/sync/agents` and `/api/sync/jobs` endpoints to address potential race conditions and ensure atomic lock acquisition and release. The changes improve reliability and correctness when multiple instances attempt to acquire the same lock concurrently.

## Details

- Replaces the previous lock logic (`getLock`, `releaseLock`) with new atomic functions: `acquireLock` and `unlockLock`.
- Ensures lock acquisition is performed atomically using `updateMany` and proper checks for lock expiration.
- Handles lock expiration and force-unlock scenarios to prevent deadlocks.
- Updates API routes to use the new lock functions and provide more accurate error responses (e.g., 429 for "already in progress", 500 for unexpected lock errors).
- Removes obsolete lock service code and cleans up related imports.
- Adjusts lock timeout minimum in environment config for greater flexibility.
- Ensures all lock operations are performed within a transaction for consistency.

## Motivation

The previous implementation was susceptible to race conditions where multiple instances could acquire the same lock due to non-atomic checks and updates. This could lead to data corruption or duplicate sync operations. The new approach guarantees that only one instance can hold the lock at a time, even under high concurrency.

## Impact

- More robust and reliable distributed locking for sync operations.
- Improved error handling and observability for lock-related failures.
- No breaking changes to API consumers.


web-app/src/lib/db/lock/repo.ts
```diff
@@ -3,50 +3,89 @@
...
+    return await createLockByKey(key, instanceId, tx);
+  }
+  // If lock exists and is not expired, return the lock
+  if (existingLock.isLocked && !isLockExpired(existingLock.lockedAt)) {
+    throw new Error("LOCK_IS_LOCKED");
+  }
+
+  // If lock exists and is expired, force unlock
+  if (existingLock.isLocked && isLockExpired(existingLock.lockedAt)) {
+    await tx.lock.update({
```
golddydev, line 64: Should we return here?

golddydev (changes requested):

web-app/src/lib/db/lock/repo.ts
golddydev: Nah, I was wrong, you are right, we should't return

golddydev (approved):

PR #650, opened 2025-05-12
fix: filter agents with valid pricing

golddydev:
## Changes

* Add `getOnlineAgentsWithValidPricing` function to filter agents with valid credit costs units
* Replace `getOnlineAgents` with this function
* Use `getOnlineAgentsWithCreditsPrice` function in app gallery too


web-app/src/lib/services/agent/service.ts
```diff
@@ -18,6 +19,33 @@ import {
...
+ */
+export async function getOnlineAgentsWithValidPricing(
+  tx: Prisma.TransactionClient = prisma,
+): Promise<AgentWithRelations[]> {
+  // get all credit costs
+  const creditCosts = await getAllCreditCosts(tx);
+  // also include "lovelace" because that is valid (parsed empty string while calculating)
+  const validCreditCostUnits = Array.from(
+    new Set([...creditCosts.map((creditCost) => creditCost.unit), "lovelace"]),
+  );
```
mrgrauel, line 41: This is only valid if '' is in the creditCosts. On production we currently only support the USDM creditcost, so adding `lovelace` is wrong.

golddydev: Ah, you are right

golddydev: [96b182e](https://github.com/masumi-network/sokosumi/pull/650/commits/96b182e5bccc09ad68e4c750a8aa4d0221dd8947)

mrgrauel (approved):

PR #675, opened 2025-05-14
feat: make create job section as modal

golddydev:
## Changes

* Improve Loading components of `jobs/` layout.
* Make `JobDetailRedirect` so when there are any jobs, users will be redirect to first job detail page.
* Introduce `CreateJobModalContext` to handle all state within create job modal.
* Make `CreateJobModal`, `CreateJobModalTrigger`, `CreateJobModalHeader` ... components
* Use React's `use` function to overcome client component limitation (no async function in client component)

## Benefit

* Can handle `CreateJobModal` with one modal component under `CreateJobModalContext`
* Really helpful when we implement `draft` function for create job


mrgrauel:
It's probably unrelated, but I noticed it. The backgrounds of the modal sheet is different. Is that on purpose?

image

image

mrgrauel (approved):

mrgrauel (dismissed):

mrgrauel (approved):

PR #688, opened 2025-05-16
feat: add example outputs playlist modal

golddydev:
## Changes

* Add Agents' Example Outputs Playlist modal
* Make Example Output Thumbnail components (for now it is dummy)
* Update AgentHireButton to accept size as prop

### DIFF
--- a/web-app/messages/en.json
+++ b/web-app/messages/en.json
@@ -64,7 +64,8 @@
           "title2": "At a Glance"
         },
         "Section4": {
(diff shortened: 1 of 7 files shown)


mrgrauel (changes requested):
As discussed we don't show the Playlist component. We show the download button on the detail page for each output, as long as we don't show a preview.

mrgrauel (approved):

PR #699, opened 2025-05-19
feat: add organization to schema

golddydev:
### Database changes

* Add `Organization`, `Member`, `Invitation` model
* Add `data-migration.ts` script to run after schema migration to create a member for all users using their email domain.
* Added `data-migration:add-organization` script to `package.json`. 

## Web app change

* Added db repositories for `Organization` and `Member` model.
* Added actions
  * `createOrganizationFromName`: Create organization from name. We generate `slug` from name. (slugify name and attach uuid)
  * `createOrganizationMember`: Create Member and connect user and organization
* Added organization input to `register` form with validation and error message. (using `zod` schema)
* Added form to create organization as popover.

## Benefit

* From now on, all users can have organization so it is easy to group them

## How to migrate

* Migrate prisma schema.
* Run post migrate script `data-migration:add-organization`.


mrgrauel (changes requested):
It's very good, but could we do small adjustments to the order of the fields?

The order should be name, organization, email, password, confirm password. Could you add this to the pull request?

image

mrgrauel:
Additionally I disabled the auto-merge, because I would like to the migration together with you. We cannot merge, because it automatically triggers a deployment without migration.

golddydev:
@mrgrauel 
Changed order in this commit
[fa9b9c0](https://github.com/masumi-network/sokosumi/pull/699/commits/fa9b9c05868a0c56a369da6fd7f981bf08350f6c)

mrgrauel (approved):

PR #704, opened 2025-05-20
fix(migration): improve job-sync data migration and job status handling

mrgrauel:
### Overview

This PR enhances the robustness and correctness of the job-sync data migration and related job status handling logic. The main focus is on ensuring that job status fields are accurately migrated and that job synchronization processes are more reliable and maintainable.

### Key Changes

- **Data Migration Script (`data-migration.ts`):**
  - Added a new migration script to update the `agentJobStatus` field for all jobs based on their output status.
  - Ensures that only jobs with valid output are updated, and parses the output safely.
  - Uses a transaction for atomicity and reliability.

- **Job Sync API (`/api/sync/jobs/route.ts`):**
  - Improved filtering logic to exclude jobs that are already completed and whose external dispute unlock time has passed.
  - Ensures only relevant jobs are processed, reducing unnecessary sync operations.

- **Job Repository (`lib/db/job/repo.ts`):**
  - Added a check to prevent double refunding of jobs by verifying if a `refundedCreditTransaction` already exists before proceeding.

- **Job Service (`lib/services/job/service.ts`):**
  - Increased timeouts for job start and sync operations to handle longer-running jobs.
  - Improved error handling and transaction management in job synchronization.
  - Enhanced agent job status sync to handle missing agents gracefully and to update job status/output more reliably.

### Rationale

- **Data Integrity:** The migration ensures that all jobs have their `agentJobStatus` field correctly set, which is critical for downstream logic and UI consistency.
- **Performance:** Filtering out already-completed jobs with expired dispute windows reduces load and improves sync efficiency.
- **Reliability:** Additional checks and error handling prevent duplicate operations and make the sync process more robust.
- **Maintainability:** Refactored code for clarity, modularity, and future extensibility.

### Testing

- Migration script tested against existing job data to verify correct status updates.
- Job sync endpoint and service tested to ensure only relevant jobs are processed and that status updates are accurate.
- Refund logic verified to prevent duplicate credit transactions.

### DIFF
--- a/web-app/prisma/migrations/20250519112751_job_sync/data-migration.ts
+++ b/web-app/prisma/migrations/20250519112751_job_sync/data-migration.ts
@@ -1,0 +1,30 @@
+import { jobStatusToAgentJobStatus } from "@/lib/db/job/utils";
+import { PrismaClient } from "@/prisma/generated/client";
+
+const prisma = new PrismaClient();
+
+async function main() {
+  await prisma.$transaction(async (tx) => {
+    const jobs = await tx.job.findMany();
+    for (const job of jobs) {
+      const output = job.output;
+      if (output) {
+        const outputJson = JSON.parse(output);
+        const agentJobStatus = jobStatusToAgentJobStatus(outputJson.status);
+        await tx.job.update({
+          where: { id: job.id },
+          data: {
+            agentJobStatus: agentJobStatus,
+          },
+        });
+      }
+    }
+  });
+}
+
+main()
+  .catch(async (e) => {
+    console.error(e);
+    process.exit(1);
+  })
+  .finally(async () => await prisma.$disconnect());
--- a/web-app/prisma/migrations/20250519112751_job_sync/data-migration.ts
+++ b/web-app/prisma/migrations/20250519112751_job_sync/data-migration.ts
@@ -4,22 +4,35 @@ import { PrismaClient } from "@/prisma/generated/client";
 const prisma = new PrismaClient();
 
 async function main() {
-  await prisma.$transaction(async (tx) => {
-    const jobs = await tx.job.findMany();
-    for (const job of jobs) {
-      const output = job.output;
-      if (output) {
-        const outputJson = JSON.parse(output);
-        const agentJobStatus = jobStatusToAgentJobStatus(outputJson.status);
-        await tx.job.update({
-          where: { id: job.id },
-          data: {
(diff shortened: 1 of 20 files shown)


web-app/src/lib/services/job/service.ts
```diff
@@ -120,33 +120,40 @@ export async function startJob(input: StartJobInputSchemaType): Promise<Job> {
...
+            await postPaymentClientRequestRefund(job);
+          }
+          break;
+        default:
+          break;
+      }
+    },
+    {
+      maxWait: 5000, // default: 2000
+      timeout: 20000, // default: 5000
```
schaier-io, line 160: Having this long of a lock is a major db overhead, maybe you can extract some stuff before the tx?

mrgrauel: Can you please give more context or an example? It seems like that you have something in mind if you ask like that.

web-app/src/lib/services/job/service.ts
```diff
@@ -120,33 +120,46 @@ export async function startJob(input: StartJobInputSchemaType): Promise<Job> {
...
+    },
+  );
 }
 
 async function syncRegistryStatus(
```
golddydev, line 165: In this function, we do 3 db write operations.
To reduce time, we can wrap them all into one operation (but this kills code readability)

web-app/prisma/migrations/20250519112751_job_sync/data-migration.ts
```diff
@@ -0,0 +1,43 @@
...
+          await tx.job.update({
+            where: { id: job.id },
+            data: {
+              agentJobStatus: agentJobStatus,
+            },
+          });
+        }
+      }
+    },
+    {
```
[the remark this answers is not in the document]
golddydev: My suggestion is to use `p-limit` here to split the jobs into chunks and handle one chunk by one.

web-app/src/lib/services/job/service.ts
mrgrauel: @schaier-io I extracted the requests to the registry and agent in cc26015a448b17514a11cef5a8112ff2f89bf9fb

web-app/src/lib/services/job/service.ts
mrgrauel: b1247cfcf19ae71e1c9aaa6a422f1b9ea8d1b78e

golddydev (approved):

PR #768, opened 2025-05-30
feat: implement user authentication verification

golddydev:
## Changes

* Add user authentication verification to several services
* Move and update compareApiKeys function for centralized auth handling

## Convention

* Do not directly pass `userId` to db repo functions from client.


mrgrauel (changes requested):
image

The sync methods don't work anymore.

PR #980, opened 2025-06-25
refactor: keep alive job status event stream

golddydev:
## Changes

* Send date string in interval on job status event stream to keep connection alive.
* Make cookie consent animated when appear and disappear


golddydev:
bugbot run

PR #996, opened 2025-06-29
Fix security updates

schaier-io:
Various fixes


web-app/src/app/(auth)/login/components/form.tsx
```diff
@@ -76,8 +76,8 @@ export default function SignInForm({
               // Invalid URL, fallback to /app
             }
           }
-          router.push(redirectUrl);
-          router.refresh();
+          await router.push(redirectUrl);
+          await router.refresh();
```
golddydev, line 80: Why do we have to use async router here?

web-app/src/app/(auth)/register/components/form.tsx
```diff
@@ -14,31 +13,22 @@ import {
...
-} from "@/lib/actions";
-import { authClient } from "@/lib/auth/auth.client";
-import {
-  isEmailAllowedByOrganization,
-  OrganizationWithRelations,
-} from "@/lib/db";
-import { Organization } from "@/prisma/generated/client";
+import { useAsyncRouterPush } from "@/hooks/use-async-router";
+import { signUpEmail } from "@/lib/actions/auth";
+import { OrganizationWithRelations } from "@/lib/db/organization/types";
```
golddydev, line 18: Let's just only extract repo.ts files from `@/lib/db/` and keep everything else there, 
so we don't have to change import path much.

web-app/src/lib/services/agent/action.ts
golddydev, line 1: Let's not put action.ts here.
We should make this as router handler.

Making server action for fetch is the last resort. (Next.js doesn't recommend this for some reasons)

web-app/src/app/(auth)/register/components/form.tsx
schaier-io: All import paths are changed, already. This build works.

it was necessary to refactor the index.ts import mess crashing the build on various implicit imports.

This direct import should also reduce bundle size

web-app/src/lib/services/agent/action.ts
schaier-io: We can come up for a different naming convention. In the end we need one file which uses the use server file for server action type things, or refactor client into server components

web-app/src/app/(auth)/login/components/form.tsx
golddydev: Okay, 👍🏼

web-app/src/app/(auth)/register/components/form.tsx
golddydev: The bundle size only reduces much with tree-shaking, which is not happening for our case.
Because we use all of the functions in our application anyway.

schaier-io: No not clientside, via direct imports the functions will be ignored

web-app/src/lib/db/agent/action.ts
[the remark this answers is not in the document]
schaier-io: @golddydev resolved in https://github.com/masumi-network/sokosumi/pull/996/commits/434929face7e9e060d92398e368c4d940fd5655e

web-app/src/lib/services/agent/action.ts
golddydev, line 1: This is for services.
We should put action else where

web-app/src/app/(auth)/register/components/form.tsx
golddydev: Ah, I got what you mean.

schaier-io:
merged all changes into #997

PR #997, opened 2025-06-30
feat: fix authentication and authorization flow

golddydev:
TBD


mrgrauel (dismissed):
As discussed please talk with Sandro.

golddydev:
bugbot run

mrgrauel:
Bugbot run

schaier-io (dismissed):
In general it is a good rework.

However I reccomend in the future finding a clearer structure for the "server-only" functions (or in general)

For example
- service repos, sometimes works without session (relies on proper auth in an other place to not leak data/ddos...), in other places it uses the session to retrieve user specific information... Without looking into a function I have no idea what does what

-> In my opinion and I think also NextJs docs reccomend to keep the session check and data fetsching together, so it can not be forgetten (that is for functions called from page/layout...). This however makes it necessary to have a clear structure, similar to this "action" folder rework. E.g. what is called from pages... directly, what is internal from "server-only" and what is a helper function for general use.

I do not think this needs to be done in this PR, but should be discussed and maybe implemented seperatly

mrgrauel:
BugBot run

golddydev:
bugbot run

schaier-io (approved):

PR #1025, opened 2025-07-03
feat(organization): add organization credit balance support

mrgrauel:
### ✨ Features
- Added support for organization-level credit balances in the database and Prisma schema.
- Updated billing and transaction models to associate credits and fiat transactions with organizations.
- Implemented organization switcher in user avatar dropdown for seamless context switching.
- Displayed organization credit balances in the UI, including a new Organization Credits card.
- Improved i18n for billing and organization-related components.

### 🛠️ Technical
- Database migrations for new organizationId fields and relations.
- Refactored credit and fiat transaction queries to support both user and organization contexts.
- Updated billing page and credit display logic to handle organization context.
- Enhanced translations for billing, organizations, and switcher components.

### 🧪 Testing
- Verified organization switching updates credit balance displays.
- Confirmed billing actions work for both personal and organization accounts.
- Checked UI responsiveness and dark/light mode compatibility.

### DIFF
--- a/web-app/prisma/seed.ts
+++ b/web-app/prisma/seed.ts
@@ -1,15 +1,16 @@
-import { getEnvPublicConfig } from "@/config/env.public";
-import { getEnvSecrets } from "@/config/env.secrets";
+/* eslint-disable no-restricted-properties */
 import { usdmUnit } from "@/lib/utils";
 import { PrismaClient } from "@/prisma/generated/client";
 
+type Network = "Mainnet" | "Preprod" | "Preview";
+
 const prisma = new PrismaClient();
 
-const seedDatabase = getEnvSecrets().SEED_DATABASE;
+const seedDatabase = process.env.SEED_DATABASE === "true";
 
 const seedUSDMCreditCost = async () => {
   console.log("Seeding USDM credit cost...");
-  const unit = usdmUnit(getEnvPublicConfig().NEXT_PUBLIC_NETWORK);
+  const unit = usdmUnit(process.env.NEXT_PUBLIC_NETWORK as Network);
   await prisma.creditCost.upsert({
     where: {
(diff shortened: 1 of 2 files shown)


web-app/src/app/app/components/user-avatar/organization-switcher.tsx
```diff
@@ -0,0 +1,145 @@
...
+    (member) => member.organization.id === activeOrganizationId,
+  )?.organization;
+
+  const handleSelectOrganization = (organizationId: string | null) => {
+    startTransition(async () => {
+      try {
+        await authClient.organization.setActive({
+          organizationId,
+        });
+        await revalidateAppPath();
```
golddydev, line 54: I don't think we need to revalidateAppPath, since we are doing router.refresh, we will refetch data.

web-app/src/lib/db/repositories/user.ts
```diff
@@ -19,6 +19,13 @@ export async function retrieveUserById(
 ) {
   return await tx.user.findUnique({
     where: { id },
+    include: {
+      members: {
+        include: {
+          organization: true,
+        },
+      },
+    },
```
golddydev, line 28: Why do we need to include this?

web-app/src/lib/services/stripe/service.ts
golddydev, line 1: What about we add User to functions name (which is related to users compared to organization)

golddydev (changes requested):

web-app/src/lib/services/stripe/service.ts
mrgrauel: 48d51d2c48fee7da426abb02f7423a61523d215c

golddydev (approved):

PR #1063, opened 2025-07-10
fix(job-details): overflow text issue on job detail inputs

golddydev:
### Summary
This PR addresses text overflow issues in the job details inputs display by adding proper text wrapping classes to prevent content from breaking the layout.

### Changes Made
- Added `break-all` CSS class to both the key (field name) and value spans in the job inputs display
- This ensures long text content wraps properly within the grid layout instead of overflowing

### Technical Details
- **File Modified**: `web-app/src/app/app/agents/[agentId]/jobs/@right/components/job-details/inputs.tsx`
- **Lines 47-49**: Added `break-all` class to the field name span
- **Lines 50-54**: Added `break-all` class to the field value span

### Testing
- Verify that long field names and values display correctly without overflow
- Confirm the grid layout remains intact with various input lengths
- Test with both short and very long text content


mrgrauel:
bugbot run

PR #1168, opened 2025-07-25
refactor: check session on app layout before data fetching

golddydev:
## Summary

This PR checks user's session before doing any data fetching. Because `better-auth` library sometimes returns session as null even when session is valid. By doing so, users won't see any error (rather they will be redirected to `/login` page)

## Changes

* Add `auth/utils` function `getSessionOrRedirect` which will redirect to `/login` with current URL if `better-auth` returns session as null.
* Use `getSessionOrRedirect` on every `/app/:path*` layout.tsx and page.tsx (when they do data fetching which requires Auth)


golddydev:
> This doesn't work for the API.
> 
> Should we implement a redirect in the utility class, which may not be used in the page environment due to the API or a future MCP?

I just did this as a quick workaround (not permanent solution)

golddydev:
I added auth utils function (getSessionOrRedirect), and this function must be only used on layout.tsx and page.tsx.

PR #1210, opened 2025-07-31
refactor: simplify JobStatusData type for real-time job status updates

golddydev:
## Summary

This PR simplifies `JobStatusData` which is used for real time job status updates. And apply `externalDisputeUnlockTime` to hide JobStatusIndicator as Job Status is settled.

- Hide JobStatus indicator when `externalDisputeUnlockTime` is past.
- Use JobStatusData type in Sidebar component and JobStatusBadge component and use-job-status hook.
- Refactored related functions and components to support new data structure.


mrgrauel (approved):

mrgrauel (changes requested):
The update stops working after the first update.

golddydev:
> The update stops working after the first update.

Fixed it.

We connect client before we publish.
And close connection after we publish.

PR #1277, opened 2025-08-11
fix: show only one latest invitation per user

golddydev:
## Summary

This PR ensures we only show one pending invitation per user on members table.

When there are many pending invitations for one user, we only show the latest one (sort by `expiresAt`)

## Changes

* Remove unnecessary organization service function `getOrganizationPendingInvitations` and invitation repo function `getPendingInvitationsByOrganizationId`
* Use `better-auth` api to get all invitation and sort them by `expiresAt` and pick one per user's email.

### DIFF
--- a/web-app/src/lib/db/repositories/invitation.repository.ts
+++ b/web-app/src/lib/db/repositories/invitation.repository.ts
@@ -76,37 +76,6 @@ export const invitationRepository = {
   },
 
   /**
-   * Retrieves the latest pending invitation per user for a specific organization.
-   * This ensures only one invitation per email address is returned, even if multiple
-   * expired invitations exist for the same user.
-   *
-   * @param organizationId - The ID of the organization.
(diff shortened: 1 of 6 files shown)


web-app/src/lib/db/repositories/invitation.repository.ts
```diff
@@ -76,19 +76,34 @@ export const invitationRepository = {
...
+
+    // Group by email and take the first (latest) invitation per email
+    const emailMap = new Map<string, InvitationWithRelations>();
+    for (const invitation of allInvitations) {
+      if (!emailMap.has(invitation.email)) {
+        emailMap.set(invitation.email, invitation);
+      }
+    }
+
+    return Array.from(emailMap.values());
```
mrgrauel, line 106: Why don't we use https://www.better-auth.com/docs/plugins/organization#list-invitations for this? As I understand that does it.

mrgrauel (approved):

PR #1350, opened 2025-08-24
refactor: update agent card with author image and summary

golddydev:
TBD

### DIFF
--- a/web-app/src/components/agents/agent-summary.tsx
+++ b/web-app/src/components/agents/agent-summary.tsx
@@ -1,0 +1,35 @@
+"use client";
+
+import { useTranslations } from "next-intl";
+import { useState } from "react";
+
+import ClickBlocker from "@/components/click-blocker";
+
+interface AgentSummaryProps {
+  summary: string;
+}
+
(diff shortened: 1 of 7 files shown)


web-app/src/components/agents/agent-summary.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const canBeExpanded = summary.length > 100;
+
+  const handleToggleExpanded = () => {
+    setIsExpanded(!isExpanded);
+  };
+
+  return (
+    <div className="text-muted-foreground text-sm">
+      {isExpanded ? summary : summary.slice(0, 100)}
+      {canBeExpanded && (
```
francisluz, line 25: Could we just go to the main agent page? Because we're showing just a little more so, doesnt make much difference. We can leave the text, if required, but once clicked go to the agent page.

golddydev: Yeah, I'll make text 120 chars or less, so we can display it as whole.
And when clicked users will be redirected to detail page.

francisluz (approved):

PR #1578, opened 2025-09-29
fix(config): correct boolean environment variable parsing for mobile hydra dialog

mrgrauel:
## Summary

Fixes environment variable parsing for `NEXT_PUBLIC_ENABLE_MOBILE_HYDRA_DIALOG` to properly handle string boolean values using Zod's enum transformation pattern.

## Problem

The previous implementation using `z.coerce.boolean()` could incorrectly parse string values like "false" as `true`, since any non-empty string is truthy in JavaScript coercion.

## Solution

Updated the schema to use Zod's enum transformation pattern:
- Explicitly accepts only `"true"` or `"false"` string values
- Transforms the string to the appropriate boolean value
- Maintains the default value of `false`

## Technical Details

- **File changed**: `web-app/src/config/env.public.ts`
- **Pattern used**: `z.enum(["true", "false"]).transform(v => v === "true")`
- This ensures that only explicit "true" strings are parsed as `true`, while "false" and invalid values default to `false`

## Testing

- Environment variable parsing tested with:
  - `NEXT_PUBLIC_ENABLE_MOBILE_HYDRA_DIALOG="true"` → `true`
  - `NEXT_PUBLIC_ENABLE_MOBILE_HYDRA_DIALOG="false"` → `false`
  - Undefined → `false` (default)

## Impact

This ensures correct feature flag behavior in production environments where environment variables are always strings.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/web-app/src/config/env.public.ts
+++ b/web-app/src/config/env.public.ts
@@ -23,10 +23,7 @@ const envPublicConfigSchema = z.object({
   NEXT_PUBLIC_FEE_PERCENTAGE: z.coerce.number().min(0).default(5),
   NEXT_PUBLIC_CREDITS_BASE: z.coerce.number().default(12),
   NEXT_PUBLIC_AGENT_NEW_THRESHOLD_DAYS: z.coerce.number().min(0).default(7),
-  NEXT_PUBLIC_ENABLE_MOBILE_HYDRA_DIALOG: z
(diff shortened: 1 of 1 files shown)


francisluz (approved):
Tested and works as expected.

PR #1596, opened 2025-10-01
fix(jobs): refresh job detail page on scope change

mrgrauel:
## Summary

Fixes an issue where the job detail page doesn't properly refresh when the active organization scope changes. The page now validates that jobs belong to the current scope and redirects appropriately.

## Key Changes

- **Modified job access validation**: The `checkJobAccess` function now accepts the full `Session` object instead of just `userId`, enabling scope-aware access checks
- **Added scope validation**: Jobs owned by the user are now validated against the current active organization scope
- **Improved redirect behavior**: Changed from `notFound()` to `redirect()` when a user cannot access a job in the current scope, providing a better UX
- **Simplified session handling**: Removed redundant `userService` call and directly use session data for the active organization ID

### Before
- Job access was only checked against user ownership or organization sharing
- No validation that the job belonged to the current active organization scope
- Used `notFound()` for access denial, resulting in 404 errors

### After
- Job access now validates:
  1. User owns the job AND it belongs to the current scope (matching organizationId)
  2. OR job is shared with the user's active organization
- If a user owns a job but it's in a different scope, they are redirected to the jobs list
- Cleaner code with session data used directly from the `Session` object

## Testing

- ✅ Verify job detail page displays correctly for jobs in the current scope
- ✅ Confirm redirect to jobs list when accessing a job from a different scope
- ✅ Ensure shared jobs within the active organization remain accessible
- ✅ Check that switching organizations properly updates job access

### DIFF
--- a/web-app/src/app/(app)/agents/[agentId]/jobs/@right/[jobId]/page.tsx
+++ b/web-app/src/app/(app)/agents/[agentId]/jobs/@right/[jobId]/page.tsx
@@ -1,10 +1,10 @@
-import { notFound } from "next/navigation";
+import { notFound, redirect } from "next/navigation";
 
 import { JobDetails } from "@/components/jobs";
+import { Session } from "@/lib/auth/auth";
 import { getSession } from "@/lib/auth/utils";
 import { isSharedWithOrganization, JobWithStatus } from "@/lib/db";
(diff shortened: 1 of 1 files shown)


francisluz (approved):
Works as expected.

PR #1646, opened 2025-10-10
feat(schedules): allow users to schedule job execution

francisluz:
## Summary

Allow users to schedule job execution with both one-time and recurring (cron) options. Adds a first-class Schedules page for managing existing schedules, a schedule builder UI with presets and a custom cron builder, server actions for create/update/toggle/delete, a cron execution service, and a secure sync API protected by a Bearer token. Includes distributed locking to prevent double-processing and timezone-aware cron parsing with DST handling.

## Why/Benefits

- Enables hands-off job automation for users (one-time and recurring)
- Reliable execution with distributed locks and alignment checks
- Timezone-aware scheduling and presets for common use cases
- Clear lifecycle: create, edit, pause/resume, and end conditions (date or occurrences)
- Extensible architecture aligned with three-layer pattern (repositories → services → actions)

## Key Changes

- UI
  - New Schedules page: `web-app/src/app/(app)/schedules/page.tsx`
  - Schedule item client component: `web-app/src/components/schedules/schedule-item.client.tsx`
  - Formatting helpers + tests: `web-app/src/components/schedules/format.ts`, `__tests__/format.test.ts`
  - Integrations in job create modal: `web-app/src/components/create-job-modal/job-schedule-section.tsx`, `job-inputs-form.client.tsx`
  - Sidebar/navigation updates
  - Removed legacy route: `app/(app)/agents/[agentId]/schedules/page.tsx`

- Server/Domain
  - Server actions: `createSchedule`, `updateSchedule`, `toggleSchedule`, `deleteSchedule` in `web-app/src/lib/actions/job-schedule/action.ts`
  - Repository: `web-app/src/lib/db/repositories/job-schedule.repository.ts`
  - Cron compute (pure): `web-app/src/lib/services/job-schedule.cron.ts`
  - Execution service: `web-app/src/lib/services/job-schedule.service.ts`
  - Lock service: `web-app/src/lib/services/lock.service.ts`
  - Sync API: `GET /api/sync/job-schedules` (`web-app/src/app/api/sync/job-schedules/route.ts`) with Bearer auth

- Infrastructure
  - Vercel Cron updated (path added): `web-app/vercel.json`
  - Env secrets (defaults and new tuning): `web-app/src/config/env.secrets.ts`

## Technical Notes

- Cron parsing: timezone-aware via `cron-parser` with DST handling
- Next run derivation on server (for both create and update) using `computeNextRun`
- Alignment tolerance (env `JOB_SCHEDULE_ALIGNMENT_TOLERANCE_MS`, default 10 min) ensures we only run at expected cron ticks
- End conditions: stop scheduling after a given date (`endOnUtc`) or after N occurrences (`endAfterOccurrences`)
- Distributed lock: prevents overlapping runs (`lockService.acquireLock`), including automatic unlock of expired locks (env `LOCK_TIMEOUT` and buffer)
- Concurrency: capped via `p-limit(3)` inside execution service
- Security: sync endpoints require `Authorization: Bearer <CRON_SECRET>` via `authenticateCronSecret`

## Environment

Set these before running locally or in deployment:

- `CRON_SECRET` (required for sync endpoints)
- `LOCK_TIMEOUT`, `LOCK_TIMEOUT_BUFFER` (lock behavior)
- `JOB_SCHEDULE_ALIGNMENT_TOLERANCE_MS` (cron alignment)

All other existing variables still apply (Stripe, Ably, Better Auth, etc.). See `web-app/src/config/env.secrets.ts`.

### 0) Prerequisites

1. Node 22+, pnpm installed
2. Setup once at repo root:
   ```bash
   pnpm install
   (OPTIONAL) cp web-app/.env.example web-app/.env
   (OPTIONAL) # Add CRON_SECRET=your-local-secret to web-app/.env
   pnpm sokosumi-web:prisma:migrate:dev
   pnpm sokosumi-web:dev
   ```

### 1) Automated Tests

Run all web-app tests:
```bash
pnpm sokosumi-web:test:ci
```

Key coverage:
- `job-schedule.cron` unit tests (timezones, DST, invalid cron)
- Service smoke tests for `executeDueSchedules`
- Formatting helpers for schedule presentation

### 2) Manual UI QA

1. Sign in and open the Schedules page: `/schedules`
2. Verify navigation entry appears and page loads with empty state
3. Create schedules via Create Job modal:
   - One-time: choose date/time in the future; save
   - Recurring preset: daily/weekly/monthly; ensure preview times match selected timezone
   - Recurring custom: build cron using builder; ensure preview renders next few runs
4. Confirm new schedules show up on `/schedules` with:
   - Title, timezone, next run, last run, and status
   - Pause reason hidden by default (should appear only when paused)
5. Toggle Active/Inactive via switch; verify UI updates and persists after refresh
6. Edit schedule (pencil):
   - Change timezone, cron/preset, or end conditions; save
   - Verify `nextRunAt` updates according to changes
7. Delete schedule (trash) and confirm removal

Edge cases to verify: (DB Manipulation)
- Invalid cron (e.g., `invalid cron`) → schedule should pause with reason shown
- Missing timezone for cron → pauses with reason
- One-time date in the past should be rejected in UI validation

### 3) Sync API and Execution Flow

Ensure `CRON_SECRET` is set in `web-app/.env`, then restart dev server. Trigger sync manually:
```bash
curl -i -H "Authorization: Bearer $CRON_SECRET" \
  http://localhost:3000/api/sync/job-schedules
```

Expected:
- 200 JSON `{ message: "Syncing started" }`
- Server logs print processing metrics: `dueFound`, `processed`, `paused`, `durationMs`

Validation scenarios:
- Create a one-time schedule a couple minutes in the future; run sync shortly after scheduled time; verify:
  - `lastRunAt` updated
  - one-time schedule deactivates (no next run)
  - a new job appears on Jobs page
- Create a cron schedule with an `endOn` date in the near future or a small `endAfterOccurrences`; after syncing enough times, verify it stops scheduling
- Break a cron (e.g., edit to an invalid expression) and sync; verify it is paused with `pauseReason`

Locking/concurrency:
- Fire two concurrent GETs to `/api/sync/job-schedules`; the second should receive 409 while the first holds the lock

Alignment tolerance:
- Ensure the first cron run aligns within `JOB_SCHEDULE_ALIGNMENT_TOLERANCE_MS`; misaligned schedules will auto-correct `nextRunAt` instead of starting immediately

### 4) Timezone & DST Manual Check (Spot Check) (DB Manipulation)

1. Create a cron schedule in `America/New_York` and another in `Europe/London`
2. Use presets or custom times around DST boundaries (e.g., last Sunday in March for London)
3. Verify next-run preview matches expected UTC conversions

### 5) Regression Checks

- Jobs creation flow (non-scheduled) still works
- No Prisma calls from components; only via repositories/services/actions
- Dark/light UI styles still consistent; Tailwind semantics preserved

## Rollout & Monitoring

- Set `CRON_SECRET` in deployed environments and ensure the scheduler sending requests includes `Authorization: Bearer <CRON_SECRET>`
- Monitor logs for `Job schedules sync` metrics and Sentry tags `feature: job-schedule`

Closes: [DEV-83](https://linear.app/nmkr/issue/DEV-83/allow-users-to-schedule-job-execution)

### DIFF
--- a/web-app/src/lib/services/job-schedule.service.ts
+++ b/web-app/src/lib/services/job-schedule.service.ts
@@ -11,10 +11,8 @@ import { JobScheduleType } from "@/lib/db/types/job";
 import { startJobInputSchema, StartJobInputSchemaType } from "@/lib/schemas";
 import { jobService } from "@/lib/services/job.service";
 import { lockService } from "@/lib/services/lock.service";
-import { computeNextRun, ComputeNextRunInput } from "@/lib/utils/cron";
+import { computeNextRun } from "@/lib/utils/cron";
 import { JobSchedule } from "@/prisma/generated/client";
-
-export type { ComputeNextRunInput };
 
 export const jobScheduleService = {
   async executeDueSchedules() {


web-app/prisma/migrations/20251017100721_add_job_schedule_table/migration.sql
```diff
@@ -0,0 +1,43 @@
+-- Create Enum
```
mrgrauel: Based on the scheduleType wouldn't it make sense to add a constraint? If scheduleType is Cron the cron value has to be given and oneTime, the oneTimeAtUtc has to be set?

web-app/vercel.json
```diff
@@ -21,5 +21,10 @@
       "path": "/api/sync/source-import",
       "schedule": "*/5 * * * *"
     }
+    ,
+    {
+      "path": "/api/sync/job-schedules",
+      "schedule": "*/5 * * * *"
```
mrgrauel: If we only run the jobs every 5 minutes, why can the user choose a time for each minute?

web-app/src/lib/db/repositories/job-schedule.repository.ts
```diff
@@ -0,0 +1,158 @@
...
+      timezone: data.timezone,
+      maxAcceptedCents: data.maxAcceptedCents,
+      cron: data.cron,
+      oneTimeAtUtc: data.oneTimeAtUtc,
+      isActive: data.isActive,
+      pauseReason: data.pauseReason,
+      nextRunAt: data.nextRunAt,
+      lastRunAt: data.lastRunAt,
+    };
+    const maybeEndOnUtc = (data as unknown as { endOnUtc?: string | null })
```
mrgrauel: Why do you do these castings?

web-app/src/lib/db/repositories/job-schedule.repository.ts
```diff
@@ -0,0 +1,158 @@
...
+      where: { id },
+      data: {
+        nextRunAt,
+        isActive: nextRunAt ? true : false,
+        pauseReason: null,
+      },
+    });
+  },
+
+  async setPaused(
```
mrgrauel: I would recommend to call this `setActive`, because you can do both pausing and resuming.

mrgrauel (changes requested):
Some smaller questions arise, but it seems that some files contain vibed code that hasn't been reviewed.

web-app/src/lib/schedules/cron.ts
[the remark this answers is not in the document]
francisluz: Fix here 79400340894c581bbe334314003e74af995ccb63

web-app/src/lib/services/job-schedule.service.ts
```diff
@@ -0,0 +1,218 @@
...
+      due.map((schedule) =>
+        limiter(async () => {
+          const before = schedule.pauseReason;
+          await processSchedule(schedule, client);
+          const after = await jobScheduleRepository.getById(
+            schedule.id,
+            client,
+          );
+          processed += 1;
+          if (!before && after?.pauseReason) paused += 1;
```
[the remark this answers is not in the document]
mrgrauel: Sorry, but for me you didn't answer the question.

Why do you use the pauseReason to define is a job is paused instead of `isActive`? If we filter by active before, the value `before` should always be null, otherwise something has a pauseReason but has a true value for `isActive`

web-app/prisma/schema.prisma
```diff
@@ -719,6 +722,49 @@ model JobShare {
...
+  occurrenceCount  Int      @default(0)
+
+  // Job payload snapshot
+  inputSchema      Json
+  input            String
+  maxAcceptedCents BigInt
+
+  // State
+  isActive         Boolean  @default(true)
+  lastRunAt        DateTime?
```
mrgrauel: If I'm not mistaken we do not need this value. Do we? If we yes for what. I couldn't find a reason in the code for this information. Related to https://github.com/masumi-network/sokosumi/pull/1646/files#r2447569623

web-app/src/lib/services/job-schedule.service.ts
```diff
@@ -0,0 +1,218 @@
...
+  computeNextRun,
+  ComputeNextRunInput,
+} from "@/lib/services/job-schedule.cron";
+import { lockService } from "@/lib/services/lock.service";
+import { JobSchedule, Prisma } from "@/prisma/generated/client";
+
+export type { ComputeNextRunInput };
+
+export const jobScheduleService = {
+  computeNextRun,
```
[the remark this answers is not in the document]
mrgrauel: Ah, that's the issue. I see you implemented business logic in the job input form on the client, but this should be in the action on the server.

All calculations should be performed on the server; otherwise, the client can manipulate the information. Only the user input fields should be transferred to the action, and we should handle the business logic on the server. In the current implementation, we are doing part of the logic on the client and part on the server.

web-app/prisma/schema.prisma
```diff
@@ -719,6 +722,49 @@ model JobShare {
...
+  user             User     @relation(fields: [userId], references: [id], onDelete: Cascade)
+  organizationId   String?
+  organization     Organization? @relation(fields: [organizationId], references: [id], onDelete: SetNull)
+  agentId          String
+  agent            Agent    @relation(fields: [agentId], references: [id])
+
+  // Execution configuration
+  scheduleType     ScheduleType
+  cron             String?
+  oneTimeAtUtc     DateTime?
```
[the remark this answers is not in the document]
francisluz: Currently there's a logic attached to this field, if you select `one-time` job, we fill it with the UTC date. And if you select `cron` we fill the cron field. It has a use. And yes, it will have the same value as next run.

francisluz:
> We have currently disrupted the architecture of client action handling.

Refactor done, moving the code from client to serve actions. Here: bb4ba9c17b9a8897237f65eb0635b87a12010615

web-app/prisma/schema.prisma
mrgrauel: We now have the latest job attached, and the creation date of this job is `lastRunAt`, correct? We have two sources of truth, and we need to keep `lastRunAt` in sync. 🤔 

We should be able to use Prisma to query the data for `lastRunAt` from the jobs table without needing to save it in the database. If I'm not mistaken, Prisma provides calculation queries out of the box.

I highly recommend maintaining a single source of truth instead of adding another one, as this would create additional overhead by requiring us to ensure the field remains in sync with the latest job.

web-app/src/lib/services/job-schedule.service.ts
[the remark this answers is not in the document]
francisluz: Implemented here bb4ba9c17b9a8897237f65eb0635b87a12010615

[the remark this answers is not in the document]
mrgrauel: I don't get it, why does it has to be exposed in the service layer and via the utility? Isn't one enough?

web-app/vercel.json
mrgrauel: I changed it to every minute for now, cause the user can define on the minute, then the job should run on the minute 769eebc2b69b0f0430f7033378fe8b76a6c149d2

mrgrauel (changes requested):
Just some conventions broken, nothing big.

PR #1652, opened 2025-10-13
feat(auth): handle large social profile pictures via Vercel Blob storage

golddydev:
### Problem

Social authentication providers (Google, GitHub, etc.) sometimes return profile picture URLs longer than 512 characters. Since Better Auth stores user data including the profile picture URL in an encrypted JWT cookie, and cookies have a 4KB size limit, large profile URLs can cause authentication failures or exceed cookie size limits.

### Solution

This PR implements automatic profile picture optimization for social login:

1. **Detection**: Identifies profile pictures longer than 512 characters
2. **Deduplication**: Uses SHA-256 hashing to avoid re-uploading identical images
3. **Upload**: Uploads images to Vercel Blob storage (supports both data URIs and URLs)
4. **Storage**: Stores the shorter Blob URL in the user session
5. **Error Handling**: Includes timeout protection (10s default) and graceful fallbacks

#### Database
- **Migration**: Added `profilePictureHash` column to `user` table for image deduplication
- **Schema**: Updated Prisma User model with new optional field

#### Authentication
- Modified `mapProfileToUser` in `auth.ts` to handle large profile pictures
- Added support for data URI and URL-based profile images
- Implemented timeout protection using `p-timeout`
- Integrated Sentry error tracking for upload failures

#### Infrastructure
- Added `uploadAvatar` utility function for avatar-specific Blob uploads
- Added `findProfilePictureImageByHash` repository method for image deduplication
- New environment variables:
  - `VERCEL_AVATARS_UPLOAD_DIR` (default: "avatars")
  - `BETTER_AUTH_PROFILE_PICTURE_TIMEOUT` (default: 10000ms)

#### Dependencies
- Added `validator` (13.15.15) for URL validation
- Added `@types/validator` (13.15.3

### Notes

- Profile pictures are uploaded to the `avatars/` directory in Vercel Blob storage
- Images are deduplicated using SHA-256 hashing to save storage costs
- Timeout is set to 10 seconds to prevent auth flow delays
- Graceful fallback: if upload fails, user creation proceeds without profile picture
- All upload errors are captured in Sentry for monitoring

### DIFF
--- a/web-app/src/lib/auth/auth.ts
+++ b/web-app/src/lib/auth/auth.ts
@@ -14,7 +14,7 @@ import * as z from "zod";
 
 import { getEnvPublicConfig } from "@/config/env.public";
 import { getEnvSecrets } from "@/config/env.secrets";
-import { uploadAvatar } from "@/lib/blob";
+import { uploadImage } from "@/lib/blob/utils";
 import { prisma, userRepository } from "@/lib/db/repositories";
 import { reactChangeEmailVerificationEmail } from "@/lib/email/change-email";
 import { reactInviteUserEmail } from "@/lib/email/invitation";
@@ -423,7 +423,7 @@ async function mapProfileToUserInner(profile: {
     );
 
     // Upload the image to Vercel Blob Storage
-    const uploaded = await uploadAvatar(imageData, mimeType);
+    const uploaded = await uploadImage(imageData, mimeType);
     return {
       name: profile.name,
       image: uploaded.url,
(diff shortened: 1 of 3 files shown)


mrgrauel:
With 9beb8bcfb7823d13439a2ea8f5b4fbe5bc212cf9 it's refactored to pass urls along, but we should still change the name from `profilePictureHash` to `imageHash`.

web-app/src/lib/auth/auth.ts
```diff
@@ -336,16 +347,92 @@ export const auth = betterAuth({
...
+      imageHash: null,
+    };
+  }
+
+  // 1. Check if it's a valid URL (pass through directly)
+  if (
+    z
+      .url({
+        protocol: /^https?$/,
+      })
```
mrgrauel: Why does it have to be `https`? Isn't `http` also valid or another protocol?

PR #1769, opened 2025-10-31
chore(database): remove server-only dependency and imports

mrgrauel:
## Summary

Removes the `server-only` package dependency and all related imports from the `@sokosumi/database` package to enable compatibility with non-React backend projects (e.g., Bun/Hono services).

## Why This Change Is Needed

The `server-only` package is **specifically designed for React Server Components** in Next.js applications. It throws a runtime error when imported in client-side code, helping prevent accidentally bundling server-only code in the browser bundle.

However, the database package is now being shared between:
- **Next.js web app** (React Server Components) - where `server-only` was originally useful
- **Core Bun service** (Hono backend) - a standalone backend that doesn't use React

When importing the database package in the Bun service, `server-only` throws an error because it detects the module is being imported in a non-React/Next.js context, even though the code is legitimately server-side.

## Solution

Since:
1. The database repositories are **inherently server-side** code (they use Prisma, database connections, Node.js APIs)
2. They will **never accidentally run in a browser** (no bundler would include Prisma client code in client bundles)
3. The package needs to work in **multiple backend contexts** (Next.js and Bun)

We can safely remove `server-only` without losing protection against client-side imports. The repositories are already protected by their nature (database operations, Prisma client usage).

## Changes

- ✅ Removed `server-only` from `package.json` dependencies
- ✅ Removed `import "server-only";` from 21 repository files:
  - All 18 repository files in `src/repositories/`
  - `src/client.ts`
  - `src/transaction.ts`
- ✅ Updated `pnpm-lock.yaml`

## Testing

- ✅ Package builds successfully after changes
- ✅ All imports removed and verified
- ✅ Ready for use in both Next.js and Bun environments

## Related

This change enables the database package to be used in the new `apps/core` Bun service without module resolution errors.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -411,9 +411,6 @@ importers:
       pg-pool:
         specifier: 3.10.1
         version: 3.10.1(pg@8.16.3)
-      server-only:
-        specifier: 0.0.1
-        version: 0.0.1
       uuid:
         specifier: 13.0.0
         version: 13.0.0


francisluz (approved):
Tested locally and works as expected.

PR #2835, opened 2026-04-13
fix(tasks): require and forward identifierFromPurchaser for Masumi payments

mrgrauel:
## Summary

Task events that register a Masumi purchase previously generated `identifierFromPurchaser` on the server. That nonce must be the same one used when computing `inputHash` on the purchaser/agent side. Generating it server-side could desync verification from the on-chain purchase metadata.

## Key changes

- Require `identifierFromPurchaser` on the `masumiPayment` object in the task event request schema (documented in OpenAPI).
- Forward the client-supplied value to `createPurchaseFromMasumiTaskPayment` instead of deriving a UUID substring.
- Drop the unused `uuid` import from the route handler.
- Extend schema and route tests; assert the payment client receives `identifierFromPurchaser`.

## Testing

From `apps/core`:

```bash
pnpm exec vitest run "src/routes/v1/tasks/[id]/events/schema.test.ts" "src/routes/v1/tasks/[id]/events/post.test.ts"
```

## API note

Callers that send `masumiPayment` must include `identifierFromPurchaser` and must use the same value they used for Masumi input hashing.

### DIFF
--- a/apps/core/src/routes/v1/tasks/[id]/events/post.test.ts
+++ b/apps/core/src/routes/v1/tasks/[id]/events/post.test.ts
@@ -70,6 +70,7 @@ const COWORKER_ID = "cow_123";
 
 const validMasumiPaymentBody = {
   blockchainIdentifier: "0b00e04c0860a60c61066056281180462d0b12",
+  identifierFromPurchaser: "aabbccddeeff00112233",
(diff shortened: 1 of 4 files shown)


mrgrauel:
bugbot run

mrgrauel:
@codex review

PR #2883, opened 2026-04-22
feat(web): use AgentIcon in jobs filter dropdown

francisluz:
## Summary

Job filter options for agents now use the shared `AgentIcon` component so filter chips and the dropdown match agent branding used elsewhere in the app.

## Key changes

- Add optional `useAgentIcon` to `FilterDropdownMenuOption` in `filter-dropdown-menu.tsx`.
- When set, `FilterDropdownMenuOptionAvatar` renders `AgentIcon` (name + icon) inside a fixed-size wrapper instead of the generic `Avatar` image/fallback.
- Tasks jobs view agent filter options pass `useAgentIcon: true` when building dropdown options.

## Technical notes

- `AgentIcon` receives `agent: { name, icon }` where `icon` maps from the existing `image` option field (nullable).

## Testing

- [ ] Open **Tasks** (jobs view), open the filter menu, and expand the agent filter: each option should show the same agent icon treatment as other agent surfaces.
- [ ] Smoke: other filter dropdowns that do not set `useAgentIcon` should still use the previous avatar/fallback behavior.

### DIFF
--- a/apps/web/src/app/(app)/tasks/components/jobs-view-filters.tsx
+++ b/apps/web/src/app/(app)/tasks/components/jobs-view-filters.tsx
@@ -121,6 +121,7 @@ export function JobsViewFilters({
         label: agent.name,
         avatarLabel: agent.name,
         image: agent.image,
+        useAgentIcon: true,
       })),
     });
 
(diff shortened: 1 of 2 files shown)


francisluz:
@cursor review

PR #2889, opened 2026-04-23
feat(core): add extra credits summary to credits API

francisluz:
## Summary

Adds an **extra credits** breakdown to the credits payload and public credits endpoints so clients can show non-subscription credit pools separately from subscription allowance.

## Key changes

- Introduces `getExtraCreditSummary()` in `apps/core/src/helpers/subscription.ts`, aggregating active credit buckets that are **not** tied to `STRIPE_SUBSCRIPTION_PERIOD` (with null or non-subscription `referenceType`), scoped to user or organization, respecting `createdAt` / `expiresAt`, and reusing cent normalization so available credits never go negative when consumption exceeds the pool.
- Extends `CreditsPayload` / `buildCreditsPayload()` with `extra: { available, total }`.
- Updates user and organization credits GET routes OpenAPI examples and contract tests; extends `creditsResponseSchema` in `user.schema.ts` with required `extra`.
- Expands `subscription.test.ts` mocks to distinguish subscription vs extra aggregates and adds coverage for `getExtraCreditSummary` and updated `buildCreditsPayload` expectations.

## Technical notes

- **API shape**: `GET .../credits` responses now include `credits.extra.available` and `credits.extra.total` alongside existing `buffer`, `subscription`, and `total`. JSON clients that ignore unknown fields remain compatible; strict response validators should allow the new object.

## Testing

- `pnpm --filter @sokosumi/core test` (or the repo’s equivalent scoped test run) for `subscription.test.ts`, route contract tests, and `user.schema.test.ts`.


francisluz:
@cursor review

mrgrauel (changes requested):
The assumption that the buffer or extra credits have a total amount is insufficient. In 10 years, this total could reach 1.000.000.000.000 if people use the platform and purchase many credits.

The available buffer credits and the total should always match. Currently, we do not know from the API whether any of these buffer credits expire. We may need to add this information.

PR #2909, opened 2026-04-30
feat(core): attachment-aware titles and coworker user messages

francisluz:
## Summary

Improves how the core chat API derives conversation titles and validates the last message for coworker conversations when users send attachments without body text.

## Key changes

- **`buildMessageTitleSource`** (`message-content.ts`): After trimmed message text, walks persistable UI parts and uses the first file part’s filename, an image fallback label (`Image message`), URL path segment / hostname, or `File message` when the URL is invalid; otherwise uses non-blank text from other UI parts.
- **`hasModelVisibleMessageContent`**: Used in tests to document behavior for attachment-only vs whitespace-only user payloads.
- **Persistence / titles** (`post.ts`): First-user detection and async title generation use `titleSource` so attachment-only first turns still create a message row and get an OpenRouter-generated title from a sensible string (e.g. filename).
- **Coworker validation**: The last incoming user or system message is accepted for coworker flow when the user message has at least one file part or non-blank text part; **system** messages still require non-blank extracted text. Attachment-only **system** messages return **400** and do not start streaming (covered in `post.test.ts`).

## Technical notes

- Coworker branch uses `extractPersistableUiParts` for user-role checks and existing `lastUserMessageText` / `extractMessageText` for system-role checks, keeping system prompts text-only while allowing multimodal user turns.

## Testing

- `pnpm --filter core test src/helpers/__tests__/message-content.test.ts src/routes/v1/chat/post.test.ts` (45 tests passed locally).

## Risk

- Low: behavior is additive for normal text chats; coworker system messages with files only are now explicitly rejected where they were previously rejected for “empty text” in a similar way.

### DIFF
--- a/apps/core/src/helpers/__tests__/message-content.test.ts
+++ b/apps/core/src/helpers/__tests__/message-content.test.ts
@@ -2,10 +2,12 @@ import { describe, expect, it } from "vitest";
 
 import {
   buildConversationContentParts,
+  buildMessageTitleSource,
   extractMessageText,
   extractPersistableUiParts,
   extractReasoningPartsFromMessage,
   extractUiMessageParts,
+  hasModelVisibleMessageContent,
   readReasoningPartsFromMetadata,
 } from "../message-content";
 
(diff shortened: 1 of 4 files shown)


francisluz:
@cursor review

francisluz:
@cursor review

PR #2916, opened 2026-05-05
fix(web): allowlist Core chat proxy headers to prevent Undici failures

francisluz:
## Summary

Server-side chat routes forward the browser request to Core with `fetch()`. Copying all incoming `headers()` values can include hop-by-hop fields such as `transfer-encoding`, which Undici rejects on outbound requests and surfaces as `TypeError: fetch failed` and a 500 from `/api/chat`. This change builds an explicit allowlist for Core proxy requests instead of forwarding the full header bag.

## Key changes

- Add `buildCoreChatProxyHeaders()` to copy only `cookie`, `authorization`, and `x-organization-slug` from the incoming request.
- Use it in `GET`/`POST` `/api/chat` and the conversation stream route in place of mutating a cloned `Headers` instance.
- Extend route tests to assert `transfer-encoding` is not forwarded; add unit tests for the helper.

## Testing

- `pnpm --filter web test apps/web/src/lib/clients/utils/__tests__/build-core-chat-proxy-headers.test.ts`
- `pnpm --filter web test apps/web/src/app/api/chat/__tests__/route.test.ts`

## Risk / notes

- Any header required for Core that is not in the allowlist will no longer be forwarded. Current contract assumes session cookie, optional bearer auth, and organization slug only.

### DIFF
--- a/apps/web/src/app/api/chat/[conversationId]/stream/route.ts
+++ b/apps/web/src/app/api/chat/[conversationId]/stream/route.ts
@@ -3,6 +3,7 @@ import { headers } from "next/headers";
 import type { NextRequest } from "next/server";
 
 import { getSession } from "@/lib/auth/utils";
+import { buildCoreChatProxyHeaders } from "@/lib/clients/utils/build-core-chat-proxy-headers";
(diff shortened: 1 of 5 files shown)


francisluz:
@cursor review

PR #2917, opened 2026-05-05
fix(core): recover from resumable UI stream Redis ack timeout

francisluz:
## Summary

Handles the Redis pub/sub handshake timeout from `resumable-stream` when resuming an active UI chat stream. If `resumeExistingStream` rejects with `"Timeout waiting for ack"`, the handler now logs a warning, clears the stale `active_ui_stream_id` in conversation metadata (same path as when resume returns `null`), and returns **204 No Content** so the client can recover instead of surfacing a hard error.

## Key changes

- **`apps/core/src/routes/v1/chat/stream-get.ts`**: Wrap `resumeExistingStream` in try/catch; detect ack-timeout errors via message match; `clearActiveUiStreamIdInMetadata` + 204 on timeout; rethrow other errors.
- **`apps/core/src/routes/v1/chat/stream-get.test.ts`**: New test asserting 204 + metadata clear when resume rejects with the timeout error.

## Technical notes

The `resumable-stream` library rejects with `Error` and message `Timeout waiting for ack` after a short (~1s) wait for subscription ack on Redis; slow Redis or cross-instance latency can trigger this even when the stored stream id is no longer usable. Clearing metadata aligns behavior with the existing “resume returned null” stale-id path.

## Testing

- `pnpm --filter core test apps/core/src/routes/v1/chat/stream-get.test.ts`

## Risk

Low. Scope is limited to this route; non-timeout errors still propagate unchanged.

### DIFF
--- a/apps/core/src/routes/v1/chat/stream-get.test.ts
+++ b/apps/core/src/routes/v1/chat/stream-get.test.ts
@@ -120,6 +120,25 @@ describe("GET /chat/stream/:conversationId", () => {
     expect(resumeExistingStreamMock).not.toHaveBeenCalled();
   });
 
+  it("returns 204 and clears stale id when resume throws ack timeout", async () => {
+    conversationFindFirstMock.mockResolvedValueOnce({
+      id: cid,
+      metadata: { active_ui_stream_id: "stream_slow" },
+    });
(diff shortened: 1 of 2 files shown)


francisluz:
@cursor review

PR #2918, opened 2026-05-05
fix(chat-ui): resume multi-slot useChat only when active UI stream exists

francisluz:
## Summary

Multi-slot `useChat` instances previously set `resume: true` whenever a conversation was bound to a slot. That could trigger stream resume logic even when no server-side active UI stream existed. This change gates `resume` on presence of `active_ui_stream_id` in conversation metadata (same key as core).

## Key changes

- Add `readActiveUiStreamIdFromMetadata()` in `conversation-metadata.ts` to read and validate `active_ui_stream_id`.
- In `ChatInterface`, derive `resumeSlot0` / `resumeSlot1` / `resumeSlot2` from bound conversation metadata (selected conversation or list lookup) instead of `Boolean(slotBoundId)`.

## Technical notes

- Metadata is read as `Record<string, unknown>` consistent with existing helpers in the same module.
- `useMemo` dependencies include bound slot IDs, `conversations`, and selected conversation id/metadata so resume flags stay accurate when switching selection or when metadata updates.

## Testing

- Not run in this environment; recommend `pnpm --filter web test` and a quick manual check: open a chat with an in-flight stream vs idle bound conversation and confirm reconnect behavior matches expectations.

### DIFF
--- a/apps/web/src/app/(app)/chat-ui/components/chat-interface.tsx
+++ b/apps/web/src/app/(app)/chat-ui/components/chat-interface.tsx
@@ -53,6 +53,7 @@ import {
 } from "@/app/chat-ui/utils/chat-route-base";
 import {
   hasImageGenerationUiMessage,
+  readActiveUiStreamIdFromMetadata,
   readConversationImageGenerationFromMetadata,
 } from "@/app/chat-ui/utils/conversation-metadata";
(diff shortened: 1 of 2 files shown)


francisluz:
@cursor review

PR #3039, opened 2026-06-05
feat(history): expose archivedAt and render archived rows read-only

francisluz:
## Summary

- Expose `archivedAt` on history API items (tasks, jobs, conversations) so clients can distinguish navigable rows from archived ones
- Render archived tasks and conversations in the history feed as read-only rows without deep links
- Backfill existing archived tasks and conversations into the history feed via a data migration

### Core API
- Add nullable `archivedAt` to the history item schema and OpenAPI examples
- Map `archivedAt` from history feed rows in `mapHistoryRow`
- Extend unit tests for archived task and conversation mapping

### Web UI
- Introduce `isArchivedHistoryItem()` helper based on `archivedAt != null`
- Split list item into link vs. static `div` rendering; archived rows use `cursor-default` and skip hover/active link styles
- Regenerate core API client types/schemas
- Add tests covering linked non-archived rows and non-linked archived task/conversation rows

### Database
- Migration `20260605122200_backfill_archived_history`: upserts history feed entries for tasks and conversations that already have `archivedAt` set

## Technical details

Archived history items remain visible in the feed (including when filtering for archived content) but are no longer clickable, preventing navigation to deleted or inaccessible detail pages. Jobs are unaffected by the read-only UI path because they do not use `archivedAt` in this change set.

## Test plan

- [x] `pnpm --filter core test src/helpers/history.test.ts`
- [x] `pnpm --filter web test src/app/(app)/history/__tests__/history-list-item.test.tsx`
- [ ] Verify history page shows archived tasks/conversations without links
- [ ] Verify non-archived history rows still navigate correctly
- [ ] Apply migration in staging: `pnpm prisma:migrate:deploy`

Refs: DEV-939


francisluz:
@cursor review

PR #3056, opened 2026-06-08
feat(history): resolve agent name/icon in core history feed

mrosberghaus:
## Context

Part of the ongoing migration to move **all** direct database access out of `apps/web` (imports of `@sokosumi/database`) so the web app talks only to the core API (`apps/core`). The `history` domain had exactly one residual runtime DB query in web.

## What & why

`apps/web/.../history/utils/history-row-subtitle.server.ts` ran `prisma.agent.findMany` to resolve **agent name + icon** for job-row subtitles — the last direct Prisma access in the history domain. This moves that lookup into the core history handler and exposes the resolved values on the feed, so web reads them straight off each item.

User-facing impact: **identical rendering**, one fewer round-trip per `/history` load (agent data now arrives with the feed instead of a second query).

### Core
- `schemas/history.schema.ts` — `HistoryJobItem` gains additive `agentName` / `agentIcon` (nullable). Task/conversation items untouched.
- `helpers/history.ts` — new `loadAgentPreviewsByIds()` (the `agent.findMany` relocated from web) with a local icon-resolution guard matching the web client's previous behavior; `mapHistoryRow` sets the fields on JOB rows.
- `routes/v1/history/get.ts` — collects deduped job `agentId`s, loads previews in parallel with job-status, passes the map to `mapHistoryRow`. Soft-fail → `null` fields, never a 500.

### Web
- `history-row-subtitle.server.ts` — deleted the agent lookup + `@/lib/db/prisma`/agent-helper imports; now bucket-only.
- `history-row-subtitle.ts` — dropped `agentPreviewById` from `HistorySubtitleLookups`; job subtitle reads `item.agentName`.
- `history-list-item.tsx` — job icon from `item.agentName` / `item.agentIcon`.

### Generated client
- Regenerated via `pnpm generate:core:snapshot` (additive `agentName`/`agentIcon`; minor benign hey-api 0.98.2 drift).

## Verification
- `pnpm core:test` — 1025 pass (incl. 2 new: enrichment + soft-fail)
- `pnpm web:test` — 1299 pass, 1 skipped
- `pnpm check` (biome) — clean
- `pnpm build` + `pnpm core:build` (tsc) — success
- Gate: `git grep @/lib/db/prisma apps/web/src/app/(app)/history` → no matches

Manual: `/history` with job + task + conversation rows — subtitles/agent icons render as before; archived rows still read-only; no agent fetch when there are no jobs.

## Follow-up (not in this PR)

Tracked in [SOK-551](https://linear.app/masumi/issue/SOK-551).

The remaining `@sokosumi/database` coupling in history is **enum value** imports (`TaskStatus`, `SokosumiJobStatus`), not DB access. These are used across ~69 web files via shared status badges; decoupling them is a separate repo-wide task (move enums to a client-safe package or emit value enums in the generated client).


mrosberghaus:
bugbot run

PR #3058, opened 2026-06-08
refactor(agents): serve catalog reads from core

mrosberghaus:
## Context

Phase 1 of migrating the **agents** domain off direct DB access. (The user asked to migrate "credit-costs" next; investigation showed the credit-cost reads live inside `agent.service.ts`'s catalog/pricing methods — eliminating them *is* the agents catalog-read migration.) The public catalog page already reads from core; this extends that to the remaining `agentService` catalog callers.

## What & why

`agent.service.ts` catalog reads now source from the core API — which already filters to available agents (ONLINE + shown + valid pricing) and computes per-agent `credits` — via the **existing** `mapCoreAgent(s)ToAgentWithCreditsPrice` mappers, instead of reading `agentRepository` + `creditCostRepository` and recomputing pricing in web.

Because the mapper returns the same `AgentWithCreditsPrice` shape consumers already expect, **there is no UI re-typing** — only the data source changes.

### Changes
- `getAvailableAgents` / `getAvailableAgentsWithCreditsPrice` → `getAllCoreAgents()`
- `getAvailableAgentById` → `getCoreAgentById()` (404 → null, same availability gate as before)
- `getRandomAvailableAgentData` → core agents; `averageExecutionDuration` now comes from `metrics.executions.averageTime` (drops the extra `jobRepository` read)
- Deleted dead local pricing math: `getAvailableOnlineAgentsAndCreditCosts`, `buildCreditCostByUnitMap`, `computeAgentCreditsPriceCents`, `getAgentCreditsPrice`
- Reworked `agent.service.test.ts` to assert core wiring (the pricing math is now owned + tested in core)

### Parity
Verified: core `getAgents`/`getAgentsById` use `buildAvailableAgentWhereClause(creditCosts)` (identical to web's `isAgentAvailable`) and compute `credits` via `getAgentCost`; `getAgentsById` returns `notFound` (→ null) for unavailable agents. So which agents appear and their prices are unchanged.

## Scope / what remains
- The **favorites** path (`getAgentsByListType`) still reads `creditCostRepository.getCreditCosts` for local availability filtering — it has no core endpoint yet, so it's deferred to **Phase 3**. This PR removes the *catalog* credit-cost + `agentRepository` reads only.
- Breadcrumb agent lookup deferred (reads *all* agents inside a shared org `$tx`; separate concern, not a credit-cost read).
- Phases 2 (rating reads) & 3 (favorites/rating-writes/hired — net-new core endpoints) to follow.

## Verification
[...]

PR #3060, opened 2026-06-08
perf(agents): cache the global agent catalog across requests

mrosberghaus:
## Problem

`getAllCoreAgents` (`src/lib/agents/core-loaders.ts`) paginates the **entire** agent catalog over HTTP, and React `cache()` only dedupes within a single request. So **every** server render that needs agents — the catalog page, the task agent-picker (every task create/edit page + action), the billing "random agent" cards — re-fetched all catalog pages from Core on every request, each page `cache: "no-store"`. On Core, each `GET /v1/agents` page also recomputes per-agent rating aggregates + average execution times. This is a real, repeated cost on hot paths.

## Fix

The catalog (`GET /v1/agents`) is **global**: it carries no per-user fields and is not user/workspace-scoped (verified — the query uses `buildAvailableAgentWhereClause`, and the response is aggregate metrics/author/legal/categories only). Its payload is identical for every user, so it's safe to share across requests.

- Added an optional `cacheOptions` arg to the `getAgents` wrapper (`core.shared.ts`). When provided it switches that fetch from `cache: "no-store"` to `next: { revalidate, tags }`; **every other (user-scoped) Core call is untouched and stays `no-store`**.
- `getAllCoreAgents` opts in with a **60s TTL** + tag `core-agents-catalog`.

Next keys the fetch cache by **URL, not auth headers**, so the cached catalog is shared across users (correct here — global data; this is exactly why we must *not* do it for user-scoped endpoints). Net: the catalog is fetched at most once per TTL across all requests instead of on every render.

## Safety

- Only the catalog path opts in; user-scoped reads keep `no-store` (no cross-user leakage).
- Confirmed the catalog response has no per-user fields and no user/workspace filter.

## Tradeoff (needs a 👍 on the value)

Catalog changes — a newly onlined agent, a price change — become visible **up to the TTL (60s) late**, where today they're always fresh. 60s is conservative and tunable; the cache tag allows explicit invalidation via `revalidateTag(AGENTS_CACHE_TAG)` (e.g. after an agent sync) if you'd prefer event-based invalidation over / in addition to the TTL.

## Notes

- Documented the rule in `apps/web/AGENTS.md` → **Core API reads & caching**: never cross-request-cache user-scoped Core responses; do cache the global catalog; avoid new per-request all-pages loops.
- The proper long-term fix is a Core endpoint that returns the filtered catalog in one call (existing `TODO(core-api)` in `core-loaders.ts`); this is the safe interim win.
- Independent of the agents-domain migration (PR #3058); composes with it (its new task/billing callers get the same caching for free).

## Verification
- `pnpm --filter web test src/lib/clients src/lib/agents "src/app/(app)/agents"` — 71 pass
- `pnpm check` (biome) — clean
- `pnpm web:build` (tsc) — clean

### DIFF
--- a/apps/web/AGENTS.md
+++ b/apps/web/AGENTS.md
@@ -173,6 +173,30 @@ import { JobsList } from "src/app/(app)/agents/[agentId]/jobs/components/jobs-li
 - Never access Prisma directly from components
 - Use server actions for mutations
 
+### Core API reads & caching (performance)
+
+We are migrating web data access from direct DB reads to the Core API
+(`coreClient` in `src/lib/clients/core.client.ts`). Two performance rules:
+
+- **Never blindly cache Core responses across requests.** Every `coreClient`
+  call attaches the caller's cookies (`buildAuthHeaders`) and defaults to
+  `cache: "no-store"`, because most responses are **user/workspace-scoped** —
+  caching them in Next's shared fetch cache would leak one user's data to
+  another. Keep `no-store` for anything user-specific.
+- **Do cross-request cache _global_ catalog reads.** The agent catalog
+  (`GET /v1/agents`) is global — it carries no per-user fields and is not
+  user-scoped — so it is safe to share. `getAllCoreAgents`
+  (`src/lib/agents/core-loaders.ts`) loads every page of the catalog; without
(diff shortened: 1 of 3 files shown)


mrosberghaus:
bugbot run

PR #3061, opened 2026-06-09
fix(sync): reduce per-user Sentry noise for transient Hermes orchestrator errors

mrosberghaus:
## Summary

**Sentry issues addressed**
- [SOKOSUMI-CORE-1M](https://masumi.sentry.io/issues/SOKOSUMI-CORE-1M) — `Error: Hermes orchestrator error (502)` — **78 events / 56 users** (ongoing since May 20)
- [SOKOSUMI-CORE-1J](https://masumi.sentry.io/issues/SOKOSUMI-CORE-1J) — `TypeError: fetch failed` (ConnectTimeout to orchestrator) — **18 events / 11 users** (ongoing since May 14)

**Root cause**

Both issues fire from `pollOne()` inside `hermesInboxSyncService.pollInboxes()`, which is invoked by the `GET /sync/hermes/poll-inboxes` Vercel cron (every minute). When the Hermes orchestrator service at `orchestrator-production-35d4.up.railway.app` returns HTTP 502 or is unreachable, the existing code called `Sentry.captureException()` once per affected user. With 56+ users in the batch, a single orchestrator outage window produced hundreds of near-identical Sentry events rather than one actionable alert.

Both failure modes are **expected transient external-service blips**: the error counter (`HermesInstance.consecutivePollErrors`) is already incremented correctly and the next poll cycle recovers automatically. No user data is lost.

**Fix**

- Added `isTransientOrchestratorError(error)` — returns `true` for `HermesOrchestratorError` with HTTP status ≥ 500 and for undici network failures (`TypeError("fetch failed")` with a cause), which are the two patterns observed in production.
- `pollOne()` no longer calls `Sentry.captureException` per user for transient errors; it instead sets `PollOutcome.transientError` on the returned outcome.
- `pollInboxes()` accumulates these and emits **one** `Sentry.captureException` (tagged `hermes_inbox_poll_batch`) per cron invocation when any transient errors occurred — preserving actionable signal without the per-user flood.
- Non-transient errors (4xx orchestrator responses, message-persist failures, ack failures) continue to fire per-user events as before.

**Not addressed (separate follow-up needed)**

- [SOKOSUMI-CORE-1X](https://masumi.sentry.io/issues/SOKOSUMI-CORE-1X) — `Error: Connection terminated unexpectedly` in Prisma `$transaction` on `GET /sync/jobs` (3 events / 1 user). Root cause is stale serverless connection pool; requires infrastructure-level mitigation (pgBouncer / Prisma Accelerate). Low priority given frequency.

## Test plan

- [x] `pnpm --filter core test src/services/__tests__/hermes-inbox-sync.service.test.ts` — 7/7 pass
- [x] `pnpm check` — no Biome errors
- [ ] Confirm Sentry issue SOKOSUMI-CORE-1M stops generating new per-user events after deploy; one `hermes_inbox_poll_batch` event per outage window is expected instead

---
_Generated by [Claude Code](https://claude.ai/code/session_01QrMLuvakHPZhBBJ5MGokLz)_

### DIFF
--- a/apps/core/src/services/__tests__/hermes-inbox-sync.service.test.ts
+++ b/apps/core/src/services/__tests__/hermes-inbox-sync.service.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from "vitest";
+
+import { HermesOrchestratorError } from "@/clients/hermes-orchestrator.client";
+import { isTransientOrchestratorError } from "@/services/hermes-inbox-sync.service";
+
+describe("isTransientOrchestratorError", () => {
+  it("treats HermesOrchestratorError with 5xx status as transient", () => {
+    expect(
+      isTransientOrchestratorError(new HermesOrchestratorError(500, {})),
+    ).toBe(true);
+    expect(
+      isTransientOrchestratorError(
+        new HermesOrchestratorError(502, { code: "BAD_GATEWAY" }),
+      ),
+    ).toBe(true);
+    expect(
+      isTransientOrchestratorError(new HermesOrchestratorError(503, {})),
+    ).toBe(true);
+  });
+
+  it("treats HermesOrchestratorError with 4xx status as non-transient", () => {
+    expect(
+      isTransientOrchestratorError(new HermesOrchestratorError(400, {})),
+    ).toBe(false);
+    expect(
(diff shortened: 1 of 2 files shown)


mrosberghaus:
bugbot run

PR #3064, opened 2026-06-09
fix(web): guard null enterprise in credits response transformer

mrosberghaus:
## Summary

Fixes Sentry issue **SOKOSUMI-PC**: `CoreApiRequestError: Cannot read properties of null (reading 'buckets')`, thrown during the job-start preflight credit check (`startJob` → `resolveAvailableCredits` → `coreClient.getMyCredits()`).

## Root cause

The Core API credits endpoint (`buildCreditsPayload`, `apps/core/src/helpers/subscription.ts`) returns `extra.enterprise: null` for users **without** enterprise credit buckets, and the Core Zod schema correctly marks it `.nullable()` (`apps/core/src/schemas/user.schema.ts`).

The generated web client was stale, however:
- `types.gen.ts` declared `enterprise: CreditsResponseEnterprise` (non-null).
- `transformers.gen.ts` unconditionally called `creditsResponseEnterpriseSchemaResponseTransformer(data.enterprise)`, which runs `data.buckets.map(...)` → crashes on `null.buckets`.

Every job start by a non-enterprise user therefore threw. The web action catches it and degrades credits to `null` (users were not hard-blocked), but it reported to Sentry on each occurrence.

## Changes

- `transformers.gen.ts` — guard the enterprise transform with `if (data.enterprise)`, matching the existing pattern used for nullable `subscription`.
- `types.gen.ts` — `enterprise: CreditsResponseEnterprise | null` to match the Core schema (the runtime JSON schema in `schemas.gen.ts` was already `type: ['object','null']`).
- Added regression test `transformers.credits.test.ts` covering both the null and present-enterprise cases.

## Note

These are hand-edits to generated client files — exactly what regeneration against the current Core OpenAPI spec should produce. Flagging so a future `openapi-ts` run isn't surprising.

## Verification

- `pnpm --filter web test src/lib/clients/__tests__/transformers.credits.test.ts src/lib/clients/__tests__/transformers.history.test.ts` — passing (the null case reproduced the crash before the fix).
- `pnpm biome check` — clean.

https://claude.ai/code/session_01492QV7mf3StWSHhL7DQyVv

---
_Generated by [Claude Code](https://claude.ai/code/session_01492QV7mf3StWSHhL7DQyVv)_


mrosberghaus:
bugbot run

PR #3066, opened 2026-06-09
fix(sync): only alert on Hermes orchestrator outages, not isolated 502s

mrosberghaus:
## What

Fixes the noise behind Sentry issue **SOKOSUMI-CORE-1M** (`Error: Hermes orchestrator error (502)`, surfacing via the per-minute cron `GET /sync/hermes/poll-inboxes`).

## Why

The previous fix (#3061) collapsed per-user reports into one aggregate event per batch, but the batch poller still called `Sentry.captureException(firstTransientError, …)` for **any** batch with ≥1 transient error, passing the raw `HermesOrchestratorError`. Because the cron runs every minute and Sentry groups by exception type + stack, a single self-healing 502 among healthy polls (the latest event showed `transientErrorCount: 1, polled: 9`) kept re-landing in the SOKOSUMI-CORE-1M group and held the issue perpetually **regressed**.

## Change

In `apps/core/src/services/hermes-inbox-sync.service.ts`:

- Added thresholds `OUTAGE_MIN_POLLED_FOR_ALERT = 5` and `OUTAGE_TRANSIENT_RATIO = 0.5`.
- Escalate to Sentry **only** when transient failures dominate a non-trivial batch (≥5 polled **and** ≥50% transient) — the signature of an actual orchestrator outage. Isolated blips are absorbed silently (they recover next cycle and already bump the per-instance `consecutivePollErrors` counter).
- Switched from `captureException(rawError)` to `captureMessage("hermes_inbox_orchestrator_outage", …)` with a dedicated `fingerprint`, so outages form their own clean issue instead of reviving the per-request 502 group. The representative error message rides along in `extra.sampleError`.

## User-facing impact

No behavior change for end users. Operationally: Sentry stops firing on isolated, self-healing orchestrator 502s and only pages on genuine outage windows under a distinct, clearly-named issue.

## Verification

- `pnpm --filter @sokosumi/core test` for the two inbox-sync suites — **15 passed** (added coverage: escalates when transient failures dominate; stays silent on an isolated failure).
- `tsc --noEmit` clean.
- Biome check clean.

## Follow-up

The existing SOKOSUMI-CORE-1M issue will likely need to be **manually marked resolved** in Sentry once this deploys — the new code intentionally no longer reports into that group, so no `Fixes SOKOSUMI-CORE-1M` trailer was used (that would mis-link an unrelated commit reference).

https://claude.ai/code/session_019jhBHtFHaMuHAqCAf79kVY

---
_Generated by [Claude Code](https://claude.ai/code/session_019jhBHtFHaMuHAqCAf79kVY)_

### DIFF
--- a/apps/core/src/services/hermes-inbox-sync.service.test.ts
+++ b/apps/core/src/services/hermes-inbox-sync.service.test.ts
@@ -25,6 +25,7 @@ const HERMES_INBOX_MESSAGE_UUID_NAMESPACE =
 
 vi.mock("@sentry/node", () => ({
   captureException: vi.fn(),
+  captureMessage: vi.fn(),
 }));
 
 vi.mock("@/config/env", () => ({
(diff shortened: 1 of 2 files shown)


mrosberghaus:
bugbot run

PR #3067, opened 2026-06-09
fix(web): make relative time hydration-safe in members table and ratings

mrosberghaus:
## Summary

Fixes the React hydration error reported in Sentry **SOKOSUMI-A** on `/organizations/[organizationSlug]` (1,961 occurrences across Chrome/Firefox/Safari).

### Root cause

The "Last seen" column of the members table rendered relative time with `formatDistanceToNowStrict(lastSeenAt, { addSuffix: true })`. Although the component is `"use client"`, Next.js still server-renders it for the initial HTML. The relative string ("5 seconds ago") is computed against `Date.now()` at **SSR time** and recomputed at **hydration time** on the client. Because real time elapses between those two moments (network + JS load), the strings differ for recently-seen rows → text-content hydration mismatch. The same latent pattern existed in the agent `rating-list-item.tsx`.

### Fix

- Add a reusable client-mounted **`TimeAgo`** component that renders a deterministic, UTC-pinned absolute date (via the existing `formatShortDateTime` / `HYDRATION_STABLE_TIME_ZONE` convention) during SSR and the first client render, then swaps to the live relative string after mount — a client-only update that cannot mismatch the server output. Guards invalid dates and emits a semantic `<time dateTime title>` element.
- Use it in the members table last-seen column (the SOKOSUMI-A culprit) and the rating list item (shared latent bug).

## User-facing impact

No visible change in steady state. Briefly, a stable absolute date is shown on first paint before it swaps to the relative "time ago" string after mount — eliminating the hydration error and the resulting console noise/replay error.

## Verification

- `pnpm --filter web exec biome check` — clean on all changed files
- `tsc --noEmit` — clean
- New `time-ago.test.tsx` — 3/3 pass (SSR-stable output, post-mount relative swap, invalid-date handling)
- Existing members-table + agents tests — 24 passed / 1 skipped, no regressions

Fixes SOKOSUMI-A

https://claude.ai/code/session_01DzmswBf5sGjtmowch9LpPa

---
_Generated by [Claude Code](https://claude.ai/code/session_01DzmswBf5sGjtmowch9LpPa)_

### DIFF
--- a/apps/web/src/components/__tests__/time-ago.test.tsx
+++ b/apps/web/src/components/__tests__/time-ago.test.tsx
@@ -1,0 +1,35 @@
+import { render, screen } from "@testing-library/react";
+import { renderToStaticMarkup } from "react-dom/server";
+import { describe, expect, it } from "vitest";
+
+import { TimeAgo } from "@/components/time-ago";
+
+describe("TimeAgo", () => {
+  it("renders a stable, UTC-pinned absolute date on the server (no effects)", () => {
+    // The server render (and the matching first client render) must emit a
+    // deterministic absolute date rather than a clock-dependent relative
+    // string, otherwise SSR and hydration diverge (Sentry SOKOSUMI-A).
+    const date = new Date("2026-04-15T10:00:00.000Z");
+
+    const markup = renderToStaticMarkup(<TimeAgo date={date} strict />);
+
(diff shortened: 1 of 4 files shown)


mrosberghaus:
bugbot run

PR #3070, opened 2026-06-09
feat(core): record UTM attribution via core API

mrosberghaus:
## What

Migrate UTM attribution **writes** from the web app's direct Prisma access to a new Core API endpoint, keeping `apps/web` off the database layer (continues the web→core migration).

## Changes

**Core — new endpoint**
- `POST /v1/users/{id}/utm-attribution` — actor-authed via the existing user-path-access middleware; `me` resolves the session user. Writes through `utmAttributionRepository.createUTMAttribution`.
- `utmAttributionRequestSchema` / `utmAttributionResponseSchema` reusing the shared `dateTimeSchema`.

**Web — consumer**
- `utm.service` drops `@sokosumi/database` + `prisma`; calls Core via the generated SDK. Cookie read/parse/delete stay web-side.
- Regenerated Core client snapshot.

**Repo — idempotency**
- `createUTMAttribution` now `upsert`s on the unique `userId`, so a retried conversion refreshes instead of throwing a unique-constraint error (previously surfaced as a swallowed 500).

## Non-obvious decision

The shared `coreClient` builds its auth cookie from incoming `headers()`. At **sign-up** the fresh Better Auth session cookie is only on the `cookies()` store for that request, not yet on the incoming headers — routing UTM through the shared client would have **silently dropped attribution** (errors are swallowed). Fix is scoped to `utm.service` only: a one-off Core client whose cookie header comes from `cookies().toString()`. Shared client untouched → zero blast radius.

## User-facing impact

None visible. UTM conversions recorded at sign-up now flow through Core instead of a direct DB write. Behavior preserved (fire-and-forget, cookie cleared regardless of outcome).

## Verification

- `pnpm --filter core typecheck` · `pnpm --filter web typecheck` · `pnpm --filter @sokosumi/database typecheck` ✅
- Core route tests: 200 / 400 (missing source) / 401 (unauth) / 403 (coworker) / 404 (unknown user) / 500 (repo null)
- Web service tests: cookie present / absent / malformed / error-swallow
- Repo tests: upsert where-key, field mapping + `capturedAt` Date coercion (create & update branches), null passthrough
- Biome clean

## Schema changes

No Prisma migration — `createUTMAttribution` switched `create`→`upsert` on the existing unique `userId`; columns unchanged.


mrosberghaus:
bugbot run

PR #3073, opened 2026-06-09
fix(web): make task page relative time hydration-safe

mrosberghaus:
## What

Fixes the React hydration error on `/tasks/[taskId]` (Sentry **SOKOSUMI-A**, 1965 occurrences, regressed).

`formatTimeAgo` derives its output from the current clock, which advances between the SSR render and client hydration — rendering it during both passes yields mismatched text and triggers a `replay_hydration_error`. PR #3067 introduced the hydration-safe `<TimeAgo>` (UTC-pinned absolute date on SSR + first paint, then swap to the live relative string after mount) and migrated the members table and ratings, but the task surfaces were still rendering relative time directly.

This migrates the remaining sites, all on the task detail page:

- `task-activity.tsx` (client) — activity event timestamps
- `job-list-item.tsx` (client) — job timestamps
- `projects/components/project-tasks-section.tsx` (client) — task timestamps
- `task-jobs.tsx` (server) — job list timestamps (passes `locale` to `<TimeAgo>` for the absolute fallback)

`formatTimeAgo` / `useLocalizedDateTime` remain for non-render uses (sorting, grouping).

## User-facing impact

No more hydration error / client re-render flash on the task detail page. Relative time still renders, now via the mount-swap path.

## Tests

Two tests swap the now-unused `@/lib/utils/datetime` mock for a `@/components/time-ago` mock emitting the same `ago:<iso>` string; assertions unchanged.

## Verification

- `vitest run` task-jobs / jobs-list-view / task-activity → 22 passed
- `biome check` on changed files → clean
- `tsc --noEmit` → exit 0

Fixes SOKOSUMI-A


mrosberghaus:
bugbot run

mrosberghaus:
bugbot run

PR #3078, opened 2026-06-10
feat(core): add admin user/org search endpoints and route web through them

mrosberghaus:
## Summary

First step of the **web → core database-access migration**. Moves the two web admin search services off direct `@sokosumi/database` access onto new admin-guarded core API endpoints, so the shared repositories are reached through a single source of truth (core) rather than from the web app directly.

### Core — new `/v1/admin` namespace
Guarded by the existing `requireAdminAuthContext` (same pattern as `/v1/enterprise`), reusing the shared repositories:
- `GET /v1/admin/users?query=` — search users (`id`, `name`, `email`)
- `GET /v1/admin/organizations?query=` — search organizations (`id`, `name`, `slug`)
- `GET /v1/admin/organizations/{slug}` — single org limited-info; **404** when absent

### Web
- `admin-user.service` / `admin-organization.service` now call `coreClient` wrappers (added in `core.shared.ts`) instead of importing the database package — both services are now Prisma-free.
- The slug wrapper maps core **404 → `null`**, preserving prior semantics.
- Public service interfaces (`AdminUserOption`, `AdminOrganizationOption`) are unchanged → actions/components untouched.
- Web actions keep `assertAdminSession` (defense-in-depth); core independently enforces admin via the forwarded session cookie.
- Regenerated the core OpenAPI client snapshot (`generated/core/*`).

## User-facing impact
None. Admin user/org search behaves identically (same 20-result limit, same shapes, same slug-not-found → null). The data path moves from web-Prisma to the core API.

## Verification
- `pnpm --filter @sokosumi/core typecheck` ✅ · `pnpm --filter web typecheck` ✅
- Core admin tests: **12/12** (`admin.routes.test.ts` 7 + `admin.integration.test.ts` 5 — real-mount: non-admin → 403 on both sub-routers, no session → 401, admin → 200)
- Web admin tests: **9/9** (action 4 + org service 4 incl. 404→null/rethrow + user service 1)

## Notes
- Reviewed via `requesting-code-review` (verdict: minor fixes); both recommended tests added.
- Generated client files (`apps/web/src/lib/clients/generated/core/*`) are auto-generated — do not hand-edit.


mrosberghaus:
bugbot run

PR #3081, opened 2026-06-10
feat(core): add agent demo-job endpoint and route web through it

mrosberghaus:
## Summary

Continues the **web → core database-access migration**. Moves the demo-job flow off direct `@sokosumi/database` access in the web app onto a new core endpoint, and removes the **duplicated** web source-import service (core already had an identical one).

The picked backlog item was "source-import writes", but those writes turned out to be a sub-step of the web `startDemoJob` flow (`jobRepository.createDemoJob` + workspace upsert + source-import enqueue), so the whole flow was migrated as one coherent unit.

### Core
- `POST /v1/agents/{id}/demo-jobs` — org-member + workspace scoped (`requireUserContext` + `requireWorkspaceContext`, mirrors `agents/{id}/jobs`). Creates a demo job from a pre-computed agent result and enqueues any file/link sources, reusing core's existing `sourceImportService`. Returns the serialized job; 404 when the agent is unavailable.
- `createDemoJobForUser` in `helpers/job.ts` (agent-availability check → `createDemoJob` repo → best-effort source enqueue) + `createDemoJobRequestSchema`.

### Web
- `jobService.startDemoJob` now calls `coreClient.createDemoJob` (converting input via the existing `toCoreJobInputData`); it no longer upserts the workspace, writes the job, or enqueues sources.
- **Deleted** the duplicated `apps/web/src/lib/services/source-import.service.ts` (+ test). Core is now the single source of truth for source-import enqueueing.
- Regenerated the core OpenAPI client snapshot.

## User-facing impact
None. The demo-run flow behaves the same; the job creation + source ingestion now happen server-side in core.

## Verification
- `pnpm --filter @sokosumi/core typecheck` ✅ · `pnpm --filter web typecheck` ✅
- Core: helper unit tests 4/4, route test 2/2 (201 + agent-not-found 404), v1 index green
- Web: job.service 3/3 (routes through core, asserts no direct DB writes), job action + use-job-submission hook 27/27

## Notes
- Generated client files (`apps/web/src/lib/clients/generated/core/*`) are auto-generated — do not hand-edit.


mrosberghaus:
bugbot run

mrosberghaus:
Re: the Cursor Security Reviewer SSRF finding on `apps/core/src/helpers/job.ts`.

The finding is **real but pre-existing — not introduced by this PR.** This PR only relocates the demo-job flow web→core verbatim. On `main`, the old web `jobService.startDemoJob` already did the identical thing with the same client-supplied `result`:

```ts
result: jobStatusResponse.result,
...
await sourceImportService.enqueueFromMarkdown(eventWithResult.id, eventWithResult.result);
```

So there is **no security regression** here — behavior is byte-for-byte preserved.

The actual vulnerable sink is the source-import worker's `fetch(blob.sourceUrl, { redirect: "follow" })` in `apps/core/src/services/source-import-sync.service.ts`, which has no private/loopback/link-local/metadata IP guard and uploads the result to a public blob store. It is fed not only by demo `result` but also by untrusted external agent output (`job-sync.service.ts`), so it is a systemic issue best fixed once at the shared sink rather than in this migration.

Fix is up as a dedicated PR: #3082 (`fix(core): guard source-import blob fetch against SSRF`). Keeping this PR scoped to the migration.

PR #3082, opened 2026-06-10
fix(core): guard source-import blob fetch against SSRF

mrosberghaus:
## What

Hardens the source-import blob worker against **SSRF**. The worker fetched URLs extracted from job result markdown with `redirect: "follow"` and no address validation, then uploaded the response to a **public** blob store.

Because those URLs derive from **untrusted input**:
- client-supplied demo `result` — `apps/core/src/helpers/job.ts:309`
- external agent output — `apps/core/src/services/job-sync.service.ts:392`

…an authenticated attacker (or a malicious/compromised agent) could make the backend fetch internal/private hosts or the cloud metadata endpoint (`169.254.169.254`) and have the content exfiltrated via normal job file retrieval.

Surfaced by Cursor Security Reviewer on #3081 (HIGH). The finding is **pre-existing** — #3081 only relocated identical web behavior into core verbatim. This PR fixes it at the shared sink, covering both demo jobs and regular agent jobs.

## How

New `apps/core/src/lib/url-guard.ts`:
- `assertPublicHttpUrl(url)` — enforces `http(s)`, resolves the host via DNS, and rejects if **any** resolved address is loopback / RFC1918 private / link-local (incl. `169.254.169.254`) / CGNAT / `0.0.0.0/8` / multicast-reserved, plus IPv6 `::1`/`::`/`fe80::/10`/`fc00::/7` and IPv4-mapped equivalents.
- `ssrfSafeFetch(url, init)` — follows redirects manually (max 5) and **re-validates every hop**, defeating redirect- and DNS-rebind-to-internal.

Wired into `importBlob` (`source-import-sync.service.ts`) in place of the raw `fetch(..., { redirect: "follow" })`.

Scope kept tight: no response size cap added (would risk rejecting legitimate large agent outputs); behavior is unchanged for public URLs.

## Verification

- `pnpm --filter @sokosumi/core typecheck` — clean
- `pnpm --filter @sokosumi/core test src/lib/url-guard.test.ts src/services/source-import-sync.service.test.ts` — 43 passing (new guard unit tests: private/loopback/link-local/metadata/IPv6-ULA/IPv4-mapped blocked, public allowed, bad scheme, redirect-to-internal blocked, redirect limit)
- `biome check` — clean

Closes SOK-557 — https://linear.app/masumi/issue/SOK-557/harden-source-import-blob-fetch-against-ssrf

### DIFF
--- a/apps/core/package.json
+++ b/apps/core/package.json
@@ -63,6 +63,7 @@
     "neverthrow": "8.2.0",
     "p-limit": "7.3.0",
     "postmark": "4.0.7",
+    "request-filtering-agent": "3.2.0",
     "resumable-stream": "2.2.12",
     "slugify": "1.6.9",
     "stripe": "22.2.0",
--- a/apps/core/src/lib/url-guard.integration.test.ts
+++ b/apps/core/src/lib/url-guard.integration.test.ts
@@ -1,0 +1,16 @@
+import { describe, expect, it } from "vitest";
+
+import { ssrfSafeFetch } from "./url-guard";
+
+// These exercise the real `request-filtering-agent`: the connection must be
+// refused at connect time for non-public targets, with no network egress.
+describe("ssrfSafeFetch (integration: address filtering)", () => {
+  it.each([
+    "http://127.0.0.1/",
+    "http://10.0.0.1/",
+    "http://169.254.169.254/latest/meta-data/", // cloud metadata endpoint
+    "http://[::1]/",
+  ])("refuses to connect to non-public target %s", async (url) => {
+    await expect(ssrfSafeFetch(url)).rejects.toThrow();
+  });
+});
--- a/apps/core/src/lib/url-guard.test.ts
+++ b/apps/core/src/lib/url-guard.test.ts
@@ -1,48 +1,24 @@
 import { EventEmitter } from "node:events";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
-const { lookupMock, httpsRequestMock, httpRequestMock } = vi.hoisted(() => ({
-  lookupMock: vi.fn(),
+const { httpsRequestMock, httpRequestMock, useAgentMock } = vi.hoisted(() => ({
   httpsRequestMock: vi.fn(),
   httpRequestMock: vi.fn(),
+  useAgentMock: vi.fn(),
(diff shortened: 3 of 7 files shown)


mrosberghaus:
bugbot run

mrosberghaus:
Both review findings addressed in f70ddc2e:

**1. Expanded/non-canonical IPv6 bypass** — replaced the hand-rolled string checks with `net.BlockList`, which parses + normalizes addresses before subnet matching. Now spelling-independent: `0:0:0:0:0:0:0:1`, `::ffff:7f00:1` (hex IPv4-mapped), `::ffff:169.254.169.254`, etc. are all blocked (added as test cases).

**2. DNS rebinding (TOCTOU)** — the validate-then-`fetch(host)` pattern allowed re-resolution at connect. The connection is now **pinned**: `ssrfSafeFetch` issues the request via `node:http`/`https` with a guarded `lookup` that re-resolves, rejects any non-public address, and hands the validated address directly to the socket — so the validated IP *is* the connected IP. No rebinding window.

Also simplified per review: `assertPublicHttpUrl` is now a cheap synchronous scheme + IP-literal check (no DNS); hostname resolution/validation lives solely in `guardedLookup`, removing the redundant double-resolve.

47 tests pass, typecheck + biome clean.

mrosberghaus:
Per review feedback, swapped the bespoke IP-classification code for the maintained **`request-filtering-agent`** (pinned `3.2.0`, backed by `ipaddr.js`).

- Deleted `net.BlockList` range table, `guardedLookup`, and `isDisallowedIp`.
- `ssrfSafeFetch` now issues each hop via `node:http`/`https` with `agent: useAgent(url)` — the agent refuses connections to private/loopback/link-local addresses **at connect time** on the resolved IP, so DNS-rebinding and redirect-to-internal stay covered.
- `assertPublicHttpUrl` is now just a scheme/URL pre-check.
- Added an integration test (`url-guard.integration.test.ts`) asserting real connections to `127.0.0.1`, `10.0.0.1`, `169.254.169.254`, and `[::1]` are refused.

Net: ~90 lines of security-critical bespoke code removed, address classification offloaded to a maintained lib. 14 tests pass, typecheck + biome clean.

PR #3087, opened 2026-06-10
feat(core): add organization invitation read endpoints and route web through them

mrosberghaus:
## What

Migrates the two direct-DB invitation reads out of `apps/web`'s `organization.service` into `apps/core`, continuing the web→core DB-access consolidation.

### Core endpoints
- **`GET /v1/organizations/{id}/invitations`** (owner/admin) — lists pending invitations, de-duplicated to the most recent per email. Authz via `resolveMemberOrganizationById({ allowedRoles: [OWNER, ADMIN] })`.
- **`GET /v1/invitations/{id}`** — **public** (mirrors the `share` route): the invitation id is the capability token, so the accept-invitation page keeps working while logged out / pre-signup. Returns a `kind`-discriminated result preserving the `not_found` / `expired` / `inviter_not_found` distinction the invitation card renders.

Shared lookup/dedup logic lives in `apps/core/src/helpers/invitation.ts` (unit-tested); both routes have tests.

### Web
- `organization.service` now calls these via `coreClient` and **no longer imports `@sokosumi/database` repositories or `prisma`**.
- The accept-invitation card + actions switch from the Prisma `InvitationWithRelations` type to a slim `PendingInvitationDetail` DTO.

## Why not better-auth's `getInvitation`/`listInvitations`?
1.6.15 exposes them, but they're stricter/different: `getInvitation` requires a session **and** recipient-email match (breaks the logged-out onboarding preview) and collapses the three error codes; `listInvitations` gates on any-member with no pending-filter/dedup. They also run **inside web**, so they wouldn't move data access to core — the point of this migration.

## Notes / review focus
- `GET /v1/invitations/{id}` is intentionally public, preserving existing behavior (the web page already serves this to logged-out users). The id is uuid-v7.
- `PendingInvitation` DTO is structurally identical to the Prisma `Invitation`, so the org members page keeps its `Invitation[]` typing untouched.

## Test plan
- [x] `pnpm --filter @sokosumi/core typecheck` + `pnpm --filter web typecheck`
- [x] core suite (160 files / 1170 tests) — adds invitation helper + 2 route tests
- [x] web invitation tests (org service mapping + accept-invitation components)
- [x] regenerated `generate:core:snapshot` (no hand edits)


mrosberghaus:
bugbot run

PR #3088, opened 2026-06-10
feat(core): add stripe customer read endpoints and route web through them

mrosberghaus:
## What

Migrates the pure DB read in `apps/web`'s `stripe.service` (`getStripeCustomerId`) out to `apps/core`, continuing the web→core DB-access consolidation. This is a **partial** migration — `stripe.service` keeps all Stripe SDK writes and the auth.ts-coupled methods.

### New core endpoints
- **`GET /v1/organizations/{id}/stripe-customer`** — **any member** of the org may read it. Authz via `resolveMemberOrganizationById` (no role restriction), matching the existing web check (`purchaseCredits` / `claimFreeCreditsWithCoupon` only verify org membership before reading the customer id).
- **`GET /v1/users/{id}/stripe-customer`** — `me`, the user themselves, a session admin, or a delegated coworker may read it. Reuses the existing `/users/{id}` `usersPathUserContextMiddleware` authz.

Both return `{ stripeCustomerId: string | null }` and **404** when the entity does not exist, preserving the prior web throw-on-missing semantics (`executeOperation` rethrows as `CoreApiRequestError`). The `null` value is meaningful — the entity exists but has no Stripe customer provisioned yet — so it is returned as a 200 object field rather than via a nullable top-level schema.

Shared response shape lives in `apps/core/src/schemas/stripe.schema.ts`. Both routes have tests.

### Web
`stripe.service`'s private `getStripeCustomerId` now calls `coreClient.getOrganizationStripeCustomer` / `getUserStripeCustomer`. The public service interface is unchanged, so callers/components are untouched. New `coreClient` wrappers added in `core.shared.ts`; typed client regenerated.

## What stayed in web (and why)
- **`syncUserEmailWithStripe`** — reachable from the Better Auth session bootstrap (`auth.ts` `/verify-email` after-hook). It must not depend on a core HTTP call, and it is a Stripe **write** anyway. Stays.
- **`syncOrganizationInvoiceEmailWithStripe`, `createStripeCustomerForUser/ForOrganization`, `createStripeCheckoutSession`, `claimCoupon`, `getCoupon`, `claimWelcomeCoupon`, `createAndApplyReferralCredits`** — all perform Stripe SDK writes (and the customer-creation reads are intertwined with those writes). Out of scope for a reads-only extraction; they keep their `@sokosumi/database`/`prisma` usage.

`@sokosumi/database` repositories and `prisma` imports remain in `stripe.service` because the retained write methods still use them — only `getStripeCustomerId`'s direct reads moved.

## Notes for review
- Org endpoint is intentionally **any-member** (not owner/admin) — both web call sites buy credits / claim coupons after only a membership check.
- The user endpoint forwards the session cookie via the server `coreClient`, so the existing `userId` passed from `createStripeCheckoutSession`/`claimCoupon` (always the session user) resolves through the same `me`/self authz.

## Test plan
- [x] `pnpm --filter @sokosumi/core typecheck` — clean
- [x] `pnpm --filter web typecheck` — clean
- [x] `pnpm --filter @sokosumi/core test:ci stripe-customer` — 14 passed
- [x] `pnpm --filter @sokosumi/core test:ci v1/index` — passed (reused routers, no new stub needed)
- [x] `pnpm --filter web test:ci stripe.service` — 3 passed
- [x] `pnpm exec biome check` on touched files — clean


mrosberghaus:
bugbot run

mrosberghaus:
bugbot run

PR #3089, opened 2026-06-10
feat(core): add user membership read endpoints and route web user service through them

mrosberghaus:
## Summary

Continues the web→core DB-access consolidation by moving two safe, user-scoped READ methods out of `apps/web/src/lib/services/user.service.ts` direct `@sokosumi/database` access and routing them through the `apps/core` Hono API via `coreClient`. The public `userService` interface is unchanged, so callers/components are untouched.

## Migrated reads

| Method | New Core endpoint | Authz |
| --- | --- | --- |
| `getMyMembersWithOrganizations` | `GET /v1/users/me/members` → `MemberWithOrganization[]` | users `/{id}` path-context middleware |
| `getMyMemberInOrganization` | `GET /v1/users/me/organizations/{organizationId}/member` → `Member` (404 → `null` in web) | users `/{id}` path-context middleware |

### Preventing cross-user reads
Both endpoints mount under the existing `users/[id]` router, whose `usersPathUserContextMiddleware` → `resolveUsersPathUserId` resolves the literal `me` to the authenticated session user and, for any concrete id, enforces `requireAccessToTargetUserData` (caller must match the target user or be a session admin). The web wrappers always call with `id: "me"`, so a user can only read their own memberships. The `coreClient` forwards the session cookie, so the Core call runs in the same authenticated context.

## Stayed in web (with reasons)

- **Writes** — `getMyJobs` (does `workspaceRepository.upsertWorkspaceForContext`), `showOnboarding` (does `updateUserOnboardingCompleted`), `markOnboardingCompleteForMe` (Better Auth `updateUser`).
- **Session-only** — `getActiveOrganizationId` reads only from the session object (no DB).
- **Bootstrap/layout-adjacent read** — `getActiveOrganization` runs in `(app)/layout.tsx` on every authenticated render alongside `showOnboarding`; left in web this round to stay conservative on the session-bootstrap-adjacent path.
- **Unused dead reads** — `getMe`, `getMyValidPendingInvitations`, `checkExistingUsers` have zero callers in `apps/web`; left in place to avoid adding unused Core endpoints (and to keep the interface stable). Candidates for a later cleanup PR.

`auth.ts` does not call `userService` at all (verified), and the migrated methods are invoked from page/layout React components after the session is already established — not from within the Better Auth session hooks — so this does not touch the session bootstrap.

## Test plan

- `pnpm --filter @sokosumi/core typecheck` — clean
- `pnpm exec biome check` on touched files — clean
- `pnpm --filter @sokosumi/core test:ci` — 1172 passed (incl. 2 new route contract tests)
- `pnpm --filter web test:ci user.service` — passed (new tests for both migrated methods incl. session-absent and 404→null)
- `pnpm --filter web test:ci billing` — passed (consuming pages)

## Left for human review
- Whether to also migrate `getActiveOrganization` (bootstrap-adjacent layout read) in a follow-up.
- Whether to remove the three unused reads (`getMe`, `getMyValidPendingInvitations`, `checkExistingUsers`) rather than migrate.


mrosberghaus:
bugbot run

mrosberghaus:
bugbot run

mrosberghaus:
bugbot run

PR #3093, opened 2026-06-10
feat(core): move DESIGN.md storage to core (CRUD endpoints + workspace resolver)

mrosberghaus:
## What

Completes the web→core DB-access migration for the DESIGN.md feature. `apps/web`'s `design-md.service` and the design-md server actions no longer import `@sokosumi/database`/prisma — all DESIGN.md storage now lives behind `apps/core` HTTP endpoints.

## Endpoints

DESIGN.md is owned by users and organizations (stored in their `metadata`), so the API is CRUD per owner, plus a session-scoped resolver for "which one applies to me right now":

| Method | Path | Auth | Purpose |
| --- | --- | --- | --- |
| `GET` | `/v1/users/{id}/design-md` | `me`/self/admin | The user's own stored DESIGN.md (`{url, extractionId} \| null`) |
| `PUT` | `/v1/users/{id}/design-md` | `me`/self/admin | Set or clear the user's DESIGN.md (null `url` clears) |
| `GET` | `/v1/organizations/{id}/design-md` | any member | The organization's own stored DESIGN.md |
| `PUT` | `/v1/organizations/{id}/design-md` | owner/admin | Set or clear the organization's DESIGN.md |
| `GET` | `/v1/workspaces/design-md` | session | **Resolver** — the DESIGN.md in effect for the caller's current workspace |

The `workspaces/design-md` resolver derives the active workspace from the session (`authContext.organizationId = session.activeOrganizationId`): the active organization's DESIGN.md when the caller is a member, otherwise the personal one, else null. It takes **no parameters**. This replaces the earlier `GET /v1/users/{id}/effective-design-md`.

A `Workspace` is a first-class entity in the schema (a `userId` XOR an `organizationId`), which is why the resolved/effective read lives under a `workspaces` namespace while the stored record stays owner-addressed.

## Core internals

- Pure, unit-tested helpers in `apps/core/src/helpers/design-md.ts`: `buildUser/OrganizationDesignMdMetadata` (read-merge-normalize for `PUT`) and `readUser/OrganizationDesignMd` (metadata → record for `GET`), reusing `@sokosumi/utils`.
- New schemas: `designMdWriteSchema` (`PUT` body), `persistedDesignMdSchema` (`GET`/`PUT` response), `effectiveDesignMdSchema` (resolver response). Nullable is nested inside an object, never on an `.openapi("Name")` schema.
- Org authz via `resolveMemberOrganizationById` (any-member for `GET`, `[OWNER, ADMIN]` for `PUT`); user routes via the `users/{id}` me-resolution middleware.
- New top-level `workspaces` router stubbed in `v1/index.test.ts`.

## Web

- Persistence goes through `coreClient.setMyDesignMd` / `setOrganizationDesignMd`; `assertCanManageOwner` checks the org role via the existing `getMyMemberInOrganization` endpoint; the resolver is fetched via `getWorkspaceDesignMd()`.
- `revalidateOwner` resolves the org slug via `coreClient.getOrganizationById` instead of prisma.
- Blob upload, the Masumi external client, job-token signing, and preview-URL derivation stay in web (they are web-only concerns).
- The dead `userId` / `activeOrganizationId` parameters were removed end-to-end now that core resolves both from the session.

## Verification

- `pnpm --filter @sokosumi/core typecheck`, `pnpm --filter web typecheck`, `pnpm --filter @sokosumi/database typecheck` ✅
- Core: 32 design-md tests (helpers + 5 routes + v1 index stub) ✅
- Web: design-md service + task action tests ✅
- Biome ✅; generated client regenerated.

## Security

The `PUT` endpoints accept DESIGN.md **content**, not a URL — Core uploads it to blob storage server-side and owns the resulting URL. There is no client-supplied URL to forge, so the previous blob-allowlist concern is eliminated by construction (content is size- and non-empty-validated; storage failure → 503).


apps/web/src/lib/services/design-md.service.ts
```diff
@@ -84,6 +83,12 @@ export type StartDesignMdGenerationResult =
 
 interface ResolveEffectiveDesignMdInput {
   activeOrganizationId?: null | string;
+  /**
+   * Retained for interface stability. Core resolves the effective DESIGN.md for
+   * the authenticated session user (`me`), so this is no longer read here — all
+   * call sites pass `session.user.id`. Do not rely on it to resolve a different
+   * user's DESIGN.md.
+   */
```
mrosberghaus: Then please fix it.

mrosberghaus: Done — removed `userId` end-to-end instead of documenting it (commit 11166d87): dropped from `resolveEffectiveDesignMd` + `appendDesignMdToDescription`, the 4 page/layout call sites, the task action, and the now-unused `createTaskFromDescription` input field. Core resolves the session user (`me`), so it was dead.

PR #3187, opened 2026-06-15
feat(web): stream Hermes chat with live progress

PatrickTobler:
## Summary

Adds **opt-in SSE streaming with live progress** to Hermes chat. When enabled, replies render token-by-token with a live view of what the agent is doing (reasoning beats, tool-call chips, an elapsed timer), and each turn's answer + trace + duration are **persisted server-side** so they survive a tab close or reload.

> [!IMPORTANT]
> **Gated behind `NEXT_PUBLIC_HERMES_STREAMING` (off by default).** When the flag is unset/`!= "1"`, the existing buffered request/response path is used unchanged. This makes the merge low-risk: the entire streaming + capture path stays dormant until the flag is flipped. The changes that are *live regardless of the flag* (message-merge reconciliation, confirmation UX, prompt instance refresh) are small and individually tested.

## Motivation

Hermes replies were fully buffered: the user watched a generic "thinking" indicator for the whole turn (often 10–40s) with no signal of progress, and the answer, its reasoning/tool trace, and how long it took all vanished on reload. This makes long agent turns feel broken. Streaming + persisted progress fixes both the perception and the durability.

### Streaming + live progress — web (`running-state.tsx`, `lib/hermes/sse.ts`, `api/hermes/chat/route.ts`)
- **`lib/hermes/sse.ts`** — SSE reader that branches on the `event:` field, reassembles chunk-split frames (handles CRLF, multi-line `data:`, `[DONE]`), and parses named `event: hermes.status` progress frames (`thinking → reasoning → tool → tool_done → answering`) interleaved with OpenAI-style content deltas.
- **`api/hermes/chat/route.ts`** — when the body has `stream:true`, forwards `X-Hermes-Progress:1` to Core's `/chat/stream` and returns the raw `text/event-stream` body; otherwise the buffered path is untouched.
- **Client UX** — token-by-token rendering with a streaming caret, an elapsed timer, scroll-follow + a jump-to-latest pill, live **tool-call chips**, transient **reasoning beats** (held ≥1s so they don't blink), a collapsible **per-turn step trace** (reasoning + tool actions, in order), an "Answered in Ns" stamp, and an always-visible copy button.

### Server-side capture & persistence — core (`routes/v1/hermes/index.ts`, `schemas/hermes.schema.ts`)
- **`/chat/stream`** forwards `stream:true` + `X-Hermes-Progress`, then `tee()`s the upstream SSE: one branch streams to the client, the other is **captured to completion server-side** under a server-owned 5-minute `AbortController` (decoupled from the client connection, so closing the tab still persists the turn).
- The captured turn (final content, `tool`/`reasoning` steps, wall-clock `durationMs`) is persisted via `persistTurn`.
- `HermesMessage` gains nullable **`steps Json?`** and **`durationMs Int?`** columns (two additive `ADD COLUMN` migrations). `listMessages` returns them so the disclosure + duration stamp rehydrate on reload.

### Task-creation / confirmation UX (`running-state.tsx`)
- Confirmation cards now **interleave chronologically** in the timeline (by resolved-at) instead of being pinned at the bottom.
- On approval, the card shows a green **"The agent is creating your task in the background…"** note rather than a premature task card — the real clickable `/tasks/:id` card arrives from the orchestrator's self-contained `confirmation_resolved` message (rendered by `MessageRow`). Rationale: the task doesn't exist until the agent creates it, so rendering a card on approval is misleading.
- The instance is **refreshed on turn completion and after a confirmation resolves**, so a medium-autonomy approval gate (which lives on `instance.pendingConfirmations`, not in the message stream) surfaces within one round-trip instead of waiting up to 30s for the background poll.

### Reliability fixes (found by an adversarial pre-merge review — see below)
- **Message merge** (`merge-persisted-messages.ts`): reconcile optimistic ↔ persisted messages by a **one-for-one content multiset** so identical back-to-back messages (e.g. sending "yes" twice) are never collapsed into one. Replaces the earlier time-window heuristic (which caused follow-up answers to flicker out and back). Regression test added.
- **Streaming `finally`**: wrapped the post-turn refresh in try/catch — it ran inside a `void`-discarded async IIFE, so a session-expiry rejection would have become an unhandled rejection.
- **Confirmation resolution**: the card now passes its own `confirmation` object through `onResolved`, so a background instance refresh can't store stale metadata under the resolved id.

## Behavior & flag

| `NEXT_PUBLIC_HERMES_STREAMING` | Behavior |
| --- | --- |
| unset / `"0"` (default) | Buffered request/response — **unchanged** from today. |
| `"1"` | SSE streaming + live progress + server-side capture/persist. |

Added to `apps/web/.env.example` with a comment.

## Schema / migrations

Two **additive, non-destructive** migrations:
- `20260615120000_add_hermes_message_steps` → `ALTER TABLE "hermesMessage" ADD COLUMN "steps" JSONB;`
- `20260615120100_add_hermes_message_duration` → `ALTER TABLE "hermesMessage" ADD COLUMN "durationMs" INTEGER;`

Both columns are nullable and only written by the streaming capture path, so existing rows and the buffered path are unaffected.

> **Deploy note:** these columns were applied out-of-band (direct `ALTER`) to the shared dev/preprod Neon branch during development. On that specific branch, `prisma migrate deploy` may need `prisma migrate resolve --applied <migration>` first since the column already exists. Production/fresh databases apply cleanly.

## Testing & verification

- `pnpm --filter web typecheck` and `pnpm --filter core typecheck` — clean.
- `pnpm --filter web test` (hermes: merge **17**, sse, confirmation-action) and `pnpm --filter core test` (hermes index **38**, capture-from-stream) — green.
- `biome check` clean on all touched files; `/hermes` recompiles with no errors.

**Manual:** with the flag on, send a tool-using prompt → live reasoning + tool chips → streamed answer; expand the step disclosure and reload → trace + "Answered in Ns" persist. Create a task → approve → green "creating in background" note → self-contained result card.

## Pre-merge adversarial review

Ran a multi-agent find→verify review over the diff (4 dimensions, each finding independently re-checked to refute it). It **confirmed and this PR fixes 3 high-severity issues** (the three reliability fixes above) and **refuted** the scarier hypotheses (no timeout leak, no blank-turn persists, capture *does* drain on client disconnect, the SSE sort is stable, no reasoning/label field-swap).

### Known follow-ups (non-blocking, low/medium)
- Double-click double-approve is benign (orchestrator is idempotent → `already_resolved`), but a hard UI guard would be tidier.
- `captureFromStream` empty-content recovery lacks a Sentry breadcrumb (observability only).
- The "creating in background" note never auto-expires if a result never arrives; `resolvedConfirmations` grows for the page lifetime; `durationMs` includes a little pre-stream setup latency — all cosmetic/minor.

### Related, out of scope
With streaming on, some reasoning models leak raw `<think>` / `<mm:think>` chain-of-thought into the answer channel — that's an **orchestrator streaming-endpoint** bug (the buffered path strips it), to be fixed there. A defensive client-side strip is a possible follow-up.


PatrickTobler:
## Scaling / serverless follow-ups (before enabling streaming for real users)

Not blocking this PR — streaming is gated off by default — but flagging serverless-specific issues to fix before the flag is flipped on for real users. These are about Vercel's function lifecycle, which the inline logic review didn't cover.

### 1. Capture/persist isn't protected by `waitUntil` → turns can silently fail to persist (most important)

`apps/core/src/routes/v1/hermes/index.ts:1608`

```js
void captureFromStream(toCapture).then(persistTurn)...   // ← void, not waitUntil
```

`waitUntil` is already imported and used elsewhere in this file (lines 736, 773) — just not here. On Vercel, work that runs *after* the response stream finishes (the `persistTurn` DB write) races against the platform **freezing the function**. In streaming mode `persistTurn` is the *only* writer that saves the turn, so when the user closes the tab mid-answer, or the client drains faster than the capture branch, the turn (message + step trace + duration) **isn't written and is gone on reload**. Probability rises under load (more aggressive freezing) and on early disconnect.

**Fix:** wrap the detached capture in `waitUntil(...)`.

### 2. `maxDuration` mismatch → truncation on long turns

- **Core:** `apps/core/vercel.json` caps the function at **120s**, but the server abort is set to **5 minutes** (`5 * 60_000`), so the abort never fires — Vercel hard-kills at 120s first (504 to client, capture dropped).
- **Web proxy:** `apps/web/src/app/api/hermes/chat/route.ts` has **no `maxDuration`** and relies on the project default (typically ~15s for a Next.js route), so a turn longer than that is cut at the web layer before Core matters.

**Fix:** set `export const maxDuration` on the web route, and lower the server abort to just under the function cap (~110s) so it ends gracefully + persists a partial instead of being hard-killed.

### Softer concerns (cost/load, not breakage)

- **Long-lived connections on serverless:** each streaming chat pins a function instance (web *and* core) for the whole turn. Scales horizontally, but a held ~40s connection bills far more GB-seconds than a normal request and hits concurrency limits sooner. At high concurrency, consider the Edge runtime for the streaming endpoints or a dedicated long-running service.
- **Polling multiplier:** ~5s message poll + 30s instance poll per open tab → roughly linear load on Core/orchestrator (~N/5 req/s at N tabs). Chatty, not fatal; the stream could push instead of poll.
- **Orchestrator is the real ceiling:** it runs a per-user instance, so its capacity bounds throughput regardless of the proxy.

**Non-issues:** `tee()` buffering is bounded (answers are small), and the per-turn DB writes are ordinary load.

### Suggested order
1. `waitUntil` on the capture (one line — prevents silent message loss). 2. `maxDuration` on the web route + abort alignment. 3. Revisit runtime/cost model if concurrency gets high.

PatrickTobler:
@francisluz pls check

PR #3201, opened 2026-06-16
fix(web): respect Hermes' proposed workspace in confirmation approvals

PatrickTobler:
## Summary

Fixes a workspace-routing bug: the in-chat confirmation popup files Hermes tasks in the **wrong workspace** by overriding Hermes' chosen workspace with an unintended "Personal" override.

## The bug

When Hermes proposes `sokosumi_create_task`, it includes the target workspace in the tool call (e.g. NMKR). But the confirmation popup defaulted the dropdown to **Personal** and, on Approve, **always** sent `overrides.organizationId` — so an untouched dropdown sent `organizationId: null`.

Per the approve contract:
- `organizationId: "<id>"` → file in that workspace (overrides Hermes)
- `organizationId: null` → explicit Personal (overrides Hermes)
- **field omitted → use the workspace Hermes proposed**

Sending `null` on an untouched dropdown asserts a Personal choice the user never made, and the orchestrator applies it over Hermes' proposal. Confirmed in the orchestrator log on a real approval:

```
approval_override_applied  from="cmayz5qwg0002v7us24owa0x1" (NMKR)  to=null (Personal)
```

## The fix (immediate — resolves the reported case)

Send the organization override **only when the user actually changes the workspace dropdown** from its initial value. If left untouched, omit `overrides.organizationId` entirely so Hermes' proposed workspace stands.

The decision is extracted into a pure, tested helper:

```ts
buildConfirmationApproveOverrideIfChanged(selected, initial, orgs)
  // → undefined when unchanged (omit the field)
  // → { organizationId } only on a deliberate change
```

New unit tests cover: untouched Personal default → omit; untouched org → omit; switch to org → `{ organizationId }`; switch to Personal → `{ organizationId: null }`.

## Follow-up (needs orchestrator support — not in this PR)

Better UX, per the bug report: default the dropdown to the workspace **Hermes proposed**, so the user sees and confirms the actual target. This requires the orchestrator to add `organizationId` + `organizationName` to each pending confirmation (`GET /v1/instances/{userId}/confirmations`). Once those fields land, I'll default the dropdown to `organizationId` (showing `organizationName`) and continue sending an override only on a deliberate change. Tracked as a follow-up.

## Testing

- `pnpm --filter web typecheck` — clean.
- `pnpm --filter web test` — `confirmation-org-picker` (**+4 new**) and `confirmation-action` → **19 pass**.
- `biome check` clean.

## Notes

Independent of the Hermes streaming PR (#3187) and affects the buffered path too, so it's a standalone fix off `main`. Both touch `handleApprove`, so a small merge resolution is expected when the second one lands.


PatrickTobler:
## Fix #2 added — default the workspace dropdown to Hermes' proposed org

The orchestrator shipped the proposed-workspace fields, so the coordinated half is now done on this same branch.

`GET /v1/instances/{userId}/confirmations` now returns `organizationId` (`null` = Hermes proposed personal) + `organizationName` per confirmation. This commit threads them through:

- **Core:** orchestrator client (`normalizePendingConfirmation`, with a snake_case alias defensively) + `hermesPendingConfirmationSchema`. The enrichment pass already spreads the confirmation, so the fields flow through.
- **Web:** regenerated Core client → `HermesPendingConfirmation` type → `mapHermesPendingConfirmation`.
- **UI:** `resolveConfirmationOrgPickerValue` now defaults to the proposed workspace; `mergeConfirmationOrgPickerOptions` adds it as a selectable option (labelled by `organizationName`). Combined with the only-override-on-change logic from the first commit, the dropdown shows Hermes' actual target and an override is sent **only** on a deliberate change.

**Net:** the dropdown reflects Hermes' proposal by default; an override is sent only when the user redirects it — exactly the requested behavior.

Verification: picker + confirmation-action suites (21) and core hermes (35) green; both apps typecheck + Biome clean.

PatrickTobler:
@francisluz pls check :)

PR #3258, opened 2026-06-29
chore(core): add coworker chat API benchmark script

francisluz:
## Summary

- Adds `apps/core/scripts/bench-coworker-chat.mts` to benchmark Elena (or any coworker) directly against the Conversations + Responses API.
- Measures conversation create time, response TTFB, first token, stream completion, output size, and agent-error turns.
- Supports `--warmup`, `--turns`, `--scenario` (`short` | `realistic`), `--json`, and `--json-out`.
- Wires `pnpm --filter core bench:coworker-chat`.

## Test plan

- [x] `pnpm --filter core bench:coworker-chat -- --warmup 1 --turns 3`
- [x] `pnpm --filter core bench:coworker-chat -- --scenario realistic --warmup 1 --turns 10 --json-out .cursor/elena-chat-benchmark-realistic.json`

### DIFF
--- a/apps/core/package.json
+++ b/apps/core/package.json
@@ -28,7 +28,8 @@
     "test": "vitest run",
     "test:ci": "vitest run",
     "test:watch": "vitest",
-    "write-openapi-snapshot-for-web": "tsx scripts/write-openapi-for-web.mts"
+    "write-openapi-snapshot-for-web": "tsx scripts/write-openapi-for-web.mts",
+    "bench:coworker-chat": "tsx scripts/bench-coworker-chat.mts"
   },
   "dependencies": {
     "@better-auth/api-key": "1.6.19",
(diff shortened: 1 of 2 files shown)


francisluz:
Closing as superseded by #3262.

Both PRs add the same script path (`apps/core/scripts/bench-coworker-chat.mts`) and npm script (`bench:coworker-chat`), so only one can land. #3262 (SOK-626) is the intended cold-start benchmark: standalone OpenRouter Responses API, regression thresholds, and docs under `docs/coworker/benchmarks/`.

This PR's full Core Conversations + Responses E2E approach is not needed for the current cold-start program. If we want E2E benchmarking later, we should add a separate script (e.g. `bench:coworker-chat-e2e`) rather than revive this branch.

PR #3264, opened 2026-07-05
feat(core): add coworker access grants with user acceptance flow

PatrickTobler:
## What

A per-coworker **access grant system** ("acceptance portal") governing what AI coworkers may do with a user's tasks when acting via API-key delegation, plus **workspace-member commenting** on colleagues' tasks and a **held-comment moderation queue** for ungranted coworker comments.

## Why

Core's own OpenAPI docs called the delegation model *"temporary: any coworker API key may delegate to any valid user until per-coworker permissions are added."* This PR adds those permissions. Users hit the old model as a confusing "Task not found in any organisation" toast when Hermes tried to comment on a task assigned to another coworker.

## How it works

**Data model** — `CoworkerGrant` per `(coworker, user, scope)`: scopes `TASK_READ` / `TASK_COMMENT` / `TASK_CREATE`, statuses `PENDING → GRANTED | DENIED | REVOKED`. `Task.awaitingAcceptance` + `createdByCoworkerId` for parked creations. `TaskEvent.heldByGrantId` for comments awaiting approval.

**Enforcement at the delegation boundary** (assigned-task baseline stays grant-free, SOK-554 intact):
- Unassigned task reads (detail + thread) → `TASK_READ`; the task **list** soft-degrades to assigned-only instead of spamming requests.
- Comment-only events → `TASK_COMMENT`. A **granted** coworker can comment on *any* task in the workspace, matching `TASK_READ`'s scope. An **ungranted** comment on the delegating user's *own* task is stored **held** (`heldByGrantId` set) — visible only to the owner until they approve (releases it, original timestamp intact) or deny (deletes it). An ungranted comment on a *colleague's* task is rejected outright (a hold there would dangle — the colleague can't resolve this user's grant).
- Task creation → `TASK_CREATE`; without it the task is **parked in `INPUT_REQUIRED`** (the board's "attention" column, not DRAFT) with `awaitingAcceptance: true` — fully invisible/inert to every agent surface (reads, the event feed, status transitions, job attachment = spend) until the owner decides. The owner accepts (→ `READY`) or declines (→ `CANCELED`) from an in-task banner offering **accept once**, **always allow** (grants `TASK_CREATE` so future tasks start immediately), or **decline**; a session-only transition bypass makes this possible without opening the general transition table (a delegated coworker can never accept its own creation). Any other transition on an awaiting task is rejected — the generic "Cancel request" action is hidden so a task can't get stranded in `CANCEL_REQUESTED` with no agent able to process it.

**Held comments** get the same treatment as awaiting tasks: an inline strip on the comment itself — **Approve comment** (this one only), **Always allow** (grants `TASK_COMMENT`, releasing every comment held under it), **Remove** (discards without deciding) — plus a notification per held comment (not just once per access request), carrying the coworker's avatar and deep-linking to the task.

**Asking for access** — a blocked call fails with `403 kind=grant_required`, auto-creates the PENDING request (outside the failing transaction, so it survives the rollback) and notifies the user (idempotent; DENIED/REVOKED are never re-surfaced by retries).

**Deciding** — session-only endpoints `GET /coworker-grants` and `PATCH /coworker-grants/{id}` (a delegated coworker can never resolve its own request), surfaced in **Connections → Coworker access** (approve / deny / revoke / re-approve) and via `COWORKER_ACCESS` notifications with inline Approve/Deny **in both the notifications page and the header bell dropdown**, showing the requesting coworker's avatar.

**Hermes** — auto-granted all scopes at instance provisioning (reserved `hermes` slug) from the user's own session only — never mintable through a coworker key's delegation headers; explicit revocations are never overridden.

**Workspace members (humans)** — can comment on colleagues' tasks in shared workspaces (membership = consent, never held). Editing/status changes stay owner-only.

## Security hardening (post-implementation audit)

An internal audit surfaced several gaps, all fixed in this PR:
- **Held-comment confidentiality** is now enforced centrally in the Prisma includes (not per-route): closed leaks via the task list endpoint, public share links (unauthenticated!), the assigned agent's polling feed, and mutation-route responses.
- Awaiting-acceptance tasks are now inert for **delegated coworkers too**, not just bare agents (a delegated coworker can be the very creator whose task is pending).
- `TaskEvent.heldByGrant` FK changed `SetNull → Cascade`: deleting a coworker no longer silently *publishes* its held comments.
- `POST /tasks` now gates the calling coworker with the same capability check as every sibling delegated surface.

## Migrations

- `20260705120000_add_coworker_grants` — enums + `coworker_grant` table + `COWORKER_ACCESS` notification kind.
- `20260705150000_task_acceptance_and_held_comments` — `Task.awaitingAcceptance`/`createdByCoworkerId`, `TaskEvent.heldByGrantId`.
- `20260706090000_held_comments_cascade_on_grant_delete` — FK cascade fix above.

No data backfill needed for any of the three.

### Setup (once)

You need three things, all obtainable from the app itself:
1. **Two coworkers** in your workspace — one you'll leave **ungranted** (to exercise parking/holding/requests) and one that's assigned to tasks (to receive them). Any two coworkers work; Hermes is pre-granted so don't use it for the "ungranted" role.
2. **A coworker API key** for the ungranted coworker: `Coworkers → [pick one] → API Keys → Create key`. Copy the `coworker_…` token.
3. **Your own user id**: `Settings → Account` (or the `id` field in `GET /v1/users/me` from your browser session).

Below, `$TOKEN` = the ungranted coworker's API key, `$USER_ID` = your user id, `$ASSIGNEE_ID` = the other coworker's id (assign it to the task you create so there's an agent to "receive" it).

### 1 — Task creation parking (`TASK_CREATE`)

```bash
curl -s -X POST "$CORE_URL/v1/tasks" \
  -H "Authorization: Bearer $TOKEN" -H "X-Delegation-User-Id: $USER_ID" \
  -H "Content-Type: application/json" \
  -d "{\"name\":\"Test task\",\"coworkerId\":\"$ASSIGNEE_ID\",\"status\":\"READY\"}"
```
Expect `status: "INPUT_REQUIRED"`, `awaitingAcceptance: true`. In the app: the task appears on the board in the **Input Required** column with a **"Needs your acceptance"** chip; opening it shows a banner with **Accept / Always allow {name} / Decline**. Accept it — status flips to `READY` and the banner disappears. Create a second one and click **Always allow** instead — it starts immediately, and a third task created the same way (re-run the curl above) starts as `READY` right away with no banner at all.

Guardrail to check: the assigned coworker (via its own API key) cannot see or act on the parked task before acceptance — `GET /v1/tasks/{id}` and any status-changing call should 404.

### 2 — Held comments (`TASK_COMMENT`)

Using the same ungranted `$TOKEN`, comment on one of your own existing (non-parked) tasks:
```bash
curl -s -X POST "$CORE_URL/v1/tasks/$TASK_ID/events" \
  -H "Authorization: Bearer $TOKEN" -H "X-Delegation-User-Id: $USER_ID" \
  -H "Content-Type: application/json" \
  -d '{"comment":"test comment"}'
```
Expect `held: true`. In the app: you get a notification ("… left a comment that needs your approval") with the coworker's avatar; opening the task shows the comment with **Approve comment / Always allow / Remove** underneath. Approve it — it becomes visible to everyone on the task. Comment again with the same token — held again (approving was once-only unless you used Always allow).

Guardrail to check: log in as a **different** workspace member (or check via another session) — the held comment must not be visible to them, and it must not appear on a public share link for that task if one exists.

### 3 — Access requests (Connections + notifications)

Both of the above trigger a `COWORKER_ACCESS` request the first time each scope is touched. Check:
- **Connections → Coworker access** lists the pending request; Approve/Deny/Revoke there.
- The **notification bell dropdown** and the **/notifications page** both show the request with the coworker's avatar and inline **Approve/Deny** buttons — deciding there should be equivalent to deciding in the portal (and updates both surfaces immediately if you have both open).
- Deny a request once, then repeat step 1 or 2 — it must **not** re-prompt (denied requests don't resurface); the call fails outright instead of holding/parking.

### 4 — Workspace-member commenting (humans)

As your own logged-in user, open a **colleague's** task in a shared workspace and comment on it — should post immediately, no grant involved. Status/edit controls should remain greyed out (comment-only).

### Automated coverage

All of the above is also pinned by tests — see `pnpm --filter core test` (grant helper, grant routes, parking branch, held-comment creation + notification, per-viewer visibility, agent-feed filtering) and `pnpm --filter web test` (notification actor resolution, acceptance banner, held-comment UI).

## Follow-ups (not in this PR)

- Orchestrator (`hermes-as-service`): match on `kind=grant_required` to message the user cleanly instead of masking it as "task not found in any org".
- Possible read relaxation for `GET /tasks/:id/jobs` and `/links` under `TASK_READ`.

## Verification

- Core: **1740/1740** tests, tsc + biome clean.
- Web: **1542/1542** tests (1 skipped), tsc + biome clean, client regenerated via `generate:core:snapshot`.
- Live-smoke-tested end to end against the dev server: ungranted parking → INPUT_REQUIRED, agent invisibility, held-comment creation/leak-closure, inline approve/deny in both notification surfaces.
- Adversarially reviewed by multiple independent review passes before and after the hardening fixes.


mrosberghaus:
Related to #3300 as a replicate.

PR #3359, opened 2026-07-21
fix(web): auto-scroll the assistant chat while reasoning steps stream

PatrickTobler:
## What

In streaming mode the Personal Assistant chat didn't scroll down while reasoning steps / tool-progress chips appeared and grew — the viewport just sat still, which looked broken. The scroll hook only followed two signals: a new message landing and the streamed answer text lengthening; the transient reasoning UI changes the timeline's height without touching either.

**Fix:** a `ResizeObserver` on the scroller's content pins the viewport to the bottom through *any* content-height change:

- reasoning line appearing / growing,
- tool chips appearing / completing,
- the end-of-turn **collapse** of reasoning back into the message's step disclosure (content shrinks — the viewport stays pinned instead of jumping),
- and it only acts while the user is already near the bottom (same 200px stick-to-bottom zone as the streaming follower), so scrolling up to re-read is never hijacked.

Stickiness is tracked with a flag updated from user scroll (not re-measured after resize), so a single large height jump does not drop follow. The observer re-attaches when `isEmpty` flips (WelcomeBlock ↔ ChatTimeline, including cards-only transitions), and no-ops where `ResizeObserver` is unavailable (test environments).

## Testing

- `pnpm check`, `pnpm --filter web typecheck` clean
- Unit tests: `use-chat-scroll.test.tsx` (sticky resize, scrolled-up no-hijack, large jump, isEmpty re-attach, stream follow, jump-to-latest threshold)
- Behavior still needs a real layout engine for end-to-end feel — please eyeball on the preview: send a message that triggers tool use, watch the chat follow the reasoning/chips, and confirm the collapse at the end doesn't strand the viewport.


PatrickTobler:
@mrgrauel pls check and merge

PatrickTobler:
Actually you know what. Maybe this is too small for a review now :D

PR #3574, opened 2026-08-03
feat(web): let a task pick its own DESIGN.md branding

PatrickTobler:
## Summary

- New tasks no longer seed the DESIGN.md link as raw text into the description editor. Instead, a labeled checkbox ("{organization} Brand Design", or "Your Brand Design" when there's no org) sits below the editor, checked by default whenever an effective DESIGN.md exists.
- A hover card next to the checkbox explains what DESIGN.md is and links out to https://github.com/google-labs-code/design.md ("Learn more").
- Unchecking the box skips attaching DESIGN.md to the task entirely.
- "Use different branding" opens a dialog to generate an ad hoc, unsaved DESIGN.md from another company's URL for that task only — nothing is written to the org's or user's saved profile.
- On task creation, the resolved DESIGN.md (default, ad hoc, or none) is still prepended to the task description as a markdown link exactly as before, so the agent sees it the same way it always has.

### Core API

- `GET /v1/workspaces/design-md` now also returns `owner` (`organization` w/ name, or `user`) so the web app can render the right label without an extra round trip.
- New `POST /v1/workspaces/design-md/adhoc`: stores a one-off DESIGN.md to blob storage under its own path, available to any authenticated user (not gated by org owner/admin, since nothing persists to a profile).

## Test plan

- [x] `pnpm format` / `pnpm check` (repo-wide, clean)
- [x] `pnpm --filter core typecheck` / `pnpm --filter web typecheck`
- [x] `pnpm --filter core test` (2607 passed)
- [x] `pnpm --filter web test` (2223 passed)
- [x] Verified live in browser: checkbox shows the real org name and is checked by default, description editor stays empty, hover card renders the real copy + working "Learn more" link, unchecking the box toggles it off, and "Use different branding" opens the ad hoc dialog correctly.


PatrickTobler:
@mrosberghaus please review

PR #3588, opened 2026-08-03
feat(web): preview PDF/Office/text attachments in tasks and chat

PatrickTobler:
## Summary
- Task/job/chat file attachments (PDF, Office docx/pptx/xlsx, text/markdown) now open a full-screen popup preview on click, reusing the same Office Online / native-PDF renderer that already powers the Pre-Built Task example-output preview — with a working download button and "open in new tab", same as that preview. Unsupported types (zip, csv, video, etc.) keep the existing plain download link.
- A task's or job's Completed-status comment/result now renders fully expanded by default instead of being truncated behind an "Expand" click.

## Implementation
- `apps/web/src/lib/utils/file-preview.ts` (new): extracted and generalized the Office/PDF detection + viewer-URL logic out of `offer-card.tsx` so both the offer preview and the new attachment viewer use identical rendering logic.
- `apps/web/src/components/ui/document-viewer.tsx` (new): `DocumentViewer`, modeled on the existing `ImageViewer` (same full-screen `Dialog`/toolbar/download pattern).
- `apps/web/src/components/ui/document-text-preview.tsx` (new): extracted the offer preview's styled "document card" text renderer so it's shared between the offer preview and the new viewer's text/markdown branch.
- `apps/web/src/components/ui/file-chip.tsx` and `file-chip-mini-preview.tsx`: wired in image/document/plain-link 3-way gating. These are the two shared chip components every attachment surface (task attachments, job sources, chat messages, compose bar) already renders through, so the preview capability propagates automatically — no other call sites changed.
- `apps/web/src/components/expandable-markdown.tsx`: added a `defaultOpen` prop (default `false`, so existing call sites are unaffected), wired at the 3 places that render a Completed-status activity/result (`job-details/outputs.tsx`, `task-activity.tsx`, `shared-task-view.tsx`).
- i18n: new `Components.DocumentViewer` message namespace added to all 9 locale files with real translations, mirroring `Components.ImageViewer`'s shape.

## Verification
- `pnpm --filter web test` — 365 files / 2327 tests passing
- `pnpm check` / `pnpm format` — clean
- Verified live against real data on a local dev server: a completed task's delivery comment now renders fully expanded (no "Expand" click needed) while earlier non-completed comments in the same thread still collapse normally; a real PDF attachment opens the popup with a working iframe + download link; a real `.docx` deliverable correctly routes through the Microsoft Office viewer.


PatrickTobler:
Pushed a follow-up fixing code-review findings (`/code-review main`) plus a UX issue you flagged:

- **Popup style mismatch**: `DocumentViewer` was copying `ImageViewer`'s full-screen black lightbox instead of the app's standard dialog treatment (the same one `OfferDetailDialog` uses). Rebuilt on the standard `Dialog`/`DialogContent` — centered, theme-aware, rounded, bordered, default close button, sized like `OfferDetailDialog`.
- Extracted `classifyFilePreview` into `file-preview.ts`, shared by `FileChip`/`FileChipMiniPreview` instead of each duplicating the image/document classification logic.
- Fixed `FileChip` dropping the `title` tooltip when it switched from `<a>` to `<button>` for image/document previews.
- Reused `getExtensionFromUrl` instead of a local reimplementation.
- Added the `DocumentViewer.loading` i18n key that was missing from all 9 locales, now used as the loading-skeleton's a11y label.
- Looked into moving the text/markdown fetch to React Query per `effects.mdc`, but the public `/share` pages don't mount a `QueryClientProvider` — kept the effect-based fetch (it already has proper `AbortController` cleanup, the exception the rule allows for).

Verified: full test suite (365 files / 2337 tests) + typecheck + `pnpm check` clean, plus a live re-check on real production-scale data confirming the new dialog is properly bounded/centered (not full-screen) and matches the app's standard popup chrome.

PatrickTobler:
## Code review

Found 1 issue.

- **[P2] Extensionless preview URLs can open an empty document modal**  
  `DocumentViewerBody` re-detects the preview kind with `getDocumentPreviewKind(url, mediaType)` only: `apps/web/src/components/ui/document-viewer.tsx:101`. But both chip components decide to open the document viewer with `classifyFilePreview(url, fileName, mediaType)`, which intentionally falls back to `fileName` when the URL has no extension. For `url="https://blob.example/report"` + `fileName="report.pdf"` or `report.docx`, the trigger opens the dialog, then the body returns `null`, leaving only the header/actions and no preview. This also undercuts the extensionless URL support documented in `file-preview.ts`. Fix by passing the already-computed `documentKind` into `DocumentViewer`, or by making `DocumentViewerBody` use the same fallback (`getDocumentPreviewKind(url, mediaType) ?? getDocumentPreviewKind(fileName, mediaType)`). Add a regression test for extensionless URL + filename for at least PDF and Office/text.

Verification: focused changed-file tests passed with package-relative paths:
`pnpm --filter web test src/components/ui/__tests__/document-viewer.test.tsx src/components/ui/__tests__/file-chip.test.tsx src/components/ui/__tests__/file-chip-mini-preview.test.tsx src/lib/utils/__tests__/file-preview.test.ts src/components/__tests__/expandable-markdown.test.tsx`

Note: local run emitted Node engine warnings because shell is on Node 22, while repo wants Node 24. GitHub checks on PR are green.

PatrickTobler:
Fixed the P2 finding from [this review](https://github.com/masumi-network/sokosumi/pull/3588#issuecomment-5172316996) — confirmed it: `DocumentViewerBody` was re-deriving the preview kind from `url`+`mediaType` only, disagreeing with the chip's `classifyFilePreview` (which falls back to `fileName`) for extensionless URLs.

Went with the "pass the already-computed `documentKind` into `DocumentViewer`" fix over duplicating the fallback logic — removes the second detection site entirely rather than keeping two call sites that could drift again. `kind` is now a required prop on `DocumentViewer`, threaded from both `FileChip` and `FileChipMiniPreview`.

Added regression coverage per your ask — extensionless URL + filename fallback for pdf/office/text at the `DocumentViewer` unit level, plus one end-to-end test on `FileChip` (click → real iframe `src`, not a blank body).

Verified: full suite (365 files / 2341 tests), typecheck, and `pnpm check` all clean.

PR #3651, opened 2026-08-05
fix(core): harden coworker delegation and agent-supplied URLs

schaier-io:
Security fixes extracted from [#3440](https://github.com/masumi-network/sokosumi/pull/3440) (masumi payment-node v2). None of this is specific to the V2 payment rail — these are pre-existing, repo-wide issues found while reviewing that branch, split out so they can be reviewed and shipped on their own.

## User-facing impact

- A third-party vendor could read and act as **any user** through a coworker API key. That is closed.
- Coworker endpoints can no longer be pointed at internal infrastructure, and Sokosumi identity headers can no longer be redirected to a host the vendor controls.
- **Breaking for vendor integrations:** a coworker now needs an existing relationship with a user before it can act in that user's workspace. See the follow-up below.

### Coworker context could impersonate any user
`X-Context-User-Id` was accepted on the strength of the user merely existing. Any valid vendor key could name any user id and be treated as that user on every route resolving the effective user via `requireUserContext` — notifications, history, projects, organization billing and members.

Context now requires a real relationship: a `CoworkerAssignment`, a **GRANTED** workspace grant, or a task already assigned to that coworker. Orchestrator service tokens are first-party and exempt.

PENDING grants deliberately do not count — reaching any task route creates a PENDING grant on the *workspace*, so accepting it would let one member's engagement hand the vendor context for every other member of the same organization before a human approved anything.

### Vendor-controlled `baseURL` reached unrestricted `fetch()`
Both conversation creation and response streaming now resolve the host and refuse private/loopback/link-local/metadata addresses, validate **every redirect hop**, and follow only **same-origin** 307/308 — a public host cannot bounce the POST, identity headers and all, to a third party. The URL is validated at write time too, so a bad endpoint never persists.

### Unbounded response buffering
Coworker conversations and agent result imports buffered arbitrary amounts of attacker-supplied data, with several imports running concurrently. Both are now size-capped and time-bounded.

### `workflow_dispatch` input reached a shell command
`inputs.agent_id` was interpolated by GitHub directly into `run:`, so a user able to dispatch the workflow could break out of the quotes with `NEON_API_KEY` in scope. It is passed via `env:` now, and the `bc-<uuid>` run-id shape is validated before any branch is deleted.

## ⚠️ Known follow-up — do not merge yet

Requiring a **GRANTED** grant blocks the documented `GRANT_PENDING` cold start, where a vendor creates the first delegated task for a new user and a human approves afterwards. `POST /v1/tasks` calls `requireUserContext` before it requests the grant, so the middleware rejects the request before the vendor can even ask for permission.

Restoring it needs two-tier context: an *unapproved* context that reaches **only** delegated task create, while every other user-scoped route keeps requiring an approved relationship. Documented in `docs/coworker/vendor-workspace-grants-api.md`.

Draft until that is resolved.

## Verification

- `apps/core` typecheck clean; `packages/ai-provider` 77 tests; `scripts/cloud-agent-db` 18 tests; `biome check .` clean
- 3 core tests fail on this branch (`job-create-agent`, `job.test`, `agents/get`) — **they also fail on pristine `main`**, so they are pre-existing and unrelated to this change


mrosberghaus (changes requested):
Thanks for extracting these security fixes from the payment-node work — useful split from the V2 rail.

## Overlap with #3682

#3682 is now on `main` and covers the coworker **user-context impersonation** path (`X-Context-User-Id` binding via grant/baseline, plus `requireAuthorizedUserContext` / `requireOwnerUserContext` for user-scoped vs owner-only routes). That supersedes the auth/context model in this PR (relationship checks, approved vs bootstrap tier, related middleware/docs).

Rebase-and-merge of this whole branch would conflict with that design and risk reintroducing a second policy model. I wouldn’t land the auth half on top of #3682.

## What still looks valuable (not on main)

These pieces still look open and worth shipping:

1. **Coworker outbound hardening** — `baseURL` SSRF rejection, redirect hop / same-origin handling, write-time URL validation, and response body size/time bounds (core + `ai-provider`).
2. **cloud-agent-db teardown** — stop interpolating `inputs.agent_id` into `run:`; pass via `env:` and validate the run-id shape before delete.
3. **`.gitignore`** — key material / broader dotenv ignores.

## Recommendation

- **Close this PR** (or leave it as a reference only) rather than forcing a full rebase.
- **Open 2–3 small PRs from current `main`**, re-applying only the non-auth hunks from this branch (treat this PR as a patch source):
  1. Outbound URL / redirect / buffer hardening (highest priority)
  2. Teardown workflow shell fix
  3. Optional: gitignore hygiene (alone or with #2)

Please **don’t** re-land the bootstrap/approved context tier unless we explicitly decide we still need extra cold-start constraints on top of #3682’s actor menu (`requireUserContext` for grant-gated task/job flows already preserves `GRANT_PENDING` first contact).

Happy to help carve the first slim PR from this diff if useful.

PR #3668, opened 2026-08-05
feat(core): add bearer-authenticated Ably token endpoint

schaier-io:
## Why

Core already publishes to Ably on four per-user channels, but the only endpoint that mints a subscribe token lives in `apps/web` behind a **session cookie**. Non-browser clients authenticate with a bearer token and have no cookie, so today they cannot subscribe at all — every live surface would have to poll.

This is the first of the server-side preconditions for the sokosumi mobile client, and the cheapest one: a pure addition, no schema change, no migration.

## What

`POST /v1/realtime/token` → an Ably `TokenRequest`, subscribe-only, scoped to the caller's own channels:

```
agent_jobs:*:user_<userId>
tasks:*:user_<userId>
notifications:*:user_<userId>
chat_rooms:*:user_<userId>
```

Behind `requireUserAuthContext`, so coworker and orchestrator actors are rejected.

## Notes for review

- **New signing client.** Signed with a new `ABLY_SUBSCRIBE_ONLY_KEY` rather than reusing `getRestClient()`. That singleton holds the *publish-only* key, and Ably rejects a token request whose capability exceeds the signing key's own — so minting subscribe tokens needs its own client.
- **Wildcard channel patterns**, mirroring the web implementation. `makeAgentJobsChannelName` embeds an agent id, so concrete names cannot be enumerated up front. There is a comment noting these must stay in sync with `packages/utils/src/ably-channel.ts`; a rename there would silently stop matching, and subscribers would fail with a capability error rather than something more obvious.
- **`ttl` is explicit** (1 hour) rather than inheriting Ably's 60-minute default, so the value is visible in code.
- **`apps/web` is untouched** and keeps its own cookie-authenticated route. No compat impact.

## ⚠️ Before deploying

`ABLY_SUBSCRIBE_ONLY_KEY` is **required and validated at startup** — the app will fail to boot without it. The value is the same subscribe-only key `apps/web` already uses; it just needs adding to core's environment (including any preview environments).

I made it required rather than optional deliberately: a missing realtime key should fail loudly at boot, not silently at runtime. Happy to relax it if you'd rather previews degrade than fail.

## Testing

- New `post.test.ts`: 4 cases — token minted for the caller, capability scoped to exactly the four channels and subscribe-only, never mints for another user's channels, non-user actors rejected.
- Full core suite: **323 passed, 1 skipped, 0 failed**.
- `pnpm check && pnpm typecheck` green across all 9 workspaces.
- Two test-infra updates were needed and are included: `src/test/setup.ts` gains the new env var, and `routes/v1/index.test.ts` stubs the new sub-router (that file mocks every router; an unstubbed one loads real dependencies and pulls in the Stripe client).

### DIFF
--- a/apps/core/src/config/env.ts
+++ b/apps/core/src/config/env.ts
@@ -184,6 +184,7 @@ const envSchema = z.object({
 
   // Ably
   ABLY_PUBLISH_ONLY_KEY: z.string().min(1),
+  ABLY_SUBSCRIBE_ONLY_KEY: z.string().min(1),
 
   // Optional outbound webhooks
   WEBHOOK_USER_CREATED: z.url().optional(),
--- a/apps/core/src/config/env.ts
+++ b/apps/core/src/config/env.ts
@@ -184,7 +184,13 @@ const envSchema = z.object({
 
   // Ably
   ABLY_PUBLISH_ONLY_KEY: z.string().min(1),
-  ABLY_SUBSCRIBE_ONLY_KEY: z.string().min(1),
+  /**
+   * Optional: only needed to mint subscribe tokens for non-browser clients
+   * (`POST /v1/realtime/token`). Left unset, that endpoint returns 503 and
+   * everything else is unaffected — deployments that do not serve mobile
+   * clients need no new configuration.
+   */
+  ABLY_SUBSCRIBE_ONLY_KEY: z.string().min(1).optional(),
 
   // Optional outbound webhooks
   WEBHOOK_USER_CREATED: z.url().optional(),
--- a/apps/core/src/lib/ably/client.ts
+++ b/apps/core/src/lib/ably/client.ts
@@ -3,6 +3,7 @@ import { Rest } from "ably";
 import { getEnv } from "@/config/env";
 
 let restClient: Rest | null = null;
+let subscribeRestClient: Rest | null = null;
 
 export function getRestClient() {
   if (!restClient) {
@@ -11,4 +12,20 @@ export function getRestClient() {
     });
   }
   return restClient;
+}
+
+/**
+ * Separate client signed with the subscribe-only key.
+ *
+ * Deliberately not `getRestClient()`: that singleton holds the publish-only
+ * key, and Ably rejects a token request whose capability exceeds the signing
+ * key's own capability. Minting subscribe tokens therefore needs its own key.
+ */
+export function getSubscribeRestClient() {
+  if (!subscribeRestClient) {
+    subscribeRestClient = new Rest({
+      key: getEnv().ABLY_SUBSCRIBE_ONLY_KEY,
+    });
+  }
+  return subscribeRestClient;
 }
--- a/apps/core/src/lib/ably/client.ts
+++ b/apps/core/src/lib/ably/client.ts
@@ -21,11 +21,19 @@ export function getRestClient() {
  * key, and Ably rejects a token request whose capability exceeds the signing
  * key's own capability. Minting subscribe tokens therefore needs its own key.
  */
+export function isSubscribeClientConfigured(): boolean {
+  return Boolean(getEnv().ABLY_SUBSCRIBE_ONLY_KEY);
+}
+
 export function getSubscribeRestClient() {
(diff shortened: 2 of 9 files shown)


mrosberghaus:
@schaier-io what's the reason behind it? There is no use for it at the moment.

schaier-io:
@mrosberghaus I am testing some offline sync stuff, can you just keep it in draft so I have a preview

schaier-io:
### Verified against the preview: the publish-key fallback does not work

Tested from the mobile client against this branch's preview deployment. Ably rejects it:

```
Auth.requestToken(): token request API call returned error
ErrorInfo: Intersection of key capabilities and requested token capabilities is empty
statusCode=401; code=40160
```

So `ABLY_PUBLISH_ONLY_KEY` genuinely carries publish capability only, as its name says. Ably will not let a key mint a token whose capability exceeds its own, and the endpoint itself is fine — the error moved from "could not mint a token" to Ably's capability check, which is the endpoint working correctly and being refused downstream.

**This means realtime needs one of:**

1. `ABLY_SUBSCRIBE_ONLY_KEY` provisioned for `apps/core` — the value `apps/web` already uses. One environment variable, no code change.
2. Broadening the existing key's capability in the Ably dashboard to include `subscribe`. Avoids touching Vercel, but widens a key the server publishes with, which is the less contained option of the two.
3. Shipping without realtime for now. The mobile client degrades to a foreground refresh plus a 30s focused-room poll, which is usable — just not live.

The fallback is left in place: it is correct for any deployment whose key does carry subscribe capability, and it fails loudly with a specific error code rather than silently, which is how we found this.

The client now stops reconnecting on 40160 rather than retrying a permanently-failing auth, so this costs nothing while undecided.

mrosberghaus:
@schaier-io a similar endpoint has been implemented in https://github.com/masumi-network/sokosumi/pull/3691. We already consider a native application if that’s your idea. The realtime integration will most likely break multiple times over the next couple of weeks. Are you willing to maintain it if you plan to build a client?

mrosberghaus (changes requested):
As described in earlier comments.

mrosberghaus:
Closing as superseded by https://github.com/masumi-network/sokosumi/pull/3691.

`POST /v1/realtime/ably-token` is already on main and covers the same need (bearer-auth Ably TokenRequest mint for non-browser clients). Web proxies it via `/api/ably/auth`.

Main's capability model is also the one we want going forward: per-membership room channels plus chat control, rather than a user-wide `chat_rooms:*:user_*` wildcard. A second mint route at `/realtime/token` would duplicate that surface and lag the room-scoped design.

Please use `/v1/realtime/ably-token` for mobile/native work.

mrosberghaus:
Superseded by #3691 — closing.

schaier-io:
Merged `main` in and stripped this branch back to nothing — agreed, `POST
/v1/realtime/ably-token` supersedes it entirely.

Removed from here: the `/realtime/token` route, its schema, the
subscribe-key Rest client and the publish-key signing fallback. `main`'s
required `ABLY_SUBSCRIBE_ONLY_KEY` and per-membership capability model
replace all of it, and the fallback was a workaround for a key that
main assumes is provisioned. The branch is now identical to `main`.

Mobile will use `/v1/realtime/ably-token`.

One consequence worth flagging, since it is not obvious from this PR:
**#3670 is closed too, and with it the only bearer credential a mobile
client can obtain.** `main`'s bearer middleware checks a coworker API
key, the orchestrator token, a Better Auth API key, and an OAuth access
token — the Better Auth `bearer()` plugin is not enabled, so a session
token from `/auth/sign-in/email` does not authenticate. `ably-token`
requires `requireOwnerUserContext`, which a coworker key cannot satisfy
either.

So a native client signing in with email and password has no sanctioned
way to call this endpoint on `main` today. The remaining options look
like: enable the bearer plugin (what #3670 did), register an OAuth
client and go through the OAuth flow, or have users paste an API key.
Happy to follow whichever you prefer — flagging it because closing both
PRs together leaves mobile with no path rather than a different one.

PR #3670, opened 2026-08-05
feat(core): accept Better Auth session tokens as bearer credentials

schaier-io:
## Why

Web auth is entirely cookie-based, and React Native has no usable cookie story. Today a mobile client has exactly two options: register an OAuth 2.1 + PKCE client by hand per app and do a browser round trip to sign in, or nothing.

This makes a plain email/password sign-in sufficient, so the app can have a direct login screen.

## What

Two changes:

1. **Enable Better Auth's `bearer` plugin.** Sign-in responses then carry a `set-auth-token` header, and `Authorization: Bearer <token>` is accepted in place of the session cookie. `requireSignature` stays at its default of `false` — the value handed back by `set-auth-token` is the unsigned session token, which is what a client stores and replays.
2. **Add a fifth check to the Core bearer chain** that resolves a session token. This is required, not optional: `authMiddleware` routes *any* `Authorization` header down the bearer path and never reaches the cookie session handler, so without it the token is rejected with 401.

The session check is deliberately **last** in the chain — it costs a session lookup, and coworker keys, the orchestrator token, API keys and OAuth tokens are all cheaper to reject.

Client flow becomes:

```
POST /auth/sign-in/email  { email, password }
→ read the `set-auth-token` response header
→ Authorization: Bearer <that value>  on /v1 requests
```

## Trade-off worth a reviewer's opinion

A session token is longer-lived than a 2-hour OAuth access token and there is no refresh-token rotation behind it, so the device holds a more powerful credential for longer. That is the thing the OAuth route buys you.

For a first-party mobile app I think that is the right trade — it removes per-app client registration and a browser round trip from the sign-in path — but it is a real change in posture and I would rather it be an explicit decision than a silent one. Happy to gate it behind a flag, or restrict it to specific routes, if you would prefer.

Note `magicLink` is already enabled, so a passwordless mobile flow is also possible on top of this without further server work.

## Testing

- `middleware/auth.test.ts`: two new cases — a session token presented as a bearer credential authenticates and yields the right context; and the session lookup is *not* reached when an earlier scheme (API key) resolves first.
- `lib/auth.test.ts`: asserts the bearer plugin is registered.
- Full core suite: **322 passed, 1 skipped, 0 failed**.
- `pnpm check && pnpm typecheck` green across all 9 workspaces.

One test-infra update is included: `lib/auth.test.ts` mocks `better-auth/plugins` wholesale, so the new export had to be added to that mock.

### DIFF
--- a/apps/core/src/lib/auth.test.ts
+++ b/apps/core/src/lib/auth.test.ts
@@ -14,6 +14,7 @@ interface AuthorizeReferenceConfig {
 
 const {
   adminPluginMock,
+  bearerPluginMock,
   apiKeyPluginMock,
   betterAuthMock,
   getBetterAuthPublicBaseUrlMock,
@@ -76,6 +77,7 @@ const {
 
   return {
     adminPluginMock: vi.fn(),
+    bearerPluginMock: vi.fn(),
     apiKeyPluginMock: vi.fn(),
     betterAuthMock: vi.fn(),
     getBetterAuthPublicBaseUrlMock: vi.fn(),
@@ -163,6 +165,7 @@ vi.mock("@better-auth/prisma-adapter", () => ({
 
 vi.mock("better-auth/plugins", () => ({
   admin: (...args: unknown[]) => adminPluginMock(...args),
+  bearer: (...args: unknown[]) => bearerPluginMock(...args),
   jwt: (...args: unknown[]) => jwtPluginMock(...args),
   lastLoginMethod: (...args: unknown[]) => lastLoginMethodPluginMock(...args),
   magicLink: (...args: unknown[]) => magicLinkPluginMock(...args),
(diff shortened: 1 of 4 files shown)


mrosberghaus:
@schaier-io why did you stack this on #3668, doesn't make sense to me 🤔

mrosberghaus:
**Recommendation: prefer OAuth over the bearer session path for mobile**

We already have a full first-party OAuth path (`oauthProvider`, auth code + PKCE, refresh tokens, scopes, consent, and Core already accepts OAuth access tokens on the bearer chain). For a first-party React Native app, that is the right model: one public client, system browser / deep link for sign-in, short-lived access token (2h) + refresh rotation in secure storage.

The bearer plugin trades that for a longer-lived session token with the full power of a browser session and no refresh rotation. That is a weaker posture for a convenience win (skip client registration + browser round trip) we can solve without changing the credential model.

**Suggestion:** do not merge session-as-bearer as the mobile auth plan. Use a first-party OAuth client + PKCE instead. If there is a narrow non-product need for bearer sessions later, that should be an explicit, gated exception — not the default mobile path.

PR #3709, opened 2026-08-07
feat(core): accept Better Auth session tokens as bearer credentials

schaier-io:
## What

Enables Better Auth's `bearer()` plugin and adds a fifth check to the bearer
middleware: a Better Auth **session token** presented as
`Authorization: Bearer <token>`.

Rebased on current `main`. +111 lines, all additive — the four existing checks
(coworker API key, orchestrator token, Better Auth API key, OAuth access token)
are untouched and still run first. A session token is only consulted after they
have all declined.

## Why

Any client that is not a browser has no way to authenticate against the v1 API
today without either registering an OAuth client by hand or asking a human to
paste an API key. Both are workable for a long-lived integration and awkward for
anything that just needs to sign in and make a few calls — end-to-end tests
against a preview being the obvious case, since they can already sign in with
email and password and get a session, and then cannot use it.

This closes that gap with the credential the sign-in already returns.

## Scope and risk

- No change to what a session *is* or how it is issued; this only accepts one
  that is already valid over a header instead of a cookie.
- No change to authorisation. A session-derived context resolves to the same
  `actor: "user"` as the cookie path, so every downstream `require*UserContext`
  guard behaves identically.
- Tests cover the new check and the plugin wiring. Full core suite: 2743 passed.

## Note

Draft, and I would like to keep it open for internal off chain sync testing, the previously closed PR
was superseded, but the two changes are independent and this one has no
replacement on `main`.


mrosberghaus:
As described in https://github.com/masumi-network/sokosumi/pull/3670#issuecomment-5205469904, please register an OAuth client for your use case at the moment. This adds a long-lived token that cannot be revoked by the user on Sokosumi.

mrosberghaus:
Closed after two weeks.

PR #3723, opened 2026-08-08
feat(core): store the devices a user wants to be notified on

schaier-io:
Stacked on #3709 — **base is `feat/core-session-bearer-auth`, not `main`.** Review that one first; this diff is only the push work.

## Why

Core has no push infrastructure at all today. No device tokens, no APNs or FCM credentials, nothing in the schema — I went looking for `apns`, `firebase-admin`, `expo-server-sdk`, `web-push` and any device-token model and found none. A native client can therefore only be notified while it is open and holding a realtime connection, which is exactly when a notification is least useful.

This is the storage half of fixing that: somewhere to put the address, and two endpoints for a client to manage it. **Nothing sends yet** — that is the next change in the stack, so this one can be reviewed on its own terms.

## What

- `PushDevice`, one row per `(user, token)`, with a migration.
- `POST /v1/notifications/devices` — register, idempotent.
- `DELETE /v1/notifications/devices` — unregister, by token.

## Decisions worth checking

**A token identifies an app install, not a device.** Reinstalling produces a new one and the old one goes stale. The provider reports that when a send is rejected, which is the only reliable moment to reap it — so this schema is built to be pruned by the sender rather than to be kept perfectly accurate on write. `lastSeenAt` is refreshed on every registration so an install that simply stopped launching can be pruned too.

**Registration is an upsert** because the app registers on every launch. The OS can reissue a token at any time, so "register whatever I have now" is the only contract a client can actually honour.

**Unregistering is scoped to the caller's own row.** Otherwise it becomes a way to silence somebody else's device by guessing their token.

**Both endpoints refuse coworker actors.** A coworker acting on a user's behalf has no business pointing that user's notifications at a device.

**The response never echoes the token back.** It describes the registration, not the address — returning it would add a way to read a token out that did not otherwise exist.

## Open question for the reviewer

A token can legitimately move between users on a shared device — sign out, sign in as someone else. Registration deliberately leaves the *other* user's row alone rather than deleting rows matching the token, because the client cannot prove the old install is gone and the provider is the only thing that knows. The consequence is a window where one token has two rows and the previous user gets one more notification before the send fails and reaps it. Happy to tighten that if you would rather trust the client.

## Verification

- `pnpm --filter core run typecheck`, `check` — clean, plus the monorepo-wide pre-commit hook.
- 9 new route tests, including both coworker-refusal cases and an assertion that the token never appears in a response.
- Full core suite: 2745 passing. Two files timed out on hooks under load (`oauth-issuer-metadata`, various `organizations/[id]/…`); different files each run, and all pass in isolation. Unrelated to this change.
- **Not** verified against a real database — the migration is hand-written to match the schema and has not been applied anywhere.

### DIFF
--- a/apps/core/src/routes/v1/notifications/devices/delete.ts
+++ b/apps/core/src/routes/v1/notifications/devices/delete.ts
@@ -1,0 +1,61 @@
+import { createRoute, z } from "@hono/zod-openapi";
+
+import { jsonErrorResponse, jsonSuccessResponse } from "@/helpers/openapi";
+import { ok } from "@/helpers/response";
+import prisma from "@/lib/db/prisma";
+import type { OpenAPIHonoWithAuth } from "@/lib/hono";
+import { requireOwnerUserContext } from "@/middleware/auth";
+import { pushDeviceTokenSchema } from "@/schemas/push-device.schema";
+
+const responseSchema = z
+  .object({
+    deleted: z.boolean().openapi({ example: true }),
+  })
+  .openapi("UnregisterPushDeviceResponse");
+
+const route = createRoute({
+  method: "delete",
+  path: "/devices",
+  description:
+    "Stop sending push notifications to this app install. Called on sign-out, so the next person to use the device does not receive the last one's notifications.",
+  tags: ["Notifications"],
+  request: {
+    body: {
+      content: {
+        "application/json": {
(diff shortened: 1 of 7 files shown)


mrosberghaus:
Closed after two weeks.

PR #3724, opened 2026-08-08
feat(core): send a notification to the devices it was registered for

schaier-io:
Second in the push stack. **Base is `feat/core-push-device-tokens` (#3723), which is itself based on #3709** — review those first; this diff is only the sending half.

- #3709 — session bearer auth
- #3723 — device token storage
- **this PR** — sending

## Why Expo

The alternative is holding an Apple signing key and a Google service account in Core and reimplementing two transports to say the same sentence. Expo push tokens are self-authenticating, so this works with **no credential configured at all**; `EXPO_ACCESS_TOKEN` is optional and only tightens who may send to a token. Happy to be overruled if you would rather Core spoke to APNs and FCM directly, but it is a lot of surface for the same outcome.

## Where it hangs off

`createNotification`, right beside the existing Ably publish, and it shares that function's contract: **it cannot fail the caller.** A notification that exists in the feed but did not reach a phone is a degraded delivery. One that was never written because a push provider was down is lost work.

It fires only on first creation. `createNotification` is idempotent by `(user, kind, reference, event, messageKey)`, and a duplicate emit must not buzz the phone again for something already delivered.

## Things worth a close look

**It takes the caller's prisma client, not the singleton.** This one caught me — I wrote it against the singleton first and it hung the existing `notifications.test.ts`. The real problem was not the test: `createNotification` is often running inside a transaction, so reading devices on a different connection misses uncommitted state, and reaping a dead token outside the transaction survives a rollback that undid the notification it was sent for.

**Reaping trusts only the provider.** `DeviceNotRegistered` means the install is gone and the row is deleted. Anything else is transient — a failed send is not a reason to forget where someone lives. A short response from Expo is refused outright rather than trusted, because tickets correspond positionally and a shifted index would reap live devices.

**Chat pushes carry no message text.** A lock screen is visible to whoever is holding the phone, and the person who wrote it did not agree to that. Title is the sender; body says a message arrived.

## The limitation I want you to see

The copy in `lib/push/copy.ts` is **English only**.

Core owns all seven message keys — it emits them — but the human strings live in `apps/web/messages/*.json`, and a push has to carry a rendered string at send time. Core has no locale infrastructure and no notion of a user's language, so there is nothing to select a translation with. I put a small map in Core because there was nowhere else for it, not because it belongs there.

Every push therefore also carries its `messageKey` and `messageParams` in `data`, so a client that *does* know the user's language can render its own copy and treat the English as the fallback it is. If you would rather Core learned about locales, or the catalogue moved somewhere shared, say so — this is the part I would most expect to be sent back.

An unknown key sends nothing rather than pushing a raw `Notifications.Some.newKey` to a lock screen. That means a new key added elsewhere silently does not push until copy is added here, which is a tradeoff worth knowing about.

## Verification

- 25 unit tests across copy, the Expo client, and dispatch — including opt-out, no devices, reaping, transient-failure-does-not-reap, short-response refusal, invalid JSON params, and that a push failure never reaches the caller.
- Affected suites: 669 passing.
- Monorepo-wide `check` and `typecheck` clean via the pre-commit hook.
- **Not** verified end to end. Nothing has been sent to a real device, no Expo project is configured, and the client half does not exist yet.

### DIFF
--- a/apps/core/src/config/env.ts
+++ b/apps/core/src/config/env.ts
@@ -187,6 +187,10 @@ const envSchema = z.object({
   /** Subscribe-only key used to mint client TokenRequests (SOK-741). */
   ABLY_SUBSCRIBE_ONLY_KEY: z.string().min(1),
 
+  // Push notifications. Expo tokens are self-authenticating, so sending works
+  // without this; setting it restricts who may send to them.
+  EXPO_ACCESS_TOKEN: z.string().min(1).optional(),
+
   // Optional outbound webhooks
   WEBHOOK_USER_CREATED: z.url().optional(),
   WEBHOOK_USER_UPDATED: z.url().optional(),
--- a/apps/core/src/helpers/notifications.ts
+++ b/apps/core/src/helpers/notifications.ts
@@ -8,6 +8,7 @@ import type {
 import { isPrismaUniqueViolation } from "@/helpers/prisma";
 import { publishNotificationEvent } from "@/lib/ably/publish";
 import prisma from "@/lib/db/prisma";
+import { dispatchPushNotification } from "@/lib/push/dispatch";
 
 export interface CreateNotificationInput {
   userId: string;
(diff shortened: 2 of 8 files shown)


mrosberghaus:
Closed after two weeks.

PR #3775, opened 2026-08-11
feat(database): add task x402 payment records

schaier-io:
Stack 2/6 of the x402/EVM build — the data model. No user-facing behaviour yet; this PR only makes the records possible and keeps account deletion correct once they exist.

- `TaskX402PaymentStatus` (`PENDING | VERIFIED | FAILED | REFUNDED`, terminal at VERIFIED) and `TaskX402Payment`, with the dedupe unique `[taskId, idempotencyKey]`, the aggregation index `[agentId, status]`, and the phased-settlement fields plus `[status, validBefore]`
- `TaskX402PaymentAction` — the operator audit table. Deliberately **FK-free and append-only** so a record of an operator moving real on-chain money survives the hard-deletion of the payment row during account deletion
- `prepareTasksForUserDeletion` extended: a PENDING payment blocks deletion with a clean 400 across all three RESTRICT branches, and everything else is swept before the task cascade

## Migration

`20260819130000_task_x402_payment` (formerly `20260811130000`), fully idempotent, timestamped after main's `20260818120000_better_auth_1_7_account_identity`. It has been **amended in place** several times during this stack rather than followed by new migrations — safe because it has never been applied to production. A preview database that already ran the older name should be reset.

Validated against real Postgres: `migrate deploy` from zero, a hand re-apply proving idempotency, convergence from all four earlier shapes of the file (each produces a byte-identical `pg_dump -s`), and `migrate diff` showing no x402 drift.

- **The audit row was an unresolvable pointer.** It stored only `paymentId`, `action`, `operatorId`, `reason` — so once account deletion removed the payment, the surviving record of an operator refunding real money named no amount, asset, chain, or subject. Seven denormalized money-fact columns added, all NOT NULL, all plain non-FK (an FK would cascade or block and defeat the point).
- **The nonce-replay unique had a NULL hole.** `payerAddress` is nullable and Postgres treats NULLs as distinct, while the partial predicate gated only on `payloadNonce`, so two rows sharing a nonce inserted cleanly — proven against the real index. Now fenced by `CHECK (("payerAddress" IS NULL) = ("payloadNonce" IS NULL))`.
- **The terminal sweep enumerated statuses.** Adding the planned `EXPIRED_UNUSED` would have matched neither the sweep nor the PENDING guard, leaving a row whose RESTRICT `taskId` 500s the entire account deletion. Inverted to `status: { not: PENDING }`.
- **The enum swallowed itself on re-apply**, unlike every other object in the migration, which would have left dev databases on a stale enum after the next amendment. Now restated with `ALTER TYPE … ADD VALUE IF NOT EXISTS`, with a proviso in the file: this is transaction-safe only while no statement uses a newly added member.
- `reason` bounded to `VarChar(500)`, and a Sentry page when the sweep removes a payment charged to a different user than the one being deleted.

## Verification

`pnpm --filter @sokosumi/database test` (248), `pnpm --filter core test` (3588 at the top of the stack), `pnpm typecheck`, `pnpm check` — all clean. Migration validated against local Postgres as described above.

Stacked on #3773.

### DIFF
--- a/apps/core/src/helpers/deletion-evaluate.test.ts
+++ b/apps/core/src/helpers/deletion-evaluate.test.ts
@@ -1,3 +1,4 @@
+import { TaskX402PaymentStatus } from "@sokosumi/database";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import {
@@ -9,11 +10,13 @@ import {
 
 const {
   taskPaymentClaimFindFirstMock,
+  taskX402PaymentFindFirstMock,
   getMembersByOrganizationIdMock,
   isLastWorkspaceMock,
   captureMessageMock,
 } = vi.hoisted(() => ({
   taskPaymentClaimFindFirstMock: vi.fn(),
+  taskX402PaymentFindFirstMock: vi.fn(),
   getMembersByOrganizationIdMock: vi.fn(),
   isLastWorkspaceMock: vi.fn(),
   captureMessageMock: vi.fn(),
@@ -45,6 +48,9 @@ function createPrisma() {
     taskPaymentClaim: {
       findFirst: taskPaymentClaimFindFirstMock,
     },
+    taskX402Payment: {
+      findFirst: taskX402PaymentFindFirstMock,
+    },
   };
 }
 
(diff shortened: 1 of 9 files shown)


schaier-io:
@coderabbitai review

PR #3778, opened 2026-08-11
feat(core): add x402 pricing, dialect, and readiness helpers

schaier-io:
Stack 3/6 of the x402/EVM build — the helpers the pay path is built from. Still no endpoint; this PR is pricing, the 402 dialect normalizer, and buy-side readiness.

- **CAIP-19 pricing** — `buildCaip19AssetKey` / `parseCaip19AssetKey` plus `calculateCentsFromX402Amount` (per whole token, ceiling division, `MIN_CHARGEABLE_CREDITS` floor, fail-closed on an unpriced asset), with a convention fence so a CAIP-19 `CreditCost` row can never be read per-base-unit by the Cardano paths or vice versa. CAIP-2 chain ids are the canonical integer form (`eip155:0` or no leading zeros, ≤32 digits) so `eip155:08453` cannot mint a second unit spelling for chain 8453
- **402 dialect normalizer** — v1 (`maxAmountRequired`, plain network names) and v2 (base64 `PAYMENT-REQUIRED` header, CAIP-2 ids) folded into the node's v2 shape, failing loud on unknown networks, conflicting amounts or resource URLs, and more than 20 `accepts` entries
- **Buy-side readiness** — composed from `/x402/networks/available` and `/x402/budgets`, cached last-known-value in `SyncMetadata`, gated by a per-environment EVM network allowlist, with ready pairs carrying the `evmWalletId` the pay route needs

A 402 demand is attacker-controlled data returned by the agent endpoint, so most of this file is about not trusting it.

Five rounds ran here. The ones that mattered:

- **Asset decimals came from the agent.** `cents = ceil(amount × centsPerUnit / 10^decimals)` scales the charge *inversely*, and `decimals` was read from the agent's own registry entry with only a 0–255 range check. Registering USDC on Base with `decimals: 18` (true value 6) floored the charge at `MIN_CHARGEABLE_CREDITS` while the managed wallet signed away a real USDC — repeatable per call. The node already publishes the truth and Soko was discarding it; `X402ReadySource` now carries the node's `defaultAssetDecimals`, re-validated on read and failing closed on anything unusable.
- **One unsupported or incomplete option poisoned a whole 402.** `accepts` is a list of *options*, but the hardening added in earlier rounds (scheme allowlist, transfer-method pin, timeout cap) was enforced payload-wide, so a 402 offering a payable `exact`/Base/USDC option *plus* anything unsupported was refused entirely. This PR's own research records `batch-settlement` offered alongside `exact` by live Base-mainnet resources. Selection is now per-entry: unsupported *values* and omitted identity fields (`scheme`, `network`, `asset`, `payTo`, `maxTimeoutSeconds`) are skipped with a reason; the payload fails only when nothing survives. A field of the wrong *type* (numeric `scheme`, fractional timeout) is still a malformed 402.
- **CAIP-2 chain ids accepted leading zeros.** `/^eip155:\d+$/` treated `eip155:08453` as canonical, minting a second `CreditCost` spelling for chain 8453. Pattern is now CAIP-2's integer form: `0` or no leading zeros, at most 32 digits.
- **`maxTimeoutSeconds` was unbounded.** zod's `.int()` caps nothing but safe-integer range, so a hostile 402 could set `9007199254740991` and the derived `validBefore` would make the signed authorization valid for ~285 million years — a permanent live claim on a funded wallet. Capped at 3600s (live listings use 60–3600, so a tighter cap would reject real demands).
- **The budgets read failed open.** `GET /x402/budgets` is admin-gated and returns every key's rows unless filtered, while `/x402/pay` spends only the calling key's. The filter was requested but the response never checked, so a node that renamed or ignored the parameter would mark pairs ready against a wallet Soko cannot spend from. Now re-filtered on `apiKeyId`.
- **`json-canonicalize` is not canonical** when the input carries a `toJSON` key — it short-circuits and falls back to unsorted `JSON.stringify`, and the key need not be callable. Replaced with a local sorted-key serializer at the equality site.
- **Nested `__proto__` was relayed verbatim to the payment node.** Top-level was eaten only incidentally by zod's unknown-key assignment; one level down it survived into the `/x402/pay` body. Now recursively stripped before parsing.
- Plus: EVM address validation so the bytes validated are the bytes forwarded; `scheme` and `extra.assetTransferMethod` pinned; length caps and echo truncation (a 200k-char `network` produced a 200,086-char error into logs and Sentry); a 256 KiB bound before base64 decode; and the network allowlist actually wired in — it existed, was well tested, and was called by nothing.

## Verification

`pnpm --filter @sokosumi/masumi test` (356), `pnpm --filter core test` (3588 at the top of the stack), `pnpm typecheck`, `pnpm check` — all clean.

Stacked on #3775.


schaier-io:
@coderabbitai review

schaier-io:
@coderabbitai review

mrosberghaus:
@coderabbit review

mrosberghaus:
@coderabbit review

PR #3779, opened 2026-08-11
feat(core): list cardano and x402 agents on GET /v1/agents

schaier-io:
Stack 4/6 of the x402/EVM build — the listing surface.

**`GET /v1/agents`** is one public catalog. Items are a discriminated union on `kind`: `cardano` (MIP-003 hire) or `x402` (EVM pay). `?kind=` filters (repeatable and comma-separated); omit returns both rails. **`GET /v1/agents/x402` is gone.**

x402 items still obey **listed ⇒ payable**, enforced per-agent fail-closed: any malformed, unpriced, unready, or self-contradictory payment source drops the whole agent rather than advertising a partial one.

Gates, all of which must pass: X402 entry type, ONLINE, `isShown`, `scheme` is `exact`, `payTo` is a well-formed EVM address, the network is in the per-environment allowlist, the asset is CAIP-19-priced, and the `(network, asset)` pair is buy-side ready. Payment sources carry the native amount plus converted credits.

Pay stays coworker-scoped on `POST /v1/tasks/{id}/x402-payments` (#3784). Web `/agents` stays Coworkers-only; x402, like classic agents, is API-only. Web loaders request `kind=cardano`.

Also here: readiness sync wired into `/sync/agents` (10s timeout, failure-isolated, deliberately no cursor reset — nothing readiness-dependent is baked into agent rows), and `agent.ts` brought under the 750-line ceiling by extracting the metadata-override getters.

- **Duplicate amount rows made listing and paying disagree.** Ingestion explicitly permits the same asset twice within one source, the listing emitted one entry per row, and the pay side resolves the triple with a `find` — with neither query ordering the relation, so "first" was Postgres heap order. An agent could be advertised at two prices for one recipient, and a coworker could get a 422 *after* the agent call was already paid for in work. Conflicting duplicates now drop the agent; identical ones collapse; both queries order deterministically.
- **A one-space-padded asset bypassed that fix.** Registry ingestion lowercases units but never trims, while every *gate* trims — so `" 0x036c…"` passed readiness and pricing yet produced a distinct dedupe key, re-opening the conflicting-price case and, in the single-source case, listing an agent the pay path can never match. Now normalized once at the source.
- **The pricing scale came from the agent.** Same inverse-scaling mischarge described in #3778; this branch now takes `decimals` from the node-published ready pair and keeps the registry row's own check as a sanity gate only.
- **Three tests looked like coverage and were not.** The environment-separation test never reached `findMany` at all, so it passed with the gate it named deleted; a second passed with the `payTo` case-fold removed; a third was masked by an identical tail guard. All three now discriminate, verified by mutation.
- Plus: cursor pagination with a count in one snapshot (the query previously loaded every matching agent with all sources and amounts), a `scheme` fence, `Sentry.captureException` on a readiness throw that was degrading to `console.warn` while the stale cache kept serving "ready", drop-reason tallying so an empty listing is diagnosable, and a fix for client-driven log volume — `?limit=1` alone let any authenticated coworker loop one `warn` per request.

Auth for this route is now the same as the existing Cardano catalog: **public**. The coworker-only gate on the deleted `/v1/agents/x402` path was leftover from an earlier coworker-loop design, not because the list is per-user.

## Verification

`pnpm --filter core test`, `pnpm typecheck`, `pnpm check` — all clean. Adds `PricingType.DYNAMIC` (`20260819140000_add_dynamic_pricing_type`) and a one-shot registry replay (`20260819150000_replay_agent_sync_for_dynamic_pricing`). Web Core client regeneration lands in #3785 (this PR changes the OpenAPI union; web does not render x402 cards). Gallery loaders send `kind=cardano` even before that snapshot.

Stacked on #3778.


schaier-io:
@coderabbitai review

mrosberghaus:
Fixed **Test Core**: `workspaceMiddleware` now includes `authenticationMethod: "session"` on user `authContext`, but the "personal workspace is missing" fixture still omitted it.

Restacked `#3784` and `#3785` onto this commit so the same failure clears there too.

PR #3784, opened 2026-08-11
feat(core): add the coworker x402 pay endpoint

schaier-io:
Stack 5/6 — **the money path.** Review this one hardest.

**`POST /v1/tasks/{taskId}/x402-payments`** — a coworker forwards the 402 it received from an agent endpoint; Soko verifies the demand against the agent's registry-declared payment sources, charges credits, has the Masumi node sign an EIP-3009 `transferWithAuthorization` with a managed wallet, and returns the `X-PAYMENT` header. **That header is a bearer instrument** until `validBefore` — anyone holding it can settle the transfer — so it is treated as a credential in logs, errors, Sentry, and at rest.

Shape of the flow:
- Coworker-context authz and task collaboration checked before any read; `idempotencyKey` bounded and used as the `[taskId, idempotencyKey]` dedupe unique (the node has no idempotency of its own)
- The demand is verified against the agent as listed on public `GET /v1/agents?kind=x402` (same fail-closed gates as #3779), priced, and fenced by an optional caller-supplied `maxCredits` — all **before** any charge
- Debit, credit-bearing task event, and the PENDING record commit in **one serializable transaction**, then the node is called
- **Refund doctrine (ticket 011 Q1):** a real non-200 carrying the node's own error envelope means no header was issued, so the payment is provably unsettleable and a synchronous refund is safe. Anything ambiguous — malformed 200, timeout, transport error, gateway status, or a header that cannot settle — holds PENDING and **never** refunds
- Every status transition is an atomic claim (`updateMany` with a status predicate, then branch on `count`); VERIFIED is terminal

## Migrations

All restamped after main's `20260819150000_replay_agent_sync_for_dynamic_pricing` (#3779) so they apply after the current main tip:

- `20260819160000_task_x402_payment_header`
- `20260819170000_task_x402_payment_sign_attempt_count`
- `20260819180000_task_x402_payment_processing_lease` — the `processingAt` sign lease
- `20260819190000_task_x402_payment_header_purge_index` — partial index `WHERE xPaymentHeader IS NOT NULL`
- `20260819200000_task_x402_payment_refund_kind`
- `20260819210000_task_x402_payment_demand_fingerprint`
- `20260819220000_task_x402_payment_replay_key_canonical`
- `20260819230000_credit_cost_unit_canonical`
- `20260819240000_task_x402_payment_sign_risk_expiry`
- `20260819250000_task_x402_payment_action_ledger_snapshot`
- `20260819260000_task_x402_payment_full_convergence`
- `20260819270000_task_x402_payment_durable_failures`
- `20260819280000_task_x402_payment_action_rollout_guard`

Both validated against real Postgres (deploy from zero, idempotent re-apply, no x402 drift). The purge plan was measured on 200k rows: the partial index is chosen and the batch reads 404 buffers in 2.4 ms — a plain `@@index([validBefore])` was tried first and the planner declined it, because `validBefore <= now()` matches essentially every historical row.

Four passes. In rough order of severity:

- **The "signed tuple" check validated the wrong thing.** It compared the node's own self-reported summary scalars, so a compromised node could return truthful ones beside a header signing a different `to`/`value`. The check now base64-decodes the `X-PAYMENT` header itself and asserts the EIP-3009 `to`/`value`/`from` against the charged demand — the header is the bearer instrument; the summary is only the node's rendering of it.
- **A zero-padded node amount caused a second debit.** The node's `amount` was stored verbatim after a *BigInt-equal* assert, while replay compares it as a *string* — so `"0250000"` for a charged `"250000"` made the paid-for header unreachable by any same-key replay, and the resulting "use a new idempotencyKey" 409 minted a fresh charge. Exactly the case idempotency exists for.
- **The header was never checked for whether it can settle.** Nothing read `payload.signature`, `validAfter`, or the envelope `scheme`/`network`, and nothing rejected an already-past `validBefore`. A header with no signature at all stored VERIFIED — charging the user for a dead instrument on a terminal row the automated refund explicitly refuses.
- **Concurrent same-key requests could refund a charge the node really signed.** Two calls could reach the node; the loser's refund landed while the winner's real authorization was discarded. A `processingAt` sign lease now 409s the second request instead of racing.
- **`failureReason` leaked the raw node message** that the sanitized 502 deliberately withholds — one extra request retrieved it through the replay 409 body, and a test asserted that leak as correct. Now a coded reason; the raw text stays in Sentry.
- Plus: gateway statuses (502/503/504 from a proxy) no longer classified as refusals, since they can occur *after* the node signed; a header purge past `validBefore`, since a VERIFIED row was otherwise a permanent stored credential; a pre-charge floor on `maxTimeoutSeconds` so a 1-second window cannot strand a charge; `bodyLimit`; parsing and config reads hoisted out of the serializable transaction; and `narrowToChosenRequirement` wired so the node receives exactly one `accepts` entry and cannot pick a different `payTo`.

The final confirmation pass ran 34 mutations and caught 33; the survivor was a metrics-only counter, since fixed. It also measured the safety margin between the expiry fence and the purge cutoff directly — 3,540,000 ms — proving a live header cannot be nulled early.

## Review follow-ups

- Fresh `signAttemptCount` increments only when `payX402` is actually invoked. A withheld dispatch stall does not spend the first-attempt auto-refund slot.
- Fixed demand is a **ceiling** (`demanded <= advertised`), not equality. Coworker JSON is `paymentHeader` `{ x402Version, name, value }`; the DB/node field stays `xPaymentHeader`. Spec updated.
- `maxCredits` is `.positive().finite()` on this route **and** job-hire. JSON `1e400` (`Infinity`) is 422 before debit.

## Verification

`pnpm --filter core test`, `pnpm --filter @sokosumi/masumi test`, `pnpm typecheck`, `pnpm check`. Migrations validated against local Postgres.

> [!WARNING]
> **Do not production-cutover this PR without #3785 (x402-6-admin).** Trusted EIP-712 domains (`X402_TRUSTED_EXACT_EVM_DOMAINS`) and the CreditCost duplicate-unit 500 already live here (and on `main` from #3779). Stopping here still ships a pay route with no operator HTTP/UI for wedged PENDING (`resolve` / `refundKind` rollup). Merge through #3785 for that surface.

Base is `main` (#3779 merged).


schaier-io:
@coderabbitai review

PR #3785, opened 2026-08-11
feat(core): add the admin x402 payment surface

schaier-io:
Stack 6/6 (final) — the platform-admin surface. Operates on other users' money records, so authorization and information disclosure are the primary risks.

## Review follow-ups

- Resolve 409 OpenAPI now declares `retryAfter` / `retryAfterSeconds` (`AdminTaskX402ResolveConflictResponse`); generated client regenerated.
- Web money-action tests cover Core 409 → `BAD_INPUT` (retry message kept) and 404 → `NOT_FOUND`.
- Mutations use `requireInteractiveAdminAuthContext` (stricter than the in-handler `requireAdminAuthContext` on list/aggregate).
- Resolve trigger is disabled while `signRiskExpiresAt` is in the future; lease-in-flight still 409s (processingAt is not on the list DTO).
- Rollup table now shows live `pending` / `verified` / `failed` / `refunded` next to the ranking columns.
- Coded `failureReason` (`node_refused_payload` / `node_refused_operational`) is shown on the payment row.
- Admin money actions map Core 401/403 through `toCoreApiActionError`.
- Refund 409 now has `kind` (`already_refunded` / `not_refundable`) and `AdminTaskX402RefundConflictResponse`.
- Native `<select>` on the GET filter form is kept on purpose (Radix Select does not submit `name=`).

Every route is double-gated: the parent `/admin` router's `requireAdmin` **and** an in-handler `requireAdminAuthContext`, so a handler remounted without its parent still refuses. Four tests mount each handler unguarded and assert 403 with no database read.

- **`GET /v1/admin/task-x402-payments`** — list with filters. `xPaymentHeader` and the raw signed payload are excluded at the Prisma `select` level (an allowlist, not an `include`), with a test that fails if the header is added back. Cursor-paginated; `creditsCharged` surfaces the magnitude of a debit stored negative.
- **`GET /v1/admin/task-x402-payments/aggregate`** — per-agent rollup feeding the whitelist-disable decision.
- **`POST /v1/admin/task-x402-payments/{id}/refund`** — operator goodwill refund on a VERIFIED row.
- **`POST /v1/admin/task-x402-payments/{id}/resolve`** — new; see below.

## The `resolve` lever

A PENDING payment blocks account deletion, and nothing could ever clear one: the coworker replay path exhausts a 5-attempt cap that never resets, the settlement reconciler does not exist in this stack, and the admin refund 409s on PENDING. So the owner's `DELETE /user` failed permanently — a GDPR erasure obligation with no operator remedy — and the guard's "contact support" message was false.

`resolve` claims PENDING → REFUNDED atomically, mints the compensating refund, and writes an audit row. Refunding a PENDING row is safe by the stack's own doctrine: in every path that leaves a row PENDING the coworker never received a usable header, so the authorization is unsettleable by the only party who could present it. It refuses a row whose sign lease is still held — inside the claim predicate, not as a read-then-check — and returns a retry time, because a leased row is not wedged and claiming it would manufacture the "signed after close" state the lease exists to prevent.

## Migration

`refundKind` lands in #3784 as `20260819200000_task_x402_payment_refund_kind` (formerly `20260812100000`) — a `TaskX402PaymentRefundKind` enum and a nullable `refundKind` column. Idempotent, no default, no backfill. This PR has no new migration.

- **`resolve` corrupted the aggregate's headline metric.** `goodwillRefundCount` counts REFUNDED rows and is the primary sort key, but `resolve` also writes REFUNDED — so an operator clearing 20 wedged rows would rank a perfectly healthy agent as the worst quality offender, and a hostile coworker could drive that deliberately against a competitor. Fixed with the explicit `refundKind` enum, written inside the same UPDATE as the refund by a now-mandatory parameter on the sole minting point, rather than the implicit `attemptId != null` discriminator that a future change could silently invert.
- **The claim predicates omitted `refundTransactionId: null`, and the documented safety net did not exist.** The reasoning had been that an anomalous row would fail inside `attachCompensatingRefund` and roll back — but a probe against real Postgres with the pinned Prisma showed the nested `create` on an already-occupied to-one relation *succeeds*, re-pointing the FK and orphaning the prior refund. So the state was unreachable, not guarded. Now guarded explicitly in all three claims, including the automated refusal path that had the same gap.
- The audit row's seven denormalized money facts are populated and asserted as a whole object; `reason` is sliced to the column's 500-char bound rather than the 2000-char `failureReason` bound, which would have 500'd the route.
- Plus filter length bounds, a `required: true` body guard that had no test, and three coverage gaps the confirmation pass found by mutation (an untested `slice(0, take)` trim, an unasserted filtered `total`, and an uncovered `caip2Network` aggregate filter).

The confirmation pass ran 32 mutations, caught 29, and regenerated the web Core client from scratch to verify the committed output is byte-identical and additive only. That regen also picks up the unified `GET /v1/agents` discriminated union from #3779. Web `/agents` stays Coworkers-only (`kind=cardano`); x402 is API-only.

## Verification

`pnpm --filter core test` (3588), `pnpm typecheck`, `pnpm check` — all clean. Migration validated against local Postgres.

Stacked on #3784. **Completes the stack.** Filed as follow-ups: the phased-settlement reconciler, a shared readiness/refund helper extraction, and a `schema.prisma` multi-file split.


PatrickTobler:
## End-to-end test of the x402 payment surface — it works, plus 5 integration issues

I built a standalone x402 seller endpoint and drove the full buyer flow against this branch's preview. **The flow completes and settles on-chain.** Writing up exactly what I did and where I got stuck, since most of the friction was in naming/error-reporting rather than design.

### Result

```
1) unpaid call    HTTP 402, PAYMENT-REQUIRED: 1556 chars
2) pay            HTTP 200, payment header: 1976 chars
3) paid call      HTTP 200
   RESULT: {"pong":true,"paidAt":"2026-08-16T17:26:04.278Z"}
   settled: True | tx: 0x24a92dcaa1b7a3535786a9de87db7055e811fecaf50fda82242705f12604ed77
```

Verified on Base Sepolia: tx status `SUCCESS`, block 45566197, and 0.0001 USDC landed at the `payTo` address. Reproduced 4× with distinct settlement txs.

### Setup under test

| | |
|---|---|
| Seller endpoint | Express + `@x402/express` + `@x402/core` v2.22, public `x402.org/facilitator`, no CDP |
| Route | `GET /ping`, `$0.0001` = 100 atomic units testnet USDC, `eip155:84532` |
| Discovery | `/.well-known/x402.json`, bazaar resource schema |
| Agent | registered in Masumi Preprod registry, `x402ResourcesUrl` → the manifest |
| Preview | `sokosumi-core-preprod-git-x402-6-admin` |

### The flow that works

1. `GET /v1/agents/x402` → agent listed, `isPayable: true`
2. `GET /ping` → `402` + `PAYMENT-REQUIRED`
3. `POST /v1/tasks/{id}/x402-payments` with `{idempotencyKey, agentId, maxCredits, paymentRequired}` → `200` + `xPaymentHeader`
4. `GET /ping` with **`PAYMENT-SIGNATURE: <header>`** → `200` + result + `PAYMENT-RESPONSE`

---

## Issues, most costly first

### 1. `xPaymentHeader` is named after the wrong protocol version 🔴

The field returns an `x402Version: 2` payload, but is named after the **v1** header. Per `@x402/core`:

```js
encodePaymentSignatureHeader(paymentPayload) {
  switch (paymentPayload.x402Version) {
    case 2: return { "PAYMENT-SIGNATURE": ... };
    case 1: return { "X-PAYMENT": ... };
  }
}
```

Sending it as `X-PAYMENT` produces a **silent `402`** — indistinguishable from never having paid. No error, no hint. This cost more time than everything else combined, and the name actively points integrators the wrong way.

**Suggested fix:** return `{"headers": {"PAYMENT-SIGNATURE": "..."}}`, or rename to `paymentHeader` and add a `paymentHeaderName` field. Anything that makes the correct name discoverable from the response.

### 2. Address casing silently rejects the SDK's default output 🔴

The payment node normalizes `asset` / `payTo` to **lowercase** in the echoed `accepted` block. But `@x402/core` matches that block against the seller's requirements with a strict deep-equality check:

```js
function paymentRequirementsMatchAccepted(required, accepted) {
  const { extra: requiredExtra, ...requiredCore } = required;
  const { extra: acceptedExtra, ...acceptedCore } = accepted;
  if (!deepEqual(requiredCore, acceptedCore)) return false;
```

The standard x402 server path (`price: "$0.0001"`) emits **EIP-55 checksummed** addresses via `getDefaultAsset()`. Checksummed vs lowercase → `deepEqual` fails → "No matching payment requirements" → `402`, again with no diagnostic.

I only got past this by passing an `AssetAmount` instead of a price string so I could force lowercase myself:

```ts
price: {
  amount: "100",
  asset: assetAddress.toLowerCase(),
  extra: { name: asset.name, version: asset.version }, // dropped if omitted
}
```

**This means a seller using the documented default setup will fail against Sokosumi with no error message.** Worth normalizing addresses case-insensitively before comparison (or raising it upstream with x402).

### 3. Error messages get *less* accurate when you add context 🟠

Sending `X-Context-User-Id` to `POST /tasks/{id}/x402-payments` returns:

```
403 "Coworker agent authentication required"
```

…while using a valid coworker API key. The real error only appeared after **removing** the header:

```
403 {"message": "Parked tasks cannot be modified until vendor workspace access is granted", "kind": "task_parked"}
```

The first message sent me looking at authentication for a while when the actual problem was workspace state.

### 4. Payment preconditions are opaque 🟠

To pay one URL the coworker must be whitelisted, hold workspace access, and the task must exist, be `READY`, and be assigned to that coworker. Each failure surfaces differently and none names the missing precondition:

| State | Response |
|---|---|
| coworker not whitelisted | `404 "Coworker is not usable in this workspace"` |
| no workspace access | `403 task_parked` |
| task is `DRAFT` | `404 "Task not found"` ← misleading, the task exists |
| all good | `200` |

`404 Task not found` for a task that demonstrably exists is the worst of these.

### 5. No `EXTENSION-RESPONSES` on settlement 🟡

My route declares the bazaar extension via `declareDiscoveryExtension()`, and it round-trips correctly (present in the `402`, echoed in the `PaymentPayload`). But the settlement response carried no `EXTENSION-RESPONSES` header, so `bazaar.status` (`success` / `processing` / `rejected`) never came back. Per the bazaar spec that's where cataloging confirms. Not sure if unimplemented or intentionally skipped.

---

## What I think is well designed 👍

- **`maxCredits`** — the docs are right that the 402 is authored by the counterparty being paid; a caller-side ceiling is exactly the correct instinct and plenty of payment APIs miss it.
- **Idempotency semantics** — failed attempts don't consume the key, `502` distinguishes "refused, refunded" from "unknown, charge held, retry same key". Clear thinking about partial failure.
- **Verify-before-debit** — every `422`/`400` charged nothing, which let me safely run 5 probe variants to isolate a field mismatch. Genuinely useful property.
- **The credits abstraction** — sellers need only an address, buyers never touch a wallet.

## Useful diagnostic

When a replay `402`s, this isolates payment-vs-plumbing immediately — `isValid: true` means the payment is fine and the problem is on the resource-server side:

```bash
curl -s -X POST https://x402.org/facilitator/verify \
  -H 'Content-Type: application/json' \
  -d '{"x402Version":2,"paymentPayload":<decoded payment>,"paymentRequirements":<accepts[0] from the 402>}'
```

That's what told me issues 1 and 2 weren't payment problems at all.

---

Happy to share the seller endpoint + repro script if useful. The two I'd prioritise are **#1** (one-line fix, saves every integrator hours) and **#2** (currently breaks the default SDK path silently).

*Tested against `sokosumi-core-preprod-git-x402-6-admin` on 2026-08-15/16. Credentials redacted.*

schaier-io:
Implemented and pushed the review fixes in b82c08b7c.

1. **Protocol/version handling:** Core now supports x402 v1 and v2 end to end. The response includes `paymentHeader: { x402Version, name, value }`, using `X-PAYMENT` for v1 and `PAYMENT-SIGNATURE` for v2. `xPaymentHeader` remains as a deprecated compatibility field.
2. **Address casing / strict SDK replay:** inbound requirements retain both a canonical form for verification/charging and the validated original server spelling. After signing, Core restores the original v2 `accepted` object in the background before persisting/returning the replay header. This preserves checksummed addresses and extension keys while all money checks remain canonical.
3. **Context auth error:** contextual coworker calls now get a direct instruction to remove `X-Context-User-Id` / `X-Context-Organization-Id`; direct coworker auth remains required.
4. **Precondition docs:** endpoint OpenAPI now documents tasks capability, assignment, DRAFT/non-DRAFT and parked-task behavior, plus the intentional non-enumerating 404 cases.

Additional review fixes included:
- verify advertised v2 signing-domain fields (`name`, `version`, `assetTransferMethod`) when present
- keep the legacy sync cursor during rolling deploys; the new versioned cursor triggers replay safely
- remove the five-minute private x402 catalog cache so agent visibility updates on navigation
- regenerate the web Core client

**Item 5:** no Sokosumi code change. `EXTENSION-RESPONSES` / `PAYMENT-RESPONSE` is produced after settlement by the seller resource middleware and facilitator. Sokosumi produces the buyer payment-signature header but never handles the paid resource response, so it cannot add that settlement response header. The reported public x402.org facilitator + Bazaar behavior belongs upstream in x402 seller/facilitator support.

Verification: 359 Masumi tests; 3,629 Core tests; added v1 purchase, v2 header-name, exact accepted-casing, and signed-domain mismatch coverage; repo check and all workspace typechecks pass.

mrosberghaus:
@schaier-io please make sure that the CI is green if you request a review. Thanks 🙏

schaier-io:
@coderabbitai review

PR #3833, opened 2026-08-14
fix(chat): unbreak room thread list recency SQL

mrosberghaus:
## Summary
- Production `GET /v1/chats/rooms/{id}/threads?limit=…` 500s with Postgres `42803` (`column "p.createdAt" must appear in the GROUP BY clause`) — Sentry **SOKOSUMI-CORE-32**.
- Root cause: recency cursor baseline joined parent→replies and selected `MAX(r.createdAt)` + `p.createdAt` without `GROUP BY`. Postgres plans that subquery even when cursor is null, so **every** full thread list load fails (including first page).
- Fix: scalar subqueries for last-reply / parent-createdAt baseline; omit cursor filter on first page.
- Side effect: room `#Marketing` unread could not clear via look / mark-all while this endpoint 500'd (dual baseline still counts thread replies until looked).

## Test plan
- [x] Neon repro of broken SQL → `42803`
- [x] Neon check of fixed scalar subquery + soft-delete fallback
- [x] `pnpm --filter core test` helpers + threads get tests (83)
- [ ] After deploy: open Marketing → Threads side panel loads (no "An unexpected error occurred")
- [ ] Mark all as read / open a thread → room sidebar unread drops

Fixes SOKOSUMI-CORE-32

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat thread recency pagination for more accurate ordering when replies are present.
  * Corrected handling of deleted replies and threads without replies.
  * Fixed first-page pagination to avoid unnecessary cursor parameters and ensure consistent result limits.

### DIFF
--- a/apps/core/src/routes/v1/chats/rooms/helpers.test.ts
+++ b/apps/core/src/routes/v1/chats/rooms/helpers.test.ts
@@ -874,16 +874,49 @@ describe("getChatRoomThreadAggregates", () => {
 
     const sql = String(queryRawUnsafe.mock.calls[0]?.[0]);
     expect(sql).toContain('"unreadReplyCount" = 0');
-    expect(sql).toContain("LIMIT $");
+    expect(sql).toContain("LIMIT $4");
     expect(sql).toContain(
       'ORDER BY "lastReplyAt" DESC, "parentMessageId" DESC',
     );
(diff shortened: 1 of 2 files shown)


mrosberghaus:
@coderabbit review

PR #3866, opened 2026-08-19
feat(auth): add marketing proof to login

PatrickTobler:
## Summary

Auth routes used an empty photo on the right. That panel now carries the product argument: headline, three proof lines, customer logos, and Florian Haller’s quote with portrait.

`AuthAside` lives on the shared `(auth)` layout, so every auth page gets the same desktop panel — sign-in, sign-up, forgot/reset password, and OAuth — not only `/signin` and `/signup`.

- Add `AuthAside` over the existing auth hero (lg+ only; phones stay form-only)
- Keep the LCP image contract (server-picked, preloaded, sizes `0px` below `lg`)
- Localize copy in `en` / `de` / `es`

## User-facing impact

On desktop auth pages, people see what Sokosumi is. Mobile is unchanged.

## Verification

- [x] `pnpm check` and `pnpm typecheck` (pre-commit)
- [x] `pnpm --filter web test src/app/(auth)/components/__tests__`
- [x] Browser: `/signin` and `/signup` at 1440 / 1280 / 1024 (aside visible)
- [x] Browser: `/signin` at 390 (aside hidden, form intact)


PatrickTobler:
@francisluz @mrosberghaus Codex reviewed this PR in-tree and the follow-up fixes are on `9c08cee31` (`fix(auth): harden marketing aside accessibility`). Marking ready for review.

## Verdict
Ready with nits. Core marketing panel is sound: locale keys aligned, background preload contract unchanged, duplicated marquee tracks form a correct `-50%` seam. Remaining items below are product/test follow-ups, not merge blockers.

## Findings and fixes

1. **Wordmark stretch** (`layout.tsx`) — `100×18` did not match the SVG ratio (`200:26`). **Fixed:** `100×13`.
2. **Logo ticker motion** (`auth-aside.tsx`) — no pause control; reduced-motion hid the clone and clipped later logos. **Fixed:** pause on hover/focus, keyboard-focusable region, wrap the primary track under `prefers-reduced-motion`.
3. **Contrast** — `text-white/45` label and `text-white/55` role sat under AA on photo backgrounds. **Fixed:** opaque white for both.
4. **Redundant alts** — portrait and Serviceplan mark repeated adjacent visible text. **Fixed:** decorative `alt=""`; ticker logos still named.
5. **Tests** — mocked `next/image` so a missing portrait still passed; no coverage for equal tracks, hidden clone, pause, or the `-50%` keyframe. **Fixed:** real file existence + marquee/pause/reduced-motion/keyframe assertions.

## Left as nits (not blocking)

- Marketing panel lives in the shared `(auth)` layout, so it also shows on forgot/reset password and OAuth consent/callback routes, not only sign-in/sign-up. Confirm that is intended.
- LCP contract test still regex-matches `<AuthAside />` in source; it does not prove the generated preload/`imagesizes` in a real response. Keep the source guard; browser coverage would be extra.

Targeted auth tests, message parity, Biome, and workspace typecheck passed after the fix commit.

PR #3875, opened 2026-08-19
feat(soko-bot): add first-party personal assistant

PatrickTobler:
First-party **Soko Bot**: a personal assistant that lives inside Sokosumi, works the Taskboard with your Coworkers, reads the mail and calendar you connect, and reports in chat. It replaces Hermes, which this PR deletes.

**In beta for whitelisted domains.** Outside `@nmkr.io` with a verified address, the routes 404, the sidebar entry is hidden, the Core API answers as if the feature does not exist, and the crons skip the bot entirely. Signup neither verifies email nor withholds the session, so the domain alone proves nothing — the verified flag is what the whitelist rests on.

## What a user gets

- **Soko Bots page** (`/soko-bots`): what a Soko Bot is, then the workspace — everyone and the assistant they built. Bots are **per workspace**.
- **Create flow**: name and mascot, a short reveal, then the bot's direct chat opens and it introduces itself.
- **Assistant console**: what it did per day, automated-check status, recent activity, skills, connected accounts, schedules, memory, picture, settings.
- **On the Taskboard**: the owner assigns it a Task; it works the Task with `update_assigned_task`, follows Tasks it created, and comments only when it holds information the Task lacks (max 3/day per Task).
- **Chat and Drive tools**: `list_chats` / `read_chat` / `post_chat` and `list_files` / `upload_file`, bounded to rooms its coworker belongs to in its own workspace, and to the owner's Drive.
- **Proactive rhythms**: a weekday stand-up and a Friday wrap, plus deterministic nudges for stuck work, all under a daily cap the owner sets.
- **Integrations via Composio**: Gmail, Outlook and Google Calendar as read-only ingest; other toolkits through `list_integration_tools` / `run_integration_tool`. It never sends mail.
- **Admin version authoring**: platform admins write a system prompt, pick a gateway model and region, choose skills and tools, test in the behaviour lab, then promote for new bots.

## Runtime

The agent loop **runs inside Core** (`apps/core/src/lib/soko-bot/in-process-runtime.ts`). An earlier revision ran it as a separate Vercel "Eve" service; that is deleted here, so Soko Bot deploys with the rest of Sokosumi and there is no third deployment to keep in step. Turns run in `waitUntil` after the response is sent, record their own event log, and are drained by `/sync/soko-bot-turns`.

## Safety rails

- Every tool call re-checks the capability grant, turn status, lease, deadline, workspace membership and administrator pause. The runtime authenticates as the owning user; it holds no token of its own.
- A **teammate** mentioning someone else's bot gets a strictly smaller ceiling than the owner: status reads only, and the context packet withholds durable memory, prior turns, the approval queue and the credit balance. The answer is published into a shared room, so nothing owner-private may reach it.
- Only the owner may assign a Task to their bot — an assigned Task is mandatory work that spends the owner's credits.
- CLARIFY and MIXED carry no write tool.
- Mailboxes are read-only by construction: there is no send, reply or delete tool.
- `SOKO_BOT_ENABLED` gates every route and sync; `SOKO_BOT_PROACTIVE_PAUSED` is a platform-wide kill switch.

## Quality

`packages/soko-bot` ships 22 lab scenarios scored by code checks and an LLM judge, including scenarios where the correct answer is silence. Composio responses can be recorded and replayed so the mail and calendar scenarios are deterministic.

## Review

Four independent reviews ran against this branch — Grok, Codex, and both axes of the repo's `code-review` skill. Everything they agreed on is fixed. Findings deliberately left are under **Known gaps**.

## Schema changes

22 migrations. Three modify existing rows and are worth reading before deploy:
- `…020000_soko_bot_per_workspace` backfills `orchestrator.workspaceId` from turn history, then **deletes** rows whose owner has no workspace.
- `…040000_soko_bot_task_watch` appends `tasks` to bot coworkers' capabilities.
- `…080000_soko_bot_beta_controls` moves every bot on `v1`/null to `v11`.

Two migrations share a timestamp prefix with main's calendar work (`20260826140000`). Both are applied in deployed environments, so the pair is allowlisted rather than renamed — renaming re-runs the SQL and wedges the database.

## Deploy notes

- New crons alongside the existing ones: `soko-bot-turns` / `-schedules` / `-events` / `-ingest` / `-avatars`. Vercel runs crons on production deployments only, so on a preview nothing self-started fires and turns settle inline.
- New env: `COMPOSIO_API_KEY`, `COMPOSIO_API_BASE_URL` (both optional) and `SOKO_BOT_PROACTIVE_PAUSED` (defaults false).
- No separate runtime deployment: Core carries the agent loop.

## Verification

- `pnpm typecheck` — 19/19 tasks
- `pnpm check` (Biome) — clean
- `pnpm test` — core 4499, web 3871, database 310, soko-bot 179; all packages green
- Verified by hand on preview: create → intro chat, workspace scoping across three workspaces, a Task assigned to a bot worked end to end, stand-up and wrap delivered, settings persisted, feedback stored, admin version authoring → lab → promote

## Known gaps

- **The behaviour lab has not been run against a real model on the current default version.** The safeguards here are code-level; whether the bot exercises judgement is unmeasured.
- Owner approval is not implemented: `request_user_decision` exists but is granted on no route, and `executeAsAccepted` resolves its own decision. The implementation plan asks for approval on paid work; this branch does not do it.
- Cancelling a turn stops further tool calls but does not abort the model call already in flight.
- Organization seat enforcement does not cover Core's Soko Bot routes, so a member without a seat can drive a bot by calling Core directly.
- `run_integration_tool` does not re-check authority immediately before the external call, so a pause can race an in-flight write.
- `eveSessionId` / `eveTurnId` still carry the old runtime's name in the schema.
- Mobile was verified by markup audit, not by rendering at phone width.


PatrickTobler:
@coderabbitai full review

PatrickTobler:
@coderabbitai review

PR #3882, opened 2026-08-20
feat(chat): DM guests from external channels

mrosberghaus:
https://linear.app/masumi/issue/SOK-846/direct-messages-with-guests-from-an-external-channel

From an External channel, either person on the room roster can click Message on another human and open a private 1:1. If that 1:1 already exists, Message opens it.

- Host can Message a Guest. Guest can Message a host. Guest can Message another Guest.
- If both people are Members of the active organization, the room is an Org Direct.
- If either person is not a Member of the active organization, the room is a Personal Direct. It lists in every workspace.
- Every human Direct lists under Direct Messages, including Personal Directs with a Guest. External is for External channels and pending invites only.
- If a Personal Direct already exists, Message teammate in the organization opens that same room.
- Leaving the External channel does not close the Direct.
- Coworker Directs are unchanged.

Check: click Message on an External-channel hover, then confirm the 1:1 lists under Direct Messages in that workspace and after a workspace switch.


mrosberghaus:
@coderabbit review

PR #3886, opened 2026-08-20
fix(core): accept client_secret_post at the oauth2 token endpoint

PatrickTobler:
## Problem

Better Auth registers OAuth clients as `client_secret_basic` and hard-rejects a `client_secret` sent in the token-request body:

> `400 invalid_client — "client registered for client_secret_basic cannot use client_secret_post"`

Existing integrators — Serviceplan's coworkers (Hannah et al., before plan-net/agentic-coworkers#2032) — send the secret in the body. Every token exchange failed, no `id_token` was issued, and their OAuth onboarding dead-ended at a "No identity" page. Likely regressed when Better Auth introduced strict `token_endpoint_auth_method` enforcement.

## Fix — **TEMPORARY** compatibility shim

`withClientSecretPostShim` wraps the Better Auth mount: a `POST …/oauth2/token` form request carrying a body `client_secret` and **no** Authorization header is rewritten into the `client_secret_basic` form (secret moves into a Basic header, RFC 6749 §2.3.1). Everything else — GETs, JSON bodies, requests already using Basic, other auth routes — passes through untouched. Secret **verification** stays entirely with Better Auth; the shim only changes transport.

**Remove once all integrators authenticate with `client_secret_basic`** (Hannah-side fix is already merged-pending in plan-net/agentic-coworkers#2032).

## Verification

- 5 unit tests (`oauth2-token-secret-shim.test.ts`): rewrite happens, body secret removed, client_id/code/verifier preserved; pass-through for existing-Authorization, non-token paths, non-POST/non-form, and secretless requests
- Live against local core: the previously failing body-secret exchange now returns `invalid_grant: invalid code` (client auth passed) instead of `invalid_client`
- `pnpm --filter core test src/routes/auth/oauth2-token-secret-shim.test.ts` green; precommit (biome + typecheck) green

https://claude.ai/code/session_01VwwHjEhB4cLiLPaYeVq1eW

### DIFF
--- a/apps/core/src/routes/auth/index.ts
+++ b/apps/core/src/routes/auth/index.ts
@@ -8,6 +8,7 @@ import { cors } from "hono/cors";
 import { TIME } from "@/config/constants";
 import { resolveCorsAllowOrigin } from "@/config/cors-allow-origin";
 import { auth } from "@/lib/auth.js";
+import { withClientSecretPostShim } from "@/routes/auth/oauth2-token-secret-shim.js";
 import { handleSetPassword } from "@/routes/auth/set-password.route.js";
 
 const oauthAuthServerMetadataHandler = oauthProviderAuthServerMetadata(auth);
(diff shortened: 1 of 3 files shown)


PatrickTobler:
**End-to-end sanity check on the rebased branch (555dbf220 base): PASS.**

Replayed byte-for-byte what prod Hannah sends — `POST /auth/oauth2/token`, form-encoded, `client_secret` in the **body**, no Authorization header — against local core running this branch:

```
response keys: [access_token, expires_at, expires_in, id_token, scope, token_type]
id_token sub : 01a02006-… (the authorizing user's Sokosumi UUID)
id_token aud : <client_id>
```

Before the shim the identical request failed with `invalid_client — "client registered for client_secret_basic cannot use client_secret_post"`. The `sub` claim is exactly what Hannah's `decode_id_token` extracts and stores as the flow identity — so with this deployed, current (unpatched) Hannah completes token exchange, stores `sokosumi_user_id`, and the select-account "No identity" dead-end is gone.

(Local note: verifying id_token issuance required deleting a stale `jwks` row encrypted under an old `BETTER_AUTH_SECRET` — dev-DB artifact, irrelevant to prod.)

PR #3898, opened 2026-08-21
feat(core): let coworker originate org 1:1

mrosberghaus:
## Summary

A coworker API key can open a coworker 1:1 with an organization member who never messaged that coworker first. That unblocks assignment DMs and any other coworker-originated Direct.

`POST /v1/chats/rooms` with `{ "kind": "direct", "memberUserIds": ["<targetUserId>"] }` create-or-gets the same org-scoped room a user would get (`directKey` `coworker:{userId}:{coworkerId}`). Then the existing `POST /v1/chats/rooms/{id}/messages` path posts as the coworker.

## Behavior

- Org-scoped only. Personal coworker 1:1s stay user-started.
- Target must be an organization member (`400` if not — not first-person `403`). Coworker id is the auth actor, not the body.
- Originate requires the coworker to be usable in that workspace (`chat` + `baseURL` + whitelist or GRANTED), same as a human adding them to a room.
- Channels, human Directs, groups, and `coworkerIds` in the body stay 403/400.
- Create-or-get unarchives a stale Direct that still holds the `directKey`. That restore applies to **all** Directs, not only coworker-originated.
- `createdByUserId` is the target human. The coworker response does not include that human's pin/mute/unread flags.
- Coworker posts into a Direct with at most two humans emit the same CHAT Direct notification as a human sender. Mute is honored. Channel coworker posts do not load the room roster for that check.

See ADR [0011](docs/adr/0011-coworker-originated-coworker-1to1.md).

## Verification

- `pnpm --filter core test src/routes/v1/chats/rooms/post.test.ts`
- `pnpm --filter core test src/routes/v1/chats/rooms/[id]/messages/post.test.ts`
- `pnpm --filter core test src/helpers/chat-direct-message-notifications.test.ts`
- `pnpm --filter core test src/routes/v1/chats/rooms/auth.test.ts`
- Husky `pnpm check && pnpm typecheck` on commit

## Summary by CodeRabbit

* **New Features**
  * Coworker integrations can create or retrieve organization-scoped 1:1 chat rooms with human participants.
  * Coworker-authored messages in eligible 1:1 rooms now notify human recipients, respecting mute settings.
  * Archived direct rooms are automatically restored when reused.

* **Documentation**
  * Added guidance defining Direct rooms and coworker 1:1 rooms.
  * Documented coworker-originated 1:1 room behavior and limitations.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -82,6 +82,16 @@ _Avoid_: Onboarding (retired intro slides + plan checkout), account setup, onboa
 The dedicated authenticated route used when the user cannot use the product yet. It is where they resolve pending organization invitations and join links, or complete identity onboarding. No app chrome. Leaving is sign out or finish.
 _Avoid_: Onboarding page, welcome, accept-invitation as a separate post-signup product (the gate owns that moment)
 
+### Chat rooms
+
+**Direct**:
+A chat room whose identity is its participant set, not a named channel. Human 1:1, multi-human group, or coworker 1:1. Scoped to an Organization, except coworker 1:1s which may also be personal (no Organization).
+_Avoid_: Conversation (retired), Channel (when meaning this), Slack DM (as the domain object)
+
+**Coworker 1:1**:
+A Direct with exactly one human member and exactly one coworker.
+_Avoid_: Channel @mention thread, mixing extra humans or coworkers into this shape
+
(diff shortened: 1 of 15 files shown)


mrosberghaus:
@coderabbit review

PR #3921, opened 2026-08-24
fix(auth): restore 1.7 implicit-link default

mrosberghaus:
https://linear.app/masumi/issue/SOK-833/decide-better-auth-implicit-link-email-verification-policy

Restore the Better Auth 1.7 default so unverified password users cannot implicit-link Google/Microsoft. Core omits `requireLocalEmailVerified`; the library default (`true`) applies. Trusted-provider linking for verified users is unchanged. ADR 0007 updated.

## Summary by CodeRabbit

- **Security**
  - Strengthened account linking: Google and Microsoft accounts can now be linked only after the local email address has been verified.

- **Documentation**
  - Updated authentication guidance to reflect the account-linking verification requirement.


mrosberghaus:
@coderabbit review

mrosberghaus:
CodeRabbit asked for a runtime Google/Microsoft implicit-link regression test against an unverified password user (outside-diff note on `apps/core/src/lib/auth.test.ts`).

Not adding that test.

`auth.test.ts` mocks `better-auth/minimal` for the whole suite and snapshots `account.accountLinking` as `{ enabled: true, trustedProviders: ["google", "microsoft"] }` with `requireLocalEmailVerified` omitted. That is the contract for this change: Core no longer overrides Better Auth 1.7’s default.

A real linking test would need an un-mocked Better Auth instance, user/account rows, and OAuth callbacks. This suite has none of that. It would mostly re-prove the library gate (`requireLocalEmailVerified ?? true`), not Sokosumi-specific logic.

The snapshot already fails if the deprecated override comes back.

PR #3947, opened 2026-08-25
fix(chat): drop title-only unfurl cards

mrosberghaus:
## Summary

Chat no longer shows title-only unfurl chips. A card is kept only when Open Graph gives an image or a description. `Sign In | Sentry` with nothing else is dropped. Description-only and thumbnail previews stay.

Authors can still remove a remaining card (from #3946). Title-only chips are not created, so there is nothing to dismiss for those links.

Rebased onto `main` after #3946. Both intents kept: skip empty chips, keep author-remove on cards that remain.

## User-facing

- Title-only link previews no longer appear under room messages
- Existing stored title-only chips are hidden with no migration
- If a preview image 404s and there is no description, the whole card hides
- Author remove control is unchanged for cards that still qualify

## Verification

- `pnpm --filter @sokosumi/utils test src/__tests__/unfurl-urls.test.ts`
- `pnpm --filter core test src/lib/__tests__/open-graph-html.test.ts src/services/chat-room-message-unfurl.service.test.ts`
- `pnpm --filter web test src/app/(app)/chat/components/__tests__/room-message-row.test.tsx`
- `pnpm --filter core typecheck` and `pnpm --filter web typecheck`

I did not click through this in a running browser.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved link previews by hiding title-only cards without an image or description.
  - Preserved descriptions when preview images fail to load.
  - Correctly handles replacement images and filters out empty previews.
  - Ensured invalid or blank preview content is not displayed or persisted.

### DIFF
--- a/apps/core/src/lib/__tests__/open-graph-html.test.ts
+++ b/apps/core/src/lib/__tests__/open-graph-html.test.ts
@@ -53,6 +53,34 @@ describe("parseOpenGraphFields", () => {
 });
 
 describe("toUnfurlCard", () => {
+  it("returns null when title exists but image and description are missing", () => {
+    expect(
+      toUnfurlCard(
+        {
+          title: "Sign In | Sentry",
+          description: null,
+          image: null,
+          siteName: "Sentry",
+        },
+        "https://masumi.sentry.io",
+        "https://masumi.sentry.io",
+      ),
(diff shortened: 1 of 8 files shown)


mrosberghaus:
@coderabbit review

PR #3954, opened 2026-08-26
fix(core): trust sniffed coworker image type

mrosberghaus:
## Summary

Coworker image upload from the developer (and admin) Display form failed with **Failed to update image** when the browser MIME did not match the file bytes.

The picker accepts `image/png` from the filename. JPEG bytes saved as `.png` then hit Core `POST /coworkers/{id}/image`, which 400'd on `sniffed !== declared`. The form maps every non-404 failure to that toast, so the mismatch never surfaced.

Core already stored the sniffed type. The mismatch check only blocked real images whose OS MIME followed the extension. Non-image bytes still 400 on magic-byte sniff.

## User-facing

- Uploading a coworker image whose browser MIME disagrees with the bytes (JPEG named `.png`, and the reverse) now succeeds when both types are allowed.
- Developer and admin Display forms share this endpoint.

## Verification

- `pnpm --filter core test src/routes/v1/coworkers/[id]/image/image.route.test.ts`
- `pnpm check` and `pnpm typecheck` (pre-commit)

I did not click through this in a running browser.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved coworker image uploads by validating supported image formats based on the file’s actual content.
  * JPEG images can now upload successfully even when the browser incorrectly labels them as PNG.


mrosberghaus:
@coderabbit review

PR #3960, opened 2026-08-26
fix(core): make x402 readiness alert a checklist

schaier-io:
## What

Rewrites the zero-ready-pairs Sentry alert in `apps/core/src/services/agent-sync.x402-readiness.ts` (`Sentry.captureMessage` in `syncX402BuySideReadiness`). The old text was one 950-character sentence with nested em-dash clauses. The new text keeps the same content as a numbered four-item checklist, ordered by the likeliest cause first (a null `defaultAssetDecimals`, per the comment above the capture).

## User-facing impact

None in the product. Operators who receive the Sentry page can now read the four checks in order instead of parsing one long sentence. The new text starts a new Sentry issue group; the old group will stop receiving events.

## No behavior change

Message string only. Tests assert the capture count, not the text (`agent-sync.x402-readiness.test.ts:297,324`). A fresh-eyes review verified every claim in the new text against `composeX402ReadySources`: the both-paths-blocked claim (`configuredBudgetPairs` is added before the wallet-binding and remaining-spend checks, and the admin fallback skips configured pairs), the four named warn logs, the location of both constants, and the non-admin-key coda.

## Verification

- `pnpm --filter core test src/services/agent-sync.x402-readiness.test.ts`: 23 passed (23)
- `pnpm check`: 3421 files, no fixes
- `pnpm typecheck`: 17 tasks successful


schaier-io:
Superseded by #3986.

That PR rewrites the same alert string as part of moving buy-side readiness off the removed `GET /x402/budgets` and onto the API key's usage credits (masumi ADR 0016). Keeping this one would leave a checklist that still tells operators to chase budget rows, which no longer exist on the node.

Closing in favour of the stack: #3985 (client read) → #3986 (readiness switch + alert rewrite).

PR #3979, opened 2026-08-26
fix: disallow crawling of app and API hosts

mrosberghaus:
## Summary

Search crawlers could index the product app and API.

- `app.sokosumi.com/robots.txt` served `Allow: /` on Mainnet.
- `api.sokosumi.com/robots.txt` and `api.preprod.sokosumi.com/robots.txt` 404, which crawlers treat as allow-all.
- `preprod.sokosumi.com` already served `Disallow: /`.

Web now always returns `Disallow: /`. Root HTML metadata always sets `noindex, nofollow`, including Mainnet. Core serves `GET /robots.txt` with the same disallow body from `mainApp`, so it stays available during maintenance.

`www.sokosumi.com` is a separate marketing site and is unchanged.

## User-facing impact

App and API hosts should stop being crawled after this deploys. Public share pages under `/share/` are also blocked by `robots.txt` and inherit root `noindex`, matching the whole-host policy.

## Verification

- `pnpm --filter web exec vitest run src/app/__tests__/robots.test.ts src/app/__tests__/layout-robots-metadata.test.ts`
- `pnpm --filter core exec vitest run src/index.test.ts`
- Husky `pnpm check && pnpm typecheck` on commit

Live hosts still serve the old files until this ships.

## Summary by CodeRabbit

* **New Features**
  * Added a `/robots.txt` endpoint that blocks search engine crawlers.
  * The crawler restriction remains active during maintenance mode.

* **Bug Fixes**
  * Disabled page indexing and link following across all environments.
  * Removed environment-specific behavior that previously allowed crawling in some deployments.

### DIFF
--- a/apps/core/src/index.test.ts
+++ b/apps/core/src/index.test.ts
@@ -126,6 +126,18 @@ describe("core index", () => {
     serveMock.mockImplementation(() => undefined);
   });
 
+  it("serves robots.txt that disallows all crawlers", async () => {
+    const fetchHandler = await loadFetchHandler();
+
+    const response = await fetchHandler(
+      new Request("http://localhost/robots.txt"),
+    );
+
+    expect(response.status).toBe(200);
+    expect(response.headers.get("content-type")).toMatch(/text\/plain/);
+    expect(await response.text()).toBe("User-Agent: *\nDisallow: /\n");
+  });
+
   it("does not generate llms markdown during startup and still serves docs routes", async () => {
     createMarkdownFromOpenApiMock.mockResolvedValue("# llms");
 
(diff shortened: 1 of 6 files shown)


mrosberghaus:
@coderabbit review

mrosberghaus:
CodeRabbit nits, checked against the repo:

1. `@/` for `../robots` — `apps/web/src/app/__tests__/manifest.test.ts` already imports `../manifest`. The suggested `` `@/app/robots` `` string is invalid. Leaving the relative import so this file matches the sibling test.

2. Mainnet mock — `robots()` no longer reads env. The mock is the regression lock for the original bug: default `NEXT_PUBLIC_NETWORK` is Preprod, so without the mock a Mainnet `Allow: /` branch would still pass. Keeping it.

3. Move `apps/core/src/index.test.ts` into `__tests__/` — that file predates this PR. Core has hundreds of sibling `*.test.ts` files and only a small set under `__tests__/`. Not relocating the existing entrypoint test in this change.

PR #3967, opened 2026-08-26
feat(chat): show latest pin under header

mrosberghaus:
## Summary

Channels already had a pin rail. This puts the **newest pin under the room header** so it is visible without opening that rail.

- One-line ticket strip: author + snippet, primary accent, “Latest pin” kicker
- Click jumps to the message in the transcript
- If there is more than one pin, a count chip opens the existing pins panel
- Channels only. Directs never show it. Empty channels stay clean
- Live pin/unpin refreshes the strip without a skeleton flash

## Verification

- `pnpm --filter web exec vitest run` on the new banner/picker tests and `rooms-client-room-header`
- Husky `pnpm check && pnpm typecheck` on commit
- Browser: local Web did not stay healthy in this worktree (portless 502, Next EMFILE / `.next/dev` deleted). Could not walk the strip in a real channel here.

## Summary by CodeRabbit

* **New Features**
  * Channel rooms now display a banner highlighting the latest pinned message.
  * Jump directly to the latest pinned message or open the complete pinned-message list.
  * Pinned-message lists now show the total number of pins.
  * Added English, German, and Spanish translations for pinned-message actions.

* **Bug Fixes**
  * Improved handling of deleted or unavailable pinned messages when selecting the latest pin.

* **Tests**
  * Added coverage for banner loading, navigation, room changes, empty lists, and deleted messages.

PR #3981, opened 2026-08-26
fix(chat): re-pin when pin banner shrinks scroller

mrosberghaus:
## Summary

On a Channel with a latest pin under the header, stick-to-bottom left the last message slightly clipped.

The pin sits outside the message scroller. When it takes height, the scroller shrinks without the content growing, so the existing content-resize pin never ran. If you were already at the live edge, the last bubble sat about one banner-height below the fold.

- Re-pin when the **scroller** resizes, not only when content grows
- Leave people who scrolled away (and search jumps) where they are
- Layout-time `holdOffBottom` is locked by a test that fails if that guard is dropped (search jump must not yank on first scroller layout)

Follow-up to the latest-pin-under-header work (#3967).

## Verification

- `pnpm --filter web exec vitest run src/app/(app)/chat/hooks/__tests__/use-stick-to-bottom.test.tsx` — 14 passed (scroller-shrink RED→GREEN, plus layout-time holdOffBottom lock)
- `pnpm test` — full workspace suite green (on the fix commit)
- Husky `pnpm check && pnpm typecheck` on both commits
- Browser: local Web/Core were not healthy in this worktree (`verify-sokosumi doctor` 404). Could not walk a pinned channel here. Reload a Channel at the live edge with a latest pin to confirm the last message is fully visible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic scrolling in chat when the available viewing area changes.
  * Preserved users’ manual scroll position instead of unexpectedly returning to the bottom.
  * Ensured “hold off bottom” behavior remains consistent during initial layout and resize events.

### DIFF
--- a/apps/web/src/app/(app)/chat/hooks/__tests__/use-stick-to-bottom.test.tsx
+++ b/apps/web/src/app/(app)/chat/hooks/__tests__/use-stick-to-bottom.test.tsx
@@ -12,26 +12,50 @@ type ResizeObserverCallback = (
 ) => void;
 
 const observerCallbacks = new Set<ResizeObserverCallback>();
+const observerCallbacksByTarget = new Map<
+  Element,
+  Set<ResizeObserverCallback>
+>();
 
 class ResizeObserverMock {
   private readonly callback: ResizeObserverCallback;
+  private readonly observed = new Set<Element>();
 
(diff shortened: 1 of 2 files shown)


mrosberghaus:
@coderabbit review

PR #3990, opened 2026-08-27
feat(web): render chat push banners in a service worker

schaier-io:
Closes SOK-875.

Top layer of the push stack. Sits on `sok-873-core-pushoptin-preference-push-subscribe-capability-gated` (#3969), which sits on `sok-699-push-design-records` (#3968). Review and merge bottom-up.

## What a reader gets

Chat mentions reach them when the app is closed. Two switches in Account settings, because there are two independent things to say yes to:

- **Account** (`pushOptIn`, stored in Core) decides whether Sokosumi may push to this reader at all. Core's publish gate reads it.
- **This browser** decides whether this particular device holds a Web Push subscription.

Turning the account row on from a browser that can push also subscribes that browser, so the common case is one click. Turning it off withdraws consent everywhere, including from browsers holding no subscription.

## Shape

- `apps/web/public/ably-push-sw.js` — the service worker. Renders the banner from a Core payload and routes a click. Plain JavaScript on purpose: the browser fetches it from `public/` as written, so it never passes through the TypeScript build. Its copies of app constants and its message catalog are guarded by tests instead of by the compiler.
- `apps/web/src/lib/ably/use-push-preference.ts` — the hook behind both switches. Owns the two axes, the permission prompt, and the interleaving between a save and the reads that would otherwise paint over it.
- `apps/web/src/lib/ably/push-activation.client.ts` / `release-push-device.client.ts` — subscribe and release this browser.
- `apps/web/src/lib/auth/sign-out.client.ts` — one place that owns releasing the push device on an explicit sign-out, so a browser left signed out stops showing the previous reader's mentions.
- `apps/web/src/lib/utils/notification-service-worker.ts` / `notification-navigation.ts` — register the worker, and route a click to the right room.

## Decisions worth a reviewer's attention

**A click is withheld from a tab that did not come forward.** Handing a tab the click target can switch the active organization, which is a session-wide write every tab shares. A tab that stayed hidden would move the reader's front tab into another workspace on its own. The worker cannot tell a cross-workspace click from a same-workspace one, so it withholds from both and opens a window instead.

**Sign-out releases the push device before it signs out.** Deactivation mints an Ably token, so it needs the session that is about to end. The release is capped at 5s: `PushSubscription.unsubscribe()` stops delivery locally and does not wait on the push service, so the cap can only shed the Ably half. A session that expires on its own releases nothing, so a returning reader finds push still on.

**The device row is owned by a save for the save's whole duration.** Granting the permission fires a permission change, and that refresh reads the browser while the activation that asked for the permission is still running. Without the ticket the reader watched the switch they just clicked turn itself back off, beside a toast saying push was on.

**An account write retires the read it outran.** The preferences query goes stale after a minute and refetches when the tab is hidden and shown again, which can land inside the wait for a prompt the reader paces.

## Verification

- `pnpm --filter web test` — 614 files, 4022 passed, 1 skipped.
- `pnpm check` and `pnpm typecheck` clean.
- Every behavioural fix in this branch was confirmed by mutation: the source was reverted with the tests kept, and the test that names the behaviour was watched to fail.

## Still to do before merge

- Manual check on the preprod preview: enable push, close the app, receive a banner, click it, then sign out and confirm the banner stops. This is the only claim in this PR not backed by a test or a primary source.

## Preview

- Web: https://sokosumi-app-preprod-bvwbllib9.preview.sokosumi.com
- Core: https://sokosumi-core-preprod-25lpfraz8.preview.sokosumi.com


mrosberghaus (changes requested):
Image

I'm unable to register for push notifications on the mainnet preview.

PR #4003, opened 2026-08-28
chore(deps): upgrade pnpm to 12.0.0

mrosberghaus:
## Why

pnpm 12 is the stable Rust rewrite. Pin the repo to `pnpm@12.0.0` and move CI onto `pnpm/setup@v2`, which is the supported installer for pnpm 11+.

## Scope

- Root `package.json` `packageManager` → `pnpm@12.0.0`
- `pnpm-lock.yaml` records `packageManagerDependencies` for the pnpm 12 native binary
- `.github/actions/setup/action.yml` replaces `pnpm/action-setup@v6` + `actions/setup-node@v7` with `pnpm/setup@v2` (`runtime: node@24`, `cache: true`, `install: false`)
- Doc/example pins in `AGENTS.md` and `.cursor/rules/pinned-dependencies.mdc`
- Deploy workflows unchanged (Node only, no pnpm)

## Tradeoffs

`pnpm/setup` defaults to running `pnpm install`. This PR sets `install: false` and keeps the explicit install step so Turborepo remote-cache setup still runs after deps land, in the same order as before.

## Blast Radius

Lint, build, and test jobs all use `./.github/actions/setup`. Local clones need Corepack or `pnpm self-update next-12` because npm `latest` is still pnpm 11. Lockfile format stays `9.0`; dependency graph resolution was not re-run beyond the package-manager front document.

## Verification

- `pnpm --version` → `12.0.0`
- `pnpm install --frozen-lockfile` (pass)
- `pnpm check` and `pnpm typecheck` via pre-commit (pass)
- Confirmed `.github/actions/setup/action.yml` has no `pnpm/action-setup` or `actions/setup-node`


mrosberghaus:
Closing for now: **pnpm 12 is not usable on Vercel yet**.

Vercel installs the `packageManager` pin with lifecycle scripts skipped, so pnpm 12’s placeholder `bin/pnpm` never becomes the native binary (`syntax error near unexpected token ')'`). That breaks preview/production installs and later turbo/`pnpm` invocations under `~/.local/share/pnpm/.tools`.

We will revisit this upgrade once Vercel supports pnpm 12 (or documents a supported install path for the Rust binary). Until then, stay on pnpm 11.

mrosberghaus:
Closed: waiting on Vercel pnpm 12 support.

PR #4028, opened 2026-08-29
fix(soko-bot): stop dropping the newest exchange from the context packet

PatrickTobler:
Asked whether bots see earlier messages in a direct message, I went looking. They do — but a trimming bug was throwing that history away, newest first.

## How history reaches the bot

A direct message goes to the model as typed; the conversation itself arrives in the context packet as `recentTurns` — the last 12 turns, each with the owner's message (≤1000 chars) and the bot's answer (≤2000). `chat-room-coworker-dispatch.service.ts` relies on this explicitly: *"Directs are the bot's own conversation: the control plane already rehydrates recent turns."*

## The bug

`fitPacketToBudget` shrinks an over-budget packet by repeatedly dropping the **last** entry of whichever collection has the largest one. That is right for every collection but this one: tasks, projects, jobs, coworkers, agents, and decisions are all ordered most-relevant-first, so their tail is the least important entry. `recentTurns` is deliberately reversed into chronological order so the model reads the conversation forwards — which puts the **newest** exchange at the tail.

So the trimmer deleted the most recent turn first. And because turn entries are the largest things in the packet (up to ~3KB each), `largestTailCollection` kept picking them: the conversation was emptied completely before a single task was dropped.

A test that builds a realistic over-budget packet (24 tasks, 12 full-length turns) shows `recentTurns` arriving at the model as `[]`.

## The fix

- Trim `recentTurns` from the front, so the oldest exchange goes first.
- Keep a floor of three exchanges, broken only when nothing else is left to trim — otherwise a busy workspace can still starve the conversation down to nothing.

## Verification

- New test asserts the three newest turns survive, in order; it returns `[]` without the fix
- `pnpm --filter core test` — 4505 passed
- `pnpm --filter core typecheck`, `pnpm check` — clean

## Scope

Only affects packets over the 48KB budget. Under the budget nothing changes, which is why this was invisible on small workspaces and total on busy ones.

### DIFF
--- a/apps/core/src/lib/soko-bot/context-packet.test.ts
+++ b/apps/core/src/lib/soko-bot/context-packet.test.ts
@@ -572,6 +572,37 @@ describe("ContextPacketBuilder", () => {
     }
   });
 
+  it("keeps the newest exchange when trimming a packet over budget", async () => {
+    // Newest first, as the query returns them; the packet renders chronologically.
+    recentTurnFindManyMock.mockResolvedValue(
+      Array.from({ length: 12 }, (_, index) =>
+        recentTurn(`turn-${11 - index}`, "y".repeat(1_200), "y".repeat(2_500)),
+      ),
+    );
+    recentTurnCountMock.mockResolvedValue(12);
+    taskFindManyMock.mockResolvedValue(
+      Array.from({ length: 24 }, (_, index) => ({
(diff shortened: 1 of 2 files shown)


mrosberghaus:
@PatrickTobler is this still wanted / needed?

PR #4032, opened 2026-08-30
feat(soko-bot): let an assistant open a direct chat with a colleague

PatrickTobler:
An assistant could talk to people only in rooms it was already in. This lets it open a direct chat with someone in its owner's organization and say something there.

## User-facing impact

- A bot told "reach out to Nina about the tiers" now does it, instead of replying that it has no way to reach anyone while holding the tools to do so.
- Opening a chat and speaking in it are one action: a bot cannot leave behind an empty room. Nobody can leave a direct chat once it exists, so an unwanted one is permanent — that is the failure this is built around.
- Bounded on purpose: only on a turn the owner asked for (never a schedule or another bot's request), at most 5 new directs per turn, only members of the turn's organization, never another assistant, and never a second room with the owner.
- Told "don't reach out to Nina yet", the bot does not — the prohibition now survives a curly apostrophe, "get in touch"/"drop a line" phrasing, and the trailing form ("reach out to Nina, but not yet"), while "don't forget to message Nina" stays the instruction it is.

## Notable details

- A name matches names and an address matches addresses. A member whose display name is `counsel@outside.example` was otherwise the match when the owner asked to contact *counsel*.
- `%` and `_` are wildcards in Prisma's `startsWith`, and the needle comes from a model reading untrusted text: unescaped, `"%"` alone matched a colleague at random and opened a permanent room with them.
- Two people named Nina produce a question, not a guess. Full names separate them; an address is offered only when the names themselves collide.
- If the first message fails to send, the room it just created goes with it.

No schema changes, no new environment variables, no Core API contract change — `open_direct_chat` is a Soko Bot tool, not an HTTP route.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 19/19 tasks
- `pnpm test` — 4539 Core tests, 191 `@sokosumi/soko-bot` tests, all green


PatrickTobler:
@coderabbitai review

PR #4034, opened 2026-08-30
fix(soko-bot): stop a direct room becoming standing permission

PatrickTobler:
Two holes a Codex review of #4032 found on `main`, both live today.

## A direct room was standing permission

`open_direct_chat` is deliberately limited to turns the owner asked for — a schedule can never approach a new colleague. But the room it leaves behind outlived that limit. `postChat` guarded only chain depth, never the turn's `source`, so every later `SCHEDULE` / `INGEST` / `EVENT` turn could find the room through `list_chats` and write to that colleague again with nobody having asked.

One instruction to reach Nina became a standing licence to keep reaching her. Nobody can leave a direct room, so the person on the other end has no way out of it.

**Rule now:** an unattended turn may not post into a direct the owner is not in.

- **The owner's own direct stays open** — briefing them there unprompted is precisely what a scheduled turn is for.
- **Channels stay open** — a person can leave or mute one, and a bot added to a project room is expected to speak in it.

The check is exact rather than heuristic: `createDirectRoomRecord` writes the counterpart as the room's sole `ChatRoomUserMember`, so a bot↔colleague direct has the colleague there and the owner's own direct has the owner.

> **Open question for review:** channels are exempt on the reasoning above. If a scheduled turn posting unprompted into a shared channel is also unwanted, the rule tightens to "rooms the owner is a member of" — one clause. I left it out because the irreversible case is the direct.

## "Don't cancel my weekly reminder" was heard as no prohibition

The negation guard's vocabulary had no verb for *undoing* something. The classifier reads `cancel|stop|remove|pause|change|move|delete` as a change to the bot's own schedules and routes them to `MANAGE_WORK`, which grants `delete_schedule` — while `hasSokoBotNegatedMutationIntent` returned false, so the runtime's negation guard never fired and the bot could cancel the reminder it had just been told not to touch.

`delete_schedule`, `create_schedule` and `update_schedule` were already in `NEGATABLE_WRITE_CAPABILITIES`; the guard simply never fired. Adding the verbs is the whole fix.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 19/19
- `pnpm test` — full suite green; new cases cover the colleague direct being refused, the owner's direct still working, channels untouched, and the schedule prohibitions

Not covered by tests: the classifier→capability→guard chain end to end. Each link is tested in isolation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unattended bot posts from appearing in colleagues’ direct rooms without the owner’s membership.
  * Continued allowing posts in the owner’s direct rooms and eligible channels.
  * Improved recognition of commands that cancel, stop, pause, delete, change, or move scheduled actions, while avoiding false positives for urgency phrases.
* **Tests**
  * Added coverage for direct-room posting restrictions and schedule cancellation language.

### DIFF
--- a/apps/core/src/services/soko-bot-runtime.service.test.ts
+++ b/apps/core/src/services/soko-bot-runtime.service.test.ts
@@ -8,6 +8,7 @@ const {
   botFindUniqueMock,
   chatRoomFindFirstMock,
   chatRoomDeleteManyMock,
+  chatRoomUserMemberFindFirstMock,
   workspaceFindUniqueMock,
   chatRoomFindManyMock,
   chatMessageFindManyMock,
@@ -77,6 +78,7 @@ const {
   botFindUniqueMock: vi.fn(),
   chatRoomFindFirstMock: vi.fn(),
   chatRoomDeleteManyMock: vi.fn(),
+  chatRoomUserMemberFindFirstMock: vi.fn(),
   workspaceFindUniqueMock: vi.fn(),
   chatRoomFindManyMock: vi.fn(),
   chatMessageFindManyMock: vi.fn(),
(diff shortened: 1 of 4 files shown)


PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PR #4035, opened 2026-08-30
feat(soko-bot): show what a bot has spent, counting every model call

PatrickTobler:
You asked to see a bot's tokens and cost, and for it not to miss any. Finding out where the tokens go turned out to be the substance of this.

## A turn is three model calls, not one

Only the agent loop was metered. Two others surround it and both discarded their usage at the call site:

- **The classifier** (`classifier.ts`) — one `generateText` on every turn the deterministic rules don't catch.
- **The judge** (`soko-bot-lab-judge.service.ts`) — the control plane comments say *"Quality score for every settled turn"*, so this runs on **every** production turn, and retries once when structured output comes back empty.

So a bot's recorded spend was short by up to three calls on a typical turn, and there was nowhere to see even the part that was recorded.

Both now report what they used, counted *before* the parse so a call whose output can't be read still counts what it burned. `gatewayCostUsd` moved out of the runtime into `lib/soko-bot/gateway-cost.ts` so all three read provider cost the same way.

## Where the numbers go

Tokens go into the turn's existing `usage` column. That works without a second source of truth because the event drain already **seeds** its aggregate from the row (`parseTurnUsage(turn.usage)`), so a classifier seed written at turn creation survives every later step write, and the judge adds on top after settlement. `usage` now means every token the turn spent.

Money is split. A new `overheadCostUsdMicros` holds the classifier and judge; `costUsdMicros` keeps its exact meaning.

> **`sokoBotUsageCents` reads `costUsdMicros` to bill.** Folding overhead into it would silently raise what every owner pays from the moment this deploys — a pricing decision, not a reporting one, so it isn't taken here. Flipping it later is a one-line change once the real numbers are visible.

## What you see

**Soko Bots page** — a usage card. **Admin bot detail** — the same figures in the existing `MetaGrid`.

Credits lead, because that is what leaves the balance. Tokens and raw model cost sit underneath, and they deliberately don't reconcile: billing applies `SOKO_BOT_MIN_TURN_CREDITS`, so a turn costing a fraction of a cent still charges the minimum. The card says so rather than hiding the gap.

## Notes

- **Migration:** `20260830120000_soko_bot_turn_total_usage` — one nullable-free column with a default, no backfill of tokens. Historical turns keep `0` overhead: the discarded usage is not recoverable and inventing a number would be worse than an honest floor.
- Totals are summed in Postgres (`$queryRaw` over the `usage` JSON plus the two cost columns) rather than in memory, since a bot accumulates turns for as long as it exists and this is read on a page load.
- Core client regenerated; `getMySokoBotUsage` is its own endpoint rather than a field on `/me`, which is polled for turn state.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 19/19
- `pnpm test` — full suite green; new cases cover the aggregate (including a bot that has never run, where Postgres `SUM` returns NULL) and the classifier reporting no usage when no call was made

**Draft:** the migration has not been run against a live database from here, and the numbers have not been eyeballed against a real bot's spend. Worth doing both before this merges.

## Summary by CodeRabbit

* **New Features**
  * Added Soko Bot usage summaries for turns, tokens, model costs, and consumed credits.
  * Added usage details to user-facing Soko Bots and administrator bot views.
  * Added an authenticated endpoint for retrieving lifetime usage metrics.
  * Added English, German, and Spanish translations.

* **Improvements**
  * Classification and judging usage is now included, with billable and overhead costs separated.
  * Usage values and costs are localized and consistently formatted.
  * Improved reliability of usage tracking during failed or concurrent evaluations.

* **Tests**
  * Added coverage for usage aggregation and classifier usage reporting.


PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PR #4037, opened 2026-08-30
refactor(soko-bot): drop the negated-intent guard

PatrickTobler:
The guard was too blunt for an assistant that is supposed to act.

## Why

It matched a prohibition **anywhere** in the owner's message and then refused **every** negatable capability for the whole turn — tasks, hires, assignments, schedules, uploads, task replies, decisions and chat alike (`soko-bot-runtime.service.ts`, the check in `executeAuthorizedTool`).

It knew that *something* had been forbidden, never *what*. So:

> *"Don't post that publicly yet, but go ahead and create the task"*

blocked the task as well. The vocabulary made this worse over time — `change` and `move` are ordinary words — but the vocabulary was never the real problem. The all-or-nothing blast radius was.

## What's gone

Everything, rather than a trim: the regex and its three shapes, the verb vocabulary and inflection helper, the anti-delay phrase, `NEGATABLE_WRITE_CAPABILITIES` and `isSokoBotNegatableWrite`, the tool-execution check, the classifier branch that forced such turns read-only, and every test covering them.

## What still holds

Honouring "don't" now rests on the prompt alone. Every guard that does **not** depend on reading intent out of prose is untouched:

- `open_direct_chat` still requires a turn the owner asked for (`source === "CHAT"`, `chainDepth === 0`)
- an unattended turn still cannot post into a colleague's direct
- the per-turn cap of 5 new directs, org membership, never another assistant, never a second owner room
- the bot-to-bot chain ceiling and the per-room hourly rate limit

## Trade-off, stated plainly

I recommended keeping the guard for the three capabilities that cannot be taken back — `post_chat`, `open_direct_chat`, `run_integration_tool` — and dropping it everywhere else, on the grounds that a task created in error is deleted in a click while a message to a colleague in a room they cannot leave is permanent. The owner chose full removal; this PR is that decision.

The consequence to watch: *"don't message Nina yet"* is now honoured only if the model chooses to honour it.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 19/19
- `pnpm test` — full suite green (4545 Core, 187 `@sokosumi/soko-bot`)

## Summary by CodeRabbit

* **Behavior Changes**
  * Updated message classification so requests not to perform actions follow standard handling.
  * Action-related capabilities are no longer blocked solely based on negated wording.
* **New Features**
  * Introduced Soko Bot version 16 as the default experience.
  * Improved handling of explicit refusals and deferrals, including clearly stating when no action was taken.
  * Added stronger confirmation requirements for irreversible or credit-consuming agent hires.


PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PR #4041, opened 2026-08-30
fix(soko-bot): start the lab's Coworker turns instead of waiting on a cron

PatrickTobler:
All three `task_event` scenarios reported *"The turn did not finish within 5 minutes."* The turns were not slow. In most cases **no turn was started at all**, and the lab had no way to say so.

## What was happening

The lab simulated a Coworker event, then waited for the one-minute events cron to notice it and hunted the bot's recent turns for one whose text mentioned the task id. Every reason a wake can be withheld looked identical from there — nothing for five minutes, then the same sentence:

- the bot is paused (`proactivePaused`)
- the platform is paused (`SOKO_BOT_PROACTIVE_PAUSED`)
- the bot is at its daily cap (`proactiveDailyLimit`, default 20)
- the turn genuinely started and ran long

## The daily cap is the one that bites, and the lab caused it

Turns whose `clientTurnId` starts with `lab:` are excluded from the proactive allowance — the comment in `proactiveGate` is explicit that *"behaviour-lab turns are ours, not the owner's budget"*.

But the lab's `task_event` path never started the turn. The **cron** did, with `sokoBotEventClientTurnId(changes)` → `event:…`. So the exclusion never applied, every scenario run spent one of the owner's twenty proactive turns, and once they were gone the Coworker scenarios stopped working — silently — until the next local day.

That matches the report exactly: the three Coworker scenarios fail while every other scenario (which starts its turn directly, with a `lab:` id) keeps working.

## Fix

`simulateSokoBotTaskEvent` now starts the turn itself with a `lab:` id and returns it. So:

- a run costs nothing from the allowance, as intended
- the lab follows a known turn id instead of scanning state for matching text
- no cron latency
- the delegation is marked seen **at the new status**, so the cron doesn't wake the bot a second time for the same event (it previously rewound the status deliberately to make the cron fire)

It reuses `buildEventMessage`, so the bot receives exactly the text the sync would have sent.

## Better information, since this class of failure will recur

- **"No turn was started"** and **"the turn did not finish"** are now different messages, and the second names the turn id to look up.
- The events sync **logs a withheld wake** — reason, `usedToday`, `limit`, how many changes were dropped — where it previously `continue`d in silence. That line is what distinguishes "bot deliberately quiet" from "sync broken" in production, not just in the lab.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 19/19
- `pnpm test` — full suite green; four new cases cover the turn being started, the `lab:` prefix, the message content, and the delegation being marked seen

**Not verified against a live bot from here.** The reasoning is from the code and matches the reported symptom precisely, but the confirming evidence would be a `[soko-bot-events] Wake withheld` line — which this PR is what adds. Worth re-running the three scenarios once this deploys.

## Summary by CodeRabbit

- **New Features**
  - Scenario Lab runs now receive and track the created turn directly.
  - Runs report when no turn starts because the bot is paused or has reached its daily limit.
  - Timeout messages include the affected turn ID and direct users to the bot’s event log.
- **Bug Fixes**
  - Wake-up denials are recorded consistently.
  - Delegation status updates are limited to the selected delegation and restored if startup fails.
  - Busy bot responses clearly indicate a conflict and refusal reason.
- **Localization**
  - Updated English, German, and Spanish messages for new run and timeout states.


PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PR #4058, opened 2026-08-31
feat(soko-bot): run the mail and calendar scenarios from the lab UI

PatrickTobler:
Six scenarios told you to go and use a terminal. They did not need to.

## The CLI restriction was mostly imaginary

The CLI's `--record` / `--replay` fixtures are a convenience for working **without** OAuth, not a requirement. Its ingest path calls `buildIngestDeltaMessageForBot` and `buildSystemBeatMessage` — ordinary service functions that read whatever the bot has connected — and then starts a normal turn. Nothing about that needs a process of its own.

## What each beat actually needs

The old guard was a blanket rule over `kind: "ingest"`. The real requirement is per beat, and narrower:

| Beat | Reads | Was |
| --- | --- | --- |
| `weekly-wrap` | **nothing** — board and memory only | CLI-only |
| `delta` | mail | CLI-only |
| `standup` | mail + calendar | CLI-only |

Two of the six were unreachable for no reason whatsoever.

## Change

`POST /v1/soko-bots/me/lab/ingest` runs one rhythm on demand:

- with the **source the cron would have used** (`SCHEDULE` for the two system beats, `INGEST` for the delta), so the turn is classified and capability-scoped exactly as the real one is — a lab run that skipped that would prove nothing
- with a `lab:` client turn id, keeping it out of the owner's proactive allowance
- returning `400` naming the connection to make when the beat reads an account the bot has not linked, instead of failing vaguely

The lab now runs those scenarios inline and shows Core's message when a connection is missing.

## Which account a run hits

Kept to the signed-in user, as decided — this stays a `/me` route, and no one can start billed turns on a colleague's assistant from a browser.

Since every run is a real turn against real Tasks and real credit, the page now says so explicitly: *"Every run is a real turn on Hermes, the assistant belonging to patrick@nmkr.io."* Sign in as the account you want to test and it targets that bot.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 19/19
- `pnpm test` — full suite green; four new cases cover the missing-connection rejection, the weekly wrap running with nothing connected, the per-beat source, and the `lab:` prefix

**Not run against a live bot from here.** The packet builders are the same ones the cron uses, so the risk is in the wiring rather than the behaviour, but the honest check is running the six scenarios once this deploys.

Two things this does not do, both deliberate: the CLI still exists and still works (it is the only way to *record* fixtures, which needs a real connected account once), and there is no account picker.

## Summary by CodeRabbit

* **New Features**
  * Added on-demand Soko Bot lab runs for stand-up, weekly-wrap, and delta scenarios.
  * Runs start directly from the Behaviour Lab and return a turn for tracking.
  * Lab pages identify the assistant and account used for each run.
  * Added localized messaging in English, German, and Spanish.

* **Bug Fixes**
  * Ingest scenarios no longer fail with a CLI-only error.
  * Added clearer handling for missing integrations, unavailable bots, and busy runs.
  * Lab runs are no longer affected by unrelated proactive-work limits.


PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

PR #4072, opened 2026-09-01
feat(core): sync job purchases through the payment diff endpoint

schaier-io:
Closes [SOK-921](https://linear.app/masumi/issue/SOK-921/sync-job-purchases-through-the-payment-diff-endpoint)

## User-facing impact

Paid jobs pick up on-chain purchase changes (funds locked, result submitted, refund withdrawn, dispute) from the payment node's diff feed. Sync cost now tracks how many purchases changed on the node, not how many jobs are open here, so purchase state lands sooner when many jobs are in flight.

No API contract change. No schema migration. The Core API client needs no regeneration.

## What changed

- **New** `apps/core/src/services/job-purchase-diff.service.ts`. It reads `GET /purchase/diff` one page at a time (50 rows) and applies each changed purchase to its job.
- **Join by id, with a fallback.** `JobPurchase.externalId` is unique and holds the node's purchase id, so a diff row maps to exactly one job. Rows that match no `externalId` are retried against `Job.blockchainIdentifier`, and a match there rewrites the stored `externalId`, so a job whose stored id went stale repairs itself instead of going dark. The fallback only accepts a row whose input hash and agent identifier are the job's own, and gives each job at most one fallback row per run.
- **Purchase phase is now a backfill phase.** It only attaches purchases to jobs that still have no `JobPurchase` row. Renamed to match: `buildJobsNeedingPurchaseBackfillWhere`, `backfillJobPurchase`, log prefix `[sync/jobs/purchase-backfill]`. Its attach-time term validation (input hash, seller vkey, agent identifier, amounts, deadlines) is unchanged.
- **New client method** `getPurchasesDiff` in `@sokosumi/masumi`, plus the exported `MasumiPurchaseDiffEntry` type.
- **Replay escape hatch.** `GET /sync/jobs?replay=true` drops the cursor and replays the whole feed, mirroring `GET /sync/agents?replay=true`.

## The cursor

`lastUpdate` plus `cursorId`, stored in `SyncMetadata` under `purchase-diff-sync:v2`. It advances only over the contiguous prefix of handled rows, which is the pattern the registry sync already uses (`agent-sync.service.ts`). A row that fails to apply parks the cursor in front of it, so the next run retries that row.

The stored `cursorId` carries one bit beyond the node's tie-break: whether the previous run reached the end of the feed.

- **Set** — the previous run stopped early (deadline, park, node error). The next run resumes on exactly that row, so it does not repeat work it already did.
- **Null** — the previous run drained the feed. The next run restarts five minutes before the stored timestamp. `lastUpdate` filters on a timestamp the node stamps when the change is made, but a row reaches us when its transaction commits, so a change stamped just before the cursor can commit just after it. Re-reading those minutes costs one no-op write each: `finalizeJobSyncResult` returns early when the job status did not change, so nothing is re-notified.

A first run with no stored cursor starts 30 days back, not at the epoch: replaying the node's whole history would re-apply terminal states to jobs that finished months ago, and a status that moves fires emails, the failure webhook, and an Ably publish. `?replay=true` is the way to ask for the full history.

## Notes for review

- **Phase order.** Backfill, then diff, then purchase-transaction polling, then agent, then refund. Refund keeps a 20s reserve and must run after backfill, which is what attaches a purchase row for escrow that landed since the last run. The diff sits inside the network budget and cannot starve refund.
- **`unfinishedFound` shrinks.** It counts jobs found by the four job selectors. Diff rows are keyed on purchases, not jobs, so jobs updated through the diff no longer appear in that number. `processed` still includes them.
- **Stall guard.** The node treats `lastUpdate` as inclusive and breaks ties on `cursorId`. If a page ever ends on the row the page started from, the run stops instead of re-reading the same rows until the deadline.
- **The feed is not scoped to our API key.** Rows for other consumers of the same node match no local job; they are skipped and still advance the cursor.
- **Deployed-node check.** VERIFIED 2026-09-01: the current PR deployments completed purchase-diff syncs against both preprod and mainnet-preview payment nodes.

## Payment source rails

The request sends no payment-source filter, so both V1 and V2 purchases come back. Verified in the node's own source, not only in its spec.

`GET /purchase`, `/payment/diff` and `/registry/diff` each default to `Web3CardanoV1` when neither `filterPaymentSourceType` nor `filterSmartContractAddress` is given. `/purchase/diff` does not. The deployed route's query parameters (`limit`, `cursorId`, `lastUpdate`, `network`, `filterSmartContractAddress`, `includeHistory`) and its `filterSmartContractAddress` description match masumi-payment-service at `5416f92fb^` exactly, and `buildPurchaseDiffWhere` there filters on `deletedAt`, `network` and `smartContractAddress` alone. If that route ever gains `filterPaymentSourceType`, this sync has to page each rail with its own cursor; the note lives on `getPurchasesDiff`.

## Payment spec snapshot

Refreshed in the same branch. The deployment had moved well past the pin while holding version 1.0.0, so the fetch guard could not catch it: 32 paths added (the `/hydra` head lifecycle and the `/reports` exports), `GET /x402/budgets` dropped (per-key usage credits replaced it and nothing here called it), 31 changed. The registry snapshot came back byte-identical. The generated payment client is regenerated from it, unedited.

## Verification

- `pnpm --filter @sokosumi/core test` — `Test Files 456 passed | 2 skipped (458)`, `Tests 4629 passed | 6 skipped (4635)`
- `pnpm --filter @sokosumi/masumi test` — `Test Files 17 passed (17)`, `Tests 396 passed (396)`
- `pnpm --filter @sokosumi/database test` — `Test Files 49 passed | 2 skipped (51)`, `Tests 349 passed | 3 skipped (352)`
- `pnpm check` — `Checked 3737 files. No fixes applied.`
- `pnpm typecheck` — `Tasks: 19 successful, 19 total`
- VERIFIED 2026-09-01: preprod current-head replay reported `found=1701`, `processed=1297`; the following normal diff reported `found=0`, `processed=0`.
- VERIFIED 2026-09-01: a new 1-credit preprod job completed as `RESULT_SUBMITTED`; its normal diff reported `found=1`, `processed=1`.
- VERIFIED 2026-09-01: an email-disabled mainnet-preview replay resumed across four time-bounded passes, applied `14106` rows, and the following normal diff reported `found=0`, `processed=0`.

## Summary by CodeRabbit

- **New Features**
  - Added purchase-difference synchronization to keep job payment statuses and details up to date.
  - Added `?replay=true` support for full purchase-feed replays.
  - Added transaction-status polling to refresh attached purchase states.
  - Improved recovery of missing or outdated purchase identifiers.

- **Bug Fixes**
  - Improved safeguards for mismatches, duplicates, stalled progress, cancellations, and failed updates.
  - Enhanced resumability, pagination, validation, and error handling.

- **Documentation**
  - Updated job-flow and payment specification references.


schaier-io:
@coderabbitai review

schaier-io:
@coderabbitai review

PR #4076, opened 2026-09-01
fix(soko-bot): stop reporting a wrong state as a missing record

PatrickTobler:
Everything here is about making the measurement layer trustworthy, because two days of lab results turned out to be measuring the harness rather than the bot.

## A wrong state was reported as a missing record

A bot read a Task with `get_task_status`, then `assign_task` on the same id answered **"Task not found"**. The Task existed; it was `FAILED`, and `requireMutableTask` folded id, ownership *and* allowed-status into one query, so every miss came back as absence. The bot concluded its own id was wrong and created a duplicate — which is what the lab scored as a failure, and what the judge scored honesty 1 for.

Status is checked after the lookup now, and a Task in the wrong state says which state it is in and which ones allow the action. The user-facing branch immediately below already did this correctly; only the Soko Bot path collapsed it.

The same conflation sat in `get_agent_input_schema`, where an offline Agent was reported as missing. `isShown` stays in that query — an unlisted Agent must still read as absent, or a bot holding a stray UUID could enumerate rows `find_agents` deliberately hides. Both tools now resolve the endpoint through `getAgentApiBaseUrl`, so an Agent reachable only through its metadata override is neither hidden by one nor rejected by the other.

## Thirteen of thirty-three capabilities had never been called

`pnpm soko-bot:tool-smoke --owner <email>` exercises every capability against a real bot through a synthetic turn that grants all of them, so coverage does not depend on which tools a model happens to reach for.

They are **real** calls: Tasks created, updated, linked, assigned, replied to and closed; a schedule created, changed and deleted; a message posted; memory written; a file uploaded; a decision raised; an Agent hired with real credits and the Job read back. An earlier version probed the guards instead and proved only that the guards work — `upload_file` was genuinely broken the whole time and no refusal-based probe would ever have shown it.

**29 ok, 0 failed, 4 skipped, 0 never exercised.** The skips are gaps in the data or side effects that land on someone else, and each says which.

What it does is stated before it does it. The owner is named on every invocation and an owner with several bots must say which; spending is `--spend`; messaging a colleague is `--dm`; running an integration tool needs the slug — `SEARCH_AND_ARCHIVE` reads as a read and archives real mail, and no naming rule separates the two.

What it creates, it takes back. One `cleanUp` runs on the happy path, both signals and both crash events: the synthetic turn deleted, the pending decision removed, the scratch Tasks archived, and the scratch schedule deleted — a leftover Soko Bot schedule is not inert, it fires and starts a turn nobody asked for. Memory is written back exactly as read, and refused entirely if the read failed, because `run` returns null for a failed read and an empty document alike. The run exits non-zero if any of that could not be done.

## The judge was accusing the bot of things it did not do

It reported a fabricated email that was in the tool result all along. `search_inbox` returned 3,091 characters, every value was clipped at 2,000, and the fifth message fell off the end.

It also never received the context packet — the rubric contained a line telling it the packet existed but that it could not see it. So every Task status and calendar entry the bot correctly reported from context read as unsourced. That is most of what scored honesty 1.

Values and the packet now share one budget for the whole transcript rather than a limit per value, sized for the fact that the payload is JSON-escaped a second time before it is sent. And the packet is evidence only where it records platform state structurally — ids, statuses, dates, senders. Task descriptions, comments, schedule prompts and the memory document are things the assistant writes itself; treating those as sourced would make anything it once invented and stored true forever.

`judgeTurnWithModel` re-grades a settled turn with a named model without storing the verdict, and `soko-bot:judge-eval` scores judges against turns whose outcome was established by reading the tool results and the packet — a judge measured by another model's opinion is not measured.

| judge | before | after |
| --- | --- | --- |
| `anthropic/claude-haiku-4.5` | 1/6 | **5/6** |
| `anthropic/claude-sonnet-4.5` | — | **5/6** |

**The default stays Haiku.** The transcript was the fault, not the model, and this judge runs on every settled turn. `judge-eval` now reports what a model charged for the set, so a case for changing it can be made with numbers rather than a hunch.

**Still unproven:** all six eval cases are turns where the bot was honest, so this measures false accusations only. Whether either model *catches* a real fabrication is untested and needs a turn known to have lied. The script says so when it runs.

## Verification

Run without pnpm — `main` pins 12.1.0 and my local version manager cannot install it — so binaries were invoked directly:

- `biome check` — 1237 files, clean
- `tsc --noEmit` on Core — no errors in anything this touches (5 remain in `compensating-refund`, `notifications`, the `agent-sync` fixtures and `stripe-invoice-credit`, all pre-existing on `main`)
- `vitest` — 4595 Core, 192 `@sokosumi/soko-bot`
- `soko-bot:tool-smoke --owner patrick@nmkr.io --spend` against the dev branch — 29 ok, 0 failed, nothing left running

Reviewed three times each by CodeRabbit and Codex; seventeen findings fixed across `28cb176`, `c9a698d` and `4727fde`.


PatrickTobler:
@coderabbitai review

PatrickTobler:
Pushed 28cb176 with the fixes. Points I did not take, and why:

**Synthetic turn bypasses turn eligibility** (tool-smoke:101-118) — correct, and deliberate. This measures whether the tools work, not whether a model can reach them; the route ceilings and proactive limits are covered by `packages/soko-bot` tests and the behaviour lab. It no longer hides that behind `as never` — the scope is the typed `{ turnId, sessionId }` `authorize` actually takes.

**judge-eval ignores verdict, other scores and latency** — it now reports the cost of the set, and the default model change it was being used to justify is withdrawn, so nothing rests on it. Broadening it into a full benchmark is worth doing when there is a dishonest turn to measure against; without one it would still only measure false accusations, which the script now prints.

Re-ran `soko-bot:tool-smoke --owner patrick@nmkr.io --spend` against the dev branch after the changes: **29 ok, 0 failed, 4 skipped, 0 never exercised**, memory restored, no turn left running.

PatrickTobler:
Pushed c9a698d — all seven of the second pass fixed except the consent one, which I'll answer rather than build:

**Owner consent in the smoke script.** Running it requires the production `DATABASE_URL` and the Composio and gateway keys. Anyone holding those can already do everything the script does, directly. A consent check inside a developer harness would be a check the operator is trusted to have satisfied before they had the credentials at all — so the honest control is that the owner is named on every invocation, spending is `--spend`, and messaging a colleague is `--dm`. All three now say who they act as before they act.

The rest are in: `getAgentApiBaseUrl` for the availability check (an override-only endpoint is hireable), memory refuses to write unless it read a string and marks with a heading the parser keeps, one `cleanUp` on both signals and both crash events that archives the scratch Tasks too and exits non-zero if a turn survives, `--integration-tool <slug>` instead of a name-based guess, `clip`'s ellipsis inside its limit with budgets sized for the second JSON pass, and the rubric limited to structured fields so a hallucination copied into a Task description is not evidence next turn.

Verified: 4595 Core tests, 192 package tests, biome clean, and `soko-bot:tool-smoke --owner patrick@nmkr.io --spend` at 29 ok / 0 failed / 4 skipped / 0 never exercised with no turn left running.

PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai full review

PatrickTobler:
Pushed 4727fde — all six of the third pass are in:

- **Leaked schedule.** The material one. A schedule that survived `delete_schedule` was not inert; it fires and starts a turn nobody asked for. `cleanUp` removes it on every exit path and only stops tracking it once the delete actually succeeded.
- **Memory restore failure now fails the run.** It was a line in "Left behind" beside an exit code of 0.
- **`find_agents` resolves the endpoint the same way `get_agent_input_schema` does**, so an override-only Agent is no longer hireable but undiscoverable.
- **`judgeTurnWithModel` requires a settled turn**, matching `judgeTurnQuality` — grading a running turn scores the judge on evidence that does not exist yet.
- **`judge-eval` counts `SokoBotJudgeFailure.usage`**, so a model that burns tokens and returns nothing no longer looks cheaper than one that answers.

Verified: 4595 Core tests, biome clean, `soko-bot:tool-smoke --owner patrick@nmkr.io --spend` at 29 ok / 0 failed / 4 skipped / 0 never exercised, no turn or decision left behind.

PR #4088, opened 2026-09-01
feat(web): give every notification subject its own row

schaier-io:
Evaluation layer for SOK-877. Disposable: one layout gets picked, it replaces the switch matrix in the layer below, and this branch is deleted.

## What changed after the last round

The previous panel showed a breadth ladder ("what to tell you about") above a row of per-subject switches, so the same subject appeared twice, in two shapes, with two ways to change it. Mentions and threads could not be chosen apart from each other either.

There is one list now.

- Everything you can ask for is a **row**, and every row appears **once**.
- Direct messages, mentions of you, threads you follow and every message in your rooms are **four separate rows**, each with its own delivery.
- Where one row already carries another, the narrower row stays visible and says so: "Covered by Every message in your rooms". It does not disappear, and it is not a second control.
- A covered row can still be set **louder** than its cover. Every message in Sokosumi, a banner only when you are named. It cannot be set quieter, and the stops below the cover are disabled with the reason on them.
- A preset is a shortcut for the rows in a group, never a second setting. A preset that would write what another preset writes is not drawn, so "Everything" and "Important" both appear only where they differ.

## The four layouts

They share the rows and the closed-row summary. Only the control on the row differs.

| | Control | Trade |
| --- | --- | --- |
| **F1** | Three stops: off, in app, banner | Everything visible at once, including what a cover has taken away |
| **F2** | A tick per channel | The only shape with somewhere to put email, which is not louder than a banner but elsewhere |
| **F3** | A switch, and the banner once it is on | Smallest control, but the column moves as rows go on |
| **F4** | F1 with covered rows nested under their cover | Containment visible before it is read, at the cost of an indent |

## Not stored yet

Threads you follow, every message in your rooms, other task updates, other run status changes, and the email channel. They are drawn so the covering rule has something to cover. What they change is remembered for the visit and never written. Everything else round-trips through the preferences API.

## Verification

- `pnpm --filter web test`: 628 files, 4227 passed, 1 skipped
- `pnpm --filter web typecheck`: clean
- `pnpm exec biome check apps/web/src`: clean
- Mutation checked: ignoring a cover's delivery, forgetting the floor, dropping the preset de-duplication, widening the push guard, and naming covered subjects in the summary each fail exactly one test

## How to try it

Open account settings, notifications. All four layouts edit the same preferences, so a change in one appears in the others. Asking for a banner turns push on and requests the browser permission from the control you pressed.


schaier-io:
Superseded by #4093, which splits jobs and tasks by whether they wait on the reader rather than giving every subject its own row. The branch sok-877-8-notification-ux-options stays on origin if we want to revisit the alternative.

PR #4096, opened 2026-09-01
feat(web): give each notification group its own presets

schaier-io:
Layer 11 of SOK-877. Based on [#4095](https://github.com/masumi-network/sokosumi/pull/4095); review that one first.

## Why

The group control set every kind in a group to one delivery. That is the wrong question for a group holding a job you have to answer next to a job that merely finished. The answer most readers want is "the ones that wait on me, loudly, and the rest quietly", and one delivery cannot say it.

## What changes

A group offers up to four answers of its own:

| | Jobs and tasks | Chat |
| --- | --- | --- |
| Everything | both rows, banner | all three rows, banner |
| Important | attention banner, updates off | mentions and direct messages banner, room messages off |
| Quiet | attention in Sokosumi, updates off | mentions and direct messages in Sokosumi, room messages off |
| Off | both off | all three off |

What a preset writes depends on the group, because what waits on the reader depends on the group. `KindSpec.important` answers that once per kind, rather than once per preset.

**A preset that writes what an earlier preset writes is dropped**, so a control never carries two stops that do the same thing. A group whose kinds all wait on the reader shows three. A group of one is still drawn as its kind with the delivery ladder, which is what its presets would have been under other names.

**Custom is not one of the answers.** It appears when the reader has set the kinds one by one, and pressing it opens the group rather than picking for them. That is the Mixed chip's job under a name that says why, and `deliveryMixed` goes with it.

`setDeliveries` replaces `setDelivery`: a preset writes a different delivery per kind, in one request, so a group cannot land half applied.

## Verification

- `pnpm --filter web test` — 4234 passed, 1 skipped, including a new `notification-delivery.test.ts` for the preset model and six new cases on the panel
- `pnpm --filter web messages:parity`, `pnpm --filter web typecheck`, `biome check apps/web` — clean
- Mutation-checked: making Quiet loud, ignoring `important`, keeping every preset instead of deduplicating them, reporting Custom always, and writing the first change's delivery for every cell are each caught by a failing test.


schaier-io:

PR #4100, opened 2026-09-02
fix(core): stop paging when a proxy answers the purchase diff

schaier-io:
## Why

After the diff-based purchase sync landed (#4072), a restarting Masumi payment node reaches `GET /purchase/diff` as a Cloudflare **521**. The body is Cloudflare's HTML error page, not the node's `{ error: { message } }` envelope, so two things went wrong:

- the whole page was `JSON.stringify`d into one log line and one Sentry title;
- the failure was classified as a hard error, so a node restart paged once a minute for the length of the outage.

No purchase data was ever lost: a failed request leaves the cursor parked, and the next run resumes at the same row.

## What changed

- **Cap the fallback dump** in `extractNodeErrorMessage` at 300 characters (`packages/masumi/src/clients/node-error.ts`). The node's own envelope message is still returned whole, because callers echo it to coworkers.
- **Carry the HTTP status** out of `getPurchasesDiff` next to the message, as `MasumiPurchaseDiffFailure { message, status }`. `status` is `null` when no response arrived.
- **Teach the shared Sentry policy** about upstream gateway statuses: new `isTransientUpstreamHttpError` covers 408/502/503/504 and Cloudflare's 520-527 origin-error range, and joins `shouldSuppressSentryForExternalError`. Those log a warning instead of paging. 429 is deliberately excluded — sustained rate limiting is a quota decision, not a wait.
- **Request 20 rows per diff page** instead of 50, so one response stays small for whatever serves it. The run deadline still bounds a run, and the loop keeps requesting while a page comes back full, so this costs more requests to drain a backlog and never fewer rows.

A node 4xx or 5xx carrying its own envelope still pages. The cursor parks on every failure either way.

## User-facing impact

None in the product. This is operational: a payment-node restart no longer pages on-call once a minute, and the diff log lines stay readable.

## Verification

- `pnpm core:test` — `Test Files 463 passed | 2 skipped (465)`, `Tests 4691 passed | 6 skipped (4697)`
- `pnpm masumi:test` — `Test Files 17 passed (17)`, `Tests 400 passed (400)`
- `pnpm typecheck` — `Tasks: 19 successful, 19 total`
- `pnpm check` — `Checked 3774 files. No fixes applied.`
- Mutation check: with the suppression line removed, 2 of the new tests fail; with the truncation removed, 1 fails.

## Notes for the reviewer

- `shouldSuppressSentryForExternalError` is shared by 23 call sites in Core, so any error object carrying a `statusCode` in those ranges is now demoted to a warning, not only the purchase diff. That is intended: a gateway 5xx is transient wherever it appears.
- The trade-off is that a long node outage now leaves only warning lines. A follow-up signal for "the diff has failed for N consecutive runs" is worth a separate ticket.
- Reviewed by a self-review pass over the full diff, not by a multi-agent adversarial review.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of temporary upstream and proxy failures, including Cloudflare errors, so affected operations can retry without unnecessary error reporting.
  - Preserved reporting for authentication and other non-transient failures.
  - Improved purchase-difference error messages and handling for invalid responses, timeouts, and connection failures.
  - Limited oversized fallback error details to prevent excessively large error output.

- **Reliability**
  - Purchase-difference requests now use smaller batches to improve compatibility with proxy response-size limits.


schaier-io:
@coderabbitai review

PR #4104, opened 2026-09-02
feat(soko-bot): let an admin move bots onto another version

PatrickTobler:
Promoting a version only decided what **new** bots are created on. Every bot carries its own `versionId`, stamped when it was created and read straight through on each turn, and `resolveSokoBotVersion` consults the promoted default only for a bot pinned to an *archived authored* version. A built-in like `v14` therefore never fell forward: promote `v16` and the existing fleet stayed exactly where it was, however many times it was promoted.

Nothing in the product moved them either. The only version picker was the scenario lab, and it is session-scoped — an admin could move their own bot and nobody else's. Thirty-one bots on mainnet, and no way to reach thirty of them.

## Per bot

A version picker on the admin bot detail page: what it runs now, a required reason, one button.

`SET_VERSION` is a new admin action rather than a second path, so it inherits what the existing operator actions already have — a mandatory reason, an operation id that makes a retry the same operation, and an audit row per attempt.

Two details that would otherwise bite:

- **The audit snapshot gains `versionId`.** Nothing else on it changes during this action, so before and after would have recorded a move from and to the same apparent state.
- **An unknown slug is refused before the intent is written.** An `ATTEMPTED` row for a version that does not exist is an outbox entry no retry could ever complete.

## Whole fleet

"Bots currently on `v14` (12 bots)" → "move to `v16`", reason, one button. It says how many bots the choice would move **before** you press it, and afterwards reports moved / already-on-version / failed, naming any bot it could not move.

It is the per-bot action **in a loop, not a bulk write.** Each bot gets its own audit entry and its own idempotency key, so re-running a partially failed migration retries only what did not land, and one stuck bot does not strand the other thirty. A single `updateMany` would be faster and would tell you nothing.

## Confirmed, not assumed

An in-flight turn is unaffected: a turn binds its version at creation, so moving a bot mid-turn does not change what is already running.

## Verification

Run with binaries directly — `main` pins `pnpm@12.1.0` and my local version manager cannot install it:

- `biome check` — 3292 files, clean
- `tsc --noEmit` — clean on Core and Web
- `vitest` — 4699 Core, 4174 Web

New tests cover the schema refinements, the move recording where it came from, the unknown-version refusal writing nothing, skipping bots already on the target while continuing past a failure, and per-bot idempotency keys. Translations added to en, de and es. The Core client is regenerated from the new OpenAPI and committed as generated.

**Not verified in a browser.** The admin routes are behind auth and I have no account on the dev database, so the two new controls have not been clicked. Types and tests cover the wiring; the counts shown before a migration are worth one human glance.


PatrickTobler:
Codex review found six issues, five High. All fixed in a884f5a8.

The worst was one I had claimed the opposite of in the PR description. The per-bot operation id was derived from a caller-supplied one, and `applyAdminAction` rethrows a recorded FAILED outcome instead of retrying — so re-running a partly failed migration replayed each failure forever. The bots worth retrying were exactly the ones it refused to touch. The base is generated per run now; idempotency comes from the target-version pre-check, which skips a bot already on the target before any work happens. That also closes the reused-id-for-a-different-migration hole and the 164-character-id-fails-every-bot bound.

The other four: every moved bot was returning `getForAdmin` (50 turns with tool calls, 500 legacy messages, schedules, audit rows) for a caller that reads none of it — the action is a void `applyAdminAction` now with a thin wrapper for the single-bot screen; more than 500 failures made the response throw on its own output after the writes had committed — the count is exact and the list is a bounded sample; and the fleet page counted versions from its own 100-bot page while Core moves everything, so it could offer to move 20 and move 270 — there is a `GET /versions/usage` that counts in the database.

Codex confirmed the rest of what I asked it to attack: unknown targets are rejected before the intent row, turns stamp their version at creation so an in-flight turn is unaffected, both new paths require admin, and the pre-existing `/me/version` is still scoped to the caller's own bots.

4702 Core tests, 4174 Web tests, biome and tsc clean.

PR #4111, opened 2026-09-02
fix(soko-bot): let the console see a turn while it is still running

PatrickTobler:
The assistant console showed no thinking at all — no orb, no step chips, just the finished answer appearing. Two things had to be true at once.

## The optimistic turn is gone

`addOptimisticTurn`, `bindOptimisticTurn` and `dropOptimisticTurn` are still defined and returned by `useSokoBotState`, and **nothing has called them** since conversation moved to the chat rooms — `console.client.tsx` destructures `{ state, refresh }`.

They used to insert a `QUEUED` turn the moment you pressed send, which did two jobs: it drew the orb immediately, and it flipped polling to its fast rate. With no composer on this surface, nothing does either.

## So the console only learns about a turn by polling

And it only polled fast **once it had already seen an active turn in the list**. Every turn it watches now is one it did not start — from a room, a schedule, an event. At a thirty-second heartbeat a turn that runs for five to twenty seconds begins and ends between two ticks, so it was never observed in an active state and never rendered as one.

## The fix

The bot's own `RUNNING` status counts as activity, not just a turn already in the list, and the heartbeat drops to eight seconds so a turn is sampled at least once while it is still going.

That is more requests on an open tab — 7.5 a minute instead of 2, only while the tab is visible. It is what watching work started somewhere else costs, and the alternative is a surface that can never show it.

The optimistic machinery is deleted rather than left dangling, including `ChatTurn.optimistic` and the branch that read it. It is the kind of dead code that answers "why is there no orb?" with a wrong answer.

## On the duplicate Task chip, reported alongside this

**Not changed, because the evidence says the UI is right.** Chips are collapsed into a `Map` keyed on `taskId`, so two delegations sharing a task render one chip — verified: same id → 1 chip, different ids → 2. That dedupe has been in place since 28 August, so it was live in the build that produced the report. Both chips in the screenshot carry the `↗` arrow, which only renders when a task link exists.

So the two chips were two different tasks, and the bot created a second one while mentioning only the first in its answer. That is the failure mode #4076 addressed earlier today — a bot told "Task not found" creating a duplicate — and mainnet did not yet have it when the report came in. Deduping the chips would have hidden a real second task rather than fixing anything.

## Verification

- A real turn on current main against a migrated local database: 15 events written (`step.started`, `actions.requested`, `action.result`, `step.completed`, `message.completed`), one delegation, one chip
- Events confirmed present in the chat-state projection Core returns
- Two new tests: a `RUNNING` bot polls fast with no turn yet in the list, and an idle console still samples quickly enough to catch a short turn
- 4199 web tests, biome and `tsc` clean

**Not verified in a browser** — I cannot sign in to the local stack, so the orb has not been watched appearing. The polling change is what makes it observable; the rendering path itself is untouched and was already correct.

## Summary by CodeRabbit

- **New Features**
  - Completed chat turns now include explanations, including turns displayed immediately during conversation updates.
  - Added lightweight activity monitoring to keep assistant status and progress current.

- **Bug Fixes**
  - Chat status and responses now refresh while the assistant is working, helping new messages and progress appear sooner.
  - Chat updates now rely on confirmed server state, improving consistency between the conversation view and the assistant’s actual status.
  - Reduced unnecessary refresh requests when the assistant is idle.


PatrickTobler:
@coderabbitai full review

PatrickTobler:
Codex rejected the first attempt and was right. Reworked in d91d8c5.

**The fix was lossy.** Treating `bot.status === "RUNNING"` as activity does not help, because that status arrives on the same poll — a five-second turn starting just after a tick still finished before the next one, and the console received an idle bot and a terminal turn. Classification and context building happen before Core creates the turn, which narrows the window further.

**And expensive for what it bought.** Every tick loaded the bot plus twenty turns with their events, delegations and decisions — two Core calls — so an eight-second heartbeat meant ~450 of those an hour on an idle open tab.

**Split in two instead.** New `GET /v1/soko-bots/me/activity`: one indexed read answering only who is running and when the last turn moved. The console probes that every 2.5s — often enough to see a short turn while it is still going — and fetches the heavy chat state only when the answer changes or a turn is in flight.

Net: discovery in a third of the time, and an idle tab stops pulling twenty turns of history to learn nothing.

Also fixed the unused `waitFor` import, which would have failed the Biome gate.

4771 Core tests, 4197 Web tests, biome and tsc clean.

PatrickTobler:
@coderabbitai full review

PatrickTobler:
Second Codex pass found six more. Four fixed in 84e53f1, two answered.

**Abort starvation (P1) — the worst of them, and self-inflicted.** `refresh()` aborted the previous `/state` read, so if that call ever took longer than one probe interval, every response while the turn was running got cancelled by the next tick and only the one after it settled arrived. That is precisely the bug this PR exists to fix, reintroduced by the fix. Overlapping reads are skipped now, not cancelled.

**No probe on mount (P2).** A turn starting between server render and hydration was invisible until the first interval elapsed.

**The probe was not as cheap as I claimed (P1).** It made a Core `/auth/get-session` round trip and *then* a second authenticated call running two queries. The session preflight is gone — Core authenticates the call itself and answers 401 — and the two reads are one nested query. Roughly a third of the work per tick.

**Biome was failing (P2)** on import order in the new route. It looked clean locally only because I piped the check through `tail -1` and read a blank line as success. Exit status decides now.

**Tests (P2) — fair hit.** Mine counted requests and would have passed against the broken version. They walk a turn through idle → running → settled and cover the slow-read case, and I verified they have teeth by restoring the aborting behaviour and watching them fail.

**One finding I am not fixing, and want to be plain about.** Core sets the bot `RUNNING` in the same transaction that creates the turn, *after* funding, classification, version resolution and context building. So a turn that spends three seconds preparing and two seconds running can still fall entirely between two probes: the answer arrives, the orb never does. Narrowing that needs either the bot marked busy before preparation — which risks a crash leaving it stuck busy — or a realtime push on a channel that does not exist yet. Both are real changes with their own review surface, and neither belongs in a bug fix. This PR makes the window much smaller, not zero.

4200 web tests, 4771 Core tests, biome and tsc clean by exit status.

PatrickTobler:
@coderabbitai full review

PatrickTobler:
Third Codex pass, four findings, all four addressed in 7d1d9ab.

**The dropped terminal refresh (P1).** Skipping an overlapping read was better than cancelling it and still wrong: the tick that saw the turn settle could find a read in flight, skip its own, *and* record the settled signature — then the slow response landed and restored the running state, and no later tick asked again because the signature was already marked seen. The orb stayed on screen for good. Same class of bug as the two before it, from the other direction.

Overlapping requests are queued one deep now, and a signature is recorded only after the read it belongs to has actually been made. The read carries a 15-second deadline, since a request that never settles would hold the flag and stop every future read.

**502 instead of 401 (P2).** The redirecting Core client turns Core's 401 into a thrown Next redirect, which the handler caught and reported as a gateway failure. Now uses `coreClientNoRedirect` — which is what makes dropping the session preflight correct rather than just cheaper.

**My round-trip claim was wrong (P1).** The nested read is one Prisma call but still two SQL statements, because the generator does not enable `relationJoins`. Corrected in the code comment. What makes it affordable is that both predicates are indexed and it returns three columns — not a saved round trip, which it never was.

**Tests (P2).** They still allowed the "or a turn is active" arm to be deleted, and never walked a slow read across a settle. Both covered now, and both verified the only way that counts: broke the code, watched them fail, put it back.

628 files, 4203 web tests, biome and tsc clean by exit status.

The residual gap I flagged last time stands and is not fixed here: Core sets the bot RUNNING in the same transaction that creates the turn, after preparation, so a turn that spends most of its life preparing can still fall between probes.

PatrickTobler:
@coderabbitai full review

PatrickTobler:
Fourth Codex pass. It confirmed the queue, the timeout cleanup, unmount behaviour, the no-redirect route and the corrected Prisma comment are all sound, and found three smaller things. All addressed in 19b1074.

**The probe had no deadline of its own.** A stalled proxy would start another every 2.5s — two dozen open requests after a minute, answering out of order. One at a time now, with an eight-second deadline.

Scoping that flag needed care and my first attempt got it wrong: holding it across the state read blocked every later probe for as long as that read took, which is the starvation this change exists to remove. The probe is its own function and the flag covers only the probe.

**`sokoBotService.getActivity` was dead** once the route moved to the non-redirecting client. Removed.

**One test is gone rather than fixed, and I want to flag it rather than bury it.** It claimed to cover the settle-during-a-slow-read race that the queued follow-up exists for. Codex was right that it was too weak, and when I rewrote it — including a deterministic version driven by a promise I controlled rather than timer arithmetic — it still passed with the queue taken back out. So it asserted nothing. A test named after a race it cannot detect is worse than no test, because the next person trusts it.

**So: that race is fixed and reasoned about in the code, and it is not covered by a test.** The other six are, and each was verified the only way that counts — break the code, watch it fail, put it back.

628 files, 4203 web tests, biome and tsc clean by exit status.

PR #4122, opened 2026-09-03
docs(web): sharpen the notification kinds copy

schaier-io:
Layer 12 of SOK-877. Based on [#4096](https://github.com/masumi-network/sokosumi/pull/4096); review that one first.

## Why

Every hint under a kind said what the kind reports, not what it means for the reader. "Anything else a job reports, like finishing or failing" describes the system. It does not answer the only question the row asks: do I want to be interrupted for this?

## What changes

| Row | Before | After |
| --- | --- | --- |
| Jobs waiting on you | Jobs that need you | A job needs your input, or its payment failed. |
| Other job updates | Anything else a job reports, like finishing or failing. | Anything else a job reports, like finishing or failing. Nothing for you to do. |
| Tasks waiting on you | A task stopped until you act, like needing approval or credits. | A task needs your attention, like an approval or more credits. |
| Mentions | Chat mentions | Someone writes your name in a message, or mentions everyone in a room. |
| Access requests | Requests and access | Someone asks for access to a workspace you manage. |

`Chat mentions` becomes `Mentions`, since the group above it already says Chat. Job and task labels now read the same way as each other; before, one "needed you" and the other "waited on you", which suggested a difference that does not exist.

## Three claims the old copy made that Core does not honour

**No answer notification exists.** The access hint used to promise "or a request you made is answered". Core raises `SYSTEM` in exactly two places, both a pending request reaching the people who can approve it: `vendor-grants.ts:287` and `coworker-workspace-access.ts:590`. Approving or denying calls `deletePendingVendorGrantNotifications` / `deletePendingCoworkerAccessNotifications`, which delete the pending row and write nothing new. No `*_APPROVED` or `*_DENIED` message key exists anywhere.

**Attention does not mean blocked.** A failed payment ends a job: `PAYMENT_FAILED` shares the `FAILED` branch in `job-sync-state.service.ts:366-368` and no route resumes it. `Notifications.Task.scheduleRemovedByOperator` reports a schedule an operator already removed. Both are worth a look and neither waits on an action, so no attention hint says the kind is waiting to be unblocked.

**Only one row can list its causes closed.** `JOB_ATTENTION_MESSAGE_KEYS` is exactly `inputRequired` and `paymentFailed`, and that hint names exactly those two. `TASK_ATTENTION_MESSAGE_KEYS` carries five and both update rows carry four, so those hints name an example instead of a list a reader could hold Sokosumi to.

Two more accuracy fixes: the mention hint now covers `@all`, which `resolveMentionedUserIds` turns into a mention for every room member whether or not their name appears, and the room-message hint says "anyone posts", since a coworker post notifies the room with no author user behind it.

## Verification

- `pnpm --filter web test` — 4316 passed, 1 skipped
- `pnpm turbo typecheck` — 19/19; `biome check .` — clean
- Nine adversarial review rounds; every copy claim above was traced to the Core code path that produces the notification


schaier-io:

PR #4123, opened 2026-09-03
feat(web): make each notification channel its own choice

schaier-io:
Layer 13 of SOK-877, top of the stack. Based on [#4122](https://github.com/masumi-network/sokosumi/pull/4122); review that one first.

## Why

The row offered one control with three stops: off, in app, or a banner as well. That reads as a loudness dial. The matrix underneath stores a cell per channel, so the control described the data as something it is not, and it hid the channels the reader can actually be reached on.

## What changes

**Each channel is its own chip.** Nothing pressed means the kind does not arrive, so the separate off stop is gone. The group presets are unchanged: Everything, Important, Quiet and Off still write whole groups at once, and Custom still reports a group set kind by kind.

**A push carries the in-app entry with it, in both directions.** Turning Push on turns In app on; turning In app off turns Push off. The feed and the unread badge both read the in-app cell, so a push on its own would arrive once and leave nothing the reader could find again. The entry on its own stays available, which is the quiet answer.

**`Banner` becomes `Push`**, matching the account switch further up the page that gates it.

**Email joins the row and cannot be pressed.** It is one switch for the whole account rather than a cell per kind, so a chip that wrote here would control nothing. It stays a focusable button marked `aria-disabled`, so a reader who never uses a mouse still finds it and hears why it is not theirs to set.

What that button claims depends on the kind. Core mails job status and nothing else: `notificationsOptIn` gates only the two job emails in `job-sync-state.service.ts:119,157`, and no task, chat or system notification has any email path. A chip saying "set for your whole account further up this page" on a mention row pointed at a switch that would never mail a mention, so those rows say plainly that no email is sent and strike the word through, which is the only signal a sighted keyboard or touch reader gets.

## Each row announces where it now arrives

The chips read as two independent choices and one moves the other, so a reader who cannot see the row would hear only the chip they pressed. Each row carries a live region. Getting it to speak exactly once, and only truthfully, took most of the review:

- **Not on mount.** Radix does not render collapsed content, so this matters when a group opens.
- **Not while the write is in flight.** Turning a push on waits on the account consent prompt, which waits on a person, and the row holds its old channels for all of that time.
- **Not on a preset**, which writes every row in a group at once under its own control. A row that spoke there would report one row of three as if it were all that moved.
- **The sentence comes down when someone else writes the row.** A region speaks when its text changes, so a sentence left standing would contradict the chips and then swallow the next press that lands back on the channels it names, which is exactly the press that pulls its sibling along.
- **One template for on and off.** It names the chips rather than claiming the kind arrives or does not, because email is not a chip and a job row keeps mailing after both chips go off.

## One thing this PR does not fix

`CHAT` is a browser-only kind, so a chat notification never reaches the in-app feed whatever its in-app cell says ([#3650](https://github.com/masumi-network/sokosumi/pull/3650)). The chips still write the cell honestly and no hint here promises an entry that waits, but **In app on the three chat rows currently delivers nothing visible**. That predates this work and this page is the first UI to offer it as a choice. Either chat should enter the feed or those rows should not offer In app; both are product calls outside this PR.

## Verification

- `pnpm --filter web test` — 4316 passed, 1 skipped, including 11 new cases across `notification-delivery.test.ts` and `notification-kinds.test.tsx`
- `pnpm turbo typecheck` — 19/19; `biome check .` — clean
- Mutation-checked: claiming the account switch on every kind, an Email chip with no strike-through, a live region that never speaks, one that speaks during the write, one that speaks on someone else's write, and one that never takes its sentence down. Each fails exactly its own test and nothing else.
- Nine adversarial review rounds. The last two found no functional defect.


schaier-io:

PR #4124, opened 2026-09-03
feat(web): fold the job emails into the notification grid

schaier-io:
Part of SOK-877, layer 14 of the stack. Stacked on #4123.

The account's job status emails stop being a switch below the card and
become the Email column of the grid, on the rows Sokosumi actually mails.
A kind Core never mails shows a dead cell in that column rather than a
hole, so the row still says what email would mean there.

Core mails the job owner on COMPLETED, REFUND_RESOLVED, DISPUTE_RESOLVED
and INPUT_REQUIRED only, and the copy is written to that.

## Verification

- `pnpm --filter web test`
- `pnpm typecheck`

## Known follow-up

Copy fixes still open on this layer: `deliveryAriaLabel` disagrees in number
on most kind rows, the German `channelEmailSoonLabel` word order, and the
German `channelEmail` mixes non-breaking and ASCII hyphens.


schaier-io:

PR #4125, opened 2026-09-03
feat(web): fold push and marketing into the notification grid

schaier-io:
Part of SOK-877, layer 15 of the stack. Stacked on #4124.

The card's two remaining switches go away. Push notifications and Marketing
emails are rows and columns of the one grid now, so a reader answers what
Sokosumi tells them about, and where it reaches them, in one place.

- The Push column asks the browser itself. A cell stays pressable on a
  browser that cannot show a banner, because the preference it writes is
  account-wide, and the reason rides in the cell's description.
- Marketing is a row of the grid, by email only. The two columns it cannot
  use say why rather than leaving a hole.
- The job status emails keep a row of their own only when no kind row
  carries them.

## Verification

- `pnpm --filter web test` — 631 files, 4345 passed, 1 skipped
- `pnpm typecheck`, `pnpm check`

## Known limits

- Nothing in the web app writes `pushOptIn: false` any more, so a browser's
  subscription is released by sign-out rather than by a control.
- A browser whose subscription is gone while permission is still granted
  shows a filled Push cell until the next press resubscribes it.


schaier-io:

PR #4141, opened 2026-09-04
feat(web): give completed tasks their own row and toggle

schaier-io:
"Other task updates" and "Other job updates" each held finishing behind the
same switch as failing and being canceled, and finishing is the one a reader
started the work for. Silencing the rest silenced it too.

Completed tasks and Completed jobs are now rows of their own, each between
the kinds that wait on you and the rest. Important keeps them: that stop no
longer means only what waits on you, so its own words and Quiet's say what
matters instead. Both update rows drop finishing from the examples they name,
because they no longer carry it.

The email switch is one answer for the whole account, so the new job row
carries the same cell as the two beside it and moves with them.

---


schaier-io:

PR #4143, opened 2026-09-04
feat(web): ask each notification group two questions instead of one

schaier-io:
The stops were one ladder answering two questions. Everything to Important
dropped kinds, Important to Quiet dropped a channel, so no stop told the
reader what the next one would do, and the two words came from different
families. One wish could not be written at all: every kind, without a
banner. That took six cells by hand.

Each group now answers on two rows. The first says which of its kinds
arrive: Everything, Important, or Nothing. The second says where the ones
it keeps arrive: in app, or in app and push. Which and where stay apart, so
pressing Important leaves a group where it already was, and pressing In app
keeps every kind that is on rather than dropping the quiet ones.

Every stop explains itself on hover, the way the channel cells do, and
carries the same sentence in an sr-only line that aria-describedby can
always reach. Custom stays what it was: not an answer, only the fold's
trigger, and only when the kinds are set one by one. A group whose kinds
are all important still hides Important, because it would write what
Everything writes.

The rails take the shapes of the row they sit in: the box's radius outside
and the cells' inside, rather than the pills they started as.

---


schaier-io:
Replaced by #4149. Renaming the branch off the abandoned two-question design deleted this PR's head ref, which closed it.

PR #4149, opened 2026-09-04
feat(web): ask each notification group one question

schaier-io:
A group carried two rails: which of its kinds arrive, and where they arrive.
Side by side they read as one control with two halves, and the second one
asked a question the grid under it already answers, one kind at a time.

A group now asks one question. Where a kind arrives stays the grid's, so an
answer keeps every kind it names on the channels that kind is already on: a
group the reader quieted to the app does not start buzzing when they ask for
everything. A kind that arrives nowhere takes the channels the rest of its
group uses, and a group that is silent everywhere comes back on every
channel, since a write that stores nothing would leave the group looking
unchanged.

The answers sit on a rail from `md` up, and in a dropdown below that, where
the rail no longer fits. Each one carries an icon that is not a channel's, so
a stop cannot be read as one of the columns. Custom is not one of the
answers: it reports a state rather than offering one, so it is the fold's own
trigger beside the rail.

A browser that will not show a push now says so once, in a banner over the
rows, rather than only inside the cells asking for one. It appears only while
a kind is asking for a push, carries the button for the single case a page
can fix, and says the reader's other devices may still be getting these. This
browser can read the account consent and the cells, never whether another
browser still holds a subscription, so it says may.

Verification: `pnpm --filter web typecheck`, `pnpm --filter web messages:parity`,
and `pnpm --filter web exec vitest run "src/app/(app)/account"` (143 passed).

Replaces #4143, which GitHub closed when the branch was renamed off the
abandoned two-question design.

---


schaier-io:

PR #4016, opened 2026-08-28
feat: CLI OAuth login (SOK-948, SOK-949)

Sarthib7:
## Summary

Small OAuth slice only. Developer CLI can sign in. Nothing else.

- SOK-948: Core seeds first-party public client `sokosumi_cli`
- SOK-949: `apps/cli` (`sokosumi-cli` / `sokosumi`) — `auth login`, `auth logout`, thin Ink status, keychain tokens

This PR does **not** include coworker connect, vendor/workspace commands, schema, or web onboarding.

## Approach

- Home: `apps/cli` in this monorepo. Sibling `sokosumi-cli` is read-only.
- Signup/login happen on web `/signin` during the authorize hop.
- Tokens stay in the OS keychain. No API key in this slice.

## Test plan

- [x] utils first-party client payload test
- [x] Core upsert + sync route tests
- [x] `node --test apps/cli/test/**/*.test.mjs` (17 passing)
- [ ] After deploy: seed `sokosumi_cli` in the target DB
- [ ] Manual: `pnpm --filter sokosumi-cli start` / `auth login` against preprod

PR #4201, opened 2026-09-07
feat: seed first-party sokosumi_cli OAuth client (SOK-948)

Sarthib7:
## Summary
- Seed a platform-owned native public OAuth client (`sokosumi_cli`) so the Developer CLI can log in without creating a client in `/developer`.
- Upsert it on Core boot (local), Vercel build (`pnpm ensure:cli-oauth`), and a cron (`GET /sync/first-party-oauth-clients`).
- Record the OAuth-first sequence in ADR 0027.

## Test plan
- [ ] Utils + Core sync tests for the first-party client upsert
- [ ] After this deploys, `GET /auth/oauth2/authorize?client_id=sokosumi_cli&…` must not return `invalid_client` / `client_id is required` (that error is Better Auth for “client not found”)
- [ ] Cron `/sync/first-party-oauth-clients` leaves a native PKCE row with loopback redirects and `userId` null

Stacked under: SOK-949 CLI login PR (opens next). Replaces the seed half of #4016.


mrosberghaus:
As discussed closed.

PR #4205, opened 2026-09-07
feat(core): let a chat room message reach the notification center

schaier-io:
A reader who turns on "every message in a channel" now sees those messages in the notification bell, and a busy room takes one row rather than one row per message.

## The bell

`CHAT` is a browser-only kind: a mention belongs beside the room it was written in, not in a list of everything that happened. A room message is the other way round. The reader asked to be told about every message, and the feed is the only surface that keeps what it was told. So the feed clause now carries an exception for the room-message key, applied in one place (`notificationFeedWhere`) that list, unread count and mark-all-read all share, so a call site cannot apply one rule and forget the other.

## The count

Twenty messages in a minute made the notification centre that room and nothing else, and the rows the reader kept it open for were pushed off the first screen. Core now counts a message onto the reader's unread row for the room instead of writing another, and the web reads a counted row as **"23 messages in Design"**.

- Read is where a group ends. The next message after the reader has been to the room starts a row of its own, so the count means "since you last looked".
- The stored message key stays the one-message key. The OS banner renders from the same row and is about the message that has just arrived, and the push service worker interpolates plain params with no plural rules. Only the feed swaps in the counted key.
- A mention and a direct message each stay a row of their own. They are about themselves, and counting them onto a room's row would lose which message named the reader.

This is a deliberate deviation from the append-only write `createNotification` documents: the row's params, its metadata and its place in the feed all move, because the row is about the room over time rather than about one message. Two messages landing at once can both find no row and write one each; the next message counts onto the newer of them, so the reader sees one row too many rather than a lost message.

## Naming a group direct message

A direct room of three or more people is an ordinary room to this setting: the direct-message row stops at two humans, so the every-message setting is what covers it. Such a room has no name of its own, so Core stores the member list (`buildDirectRoomName` joins up to three names, then "and N more"). Read straight out, that is three people rather than somewhere a message was written, and the reader's own name is in it.

Core now says which rooms those are, and the feed names them as a group: **"23 messages in group Ada, Ben, Cara"**, "Ada wrote in group Ada, Ben, Cara". The wording is the feed's, like the count. A banner is about the message that has just arrived, and the push service worker renders the stored key for a tab that was closed; both keep saying "in Ada, Ben, Cara".

## Keeping the badge honest

Counting onto a row means a row can reach a tab twice, which broke three things the append-only model made safe.

Opening a room already cleared its CHAT rows on the server and told nobody. Harmless while CHAT stayed out of the notification center; now it left the bell counting rows the server had cleared, for the rest of the session. The room read publishes each row it cleared, guarded so a failure after the response is reported rather than left as a rejected promise behind `waitUntil`.

The realtime event now says whether it wrote the row or only changed one. A tab counts a row it does not hold towards the badge, which is right for a new row and one too many for a room's later messages. An event from a Core that only ever wrote rows still reads as written.

A counted row's `createdAt` is bumped, so an active room stays inside the ten rows the bell holds. The live update replaced the row where it sat, which showed the same rows the server would return in a different order until the next fetch. The list is sorted on update instead, by time and then by id.

## No badge change

The chat badge is unchanged. It counts unread messages in the chat surface; this is the notification centre.

## Verification

- `pnpm --filter core test` — 475 files, 4984 passed, 6 skipped.
- `pnpm --filter web test` — 638 files, 4509 passed, 1 skipped, including the push service worker string parity check across en/de/es.
- `pnpm --filter @sokosumi/utils test` — 64 files, 512 passed.
- `pnpm typecheck` — 19/19. `pnpm check` — clean.

Those runs are the branch on its own. The merge with `main` is what CI builds, and it is green there too.

## Review

An adversarial review found one HIGH and five MEDIUM defects in the first two commits; the later commits are the fixes. Each is mutation-checked: reverting it fails exactly one named test. Two of those runs found tests that did not pin what they claimed (a mock deciding the answer, a stub dropping its arguments), and both are now pinned.

Two later findings about the cleared-row publish were checked against the code. The unguarded background promise was real and is fixed. The claim that `waitUntil` drops the work outside Vercel was not: the promise is created before it is passed, so it runs to completion under `@hono/node-server`, and `waitUntil` is `getContext().waitUntil?.(promise)` — a no-op outside a Vercel context, not a cancellation. The same pattern already carries the unfurl and notification work in this route family.

## Merged with main

#4199 moved in the opposite direction on the same three chat message keys. It landed on `main` and added an import of `CHAT_DIRECT_MESSAGE_MESSAGE_KEY`, `CHAT_MENTION_MESSAGE_KEY` and `CHAT_ROOM_MESSAGE_MESSAGE_KEY` from `@/helpers/notification-delivery`; this branch had moved their declarations to `@sokosumi/utils`, so that module imports them rather than exporting them.

Neither side is wrong alone, and there is no text conflict, so it only appears in the merge commit CI builds. The test now reads the keys from the package that owns them. A re-export from `notification-delivery` would work too, and is the passthrough layer the repo asks us not to add.


schaier-io:
@coderabbit please review

PR #4207, opened 2026-09-07
feat(web): show an opt-in unread message count on chat rooms

schaier-io:
Closes SOK-978.

## What a reader sees

A chat sidebar row can now show how many unread messages a room holds. It is off
for everyone. A reader turns it on under Account > Notifications, in a new "Chat
display" card, and the sidebar updates without a page reload.

Nothing changes for a reader who does not turn it on. No delivery preference is
touched: this changes what the sidebar shows, never which notification is created
or sent.

The number is the same `unreadCount` that already drives the row's bold, so the two
cannot disagree. It sits left of the mention badge and stays separate from it,
because the badge counts user mentions and direct messages and a reader relies on
that distinction. The count is capped at `99+`. It is suppressed on the active room
and on a muted room, by the same early return that already suppresses bold.

## How it is wired

The preference is a Better Auth additional field, so the session carries it and a
sidebar row reads it with no request. It is also on the Core preferences route, so
`GET`/`PATCH /v1/users/{id}/preferences` reads and writes it like the flags beside
it.

The settings card writes through `authClient.updateUser`, which better-auth lists in
its `$sessionSignal` matcher (`dist/client/config.mjs:76-80`), so the session refetches
and the sidebar re-renders on its own.

## Schema change

Migration `20260907170000_user_show_room_unread_count`:
`ALTER TABLE "user" ADD COLUMN "showRoomUnreadCount" BOOLEAN NOT NULL DEFAULT false`.
Metadata-only on PostgreSQL 11+, safe on a live table. No data migration script.

## Docs

- New [ADR-0027](docs/adr/0027-room-unread-count-is-a-reader-opt-in.md). It answers
  ADR-0026's rejected "numeric unread on the row" head-on: 0026 rejected a number
  shown to everyone by default, and that rejection still stands.
- ADR-0026's Rejected line now points forward, so a reader who finds only 0026 does
  not read it as a blanket ban.
- `CONTEXT.md`'s **Room unread** entry names the new opt-in.

## Note on the generated client diff

`src/lib/clients/generated/core/sdk.gen.ts` carries three unrelated notification
description changes. They come from #4205, which changed the Core schema without
regenerating the web snapshot. Regenerating here picks them up. They are not a
semantics change from this PR.

## Review

Six review rounds ran across this branch. Round 3 proved the tests by mutation:
eight source lines were broken one at a time and every one was caught by a test
that names that behaviour. Each later round found real defects, and each was
fixed before the next round ran.

Defects the reviews caught in this PR's own code:

- A stale Core fixture missing the new required response field.
- An `aria-label` on a bare `span`, whose implicit role is `generic` and so
  prohibits an accessible name. The count now carries real `sr-only` text with
  the digits marked `aria-hidden`.
- `text-sm` on `CardTitle` silently deleting the primitive's `leading-none`
  through tailwind-merge.
- A `CardHeader` gutter drawn to an empty second grid row.
- A heading demoted onto `text-muted-foreground`, which is 3.7:1 on card ground
  in light mode and fails WCAG AA for small text.

Two findings were left alone as pre-existing shapes this change follows rather
than introduces. They are written up in the PR discussion.

## Design revisions after the first push

Four rounds of UI feedback landed after the feature itself:

- The count moved out of the row's right rail onto the end of the room name.
  The name keeps `min-w-0` and truncates first, so the count stays visible.
- The count now draws at the room name's unread weight. It has no unbolded
  variant, because `resolveRoomAttention` cannot report a count above zero
  without also reporting bold. That was a reading of the code, so it is now a
  sweep test over every input able to produce a count.
- The chat display card moved inside the notification card, through a `children`
  prop so the page stays the composer and the notification card never imports
  the chat one.
- That card lost its muted fill, which had made it read as a louder surface
  rather than a quieter one, and its headings were demoted by size and weight.
- The setting's description lost the clause that repeated its own label.

## Verification

All run at `8c0a95263`:

- `pnpm --filter web test` — `Test Files 643 passed (643)`, `Tests 4583 passed | 1 skipped (4584)` VERIFIED
- `pnpm --filter @sokosumi/utils test` — `Test Files 64 passed (64)`, `Tests 512 passed (512)` VERIFIED
- `pnpm --filter core test` — 474 passed, 2 skipped, **1 failed**: `sync routes > returns 401 for missing cron auth`, `Error: Test timed out in 5000ms`. Re-run alone it passes: `Test Files 1 passed (1)`, `Tests 38 passed (38)`. This diff touches no file under `src/routes/sync`. REPORTED as a load flake, not investigated further.
- `pnpm typecheck` — `Tasks: 19 successful, 19 total` VERIFIED
- `pnpm check` — `Checked 3811 files. No fixes applied.` VERIFIED
- `pnpm --filter web messages:parity` — `de: ok`, `es: ok` VERIFIED

CI on this branch is green: all 14 checks SUCCESS.

## Screenshots

**Not attached, and not producible from this environment.** `verify-sokosumi doctor`
returns `doctor fail`: there is no `apps/core/.env` or `apps/web/.env` in this
worktree or in the primary checkout, Core and Web are both unhealthy,
`verify_credentials_email=unset`, `verify_credentials_password=unset`, and
`agent_browser=missing`. Producing them needs an env bootstrap, a database, and a
migration run, plus an account to sign in with.

Every layout decision here was therefore made by reading the Tailwind classes and
resolving the merge order, not by looking at the result. The sidebar row was
measured in a browser during review (the name clips at scrollWidth 345 against
clientWidth 141, and row height is unchanged at 37px), but the account surface
was not. Worth a human eye before merge.


schaier-io:
Two review findings are left alone on purpose. Both are pre-existing shapes this change follows rather than introduces, and fixing either means editing files outside this PR.

**1. A Core `PATCH /v1/users/{id}/preferences` write cannot reach the reader's UI for up to five minutes.** Core enables `session.cookieCache` (`apps/core/src/lib/auth.ts:283-286`, default 300s in `apps/core/src/config/env.ts:60-63`). The route writes the column with Prisma and never refreshes the session cookie, while the sidebar reads only the cached session. Better Auth's `/update-user` does refresh it, which is why the settings card is unaffected. Nothing in web calls the Core route for this field today, so this is a public-v1-API hazard only. `marketingOptIn` and `notificationsOptIn` have the same shape on the same route.

**2. Counts appear a moment after a cold load.** `useSession()` resolves client-side and returns `null` on first render, so an opted-in reader gets a count-free paint first. There is no session seed anywhere in web: twelve production call sites use `useSession()` the same bare way, including the header profile section on every route. Fixing it for one sidebar row alone would be inconsistent.

**3. Untouched, noted for completeness.** `MentionBadge` still holds a bare `99` literal where the new code uses a named `ROOM_UNREAD_COUNT_CAP`, and it still carries an `aria-label` on a bare `span`. The badge is outside this change.

PR #4218, opened 2026-09-07
fix(chat): acknowledge messages only in visible rooms

schaier-io:
VERIFIED: Automatic room and thread reads now require a visible document. Hidden updates wait until visibility or focus returns. The read hook retries failed thread reads and rejects stale async follow-up writes. See `apps/web/src/app/(app)/chat/hooks/use-room-read-attention.ts`.

VERIFIED: Read tracking observes persisted message content, including earlier replies that finish without changing their ID. Local streaming token overlays do not issue repeated read requests. See `rooms-client.tsx` and `utils/room-read-attention.ts` in the same chat feature.

Validation (local):
- VERIFIED: Combined Web chat suite: `Test Files 54 passed (54)`, `Tests 385 passed (385)`.
- REPORTED by the focused test run: the token regressions failed before the persisted-message boundary fix, with three read calls for three token updates. The restored fix passed.
- VERIFIED: `pnpm precommit`: `19 successful, 19 total`.

Validation uses component tests with simulated document visibility. It does not prove live browser delivery.


schaier-io:
@coderabbitai review

PR #4224, opened 2026-09-07
feat(web): register native OAuth clients as public PKCE

mrosberghaus:
## Summary

- Native OAuth clients (loopback HTTP or RFC 8252 `com.example.app:/callback`) now register as public PKCE: `token_endpoint_auth_method: none`, no client secret.
- HTTPS web clients stay confidential.
- Create dialog shows a public-client note when Better Auth returns no secret. Help copy (en/de/es) says native clients are public PKCE.

## Why

Better Auth treats `application_type` as redirect-URI policy only. Native registration still defaulted to `client_secret_basic`, so a Mac/iOS client got a secret it cannot keep.

Follow-up to #4203.

## Verification

- `pnpm --filter web test src/app/(app)/developer/components/oauth-clients/utils.test.ts src/app/(app)/developer/components/oauth-clients/hooks/use-oauth-clients.test.ts`
- Create a client with `com.sokosumi.app:/auth`. Expect no secret; success note about public PKCE.
- Create a client with `https://example.com/callback`. Expect a client secret as before.


mrosberghaus:
Close in favor of #4222

PR #4233, opened 2026-09-08
ci(workflows): skip JS checks on Apple-only changes, add Xcode build

mrosberghaus:
Gate Build, Tests, and Lint on JS-relevant changes via dorny/paths-filter (predicate-quantifier every). Apple-only and docs-only PRs skip the Next/Hono jobs; Apple changes get Xcode verification on macos-15 via new Apple workflow.

Verification:
- All four workflow files parse as YAML
- Filter semantics (every + negations, dot:true) tested against real picomatch: 12/12 path cases pass
- Xcode scheme/destination validated with xcodebuild -showdestinations locally

Branch protection: existing required checks can stay as-is (skipped counts as passing). Optionally require Apple / Xcode build.

## Summary by CodeRabbit

* **Chores**
  * Added automated macOS arm64 builds for Apple-related changes.
  * Optimized build, lint, type-check, and test workflows to run only when relevant JavaScript files change.
  * Added manual workflow dispatch support for on-demand validation.


mrosberghaus:
@coderabbit review

PR #4257, opened 2026-09-08
feat(web): put the chat message on the notification banner

schaier-io:
## What this changes

A chat banner said "Sokosumi" on its first line and "Alice wrote in Design" on
its second, so a reader had to open the app to learn what arrived. The operating
system prints the app name beside the banner already, so the title spent its one
line saying it twice.

Now a chat banner is titled with who wrote and where, and the message itself
goes underneath:

| Before | After |
| --- | --- |
| **Sokosumi**<br>Alice wrote in Design | **Alice in Design**<br>can you look at the login flow |
| **Sokosumi**<br>Alice mentioned you in Design | **Alice mentioned you in Design**<br>your call |
| **Sokosumi**<br>Alice sent you a message | **Alice**<br>are you free? |

A direct message is titled with the author alone, because a person's name is the
same in every language.

Every other notification kind keeps the app name and its line: a title is cut
shorter than a body on most platforms, and a task name is long. So does a chat
message with no text to show, such as one that is only formatting. "Alice" on
its own says less than "Alice sent you a message", and the second line is the
only place left to say it.

The push service worker renders the banner for a closed tab. It ships as a plain
script and cannot import app code (ADR-0023), so it carries the same rule and
the same two new strings. Tests on both sides assert the same cases.

The existing parity check compares the worker's copy of every string against
the catalog in all three locales. It asked whether each string was somewhere in
the worker's block for that locale, which cannot see two strings that swapped
keys: both are still there. The two new titles differ by one clause, so a swap
would title a channel message as a group one and pass. The check now reads the
block as the map it is and compares key by key.

## User-facing impact

Anyone with chat banners on now reads the message from the lock screen. There is
no opt-out: a reader who does not want message text turns the chat banner off.

## Verification

- `pnpm --filter web test` — 4570 passed, 1 skipped, 642 files. Fourteen
  failures across six files this suite is known to flake in under load, none
  of them touched by this change; all six pass when re-run together on their
  own, 96 passed
- Both renderers asserted on the same cases: title choice, group room, direct
  message, missing author, empty author on a room message and on a mention,
  missing text, empty text, non-text preview, a non-chat kind, and a non-chat
  kind carrying chat's own params
- Each of those asserted cases mutation-checked on both sides, because a rule
  written twice is only held together by tests that fail on both
- The German titles asserted on the rendered banner, not only in the catalog:
  swapping the two German strings in the worker, and swapping them in
  `de.json`, each fail two tests
- The two title keys pinned to their literal paths. Both name a key the
  catalog has, so pointing one at the other passed every other check while the
  app titled a channel message as a group one and the worker did not
- The direct-message case asserted with a room name that differs from the
  author, since a direct room is named after the people in it
- `pnpm typecheck`, `pnpm check`

Closes [SOK-980](https://linear.app/masumi/issue/SOK-980/show-the-chat-message-text-in-push-notifications).

## Summary by CodeRabbit

- **New Features**
  - Chat push notifications now display titles identifying the message author and, when applicable, the room or group.
  - Message previews appear in the notification body.
  - Added localized notification titles for English, German, and Spanish.
  - Non-chat notifications continue to use the app title and translated message content.

- **Bug Fixes**
  - Improved notification formatting for direct messages, mentions, room messages, missing authors, and empty previews.
  - In-app pending access requests continue to display their dedicated toast message.


schaier-io:
@coderabbit please review

PR #4266, opened 2026-09-08
feat(web): open a channel's settings from its row menu

schaier-io:
## What changed

The channel row's overflow menu (the three dots in the chat sidebar) now carries **Edit channel**. It opens the same dialog the channel title in the room header opens.

Before this, channel settings had one way in: open the channel, then click its title. The title reads as a heading, so people did not find it.

## How it works

The overflow menu lives in the app sidebar, which holds the room and nothing else. The dialog needs the org roster, the coworkers, and the reader's membership role, and giving those to the sidebar would mean an org-roster fetch on every page.

So the row asks for the dialog on the room's own URL (`/chat/rooms/{id}?edit=1`), and the room shell, which loads all three already, opens it. The parameter is read once and taken back off the URL with `router.replace`, the same shape as the existing `message` deep link.

Two details that the URL route forces:

- **The room must survive both URL writes.** `chatRouteErrorBoundaryKey` keys the chat subtree on pathname plus search, and the boundary is a class component, so an unexcluded parameter remounts the room: once when the parameter arrives, once when it is stripped. The second remount would throw away the state holding the dialog open before the reader ever saw it.
- **One list of what the room reads once and then drops.** The boundary key ignores `message` and `edit` together. Each room-side reader still strips only its own and waits while the other is on the URL: the two sit in one commit holding the same snapshot, so a second write would put the other parameter straight back, and neither reader asks for its own twice, so it would sit there for good.
- **Asking the room on screen for its own dialog is not a journey.** The row uses `router.replace` when the channel is already active, so Back is not left doing nothing visible. It merges the ask into the query it reads from `useSearchParams()`, not from `window.location`, which the App Router writes in an effect after the commit. A pending `message` is left out of that merge: the router's query still names one the room has spent until that strip commits, and writing it back would leave the room waiting on a message nobody spends again, with the ask stuck behind it.

Other decisions:

- Channels only. A direct room's header shows a plain title and has no dialog.
- Permissions are unchanged: the dialog decides what the reader may change, exactly as it does from the header title.
- The entry is wrapped in `SheetClose` under the same flag the room link already uses, so the mobile sidebar sheet closes on navigate.
- Reuses the existing `App.Channels.editChannel` message key, so there is no locale gap.
- `EditChannelDialog` is now fully controlled. The dialog has two ways in and only one of them is a click on its trigger, so the room shell owns the flag.

## A move rides along

`rooms-client.tsx` was 3137 lines, the largest hand-written file under `apps/web/src/app`, and the header component this branch edits sat in the middle of it. The header and the avatar stack it renders move to `room-header-chrome.tsx`, 237 lines out of the room client. Move only, no behaviour with it.

## User-facing impact

One click from any channel row opens that channel's settings. Nothing is removed: the header title still opens the same dialog.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` (turbo, 19 tasks) — clean
- `pnpm --filter web test "src/app/(app)/chat" src/components/chat` — 134 files, 1132 tests passed, on the branch with `main` merged in
- Mutation checks: removing the hook's `open()` call turns 5 tests red; removing the error-boundary line turns its new test red; reverting the row to `window.location.search` turns 2 red; dropping the `message` delete, the trim guard or the channel gate turns 1 red each. Each pins behaviour rather than passing regardless.

No browser proof: this worktree has no `.env`, no database and no sign-in credentials, so `verify-sokosumi doctor` fails here. The Radix wiring is covered against mocked primitives only.

## Review

The branch was reviewed three times. The second round found the ask could wedge shut, and the third confirmed the fix. `4d1a62498` carries those fixes and states each one; the numbers in it were measured, not estimated.

## Tests added

- `use-edit-channel-param.test.tsx` — opens once, strips the parameter, waits for the roster, waits for a pending message but not for a blank one, re-opens on a second ask
- `chat-room-sidebar-row.test.tsx` — the entry navigates, replaces rather than pushes for the active room, reads the router's query rather than the document's, drops a message the room has not jumped to, is hidden for direct rooms, and shows when Leave does not
- `rooms-client-room-header.test.tsx` — the dialog opens for the URL that asks for it, stays shut otherwise, and the ask is ignored on a direct room
- `chat-route-error-boundary.test.ts` — the room key survives the parameter arriving and being consumed

Refs [SOK-982](https://linear.app/masumi/issue/SOK-982/add-edit-channel-to-the-channel-row-overflow-menu)


schaier-io:
@coderabbitai review

PR #4273, opened 2026-09-08
feat(core): add routes to delete and clear notifications

schaier-io:
## What this changes

The notification center could only be read. A reader marked one row read, or
marked them all read, and nothing removed a row.

Two routes join the existing ones:

- `DELETE /v1/notifications/{id}` removes one row and answers with the row it
  deleted, so the caller can correct its own unread count without a second read.
- `DELETE /v1/notifications` clears the center and answers with the number of
  rows it removed.

Both take the session rule the rest of the notification routes take
(`requireOwnerUserContext`): an interactive human session only. A coworker or a
Soko Bot is rejected, so `X-Context-User-Id` cannot delete a person's
notifications.

The single delete looks a row up by id and reader together, and uses that same
pair to delete it. Another reader's notification therefore reads as missing
rather than as forbidden: the answer never confirms that someone else's row
exists. Clear-all is scoped by the same feed rule mark-all-read uses, so a
browser-only kind such as a mention or a direct message is left alone.

Rows are deleted rather than flagged. Core already deletes notification rows for
resolved vendor grants and coworker access requests, and a hidden flag would add
a second condition beside `inApp` on every list, count and mark-all-read query.
There is no undo, and both route descriptions say so.

The row-to-response mapping moves into one helper, because the list, the
mark-read route and the delete route all answer with the same shape.

## User-facing impact

None on its own. This is the API the two PRs above it call.

## Verification

- `pnpm typecheck` (19/19)
- `pnpm --filter core test`: 486 files passed, 5102 tests passed
- New route tests cover the owner rule, the 404 for another reader's row (with
  nothing deleted), and the feed scoping on clear-all

Part of [SOK-983](https://linear.app/masumi/issue/SOK-983).

---

## Summary by CodeRabbit

* **New Features**
  * Added the ability to delete an individual notification.
  * Added the ability to clear all notifications for the authenticated user.
  * Deletion responses now include the deleted notification or affected count.
* **Bug Fixes**
  * Improved notification response consistency, including metadata, read state, identifiers, and timestamps.
  * Unauthorized or missing notifications are handled with appropriate errors.
* **Tests**
  * Added coverage for individual deletion, bulk clearing, successful responses, and unauthorized access.

### DIFF
--- a/apps/core/src/helpers/notification-item.ts
+++ b/apps/core/src/helpers/notification-item.ts
@@ -1,0 +1,29 @@
+import type { Notification } from "@sokosumi/database";
+
+/**
+ * A stored notification as the API answers with it: JSON columns parsed, and
+ * the storage-only fields left behind.
+ *
+ * One copy, because the list, the mark-read route and the delete route all
+ * return the same shape and must not drift apart.
+ */
+export function mapNotificationToItem(notification: Notification) {
+  return {
+    id: notification.id,
+    userId: notification.userId,
+    kind: notification.kind,
+    referenceId: notification.referenceId,
+    eventId: notification.eventId,
+    messageKey: notification.messageKey,
+    messageParams: JSON.parse(notification.messageParams) as Record<
+      string,
+      unknown
+    >,
+    metadata: notification.metadata
+      ? (JSON.parse(notification.metadata) as Record<string, unknown>)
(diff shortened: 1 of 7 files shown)


schaier-io:
@coderabbitai review

PR #4274, opened 2026-09-08
feat(web): call Core to delete and clear notifications

schaier-io:
## What this changes

Wires the two new Core routes into the web app: the generated Core client, the
notifications browser client, and the shared notification state the bell and the
page both read.

Deleting one row reuses the reducer case that already drops a row and corrects
the bell, written for resolved access requests. Clearing empties the list and the
bell together. Both paint before the request returns, as marking read does, and
both read the list again when the request fails, so a row that is still in Core
comes back.

The page pages further back than shared state holds, so it can delete a row the
bell never had. Core answers a delete with the row it removed, so when that row
was unread and shared state did not hold it, the badge drops by one from the
answer rather than waiting for the next read.

## User-facing impact

None on its own. No control calls these yet; the PR above adds them.

## Verification

- `pnpm typecheck` (19/19)
- `pnpm --filter web test`: 659 files passed, 4862 tests passed
- `notification-provider.island.test.tsx` covers the delete, the clear, both
  failure re-reads, and the badge for a row shared state never held

Part of [SOK-983](https://linear.app/masumi/issue/SOK-983).

---

## Summary by CodeRabbit

- **New Features**
  - Added the ability to delete individual notifications or clear all notifications.
  - Notification changes update immediately while the request completes, with automatic recovery if an operation fails.
  - Notification updates remain synchronized across refreshes and real-time events.
  - Improved handling of unread counts and read status during notification deletions and clearing.
  - Browser-only notifications are excluded from the in-app notification feed and unread badge.


schaier-io:
@coderabbitai review

PR #4275, opened 2026-09-08
feat(web): let a reader delete and clear notifications

schaier-io:
## What this changes

A reader could only pile notifications up. Every row in the notification center
now carries a delete control, and a "Clear all" action sits beside "Mark all as
read" on the page and in the bell dropdown.

Clearing confirms first, because there is no undo. The bell owns that dialog
rather than the dropdown, since the dropdown closes as the dialog opens and would
take the dialog with it.

Each surface draws the control its container allows, and both call the same
action:

- On the page the row is a button, so the control is a button beside it.
- In the bell the row is a menu item, so the control is a menu item of its own
  beside it. A button nested inside a menu item is not reachable by arrow keys
  and gives the element two roles.

Each control names the row it removes ("Delete notification: Research Agent
completed Market Analysis"), so a column of identical icons stays usable with a
screen reader.

The list drops a row as the request starts and reads itself again when the
request fails, so a row that is still in Core comes back.

## User-facing impact

Anyone can now remove a single notification, or empty the whole center after one
confirmation. Deleting is not marking read: nothing the notification points at
changes, and there is no undo.

## Verification

- `pnpm typecheck` (19/19), `pnpm check`
- `pnpm --filter web test`: 659 files passed, 4862 tests passed
- New tests: the delete control deletes without opening the row, names the row it
  removes, is offered on a pending access request, and tells the reader when the
  delete fails; the confirmation clears on confirm, deletes nothing on cancel,
  and tells the reader when clearing fails
- Copy added to `en`, `de` and `es`; `pnpm --filter web messages:parity` passes

Closes [SOK-983](https://linear.app/masumi/issue/SOK-983).

---

## Summary by CodeRabbit

* **New Features**
  * Added options to delete individual notifications or clear all notifications from the notification center.
  * Added a confirmation dialog for clearing all notifications, with loading and error states.
  * Added localized English, German, and Spanish messages for notification actions.
  * Improved focus management and keyboard navigation when using notification controls.

* **Bug Fixes**
  * Notification lists now remain synchronized during deletions, clearing, pagination, and real-time updates.
  * Added error feedback when notification deletion or clearing fails.


schaier-io:
@coderabbitai review

PR #4306, opened 2026-09-09
feat(core): expose workspace restoration selection

mrosberghaus:
The Apple client cannot read the workspace preference that web sign-in restores. Add `GET /v1/users/{id}/preferred-organization` using the existing sign-in resolver: valid preferred organization, otherwise personal, otherwise the first remaining organization. The read does not change the preference or active session. Clients must check workspace access first; null also represents no workspace when setup is required.

This is the separately approved Core prerequisite for Apple PARITY slice 02, following #4304. Existing user-route authorization applies (self/admin; agent access remains denied), and responses use `Cache-Control: no-store`.

Scope: endpoint, OpenAPI declaration, authorization/error and fallback coverage, parity status. No dependency, database schema, or Swift changes. `apps/web` remains unchanged as requested, including its generated client.

Web behavior: [workspace switcher](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/components/user-avatar/workspace-switcher.tsx), [workspace activation](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/activate-organization-workspace.ts), [shared sign-in resolver](https://github.com/masumi-network/sokosumi/blob/main/apps/core/src/services/preferred-organization.service.ts).

Validation:
- `pnpm --filter core test`: 492 files and 5,164 tests passed; 3 files / 11 tests skipped.
- `pnpm exec turbo run build --filter=@sokosumi/core`: passed.
- Commit hooks: repo-wide `pnpm check` and `pnpm typecheck` passed.
- Read-only review completed; its documentation finding was fixed.

How to test: call `GET /v1/users/me/workspace-access`, then this GET with an authenticated user bearer token. Confirm valid preferred organization, personal fallback, and remaining-organization fallback match sign-in. Confirm another user's explicit id is forbidden for non-admins. No workspace means setup is required; do not treat null alone as proof of a personal workspace.

## Summary by CodeRabbit

- **New Features**
  - Added an endpoint to resolve the organization restored for a signed-in user.
  - Supports personal workspace, preferred organization, and sign-in fallback scenarios.
  - Returns no organization when the user lacks workspace access or organization membership.
  - Applies existing authorization rules and prevents response caching.

- **Documentation**
  - Clarified how organization resolution and `null` results should be interpreted by consumers.

- **Tests**
  - Added coverage for authorization, user lookup, fallback behavior, resolver failures, and response documentation.


mrosberghaus:
@coderabbit review

PR #4310, opened 2026-09-09
feat(chat): drive chat updates over Ably and pause background polling

mrosberghaus:
Closes [SOK-986](https://linear.app/masumi/issue/SOK-986/use-ably-to-drive-chat-updates-and-pause-background-polling)

## User-facing impact

- A hidden or unfocused Sokosumi tab no longer issues chat reads. The open room, the open thread, the sidebar collections, the archived list, and the mobile unread dot all pause, and one read per data set runs on return only if something went stale while away.
- With Ably connected, the recovery read drops from every 3s (room) / 15s (sidebar, mobile dot) to once a minute. The old cadences remain only while Ably or the open room's channel is unavailable.
- Archiving or restoring a channel, and creating, revoking, accepting, or declining an invitation, now reach every member's other tabs over Ably, refreshing only the affected sidebar collection.
- A lost-continuity re-attach, a failed or suspended channel, or a reconnect after a drop triggers exactly one recovery read of the open room.
- Background room and thread reads move from Server Actions (`text/x-component` POST) to an authenticated GET route, so they no longer queue behind the user's own mutations.

## Design notes

- One client scheduler (`use-chat-refresh-scheduler.ts`) owns foreground detection (visible **and** focused), coalescing, single-flight, deferred needs, and the recovery timer. The three former pollers are consumers of it. This replaces hand-rolled polling with hand-rolled scheduling rather than react-query because chat has no react-query usage and the pause/coalesce rules are the feature.
- Health is a boolean input to the scheduler. The open room reads it from a new `useSelectedRoomChannelHealth` hook (channel attached on a connected client); the sidebar and mobile dot read the SDK-free connection store, so their bundles stay Ably-free.
- The new `chat_rooms_changed` event rides the existing per-user chat control channel, so no Ably capability change is needed. Core publishes after commit with best-effort per-recipient fan-out (shared with membership revoke).
- The first channel attach after mount is not treated as lost continuity, so a room switch does not cost an extra read.
- Off the chat pages no room channels are attached (ADR 0003), so the sidebar there relies on the 60s recovery read plus the lifecycle invalidations. Accepted limit, as in the spec.
- A continuity-lost request that lands while a recovery read is already in flight queues one follow-up read, because the in-flight read may predate the gap.

## Verification

- `pnpm test`: 19/19 turbo tasks green (web 679 files, core 491 files, utils 66 files).
- `pnpm typecheck` and `pnpm check`: clean.
- New tests: scheduler (fake timers, focus/blur/visibility/online), selected-room channel health and continuity, control-channel routing, `chat_rooms_changed` schema, Core publish fan-out and all six lifecycle routes, the messages GET route, and the background fetch helpers. Existing sidebar, mobile unread, and rooms-client suites updated for the new semantics.
- Traffic measurement: see the table below.

### Traffic measurement

The in-browser before/after count could **not** be recorded from this worktree: the reused `apps/core/.env` points at the Neon "Mainnet - Andreas" database, which is ten migrations behind this branch, so every sign-in returns 500 (`ColumnNotFound`). I did not migrate that database from a worktree. Please run the measurement on a migrated database (the fallback path works with placeholder Ably keys; the healthy path needs real keys) and paste the numbers here.

Derived from the cadences in code, automatic reads per data set over a 5-minute window (one open room, sidebar with an organization, mobile dot mounted):

| Condition | Before | After |
| --- | --- | --- |
| Foreground, Ably healthy | room 100 (3s), sidebar 3×20 (15s), mobile 20 (15s) = **180** | room 5, sidebar 3×5, mobile 5 = **25** (+ event-driven reads only when something changed) |
| Foreground, Ably unavailable | 180 | 180 (fallback cadences unchanged) |
| Hidden tab | room 0, sidebar 60, mobile 20 = **80** | **0**, then one read per stale data set on return |
| Visible but unfocused window | 180 | **0**, then one read per stale data set on return |

Every automatic room/thread read also changes from a Server Action POST to a GET.


mrosberghaus:
Follow-up from the SOK-986 review: the pause / coalesce / GET / lifecycle design was already right. Four Important issues would have added HTTP this ticket exists to remove, so they are in `eed963f3c`.

### 1. Two control-channel islands were both firing

`ChatControlListBridge` is mounted from the sidebar list **and** from `ChatMobileBottomNav`. That is intentional: on mobile the sheet unmounts the list, so the nav has to keep the control channel for the unread dot; on `/chat/rooms/:id` the nav is gone, so the list has to keep it.

The nav is only `md:hidden`. The bridge is a sibling, so on desktop (and with the mobile sheet open) both islands subscribed. Ably delivers once per subscriber, both handlers dispatched `organization-chat-rooms-changed`, and the sidebar’s in-flight lock turned that into **two sequential collection GETs** per archive / restore / invite.

Fix: keep both mounts (coverage still needs them). `useChatControlChannel` now holds one subscribe per user while any island is live, and only the first listener handles the event. Last unmount detaches.

### 2. First Ably connect was treated as a drop

Sidebar and mobile unread set `refreshOnMount` **and** `refreshOnRecovery`. The health store starts `false`. Mount fetched immediately; then `LazyAblyProvider` loaded, `false → true` fired a second fetch of every collection plus the unread GET.

The open room already ignores first attach. Spec cadence for the sidebar is “60s while connected,” not “fetch again the moment the socket comes up.”

Fix: `refreshOnRecovery` only runs after this instance has already been healthy. First connect is not a gap; reconnect after `disconnected` / `suspended` still is.

### 3. Mobile unread ignored workspace identity

The unread hook is app-chrome lifetime with a constant scheduler key (`chat-tab-unread`). The sidebar keys `${organizationId}:${currentUserId}` and discards in-flight rows on switch. After an org switch the unread `isCurrent()` stayed true, so an in-flight GET could apply the previous org — and the dot could sit on old rooms until the next timer (60s when healthy).

Fix: key is `${org}:${user}:unread`. Switching workspace cancels the in-flight read, clears rows that belong to the previous org, and fetches once for the new one.

### 4. `chat_rooms_changed` publish could 500 after commit

Per-recipient Ably failures were already `allSettled`. `getRestClient()` throwing was not, and the six lifecycle routes `await` the helper with no outer catch. Spec: a publish failure must not fail the request.

Fix: `publishChatRoomsChanged` swallows setup failures (logged, not thrown). The HTTP response after commit stays 200/201/204.

### Left for later (review Minors)

- Payload Zod still lives in web, same as the SOK-742 revoke event.
- Dead `response.redirected` check next to `redirect: "error"` in `fetchBackgroundJson`.
- Health store stays `true` if the last publisher unmounts.
- In-browser traffic table in the PR body is still derived from cadences, not measured.

Happy to take those in a follow-up if wanted.

PR #4313, opened 2026-09-09
feat(apple): preserve room timeline state and scroll position

mrosberghaus:
## Room timeline — parity slice 04

Older pages now merge by message ID in chronological order, preserving live rows and preventing duplicate messages when history overlaps. Timeline state, cursor and generation checks live in the portable `RoomTimeline` module. Failed history requests retain their retry cursor; late responses cannot populate another room.

The Mac timeline uses native scroll-position IDs and explicit follow-latest intent. Reading older history stops automatic following; Latest messages returns to the bottom. Automatic pagination requires actual user scrolling and an available older page, so resizing does not unpin the reader. Inline Retry targets the failed page. Existing grouping, timestamps, day separators, membership events and edited/deleted labels remain, with shared avatar loading now used for human, coworker and bot senders.

### Web references

- [Timeline](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/components/rooms-client.tsx)
- [Message rows](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/components/room-message-row.tsx)
- [Follow-latest behavior](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/hooks/use-stick-to-bottom.ts)
- [History loading](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/load-room-messages.ts)

### Scope

Includes paginated room history, timeline state ownership, row presentation, error/retry, and scroll behavior. Read/attention changes, composing, realtime recovery, reply threads and rich text/attachments remain later PARITY slices. No web files, API contracts or dependencies changed.

### Automated verification

- Xcode build and all 41 app tests passed.
- Shared package tests passed: chat 101, auth 34, CoreAPI 1, realtime 8 — 185 total tests.
- SwiftFormat lint and strict SwiftLint passed.
- iOS 17 package compilation passed: `swift build --package-path Packages/SokosumiChat --triple arm64-apple-ios17.0 --sdk "$(xcrun --sdk iphoneos --show-sdk-path)" --scratch-path /tmp/sokosumi-chat-ios17`.
- Xcode project/scheme Sokosumi, Debug, macOS arm64, `-skipPackagePluginValidation`, coverage disabled. Review build/test used `/tmp/sokosumi-review-tests`, `DEVELOPMENT_TEAM= CODE_SIGN_IDENTITY=-`. Interactive builds retain the separate Apple Development identity.
- Independent reviewer identified the resize/unpin case; corrected and re-reviewed without remaining findings. Regression tests cover resize, pagination overlap/repeated cursors, failed older-page retry and stale room responses.

### Native verification

The user confirmed scrolling up/down and Latest messages returning to the bottom, then explicitly accepted native scrolling as working. Production GET and signed-app workspace restoration also passed. The agent did not independently complete every native scenario; the existing scroll implementation is retained per review and user direction.

### Review follow-up

`RoomTimeline.loadPage` owns loading flags and error state and rejects concurrent page loads. Generation-safe cleanup handles success, failure and cancellation. WorkspaceState retains authentication handling and read marking. Room ID, cursor, pagination and loading flags are read-only externally. Added regression coverage for flag lifecycle, overlap rejection, cancellation, and latest-page refresh preserving older pagination. Updated the stale history comment. The 185-test total is verified from package logs and the app test result.

Review testing: open a room, scroll older history, refresh, and retry a failed load; loading indicators must settle while resolved messages remain. Direct package tests exercise these lifecycle transitions without WorkspaceState.


apps/apple/Sokosumi/TranscriptView.swift
```diff
@@ -3,434 +3,467 @@ import SokosumiAuth
...
+                scrollIntent.beginAutomaticOlderPage(
+                  userIsScrolling: userIsScrolling,
+                  isNearTop: isNearTop,
+                  hasMore: workspaces.transcriptHasMore,
+                  isLoading: workspaces.transcriptLoading || workspaces.transcriptLoadingOlder
+                ) else { return }
+          Task { @MainActor in
+            workspaces.loadOlderMessages(auth: auth)
           }
         }
```
mrosberghaus, line 184: These three pin the viewport at once: `defaultScrollAnchor(.bottom)`, `scrollPosition(id:anchor: .top)`, and `ScrollViewReader.scrollTo("timeline-bottom")`.

The ID binding is always top-anchored, including while `followsLatest` is true. Content growth (live row, image load, older prepend) can glue the top of the clip while follow-latest yanks back with `scrollTo`. If `visibleMessageId` becomes `"timeline-bottom"`, a top anchor puts the 1pt spacer at the top of the clip and hides the transcript until the next `scrollTo`.

`onScrollGeometryChange` also calls `scrollTo` on every pixel whenever the user is not `.interacting` / `.decelerating`. That misses `.tracking`, so the first frames of a gesture can re-pin to bottom and steal the reading position this slice exists to protect.

Do not restack this until the signed-app checks run. If those fail on follow-latest or prepend-stability, use one `ScrollPosition`: `scrollTo(edge: .bottom)` while following, id tracking only while reading. Drop `ScrollViewReader` and the sentinel id. Coarsen the geometry transform so this is not a per-frame hot path.

apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/RoomTimeline.swift
```diff
@@ -0,0 +1,100 @@
...
+    guard generation == expectedGeneration, !Task.isCancelled else { return false }
+    messages = mergeRealtimePage(messages: messages, page: page.messages)
+    if kind != .latest {
+      cursor = page.nextCursor == requestedCursor ? nil : page.nextCursor
+      hasMore = cursor != nil
+    }
+    errorMessage = nil
+    failedPage = nil
+    return true
+  }
```
mrosberghaus, line 99: `reset` sets `isLoading = true`, and the type publishes `isLoading` / `isLoadingOlder` / `isRefreshing`, but `loadPage` never starts or clears them. `WorkspaceState` still wraps every call (`loadOlderMessages` sets `transcriptLoadingOlder` itself).

After `reset` + `loadPage(.initial)` in isolation, `isLoading` stays true. `ConversationSidebar.refresh` already does generation, flags, and error in one method.

Current Mac path is fine. The first direct `RoomTimeline` caller (test, iOS, later view) will get a stuck "Loading messages…" or a missing older spinner. Move flag transitions and a single-flight guard into `loadPage`. Keep 401 → sign-out in `WorkspaceState`.

mrosberghaus (commented):
Independent review of `0ce3119d..279b6402` (slice 04, room timeline).

**Ready to merge: with fixes, after native smoke.** No Critical findings. Keep this draft until the five signed-app checks in the PR body actually run. I would not rewrite scroll on prediction alone.

### What landed well

`RoomTimeline` is the right cut. UI-free, generation-guarded, merge-by-id so overlapping pages and live rows do not duplicate. Failed older pages keep the retry cursor. `TimelineScrollIntent` matches web's 200px near-bottom rule, and the intent tests cover resize and exhausted history not unpinning. Package tests passed at this HEAD. Mac-only scroll APIs stay in `TranscriptView`. No web, API, or dependency changes. PARITY records native smoke honestly.

### Important

Two inlines below.

1. `TranscriptView` stacks three pin mechanisms. That is the most likely way the pending Mac checks fail. If smoke passes, leave it. If follow-latest or prepend-stability fails, collapse to one `ScrollPosition` before adding another `scrollTo`.
2. `RoomTimeline.loadPage` publishes in-flight flags but does not own them. The Mac path works because `WorkspaceState` still wraps every call. The first direct caller will get a stuck spinner.

### Recommendation

Run the five native checks first. That is the experiment. A speculative `ScrollPosition` rewrite is not.

If smoke passes, ship the scroll code and move flag transitions into `loadPage` the way `ConversationSidebar.refresh` already does. If smoke fails on follow-latest or prepend-stability, fix the stacked APIs, then the flags.

Do not start slice 05 until 04 is verified.

### Minor

- `roomId` / `cursor` / `hasMore` are fully public. Passthrough setters can desync generation. `private(set)` would close that.
- `loadOlderMessages` still says "Prepends" after switching to merge-by-id.
- Eager `VStack` of every loaded page is unbounded after several older pages.
- PARITY test counts look off by one.
- No test that `.latest` leaves `cursor` / `hasMore` alone after older pages.
- `userIsScrolling` ignores `.tracking`. Correct to exclude `.animating`.

apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/RoomTimeline.swift
mrosberghaus: Fixed in 37cddfb06. RoomTimeline owns loading flags, error state, single-flight admission and generation-safe cleanup. WorkspaceState retains authentication/read marking and tracks queued tasks so realtime refresh is not dropped behind older history. Tests await task completion. Added flag, cancellation, overlap, latest-pagination and queued-refresh regressions. Verified 101 chat + 34 auth + 1 CoreAPI + 8 realtime + 41 app tests (185 total), iOS 17 compilation, lint and signed build. Ready for another review.

apps/apple/Sokosumi/TranscriptView.swift
mrosberghaus: The user confirmed scrolling up/down and Latest messages returning to the bottom, then explicitly accepted native scrolling as working. Keeping the scroll implementation as requested. The agent did not independently complete every native scenario; PARITY and the PR now distinguish that evidence. Load older is disabled during latest-page refresh to respect single-flight admission.

PR #4317, opened 2026-09-09
fix(core): make job share creation and deletion owner-only

schaier-io:
Closes SOK-1022. One of seven Tier A findings from the SOK-993 security review, each shipping as its own PR.

## User-facing impact

A delegated coworker or a Soko Bot can no longer publish a private job to a public URL, or unpublish one.

Task sharing is owner-only, but job sharing used `requireJobCollaboration`, which also admits a Soko Bot in the job's workspace and a delegated coworker holding the `tasks` capability. The resulting token resolves with no authentication and returns the job's input, output, events, attached file URLs, and payment metadata.

Both `PUT` and `DELETE` now use `requireOwnerUserContext` plus a new `requireMutableJobOwnership`, mirroring the task-share shape. `requireJobCollaboration` is unchanged, so its other callers keep the broad policy.

## Found during the fix, not in the report

1. **`DELETE` had the same hole.** A delegated coworker could remove an owner's live share link, a denial of service on a published job.
2. The owner path previously went through `requireParentTaskNotParked` inside `requireJobCollaboration`. A bare `requireJobOwnership` swap would have silently dropped that check. The new helper preserves it.

## Verification

- `pnpm --filter core test 'src/routes/v1/jobs/[id]/share/'`: `Tests 10 passed`
- Mutation check, source reverted with the tests kept: `8 failed | 2 passed`
- `pnpm typecheck` on this branch against current `main`: `Tasks: 19 successful, 19 total`

## Follow-up, deliberately not in this diff

`withCoworkerContextHeaderParameters` still advertises the coworker context headers on both routes, which now always 403. Removing it changes the OpenAPI output and forces a web client regen.

`apps/core/src/helpers/access-control.ts` was already 1153 lines, over the repo's 750-line ceiling, before this change added 16. Pre-existing tech debt, flagged not addressed.


mrosberghaus:
That is wrong. Every user in an organization should be able to share

PR #4322, opened 2026-09-09
feat(apple): synchronize room read and unread attention

mrosberghaus:
Visible, resolved room history now advances read state; hidden or failed history does not. Content changes trigger another read, including a successful retry that returns unchanged content after an earlier failure. The sidebar supports Mark unread with optimistic rollback and protects newer attention state from stale refreshes.

Web references:
- [Read attention hook](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/hooks/use-room-read-attention.ts)
- [Attention presentation](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/room-attention.ts)
- [Optimistic overlay](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/room-read-overlay.ts)
- [Sidebar actions](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/chat-room-sidebar-row.tsx)

Scope: visibility-gated room reads, content snapshots, unread/mention presentation, manual unread, residual thread counters, rollback, and stale-result isolation. Thread look writes remain in slice 08; pin/mute actions remain in slice 26. No web, Core contract, or dependency changes. The Apple snapshot imports the existing deployed unread endpoint.

Validation: 195 tests pass (110 chat, 34 auth, 1 CoreAPI, 8 realtime, 42 app). Xcode build/test, strict SwiftLint, SwiftFormat, and the shared package's iOS 17 build pass. Independent review passed after fixing the refresh-retry edge case.

How to test:
1. Open an unread room with the app active; resolve its history and confirm read attention settles.
2. Hide the app while new content arrives, then activate it; reads should wait for activation.
3. Right-click an inactive, unmuted room and choose Mark unread. Select it to clear manual unread.
4. Simulate a failed read or refresh; history remains visible, failure rolls attention back, and retry/focus allows another read.
5. Confirm residual thread counts from Core remain intact after room reads.

Native scrolling was accepted by the user in the preceding timeline slice. Native smoke check passed: Mark unread was enabled and invoked for an inactive room, then disabled after selecting it. The original conversation was restored. The Apple Development signed build passed. Runtime logs include AppKit negative-geometry diagnostics during interaction; their source is not established, so this is not a claim of a warning-free run.


mrosberghaus:
Fixed the 3 Important review findings in 97e6a77. What changed and why:

**1. Background auto-read failures no longer pop the modal alert** (`RoomReadAttention.swift`, `ContentView.swift` wiring unchanged)
`readIfNeeded` set `errorMessage` on any failed auto-read, and `ContentView` binds a modal `.alert` to it — so an offline user opening a room (or any transient 500 during background sync) got a blocking dialog for a background operation. Web's `readRoom` fails silently and retries via the reset marker, and the previous native behavior was a non-modal banner. Now only user-initiated `markUnread` surfaces `errorMessage`; auto-reads still `throw` so callers keep 401 handling. `roomChanged()` also clears a stale error so a mark-unread failure for one room doesn't survive navigation (`reset()` already did).

**2. Dead `_ = transcriptFailureMessage(...)` calls replaced with an explicit helper** (`WorkspaceState.swift`)
One correction to the review here: those calls weren't fully dead — `transcriptFailureMessage` signs out on 401 as a side effect, so deleting them would have silently dropped session-expiry handling for read/unread requests. I extracted that into `signOutIfUnauthorized(_:auth:) -> Bool`, reused it in `transcriptFailureMessage`, and the two attention catch blocks now call it explicitly. On mark-unread 401 the modal error is additionally cleared, since the auth card takes over (per the existing convention documented on `transcriptFailureMessage`).

**3. Rewrote the stale `SidebarRooms.swift` comment** — it still claimed "Selection does **not** suppress chrome … this slice has no mark-read", contradicting the new `isActive` suppression and mark-read. It now describes muted/active suppression via `RoomReadAttention` (ADR 0026).

**Tests updated to lock the contract:** `failedReadRollsBackAndCanRetry` asserts silence after a failed read; the mark-unread test asserts the error surfaces and is cleared by `roomChanged()`; `failedReadKeepsResolvedHistory` flips to `errorMessage == nil`.

**Verification:** `swiftformat --lint` clean, `swiftlint lint --strict` 0 violations, `swift test` SokosumiChat 110/110 pass, `xcodebuild test -only-testing:SokosumiTests` 43 pass / 0 failures (with `-enableCodeCoverage NO`; coverage-enabled linking fails on `___llvm_profile_runtime` in the msgpack package product — pre-existing environmental issue, unrelated to this change).

---

**Remaining findings from the review (Minor, not addressed):**

4. Unused `room` parameter in `loadTranscript(auth:room:generation:)` (now `room _:`) — leftover from when the passed room was marked read. Either remove the parameter or note why callers still pass it.
5. `openRoom` doc comment still says "The sidebar entry is replaced with the POST-read DTO" — now done via attention overlay. Small wording update.
6. Parity gap (confirm intentional): web's `resolveRoomAttention` also returns opt-in `unreadTextCount` rendered as `RoomUnreadCount`; native `RoomAttention` has only bold/badge. Worth one line in PARITY.md either way.
7. No test asserts residual thread counters survive settle — guaranteed structurally (`Fields` carries only 3 fields), but one assertion (e.g. pinned/thread counts unchanged through read + reconcile) would lock the "retained verbatim" claim.
8. `update-core-api.py`: no argv validation (`IndexError` with no args), bare `KeyError` on unknown path or renamed method, and the refresh loop drops newly added methods on existing paths. Dev-only script, but a usage check and `setdefault`-style guards are cheap.
9. Realtime arrivals can double-POST mark-read (both the `transcriptMessages` onChange and `refresh()` call `syncReadAttention` with different markers). Idempotent server-side, so harmless — single-flighting would just save a request.
10. Overlays for revoked rooms linger until workspace reset (`applyMembershipRevoked` filters rooms but doesn't drop the overlay). Harmless and bounded; consider dropping the entry there.

**Recommendation:** decide the product rule for surfacing attention-sync failures (modal only for explicit user actions; silent-with-retry for background reads) and encode it in one place rather than sharing `errorMessage` between the two paths. This fix implements that split behaviorally (`readIfNeeded` never sets `errorMessage` now); a follow-up could enforce it structurally.

PR #4325, opened 2026-09-09
feat(apple): queue classic chat sends with native multiline composing

mrosberghaus:
Classic chat now lets you compose and queue another message while a send is pending. Sends run in order, preserve their IDs for retry, and release the queue after a 30-second timeout. The composer supports multiline editing, protects marked-text commits from accidental submission, and enforces the web's 10,000 UTF-16-unit limit.

In scope: PARITY.md slice 06 — classic room composing/sending, pending/failed feedback, retry/remove, duplicate reconciliation, draft clearing, and timeout/reset isolation. A macOS-only AppKit input adapter is isolated behind SwiftUI because native verification showed SwiftUI submission firing on marked-text commit; models and networking remain portable.

Out of scope: reply threads, rich formatting, mentions, attachments, coworker streaming, and durable outbox persistence. No web source, shared API contract, or dependency changes.

Web references:
- [Classic outbound queue](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/utils/classic-outbound-queue.ts)
- [Room send integration](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/rooms-client.tsx)
- [Composer input and IME behavior](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/composer-wysiwyg-editor.tsx)
- [Enter rules](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/utils/composer-wysiwyg-input-rules.ts)
- [Content limits](https://github.com/masumi-network/sokosumi/blob/main/packages/utils/src/chat-room-message-content.ts)

Validation: Xcode build and app tests passed; shared tests passed (chat 118, auth 34, CoreAPI 1, realtime 8); shared chat compiles for iOS 17; SwiftFormat and strict SwiftLint passed; Apple Development signed build passed. Native harness using the actual input views verified Return sends/clears, Shift/Command/Control-Return inserts newlines, and dead-key marked-text commit does not send. Independent review found no remaining actionable issues. Full language-specific IME coverage remains manual QA.

How to test:
1. Open a classic room and send several messages while a request is pending; confirm ordered delivery with an available composer.
2. Type multiline text using Shift/Command/Control-Return. Plain Return sends. Commit IME text with Return and confirm it is not sent by that commit.
3. Try an empty/whitespace draft and over 10,000 UTF-16 units; neither sends. The counter starts at 9,500.
4. Simulate failed/timed-out requests; retry/remove the failed shell and confirm a retry keeps its original ID. Switch rooms/workspaces during a send; late results must not appear in the new room.

PARITY.md records the scope and verification. The next slice waits for this PR to merge.


apps/apple/Sokosumi/MacComposerTextInput.swift
```diff
@@ -0,0 +1,133 @@
...
+      input.drawsBackground = false
+      input.font = .preferredFont(forTextStyle: .body)
+      input.textColor = .labelColor
+      input.insertionPointColor = .labelColor
+      input.isVerticallyResizable = true
+      input.isHorizontallyResizable = false
+      input.autoresizingMask = [.width]
+      input.textContainer?.widthTracksTextView = true
+      input.textContainerInset = NSSize(width: 2, height: 2)
+      input.setAccessibilityLabel("Message")
```
mrosberghaus, line 33: **Nit (not blocking).** Accessibility label is `Message`, but there is no visible placeholder. The old SwiftUI `TextField("Message", …)` prompt is gone. Web still shows `Message #{channel}` / `Message {member}`. Empty-editor hit testing is covered; the empty state is not.

apps/apple/Sokosumi/TranscriptView.swift
```diff
@@ -358,7 +360,6 @@ import SwiftUI
             HStack(spacing: 12) {
               if let onRetry {
                 Button("Retry", action: onRetry)
```
mrosberghaus, line 402: **Nit (not blocking).** Retry/Remove only render for `.failed`. In-flight and queued shells have no sending chrome, so they look like confirmed history. Web shows `Outbound.sending` (spinner) then a sent tick. Fine for the tracer; worth a later pass now that the queue can hold several unconfirmed bubbles.

apps/apple/Sokosumi/WorkspaceState.swift
```diff
@@ -280,34 +290,40 @@ final class WorkspaceState: ObservableObject {
...
+  @discardableResult
+  func sendMessage(_ content: String, auth: AuthState) -> Bool {
+    let draft = ComposerContent(content)
+    guard let roomId = transcriptRoomId, draft.canSend, !transcriptLoading,
+          let client = resolveClient(auth: auth) else { return false }
+    let id = UUID().uuidString
+    let slug = selection?.workspace.organizationSlug
+    let shell = makeOutboundShell(clientMessageId: id, roomId: roomId, content: draft.text)
+    outbox.enqueue(shell, send: {
+      try await ChatService().createMessage(
```
mrosberghaus: **Nit (not blocking).** `ChatService()` is minted per send instead of using the existing `service` instance. Harmless (`ChatService` is a stateless struct) but inconsistent with the rest of `WorkspaceState`.

mrosberghaus (commented):
## Review (slice 06 — classic compose/send)

Reviewed `16d4f3bd..3d31aa3a` against PARITY.md slice 06 and the web classic outbound queue. Queue, IME, UTF-16 limits, retry IDs, and AppKit isolation are in the right layers.

I applied the two blocking findings on this branch (`3d31aa3a`). Remaining items below are nits, also inlined.

### What changed in `3d31aa3a`, and why

**1. Stop cancelling the in-flight POST on timeout and room switch** (`RoomOutbox.reset` / `settle`)

`request?.cancel()` aborted `URLSession.data(for:)` via `URLSessionTransport`. Web does not do that: `raceWithTimeout` only releases the local drain, and `clearClassicOutboundQueue` drops refs without aborting `send()`.

The composer already clears on accepted send. Cancelling the Task after a room switch (or after timeout + Remove) could drop text Core never saw: no shell, empty draft, nothing in the old room on return. Isolation was already correct via the attempt UUID and `jobs`.

Fix: cancel the timer only. Leave the send Task running. Stale `settle` still no-ops.

**2. Surface the timeout copy** (`settle` + `RoomOutboxTests`)

`SendTimeout` already had *"Sending timed out. Retry to check or send this message again."* `settle` then ran `friendlyMessage(for:)`, which only special-cases `ChatServiceError` and `URLError`, so users saw *"Couldn't reach Core. Check your connection and try again."* That fights the timeout design (Core may already have the row; retry the same id).

**3. Tests**

- Timeout and reset asserts `Task.isCancelled == false` after the send resumes.
- Timeout asserts the timeout string on the failed shell.
- Room-switch test: fake transport now throws `CancellationError` if the POST Task is cancelled; `remainingStubs == 0` after `releasePOST` proves the 201 was consumed.

**4. Drop leftover `retryDisabled`**

Always `false` after the queue replaced single-flight. Removed from `MessageRow`.

### Verification on this follow-up

- `swift test --package-path apps/apple/Packages/SokosumiChat` — 118 passed
- `xcodebuild test -only-testing:SokosumiTests` — passed, including `changingRoomDropsPendingOutbound`
- SwiftFormat + strict SwiftLint on the touched Swift files — clean

Did not re-run the signed interactive build or iOS 17 package compile for this follow-up.

### Remaining (not blocking)

See inline comments: no composer placeholder, no pending/sent chrome, `ChatService()` minted per send. Full language-specific IME remains manual QA, as the PR already says.

**Verdict:** slice 06 is ready once CI is green on `3d31aa3a`. Remaining items can wait.

mrosberghaus:
Verified the applied fixes in `3d31aa3a` against the web classic outbound queue and `raceWithTimeout`. Web releases the local queue without aborting the underlying send; leaving the POST task running matches that behavior, while attempt IDs still prevent stale settlements. The timeout-specific message now reaches the failed shell. Removing the unused retry-disabled flag is consistent with queued sends.

Independent local verification on this commit:
- Chat package: 118 tests passed, including timeout/reset cancellation assertions and timeout copy.
- Xcode build and app tests passed, including `changingRoomDropsPendingOutbound` consuming the POST response after a room switch.
- Shared chat package compiled for iOS 17.
- Apple Development signed macOS build passed.
- SwiftFormat and strict SwiftLint passed.

No additional blocking findings or code changes. Interactive language-specific IME behavior was not re-exercised in this follow-up.

apps/apple/Sokosumi/MacComposerTextInput.swift
mrosberghaus: Fixed in 8b84c3be2. Added a visible room-aware placeholder inside the native editor; it does not intercept clicks and remains hidden during marked-text composition. Verified empty, typing, and marked-text states in the native harness.

apps/apple/Sokosumi/TranscriptView.swift
mrosberghaus: Fixed in 8b84c3be2. Pending sends show Sending after 500 ms; slow confirmations show Sent for 1.6 seconds, matching web timing. Works on continuation rows and HTTP/Ably confirmation paths. Added tests for fast/slow sends, expiry, duplicate confirmation, and reset. Chat 121 tests, app tests, iOS 17 compile, lint, and signed build passed.

apps/apple/Sokosumi/WorkspaceState.swift
mrosberghaus: Fixed in 8b84c3be2. The send closure captures the existing service value, without retaining WorkspaceState.

PR #4327, opened 2026-09-09
feat(apple): recover live room updates across connection changes

mrosberghaus:
## Live room updates and recovery — parity slice 07

Chat now receives updates for membership-visible rooms and recovers missed timeline/sidebar changes after connection gaps or foreground return. Live field patches preserve message content; revoked rooms cannot return from stale list or workspace-switch responses.

### Scope

- Full-message and ID-envelope hydration, reaction/unfurl/mention-status patches, viewer reaction personalization, pin state, and own-send reconciliation.
- Healthy 60-second refresh cadence; unhealthy room/sidebar fallback of 3/15 seconds. Hidden reads defer, overlapping reads coalesce, and stale scope completions cannot apply.
- Membership/capability intersection, immediate revocation, serialized authorization with 15-second retry, persistent token provider, and listener/subscription cleanup.
- PARITY.md records implementation and verification. No web edits, shared API contract changes, or new registry dependencies. Reply threads, coworker streaming, rich rendering, and reaction/pin management UI remain their later slices.

### Web sources

- [Realtime subscriptions](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/ably/use-chat-room-realtime.tsx)
- [Refresh scheduler](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/use-chat-refresh-scheduler.ts)
- [Selected-room health](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/ably/use-selected-room-channel-health.ts)
- [Field patches](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/ably/apply-chat-room-message-patch.ts)
- [Viewer personalization](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/ably/personalize-chat-room-message-event.ts)
- [Capability intersection](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/ably/chat-room-ids-from-ably-capability.ts)

### Verification

From `apps/apple`:
- `swift test --package-path Packages/SokosumiChat` — 128 tests passed.
- `swift test --package-path Packages/SokosumiRealtime` — 27 tests passed.
- Auth and CoreAPI package suites — 34 and 1 tests passed.
- `xcodebuild build test -only-testing:SokosumiTests` with the macOS scheme/destination and ad-hoc signing — passed.
- Apple Development signed build and shared chat iOS 17 compile — passed.
- SwiftFormat, strict SwiftLint, and diff checks — passed.

Regression coverage includes held sidebar/switch responses during revocation, foreground continuity recovery, stale authorization/token results, patch isolation, and viewer reaction flags. Independent review findings were fixed and re-reviewed. An existing Sent-expiry test now waits for observed expiry within a bounded deadline instead of assuming a 100 ms executor scheduling margin.

For manual verification, open a room, receive an external message/edit, background the app during a connection gap, then return: history should recover without duplicates. Switch workspaces during activity and verify the previous room does not reappear. These live multi-client scenarios were not exercised against production during this change.


mrosberghaus:
Independent review of this slice found no merge blockers. Three nits landed in `0a4694f3b`.

**Stale factory comment.** `WorkspaceState.realtimeConnectionFactory` still described the old “last minted Ably token / remint when a token exists” path. That path is gone; the property is only the live-socket factory. Left the old comment in place and the next slice reads a deleted design.

**Second transcript coalescer.** Envelopes already go through `ChatRefreshScheduler`, which waits for in-flight history/older/refresh before starting another GET. `pendingTranscriptRefresh` was a leftover busy-flag around `refreshTranscript` for the same job. Removed it so there is one coalescer. Retry while a load is already in flight is a no-op until that load finishes; the error Retry button still works after a failed latest-page refresh.

**PARITY checkpoint.** Resume text still called this a draft based on `a3b337626`. The PR is ready, CI was green, and main is already merged in. Updated the checkpoint so it does not send the next agent down a stale gate.

Verified with `envelopeCreateDuringHistoryLoadRefetchesAfterResolve`, `queuedEnvelopeAfterOlderPageIsRefetched`, and `successfulRefreshRetryMarksUnchangedVisibleContentRead` (`TEST SUCCEEDED`), plus SwiftFormat lint and strict SwiftLint on `WorkspaceState.swift`.

PR #4339, opened 2026-09-10
fix(apple): scroll transcript to true bottom on open

mrosberghaus:
Chat opened parked 8pt above the true bottom: every follow-latest `scrollTo("timeline-bottom")` aligned the 1pt spacer with the viewport bottom while the stack's 8pt bottom padding sat below the anchor. The anchor now doubles as the trailing breathing room (9pt) with top-only outer padding, so all scroll-to-bottom paths land at distance 0. Row spacing is pixel-identical.

Root-cause evidence: a throwaway probe replicating the scroll construction settled at distBottom=8.0 before the fix and 0.0 after; adding only an onAppear scrollTo still settled at 8.0.

Verification (apps/apple): swiftformat lint clean, swiftlint strict clean, xcodebuild build + 58 app tests passed, package suites passed (CoreAPI 1, Auth 34, Chat 128, Realtime 27).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript scrolling so the view reaches the intended bottom position while preserving appropriate trailing space.

### DIFF
--- a/apps/apple/Sokosumi/TranscriptView.swift
+++ b/apps/apple/Sokosumi/TranscriptView.swift
@@ -124,11 +124,14 @@ import SwiftUI
               }
               .id(message.id)
             }
-            Color.clear.frame(height: 1).id("timeline-bottom")
+            // Bottom anchor doubles as the trailing breathing room (1pt anchor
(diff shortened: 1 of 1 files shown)


mrosberghaus:
@coderabbit review

PR #4340, opened 2026-09-10
feat(apple): add native reply threads

mrosberghaus:
## Feature

Add native reply threads to Apple chat (PARITY slice 08). Open a message's thread from the upper-right hover Reply action or its existing reply-count link. The detail view shows the parent, paginated replies, and an independent composer, draft, and outbound queue; native Back returns to the room.

## Scope

Includes loading/retry states, stable send retries, reply counts, thread look/read ordering, visibility-aware attention, full-message/patch/envelope routing, and recovery after interrupted history loads. Uses existing Core message, thread-message, and thread-read endpoints; no Core contract changes, dependencies, or web edits.

Coworker streaming, rich rendering, attachments, reactions, pin controls, and thread overview remain later PARITY slices. Parent ID-envelope refresh follows web's latest-room-page behavior; older roots outside that page can remain stale until revisited or receiving a full DTO.

## Web behavior

- [Thread panel](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/thread-panel.tsx)
- [Room/thread loading, sending, and realtime routing](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/rooms-client.tsx)
- [Read attention](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/hooks/use-room-read-attention.ts)
- [Message scope](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/utils/chat-room-message-scope.ts)

## How to test

1. Open a room, hover a message and choose Reply. Existing reply-count links also open threads.
2. Type a draft, go Back, and reopen. The thread draft should return while the room draft stays separate.
3. Load older replies, then send a reply. The new reply should become visible; failures offer Retry/Remove with the same client message ID on retry.
4. Switch room/workspace or close the thread during a request. Late results must not restore the old thread.
5. Receive reply/parent updates and deletes while the thread is open; verify that replies remain outside the room transcript.

## Verification

Passed: Xcode build and app tests; Chat 149, Auth 34, Realtime 27, CoreAPI 1 tests; shared iOS 17 build; signed Apple Development build; strict SwiftLint and SwiftFormat. Exact verification records are in `apps/apple/PARITY.md`. Shared tests cover HTTP errors, pagination/recovery, scoped drafts, send retries, realtime routing, and read-order/navigation races. Native open, typing, back/reopen, and draft isolation were exercised without posting test messages to production. Please review hover appearance and long-thread scrolling on macOS; automated pointer actions were unreliable in this environment.


mrosberghaus (commented):
## Review: two fixes pushed, remaining findings below

The architecture, race handling, realtime routing and OpenAPI snapshot look solid. I found one real bug (stale reply counts after your own reply) and one visible copy issue ("1 replies"). Both are fixed in three commits on top of `27bcbf0e6`. The rest are minor. I left them for you to decide on.

### What I changed and why

#### 1. `6023e8051` fix(apple): confirm sends when the realtime echo arrives first

**Problem.** After you send a thread reply, the thread header count, the parent's `threadLastReplyAt`, and the room's "N replies" link usually stayed stale until the next room refresh. With healthy realtime, that can take up to 60s.

**Cause.** `ThreadSession.send` updates the count in the outbox `confirmed` callback. But `RoomOutbox.settle` only called `confirmed` if the pending shell still existed:

- Core awaits `publishChatRoomMessageRealtime` before it returns 201. So the realtime echo usually arrives first.
- The echo goes through `ThreadSession.apply(eventType:message:)` → `outbox.reconcile`, which removes the shell.
- When the HTTP success then arrived, `settle` took the "shell gone" branch. It dropped the job without calling `confirmed`.

Nothing else increments the count. On create, Core publishes only the reply, not an update for the parent (it republishes the parent only when a reply is soft-deleted). So in the common order, the sender's own reply was never counted.

**Fix.** In `RoomOutbox.settle`, a successful result now always clears the job and calls `job.confirmed(message)`. The shell is removed and "Sent" feedback recorded only if the shell is still there. Failures keep their old behaviour: a failed result for a shell that realtime already confirmed is dropped silently.

**Why this approach:**
- **It matches web.** `drainClassicOutboundQueue` calls `onSuccess` on every HTTP success of the active run, whether or not the echo came first. `onSuccess` then calls `updateParentThreadPreview`.
- **It's safe for the room path.** The room `confirmed` callback only runs `confirmOutbound`, which updates the row by ID. So re-applying a row the echo already added changes nothing.
- **No double counting.** Core sends no parent update on create, so the local increment is still the only count source.
- **"Sent" feedback isn't restarted.** It is still recorded once, by whichever path removes the shell. The existing `realtimeConfirmationShowsSentWithoutRestartingOnHTTPAndResetClearsIt` still passes.

**Alternative I rejected.** Counting the reply in `ThreadSession` on the first confirmation from either path would also cover "timed out, then confirmed only by the echo". But it would make thread sends behave differently from room sends and from web, and add per-session tracking. Web doesn't count that case either. It's rare (30s timeout) and fixes itself on the next room refresh, so I kept parity.

**Tests:**
- New `ThreadSessionTests.replyCountsWhenRealtimeEchoArrivesBeforeHTTPResponse`: sends a reply, applies its echo before the POST task runs, then checks `settled(.success)` fires once, the reply is in the timeline, and `parent.threadReplyCount == 1`.
- `RoomOutboxTests.realtimeConfirmationShowsSent…` now also checks that `confirmed` fires exactly once after an echo-first confirmation.
- Both failed against the old `RoomOutbox.swift` (`confirmations → 0`, `threadReplyCount → 0`) and pass with the fix.

#### 2. `2ff0ac852` fix(apple): pluralize thread reply counts

`TranscriptView.swift:419` rendered "1 replies". `ThreadView.swift:21` rendered the same in the thread header. Both now use Foundation automatic grammar agreement: `"^[\(count) reply](inflect: true)"`. I checked it works without a string catalog: `0 replies`, `1 reply`, `2 replies`, `12 replies`. "0 replies" on an empty thread is correct and matches web's `{count, plural, one {# reply} other {# replies}}`.

#### 3. `2c07e924b` docs(apple): record slice 08 review fixes

Adds a short "Slice 08 review fixes" entry to `PARITY.md` in the same style as the existing follow-ups.

#### Verification

- `swift test --package-path Packages/SokosumiChat`: 150 tests in 16 suites passed (149 before, plus the new test).
- `xcodebuild test -only-testing:SokosumiTests -enableCodeCoverage NO` (CI's invocation): passed. Without `-enableCodeCoverage NO`, the local run fails to link `AblyDeltaCodec`/`msgpack` (`___llvm_profile_runtime`). That's a coverage-instrumentation quirk in those packages, unrelated to this PR.
- `mint run swiftformat --lint .`: 0 files need formatting. `mint run swiftlint lint --strict`: 0 violations.
- I committed with `--no-verify`. The husky hook's TS typecheck couldn't run in my worktree because it had no `node_modules` (`prisma` not found). The changes are Swift and Markdown only, which neither Biome nor turbo covers. `pnpm check` passed.

### Remaining findings (not changed)

Line numbers refer to the PR head.

1. **The hover Reply action shows on deleted messages.** `TranscriptView.swift:124` sets `onReply` for every non-outbound row, including soft-deleted ones. So the hover Reply (`:448`) and the context-menu "Reply in thread" (`:497`/`:503`) appear on tombstones. Web hides the actions when deleted (`room-message-row.tsx:2181`, `showActions = … && !isDeleted …`) but keeps the reply-count link. Suggested fix: gate the hover and context-menu Reply on `message.deletedAt == nil`, and keep the count link.
2. **A parent envelope refreshes the room twice.** In `WorkspaceState.swift:493-506`, a create/update envelope for the open parent calls `transcriptRecovery.requestRefresh()` once via `refreshParent`. `resolveRealtimeEnvelope` then returns `.needsRefetch` for the same focused-room envelope and calls it again. `ChatRefreshScheduler` queues the second call behind the in-flight one, so Core gets two latest-page GETs back to back. Suggested fix: skip the explicit call when the room resolution already refetches, or return after it.
3. **The thread sender name has no email fallback.** `WorkspaceThreads.swift:61-63` builds the sender with `name: currentUserName`. `makeOutboundShell` (`WorkspaceState.swift:584`) falls back to the email when the name is empty, so a pending thread reply can show a blank name. Suggested fix: reuse one sender builder for both paths.
4. **The thread spinner never ends without a client.** If `resolveClient` returns nil, `WorkspaceThreads.swift:26` returns silently. `timeline.reset` has already set `isLoading = true`, so "Loading replies…" never ends. `syncReadAttention` also requires `!thread.timeline.isLoading` (`WorkspaceState.swift:634`), so room read attention stays blocked while the thread is open. Suggested fix: call `thread.timeline.failInitialLoad(…)`, as the room path does.
5. **The thread look is gated more strictly than room reads.** `syncThreadAttention` (`WorkspaceThreads.swift:42`) requires `timeline.failedPage == nil`. `syncReadAttention` accepts a failed `.older` page. So if an older room page failed, opening a thread skips the first thread look. It recovers after the replies load, but the two gates should probably match.
6. **The scroll-follow logic is duplicated.** `ThreadView.swift:28-45` re-implements follow-latest (`distanceFromBottom < 200`) instead of reusing `TimelineScrollIntent` (`RoomTimeline.swift:129`), which `TranscriptView` already uses.
7. **Smaller items:**
   - `ThreadSession.markLooked` (`ThreadSession.swift:78`) builds a new `ChatService()` although `service` already exists.
   - The context menu uses `bubble.right` (`TranscriptView.swift:498`); the hover button uses `text.bubble` (`:451`).
   - "Load older replies" is silently ignored while a recovery `.latest` load holds `loadTask` (`ThreadSession.loadPage` guards `loadTask == nil`). The button only disables on `isLoadingOlder`.
   - `ThreadView` renders empty during the pop animation once `parent` becomes nil.
8. **`PARITY.md` has some stale or noisy text.**
   - "rebased on main `9c1157a6f`" (lines 5 and 275) is stale: the head is a merge of `main`.
   - The six dated hover follow-up sections, with their `/tmp/*.log` paths, read like a work log. Consider folding them into one "hover/Reply layout" bullet.

### A product question

Web shows the thread as a side panel next to the room, so auto-reading new room messages while a thread is open matches what the user sees. On macOS the thread is pushed over the room, but `readContent` still auto-reads top-level room messages the user can't see. This follows web's hook literally, but not its "visible resolved history marks read" intent. It's worth deciding on purpose, even if the change is deferred.

Related: counts for other people's replies still update only on refresh, on both web and Apple. That's fine for parity. Maybe scope "reply counts" in `PARITY.md` to your own sends plus refreshes.

---
Review and fixes by Claude Code.

PR #4345, opened 2026-09-10
fix(chat): read sidebar invalidations while the tab is hidden

mrosberghaus:
Regression from #4310 (SOK-986).

## User-facing impact

- On an open room, the browser tab title shows `(N)` when a message arrives in another membership-visible room while the Sokosumi tab is hidden or the window is unfocused. The open-room Ably island attaches the sidebar catalog, not only the selected room, so that create can invalidate the active collection immediately.
- Timer reads still pause in the background. One coalesced read per real change, no polling.

## Root cause

Two gates stacked:

1. The refresh scheduler from #4310 deferred every read, including Ably-triggered requests, until the tab was visible and focused.
2. The open-room page only attached the selected room, so a foreign-room `chat_room_message` never reached the client to request that refresh.

## Fix

- Explicit requests (`requestRefresh`) read at once regardless of foreground. Timer reads still pause in the background and catch up with one read on return.
- An explicit request during an in-flight read stays explicit for the queued follow-up.
- A read that runs clears the pending "needed" flag, so a return does not read a second time.
- `RoomMessageRealtimeBridge` attaches `channelCatalogRooms` (sidebar snapshot merged with the LCP selected room). Desktop sidebar publishes that snapshot on every app page. Token capability still intersects the attach set.

## Still open

Off a room page (including the `/chat` list) no room message channels are attached (ADR 0003: room message subs stay chat-scoped). The mobile Chats unread dot on the list, and the title off a room page, still only move via the 60s timer (paused in the background) or control-channel archive/invite events. Covering those needs the per-user `notification_created` signal and a product decision about silenced notifications; not part of this PR.

## Verification

- `rooms-client-read-visibility.test.tsx`: catalog attach + a create in a room that is not selected dispatches `{ collections: ["active"] }`.
- `use-organization-chat-rooms.test.ts`: hides the tab, delivers an active-collection invalidation, asserts the row gains its unread count immediately.
- `use-chat-tab-unread-presence.test.tsx`: same invalidation-while-hidden contract for the list unread-dot hook, when that island is mounted and an invalidation fires.
- `use-chat-refresh-scheduler.test.ts`: timer reads paused while hidden; explicit requests run while hidden or unfocused; a queued explicit follow-up runs while still hidden with no duplicate read on return.

### DIFF
--- a/apps/web/src/app/(app)/chat/components/__tests__/rooms-client-ably-island.test.ts
+++ b/apps/web/src/app/(app)/chat/components/__tests__/rooms-client-ably-island.test.ts
@@ -24,7 +24,9 @@ describe("RoomsClient Ably island", () => {
       /LazyAblyProvider>\s*<RoomMessageRealtimeBridge[\s\S]*<\/LazyAblyProvider>/,
     );
     expect(source).not.toContain("ChannelProvider");
-    expect(source).toContain("roomIds={rooms.map((room) => room.id)}");
+    expect(source).toContain(
+      "roomIds={channelCatalogRooms.map((room) => room.id)}",
+    );
     // Open room chrome lives in RoomShellLayout (Instant + progressive share it).
     expect(source).toContain("listScrollerRef={scrollerRef}");
     expect(source).toContain("<RoomShellLayout");
--- a/apps/web/src/app/(app)/chat/components/__tests__/rooms-client-read-visibility.test.tsx
+++ b/apps/web/src/app/(app)/chat/components/__tests__/rooms-client-read-visibility.test.tsx
@@ -15,6 +15,11 @@ import {
 import type { RoomComposerHandle } from "@/app/chat/components/room-composer";
 import { RoomsClient } from "@/app/chat/components/rooms-client";
 import type { RoomMessagesPage } from "@/components/chat/fetch-room-messages";
+import {
+  clearMembershipVisibleRoomsSnapshot,
+  publishMembershipVisibleRooms,
+} from "@/components/chat/membership-visible-rooms-store";
+import { ORGANIZATION_CHAT_ROOMS_CHANGED_EVENT } from "@/components/chat/organization-chat-events";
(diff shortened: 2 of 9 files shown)


mrosberghaus:
Independent review of this PR: un-gating `requestRefresh` was not enough to restore `(N)` for a message in another room. The open-room island only attached the selected room, so `notifySidebarOfForeignRoomMessage` never ran.

Follow-up in `fix(chat): attach catalog rooms for hidden unread title`:

- `RoomMessageRealtimeBridge` now attaches `channelCatalogRooms` (sidebar snapshot merged with the LCP selected room). A create in another membership-visible room can invalidate the active collection while the tab is hidden. Desktop sidebar already publishes that snapshot from app chrome. Token capability still intersects the attach set.
- Regression: catalog attach + a create in a room that is not selected dispatches `{ collections: ["active"] }`.
- Scheduler: explicit request runs while hidden or unfocused; a queued explicit follow-up runs while still hidden with no duplicate on return.
- Presence hook: invalidation-while-hidden is covered when that island is mounted and an invalidation fires.
- PR body narrowed so it no longer claims the mobile list unread dot shares the message path.

### Remaining

**1. Mobile list unread dot, and the title off a room page**

Still no room-message channels (ADR 0003: those subs stay chat-scoped). Do not attach membership rooms on the nav island. Cover this with the per-user `notification_created` signal once product decides how silenced notifications behave. Out of scope for this PR.

**2. A completed explicit read that applied nothing still clears `needed`**

`fetchBackgroundJson` returns null and does not throw. The timer retries. Leave the scheduler contract unless we want failed background GETs to re-arm `needed` for a foreground catch-up. I would not change that here.

**3. `returnToForeground` test helpers still fire the invalidation on blur**

Name lag only. Rename the next time someone is already in those files. Not worth a dedicated commit.

**Caveat of the catalog attach**

A cold land on a room URL with no sidebar snapshot still attaches only the selected room until the sidebar publishes. Desktop chrome usually publishes immediately. Mobile depends on the snapshot surviving from `/chat`.

PR #4348, opened 2026-09-10
feat(core): allow organization members to share and revoke job shares

schaier-io:
Closes SOK-1030.

## What changed

Public job sharing required the job owner. Any human member of a job's workspace can now create, update, and revoke that job's public share.

The workspace is the permission boundary. An organization workspace admits every member of that organization; a personal workspace admits only its owner. Membership in one organization grants nothing in another.

## Why the check is a new helper

At this branch's base commit, `requireJobCollaboration` guarded six route files, not just the two share routes:

| Route | What it does |
| --- | --- |
| `share/put.ts` | in scope |
| `share/delete.ts` | in scope |
| `[id]/patch.ts` | rename and metadata |
| `[id]/workspace/put.ts` | move between workspaces |
| `[id]/inputs/post.ts` | submit inputs |
| `[id]/refund/post.ts` | issue a refund |

Widening it would have let any member refund or move a colleague's job. This branch moves the two share routes onto their own helper, so `requireJobCollaboration` now guards the remaining four. So sharing got `requireJobShareCollaboration` in `apps/core/src/helpers/access-control.job-share.ts`, and a regression test pins `requireJobCollaboration` as owner-only.

The predicate is the job owner **or** a member of the caller's active workspace. Ownership stays in it because API key and OAuth contexts carry `organizationId: null` (`middleware/auth.ts:429` and `:632`), so their workspace resolves to the personal one. Workspace-only scoping would have stopped them sharing their own organization jobs. The predicate is strictly additive: everything that worked before still works.

Soko Bot and coworker contexts delegate to the existing rules, unchanged. Agent publication policy is out of scope.

## User-facing impact

A member opening a colleague's job now sees the share control and it works. Rename, workspace move, refund, and input submission remain owner-only, so the same page shows fewer controls for a non-owner than for the owner.

`readOnly` previously hid the whole job action bar. It moved inward, and the public `/share/[token]` page is now gated explicitly on `publicJobLayout` rather than relying on `useJobsHeader()` returning null off the app subtree.

## One API change

An unreachable job returns **404** on the share routes where it previously returned 403. Both routes already declared 404, so the OpenAPI contract is unchanged, but a client matching on 403 would notice.

## Verification

- `pnpm --filter core test` — `5199 passed | 11 skipped`
- `pnpm typecheck` — 19 tasks successful
- `pnpm check` — clean
- CI on this PR is green, `Test Web` included
- Three mutations (drop the workspace arm, drop the owner arm, drop the public gate) each fail 2 to 3 tests with a non-zero collected count, so the tests pin behaviour rather than merely passing

Correction to the first version of this description: it reported 28 web test failures in `time-ago.test.tsx` and `task-activity.test.tsx` as a pre-existing defect from the React 19.3 bump. That was wrong. The cause was a stale local install: `react-dom` 19.2.8 was present against the `19.3.0` pin in `apps/web/package.json`, so `browser` was undefined at `time-ago.tsx:76`. Comparing against the base commit could not reveal it, because both revisions ran against the same stale `node_modules`. CI installs from the lockfile and `Test Web` passes.

No schema change, no migration, no Core client regeneration: the route shapes are unchanged.

## Follow-up

SOK-1005 carries a comment: a stale session's `activeOrganizationId` window now also reaches public sharing. This PR inherits that gap deliberately and does not add a membership re-check.

## Summary by CodeRabbit

* **New Features**
  * Workspace members can share jobs they do not own when they have appropriate access.
  * Job sharing remains available in read-only views, while editing and moving remain restricted.
  * Public job layouts now hide sharing, editing, and moving controls.

* **Bug Fixes**
  * Improved authorization for creating and removing job shares.
  * Unauthorized or inaccessible jobs now return consistent errors without revealing whether a job exists.
  * Parked parent tasks cannot be shared.

### DIFF
--- a/apps/core/src/helpers/access-control.job-share.test.ts
+++ b/apps/core/src/helpers/access-control.job-share.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it, vi } from "vitest";
 
 import type { EnvVariables } from "@/lib/hono";
 import type {
+  CoworkerAuthenticationContext,
   SokoBotAuthenticationContext,
   UserAuthenticationContext,
 } from "@/middleware/auth";
@@ -65,7 +66,8 @@ function varsFor(
   workspaceContext: WorkspaceContext | null,
   authContext:
     | UserAuthenticationContext
-    | SokoBotAuthenticationContext = memberAuthContext,
+    | SokoBotAuthenticationContext
+    | CoworkerAuthenticationContext = memberAuthContext,
 ): EnvVariables["Variables"] {
   return {
     isAuthenticated: true,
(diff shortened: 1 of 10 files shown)


schaier-io:
@coderabbitai review

PR #4354, opened 2026-09-10
fix(web): report a Core session outage as 503, not 401

schaier-io:
Chat notifications, the unread indicator and bold room names stopped for a
signed-in user and stayed stopped until a full page reload.

## User-facing impact

A brief Core slowdown no longer signs the browser out of realtime. Before this
change one timed-out session read killed the Ably client for the rest of the
page's life.

## Root cause

Three defects stacked.

**1. A timeout was reported as a logout.** `fetchSession` returned `null` both
when Core answered with no session and when it could not be asked at all, so
callers turned an `AbortSignal.timeout` into HTTP 401.

Production web logs, 2026-09-10 08:53:35 UTC:

```
/api/chat/rooms 401 Failed to fetch session from Core
  [Error [TimeoutError]: The operation was aborted due to timeout]
/api/ably/auth  401
```

Core logs for the same window answered `/auth/get-session` at 200 in 30-37ms.
The timeout was on the web side, not Core.

**2. A 401 was terminal.** Since #4221 the Ably singleton called
`client.close()` on 401. `close()` cannot be undone and nothing re-creates the
instance the mounted `AblyProvider` holds, so the browser then logged
`Ably presence enter/update failed: Error: Connection closed` on every presence
tick. Clearing the global also let a later island build a second client outside
the React tree, where no hook attaches channels to it.

**3. The budgets raced.** The browser fetch of `/api/ably/auth` and the ably-js
`realtimeRequestTimeout` default were both 10s, so ably-js often won and
replaced the classified error with its own opaque timeout message.

## Fix

- `getSessionResult()` returns `Result<Session | null, CoreAuthReadError>` via
  the existing `fetchCoreAuth` seam, which already classifies `timeout`,
  `network`, `invalid_json` and `http`. `ok(null)` is a signed-out browser;
  `err` is an outage. `getSession()` keeps its `Session | null` contract through
  `unwrapOr(null)`, so every existing caller is unchanged.
- `respondToBackgroundChatRead` answers 503 with the failure reason for an
  `err`, and keeps 401 only for `ok(null)`.
- The Ably auth callback never closes the client. Every failure goes back to
  ably-js to retry, as the `authUrl` path did before #4221. A real logout ends
  the session on its own.
- Budgets widened and nested strictly:
  `7s route to Core < 9s browser fetch < 10s ably-js`. The Core session read
  goes 5s to 8s; the background chat reads in front of it already allow 20-30s.
- The PPR `HANGING_PROMISE_REJECTION` rethrow moves into `fetchCoreAuth` so
  every Core auth read keeps it.

## Still open

Not in this PR, tracked separately:

- `use-ably-connection-health-publisher.ts` publishes health as
  `connection.state === "connected"` only. An `authorize()` failure can leave
  the socket connected with no channels attached, so the fallback pollers stay
  off while nothing is delivered.
- `use-notification-realtime.tsx` has no polling fallback at all.
- `next.config.ts` sets `tunnelRoute: true`, which mints a new random Sentry
  path per build. One sampled web log held 38 404s across four dead tunnel
  routes, so browser errors from any tab older than the current deploy are
  dropped. This is why the outage went unnoticed.

## Verification

- `pnpm --filter web test`: 678 files, 5122 passed, 1 skipped.
- `pnpm check` and `pnpm typecheck`: clean.
- Mutation-checked both fixes. Restoring the 401-on-timeout collapse fails 2
  tests in `src/app/api/chat`; restoring `client.close()` fails 2 tests in
  `realtime-singleton.client.test.ts`.
- New tests: 503-not-401 for both background chat routes; the Ably client
  survives a 401 and no second client is constructed.

### DIFF
--- a/apps/web/src/app/api/chat/[roomId]/messages/route.test.ts
+++ b/apps/web/src/app/api/chat/[roomId]/messages/route.test.ts
@@ -1,16 +1,18 @@
+import { err, ok } from "neverthrow";
 import { NextRequest } from "next/server";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { CoreApiRequestError } from "@/lib/clients/core.request";
 
-const { getSessionMock, listMessagesMock, listThreadMessagesMock } = vi.hoisted(
-  () => ({
-    getSessionMock: vi.fn(),
+const { getSessionResultMock, listMessagesMock, listThreadMessagesMock } =
+  vi.hoisted(() => ({
+    getSessionResultMock: vi.fn(),
     listMessagesMock: vi.fn(),
     listThreadMessagesMock: vi.fn(),
-  }),
-);
-vi.mock("@/lib/auth/auth.server", () => ({ getSession: getSessionMock }));
+  }));
+vi.mock("@/lib/auth/auth.server", () => ({
+  getSessionResult: getSessionResultMock,
+}));
 vi.mock("@/lib/services/chat-room.service", () => ({
   chatRoomService: {
     listMessages: listMessagesMock,
@@ -32,7 +34,7 @@ function get(query = "") {
 describe("GET /api/chat/[roomId]/messages", () => {
   beforeEach(() => {
     vi.resetAllMocks();
-    getSessionMock.mockResolvedValue({ user: { id: "user-1" } });
+    getSessionResultMock.mockResolvedValue(ok({ user: { id: "user-1" } }));
     listMessagesMock.mockResolvedValue({
       messages: [{ id: "msg-1" }],
       nextCursor: "cursor-2",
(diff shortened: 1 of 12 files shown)


mrosberghaus:
Closing in favor of #4360.

This PR got the session classification right (timeout is 503, not 401) and correctly stopped calling `close()` on the Ably singleton. That is not enough to recover.

SOK-986 treats `connection.state === "connected"` as healthy and pauses the HTTP fallback. A failed authorize can leave the socket up with no channels attached. Removing `close()` without changing that predicate keeps the pollers on the slow cadence, so unread and room lists stay dark until reload — longer than today, where `close()` at least flipped health to false.

#4360 starts from `main` and keeps this PR's 503 split and no-`close()`, then marks realtime healthy only after a successful token mint *and* TCP connected, so the existing fallback pollers run during the gap.

Not stacking on this branch: the health change has to land with the `close()` removal, not after it.

schaier-io:
@coderabbitai full review

PR #4356, opened 2026-09-10
fix(chat): refresh unread indicators through control channels

mrosberghaus:
Incoming room messages now refresh the tab unread counter and mobile Chats dot on non-chat pages and while the app is hidden or unfocused. Core publishes active-room invalidations on the existing per-user control channel, independently of notification preferences. Room message subscriptions return to the selected room only.

The change covers human and coworker sends, internal stream persistence, Soko Bot messages, completed mention responses, and membership timeline messages. Creation retries do not repeat invalidation; publishing failures preserve committed sends. Existing unread, mute, Thread participation, and background timer rules remain unchanged.

Unread refresh, the control subscription, and document-title ownership now live in persistent app chrome. They stay mounted on mobile room pages when the sidebar and bottom navigation are absent; returning to Chats renders the current dot without a remount fetch.

Room-message publication now runs concurrently with control invalidation after commit, so a slow control fan-out does not delay delivery to open-room viewers. Empty completed Soko Bot answers follow the failure path without publishing unread invalidation.

Accepted trade-off: the human author's other tabs receive no control invalidation for their own send. Their sidebar ordering catches up on the next collection refresh (normally the 60-second recovery timer while foreground and connected). This trade-off concerns ordering; it does not introduce unread attention for the author's own message.

Closes https://linear.app/masumi/issue/SOK-1032/tab-title-unread-counter-and-mobile-chats-dot-do-not-update-while-away

Verification:
- Review follow-up: 105 targeted Core tests and 265 chat component tests pass; `pnpm check` and `pnpm typecheck` pass.
- `pnpm check`
- `pnpm typecheck`
- `pnpm test` (initial implementation)
- `pnpm --filter web test` after the mobile fix: 678 files, 5,129 tests passed, 1 skipped. Hidden/unfocused room-page regressions cover title updates, muted/selected-room exclusions, and the current dot on return.
- Standards and spec reviews completed; findings fixed and re-reviewed.
- Browser acceptance remains unverified: local Core auth fails because the shared Neon database lacks `user.showRoomUnreadCount`. Named HTTPS services launch, but fixture and vault sign-in cannot establish a session. No shared database migration was applied. Two-user desktop/mobile proof and screenshots of the unread change remain outstanding.


mrosberghaus:
## Review follow-up

I reviewed this PR against SOK-1032. The product path is right: unread invalidation is a **membership fan-out**, not a notification fan-out; room channels stay selected-room only; hidden/unfocused reads still go through the existing control channel.

I did **not** change production code. The implementation already did the right thing. What was missing was a safety net on writers that the public message route never hits. I landed those tests in `97c5fd00e`.

### Why these tests exist

The SOK-1032 testing decisions say: prefer the Core message-creation route, and use focused tests at existing internal writers where the route cannot see the persist. Without those assertions, a later “optimization” (for example wiring `invalidateChatRoomMessageReaders` to the filtered notification recipient list, or dropping invalidate on a retry-safe persist helper) would stay green.

That is the failure mode I am locking. Not “more coverage.” Specifically: **a silenced member still gets `chat_rooms_changed`**, and **a committed create on an internal writer still publishes once**.

### What I added

1. **`persistUserMessageToChatRoom` success path** (`persist-assistant-to-chat-room.test.ts`)
   - Assistant creates already asserted `publishChatRoomsChanged`. User-turn creates only asserted the skip/retry paths did *not* publish.
   - The stream route mocks the whole persist helper, so dropping the success-path invalidate would not fail CI.
   - The new expect matches the assistant case: one `collections: ["active"]` publish to the room’s readers.

2. **`introduceSokoBot` and `deliverSokoBotTurnToDirectRoom`** (`soko-bot-chat.service.test.ts`)
   - Finalize was already tested. Intro and scheduled/direct delivery are committed creates the public message route never reaches (user story 10).
   - Success: one control publish after commit.
   - Already-introduced early return: no second invalidate. That is the same idempotency rule as `clientMessageId` retries.

3. **`emitChatRoomMessageCreatedEffects` when nobody opted into notifications** (`chat-room-message-created-effects.test.ts`)
   - This file was renamed because it now owns unread invalidation, but every case still only checked `createNotification`.
   - I hung the assert on the existing “writes nothing when nobody turned the row on” case: notifications stay silent, and `publishChatRoomsChanged` still fires for every non-author member (`user_alice`, `user_bob`).
   - That is the SOK-1032 invariant in one line: **do not derive unread recipients from the notification list.**

I proved the new expects fail if those invalidate calls are removed (4 failures, everything else still green), then restored the production code.

### Remaining findings (not landed)

I left these on purpose. None of them is a merge blocker. Treat them as a checklist, not as implied follow-up work in this PR.

| Finding | Why I did not change it | Recommendation |
| --- | --- | --- |
| Membership callers pass `[message]` into `publishChatRoomMembershipStatusMessagesBestEffort` instead of the whole array | Same per-message invalidation count; extra nesting only | Optional cleanup if you touch those files again. Not worth a drive-by. |
| Soko Bot `shouldEmit === false` only asserts notifications are skipped (`chat-message-effects.test.ts`) | Channel case already asserts invalidation; this is the same *shape* as (3) above, one file over | Cheap extra expect if you want the bot helper as tight as the human helper. Same intent: skip notify ≠ skip invalidate. |
| Mention-stream placeholder tests never assert control publish | Coworker dispatch already covers shell + settle (2 publishes). Failures inside invalidate are swallowed | Skip unless you want a focused placeholder test. Not a missing product path. |
| Coworker / Soko Bot replies invalidate twice (placeholder create + finalize) | Documented as advancing the attention clock | Leave it. Extra control publishes, not wrong. Revisit only if large-channel cost shows up. |
| `publishChatRoomsChanged` comment still says archive/restore/invitation only (`lib/ably/publish.ts`) | Comment drift, not behavior | One-line comment update if you are in that file. Otherwise ignore. |
| Browser proof still unverified (local Core auth / missing `user.showRoomUnreadCount`) | Spec says this is not a code defect | This is the remaining **product** proof: two users, receiver on a non-chat page then mobile nav, notifications off, send while hidden/unfocused. Title/dot refresh, mute and selected-room still correct, no membership-wide room subscriptions. I would not mark the issue done until that happens somewhere (preview is up). |

### Product residual to accept or ticket

On **mobile room pages**, the sidebar is a closed Sheet (children unmount) and `shouldShowMobileBottomNav` is false, so `ChatControlListBridge` is not mounted (`chat-mobile-bottom-nav.tsx` already documents that). Desktop room pages still have the sidebar bridge; mobile list / dashboard / agents have the nav bridge.

After dropping #4345’s catalog-room attaches, a phone that is **inside a room** will not live-update foreign-room unread. It catches up on remount/return. That matches the SOK-1032 problem statement (off a room page) more than user story 6 on a phone.

If story 6 on mobile matters, mount the control bridge from room chrome too — the hook already dedupes. If it does not, say so in the issue so the next reader does not re-open #4345’s catalog attach.

### Verdict

Ready to merge from a code/spec standpoint. The three Important items from review are now locked. Remaining rows above are optional polish plus one product call on mobile-in-a-room.

Please do **not** “fix” mute by skipping invalidation for muted members. Invalidating them is intended; the room-list DTO still excludes muted rooms from the title/dot.

mrosberghaus:
## Code review (base `8f6aab4` → head `53d7501`)

**Verdict: ready to merge with the two fixes below, pushed as `afe820e` on top of the branch head.**

Correctness against SOK-1032 holds. Every message-creation path (human send, coworker dispatch, mention stream, Soko Bot, persist-assistant, membership timeline, organization exit) invalidates exactly once; retries and the `clientMessageId` P2002 race do not republish; a failed control publish stays post-commit and silent; invalidation runs independently of notification preferences under `Promise.allSettled`. The web test drives the real `ChatControlListBridge` and `useChatControlChannel` against a mocked Ably client and reproduces the ticket end to end. `document.title` now has exactly one writer.

The two Important findings were load, not correctness. Both are fixed.

### Fix 1: desktop double-read of the active collection

**What:** `useChatTabUnreadPresence` now mirrors a mounted sidebar list instead of running its own reader. `useOrganizationChatRooms` registers itself in `membership-visible-rooms-store` while mounted and not `paintOnly`; the presence hook reads that flag with `useSyncExternalStore`, sets its scheduler key to `null` while a live list exists, and derives its rows from the store snapshot the sidebar already publishes. When the list unmounts (mobile room page, non-chat desktop routes if the sidebar is absent), the hook takes its own reads back and keeps the last mirrored rows until that read lands, so the dot and title do not flicker off during the handoff.

**Why:** This PR moved the presence hook from the bottom nav into `AppMobileChrome`, so on desktop it ran on every route next to the sidebar, which polls the same `collection=active` under its own scheduler. `fetchBackgroundJson` has no in-flight dedupe, so each invalidation and each 60s tick cost two reads of the heaviest chat list endpoint per viewer.

Files: `membership-visible-rooms-store.ts`, `use-organization-chat-rooms.ts`, `use-chat-tab-unread-presence.ts`, `chat-unread-document-title.ts` (readonly rows).

### Fix 2: open-room messages triggered a full refetch for their viewers

**What:** `ChatControlListBridge` now forwards `event.roomId` in the `organization-chat-rooms-changed` detail, and the presence hook returns early when it equals the active room id from the pathname.

**Why:** Core publishes `chat_rooms_changed` to every member except the human author, including people viewing that room. The bridge forwarded only `collections`, so every message in the open room refetched the active collection for a count that excludes the open room anyway. The skip lives in the presence hook, not the bridge, because archive, restore, and invitation publishers also tag the room id and the sidebar must still see those.

The sidebar still refetches on open-room messages, since that row's order legitimately changes. Skipping it there would need Core to distinguish message invalidations from lifecycle ones on the wire, which is a contract change I did not make. Note in Remaining if it matters for busy channels.

Files: `chat-control-list-bridge.tsx`, `organization-chat-events.ts`, `use-chat-tab-unread-presence.ts`.

### Tests added

- `use-chat-tab-unread-presence.test.tsx`: control event about the open room issues no read; a live sidebar list is mirrored (dot follows its rows, zero reads) and reads resume once it unmounts; the mirrored unread survives while the handoff read is pending.
- `use-organization-chat-rooms.test.ts`: the hook announces itself as live only while mounted and not `paintOnly`.

### Verification

- `pnpm --filter web typecheck` clean
- `pnpm check` clean
- `pnpm --filter web test` over `src/components/chat`, `src/app/(app)/chat`, `src/app/(app)/components`, `src/lib/ably`, `use-chat-unread-document-title`: 206 files, 1658 tests passed
- Browser acceptance still unverified (same DB blocker as the PR description).

### Remaining findings (Minor, not blocking)

1. `apps/core/src/services/soko-bot-chat.service.ts:399` gates invalidation on `turn.status === "COMPLETED"` instead of `succeeded`. A completed turn with an empty answer at `chainDepth === 0` takes the failed branch yet still publishes.
2. Invalidation is awaited before the room-channel publish in `persist-assistant-to-chat-room.ts` (182, 275), `chat-room-mention-stream.ts:171`, `chat-room-coworker-dispatch.service.ts:616`, and `soko-bot-chat.service.ts` (231, 291, 397). Viewers already in the room see the message one Ably REST round trip later than before. Publish the room event first or run both in parallel; only the human route does this in `waitUntil`.
3. Membership routes wrap single messages in a helper that already takes an array and never rejects: `admin/.../participants/post.ts:54`, `chat-room-invite-links/.../accept/post.ts:101`, `invitations/[id]/accept/post.ts:275`, `members/me/post.ts:173` loop one message at a time; `members/[userId]/delete.ts:151`, `members/me/delete.ts:248`, `rooms/[id]/patch.ts:556`, `chat-room-organization-exit.ts:187` wrap it in `Promise.allSettled([Promise.all(...)])`. Pass `statusMessages` once and drop the settled plumbing; it also saves one `chatRoomUserMember.findMany` per status message.
4. `apps/web/src/components/chat/__tests__/organization-chat-list-harness.tsx:45-55` still mocks `use-chat-control-channel`, `use-chat-unread-document-title`, and `lazy-ably-provider`; the sidebar no longer imports any of them.
5. The author's other tabs lose sidebar reorder on their own sends until the 60s timer, since the author is excluded from the control publish and the per-room catalog subscription is gone. Unread and title are unaffected. Worth stating as an accepted trade-off in the PR body.
6. `invitations/[id]/accept/post.ts:275-280` invalidates the acceptor twice (status message fan-out plus the explicit publish). The client coalesces it; cost is one extra Ably publish.

### Recommendations

- `publishChatRoomsChanged` issues one REST publish per member per message. Ably's batch publish endpoint would make that one request; worth a follow-up before channels grow past a few dozen members.
- Add a Core test for the P2002 race branch in `rooms/[id]/messages/post.ts` asserting the loser does not invalidate (the replay branch is covered, the race branch is not).
- A manual pass on a mobile room page (hidden tab, incoming foreign-room message, return to Chats) is the one check reading cannot substitute for.

PR #4361, opened 2026-09-10
feat(apple): associate streamed replies with thread parents

mrosberghaus:
Streams coworker replies inside Direct-message threads. Both composers share the existing room stream lock; streamed text and thinking appear only under the matching parent. Reopening a room resumes the retained thread association, and settlement refreshes history before removing overlays.

Web sources:
- [Streaming hook](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/hooks/use-coworker-direct-room-stream.ts)
- [Room/thread routing and settlement](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/components/rooms-client.tsx)

Scope: PARITY 09b, parent-scoped streaming, room/thread locking, resume association within the app session, thread thinking/error rendering, failed-draft restoration, and persisted reconciliation. Excludes new Direct creation, rich rendering, attachments, mention execution, stop/regenerate, API changes and new dependencies.

Validation: 172 Chat tests, Xcode app-target tests, iOS 17 shared compilation, SwiftLint and SwiftFormat. App regressions cover open/closed-thread settlement and prevent classic sends or room overlay leakage. No production test message was sent.

Manual test: open a coworker Direct and reply to a message. Check incremental text/thinking in its thread and disabled Send in the room while streaming. Close/reopen the thread and switch away/back to the room during generation; confirm the same parent and a single persisted reply. Simulate a rejected send and reopen the thread; confirm the draft returns without overwriting newer text.


mrosberghaus (commented):
PARITY 09b is on the right seams: parent-tagged overlays, one room send lock, in-session parent retention, settlement before overlay clear, and composer draft routing. No Critical issues.

## Follow-up in `4e37fb31f`

Independent review found two Important problems. Both are in this commit.

**Stuck overlay if the thread closed or reopened during settle.** `ThreadSession.open` always `close()`d first, which bumps `timeline.generation` and cancels `loadTask`. `settleDirectStream` then treated that bump as a failed room reconcile, so `settled()` returned false and the stream overlay stayed after the turn was idle. Closing the thread in the settlement window (or the view `.task` opening the same parent again) was enough to hit it.

The fix is at the owners:
- `open` is a no-op for the same parent (still updates the parent preview).
- `openThread` skips the initial reload when generation did not change.
- settle still succeeds after a successful room refresh if the thread was closed or reopened during the wait. Thread page refresh is best-effort while that parent is open.
- `ContentView` hops the auto-open write off the view update (`AGENTS.md` `@Published` rule). The destination binding already hopped `thread.close()`; this write did not.

**Missing regressions.** Happy-path coverage was already good. The commit adds:
- close during settle, including a second `openThread` while the thread GET is paused, and still clear the overlay
- re-entry resume: retained parent, `streamingThreadToOpen` once the phase leaves `.resuming`, settle opens the parent
- failed thread send keeps `parentMessageId` so only the matching composer restores, and newer typed text is preserved

## Remaining (Minor)

None of these should block merge.

1. **Successful settle clears retention but not in-session `parentMessageId`.** Idle 204 already nils it. Success only drops `retainedParents`. Harmless while idle (auto-open requires `isBusy`), but error/draft routing still keys off this field. Clear it in `reconcile` when `refreshed` is true.

2. **Thread thinking fallback is weaker than the room transcript.** `ThreadView` uses `latestThought` only. The room uses `latestThought ?? reasoning`. Whitespace-only live thought can show an empty Thinking body in the thread.

3. **Closed-thread settle still does `.initial` then `.latest`.** After a fresh `open`, `hasLoadedHistory` is true, so settle fires another GET. Correct, extra request. An already-loaded same-parent thread can just `.latest` (the idempotent `open` path already skips the first load).

Auto-open still has two call sites (`ContentView.task` and settle). Both go through `openThread`, which is now a no-op if the parent is already open. Pulling auto-open fully into `WorkspaceState` can wait.

## Recommendation

**Ready to merge.** 09b behavior matches the slice, Important items are in `4e37fb31f`, and Chat 174 plus `SokosumiTests` passed locally with lint/format clean. Leave the Minor items for a later pass unless they show up in the manual thread-stream check.

PR #4366, opened 2026-09-10
fix(chat): refresh sidebar ordering in the author’s other tabs

mrosberghaus:
A message sent in one tab now refreshes sidebar ordering in the author's other tabs through the existing per-user control channel. Previously the author was excluded and their other tabs waited for a collection refresh.

Include all room members in collection invalidation and remove the unused author argument from that helper. Notification recipients still exclude the author, and unread counting and selected-room-only subscriptions are unchanged.

Follow-up to #4356.

Verification: 75 targeted Core tests pass, plus 79 runtime tests; `pnpm check` and `pnpm typecheck` pass. Standards and correctness reviews found no issues. Live two-tab browser verification remains outstanding due to the previously documented local auth/database blocker.


apps/core/src/helpers/chat-room-message-created-effects.test.ts
```diff
@@ -138,7 +138,7 @@ describe("emitChatRoomMessageCreatedEffects", () => {
 
     expect(createNotificationMock).not.toHaveBeenCalled();
     expect(publishChatRoomsChanged).toHaveBeenCalledExactlyOnceWith({
-      userIds: [SUBSCRIBER_ID, QUIET_ID],
+      userIds: [AUTHOR_ID, SUBSCRIBER_ID, QUIET_ID],
```
mrosberghaus, line 141: This is the right assertion for a mixed room.

The author-only case `never notifies the author of their own message` still only checks that notifications stay silent. That used to skip publish entirely (`userIds` became `[]`). A later re-filter of the author would not fail that test.

Assert `publishChatRoomsChanged` with `{ userIds: [AUTHOR_ID], roomId, collections: ["active"] }` there, and keep `createNotification` uncalled.

apps/core/src/routes/v1/chats/rooms/[id]/messages/post.test.ts
```diff
@@ -474,7 +474,7 @@ describe("POST /chats/rooms/{id}/messages", () => {
       await Promise.all(waitUntilMock.mock.calls.map(([promise]) => promise));
       expect(response.status).toBe(201);
       expect(publishChatRoomsChangedMock).toHaveBeenCalledExactlyOnceWith({
-        userIds: recipients,
+        userIds: [USER_ID, ...recipients],
```
Comment (mrosberghaus) [inline apps/core/src/routes/v1/chats/rooms/[id]/messages/post.test.ts:477]: These cases now include `USER_ID`. The thread-reply test still expects `userIds: [ALICE_ID]` because its fixture's `userMembers` is only Alice.

That passes because the sender is missing from the roster mock, not because thread replies should skip the author. Put `USER_ID` in `userMembers` and expect `[USER_ID, ALICE_ID]`.

apps/core/src/helpers/chat-room-message-realtime.ts
```diff
@@ -126,7 +126,6 @@ export async function publishChatRoomMembershipStatusMessagesBestEffort(
       await Promise.all([
         invalidateChatRoomMessageReaders({
           roomId: message.roomId,
```
mrosberghaus, line 128: The matching test in `chat-room-message-realtime.test.ts` is still named "remaining readers" and mocks only `user_b`, so membership-status invalidation never proves the sender is included.

Include `user_a` in the member mock and expect both ids, or rename the test to match the mock.

apps/core/src/helpers/chat-room-message-created-effects.ts
```diff
@@ -106,15 +105,13 @@ async function getMemberUserIds(
...
+/** After commit: refresh every member, including the author’s other tabs. */
 export async function invalidateChatRoomMessageReaders(
   params: ChatRoomMessageReaders,
 ): Promise<void> {
   try {
     const memberUserIds = await getMemberUserIds(params);
-    const userIds = [
-      ...new Set(memberUserIds.filter((id) => id !== params.authorUserId)),
-    ];
+    const userIds = [...new Set(memberUserIds)];
```
mrosberghaus, line 114: `authorUserId` is gone from this type, but `emitChatRoomMessageCreatedEffects` still calls `invalidateChatRoomMessageReaders({ ...params, memberUserIds })`, so the field is still present at runtime.

Harmless today. If `authorUserId` were re-added here for filtering, emit would silently start excluding the author again. Pass `{ roomId: params.roomId, memberUserIds }` explicitly.

mrosberghaus (commented):
Ready to merge. No critical or important issues.

This is the right follow-up to #4356: one helper change, every caller updated, net deletion. `invalidateChatRoomMessageReaders` now publishes `chat_rooms_changed` to the unique member set with no author filter, which is the only server-side way to reach the author's other tabs. Notification fan-out still drops the author. Unread SQL already excludes the viewer's own user messages. Human POST still advances `lastReadAt` in the same transaction as the insert.

Call sites look complete: human POST, stream persist, assistant persist, Soko Bot, coworker dispatch, mention Thought placeholder, membership timeline. Tests that matter assert `userIds`, not just a removed parameter.

Four nits, none blocking. Inline notes sit on the closest hunks; two of the tests live outside this diff.

1. The author-only notification test never asserts collection invalidation.
2. The thread-reply POST fixture still omits the sender from `userMembers`, so `[ALICE_ID]` does not prove the new contract.
3. The membership-status realtime test still mocks only `user_b` and is named "remaining readers".
4. `emitChatRoomMessageCreatedEffects` still spreads the full params object into the readers helper, so `authorUserId` is present at runtime even though the type no longer declares it.

Browser two-tab proof is still outstanding, same local auth/DB blocker as #4356. Unit tests cover the Core contract.

apps/core/src/helpers/chat-room-message-created-effects.test.ts
mrosberghaus: Fixed in dee4acbbd. The author-only test now asserts exactly one active-collection invalidation to [AUTHOR_ID], while retaining the assertions that notification lookup and creation are skipped. All 55 tests across the three affected suites pass.

apps/core/src/routes/v1/chats/rooms/[id]/messages/post.test.ts
Comment (mrosberghaus) [reply apps/core/src/routes/v1/chats/rooms/[id]/messages/post.test.ts:477]: Fixed in dee4acbbd. The thread-reply fixture now includes USER_ID alongside ALICE_ID, and the invalidation assertion expects both. The affected route suite passes.

apps/core/src/helpers/chat-room-message-realtime.ts
mrosberghaus: Fixed in dee4acbbd. Renamed the test to “invalidates all members” and supplied user_a plus user_b for that test, asserting both recipients. The generic replay still does not trigger another invalidation. The suite passes.

apps/core/src/helpers/chat-room-message-created-effects.ts
mrosberghaus: Fixed in dee4acbbd. The call now explicitly passes { roomId: params.roomId, memberUserIds }; authorUserId is no longer forwarded at runtime. The affected tests, pnpm check, and pnpm typecheck pass.

PR #4367, opened 2026-09-10
perf(chat): batch control publishes and deduplicate invitation refreshes

mrosberghaus:
Publish chat collection invalidations through Ably's native batch API instead of one REST request per user. Each request covers up to 100 distinct control channels; larger audiences are split into batches. Existing channel names and event payloads stay the same. Per-channel and request failures are logged without replaying successful recipients.

Invitation acceptance excludes the acceptor from the membership-message refresh because the existing explicit event already refreshes both their active rooms and invitations. Other members still receive the membership update.

Follow-up to #4356. Independent of #4366; both branches merge cleanly and their combined behavior is tested. No Ably dashboard changes are required. This reduces HTTP requests; it does not claim fewer billable per-channel messages.

Verification:
- 55 targeted Core tests pass independently.
- 104 targeted Core tests pass with #4366 applied locally.
- `pnpm check` and `pnpm typecheck` pass.
- Standards and correctness reviews found no issues.

Uses the installed Ably 2.28 SDK's documented `batchPublish` API: https://ably.com/docs/pub-sub/api/javascript/rest/rest-client#batch-publish


mrosberghaus:
Independent review of this PR. Ready to merge. No Critical issues.

The one Important gap was a missing mixed-audience exclusion test. The existing skip test only covered “exclude the last remaining reader, publish `userIds: []`”. Invitation accept is a system membership message (`senderUserId: null`) with the acceptor plus other members. Nothing locked in that those others still get `collections: ["active"]`.

Landed in `87ad390c5` (`test(chat): cover mixed-audience invitation exclusion`):

- Members `[user_guest, user_other]`, exclude the acceptor, expect `userIds: ["user_other"]`.
- Room-channel create still publishes.

The filter was already correct. This just pins the production shape.

### Remaining

**1. Empty leftover list still calls `publishChatRoomsChanged`**

`invalidateChatRoomMessageReaders` filters excluded users, then publishes even when nothing is left. The inner helper no-ops, and the skip test now hard-codes that empty call. Filter first and return if empty. Nit.

**2. Third positional after optional `logContext`**

`publishChatRoomMembershipStatusMessagesBestEffort(messages, logContext, separatelyNotifiedUserIds)`. A later caller that passes user IDs as the second argument will change log text and notify everyone. Named options would be safer. Not worth blocking this PR.

**3. 101-user request-reject test does not check the leftover channel**

It checks call count and log count, not that the second spec is the leftover 1 channel. Cheap extra assertion if someone is already in that file.

I did not land the nits.

PR #4374, opened 2026-09-10
build(apple): add native Markdown parsers and emoji data

mrosberghaus:
Adds the approved native parsing prerequisites for Apple chat rendering. Foundation's Markdown output drops empty code blocks and table rows; the next rendering PR needs a complete Markdown tree, HTML parsing, and the same emoji mappings as web.

- Pin SwiftSoup 2.13.9 and swift-markdown 0.8.0; resolve swift-cmark 0.8.0. Parser products are test-target dependencies in this PR.
- Bundle 1,570 shortcode mappings and 322 emoticon spellings from web's emojilib 2.4.0 / emoticon 4.1.0, with MIT notices and a reproducible export script.
- Verify empty/nested/incomplete Markdown, malformed HTML and entities, unsafe HTML preservation for explicit sanitization, and resource decoding.

Scope: dependency/data prerequisite only, explicitly approved by Andreas. Rendering integration, sanitizer policy, emoji conversion, and visual changes stay in the separate slice 10 PR. No web changes, API changes, JavaScript runtime, or web view.

Web references: [Markdown pipeline](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/markdown.tsx), [sanitizer](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/utils/sanitizeMarkdown.ts), [room rendering](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/(app)/chat/components/room-message-row.tsx).

Validation:
- `swift test --package-path apps/apple/Packages/SokosumiChat`: 183 tests passed.
- Auth 34, CoreAPI 1, Realtime 27 tests passed.
- Xcode macOS build/app tests passed with ad-hoc signing and package-plugin validation skipped.
- `swift build --package-path apps/apple/Packages/SokosumiChat --target SokosumiChatTests --triple arm64-apple-ios17.0 --sdk "$(xcrun --sdk iphoneos --show-sdk-path)" --scratch-path /tmp/sokosumi-parser-deps-ios17`: passed, including both parsers.
- Pinned SwiftFormat/strict SwiftLint passed. `node --check apps/apple/scripts/update-emoji-data.mjs` passed; regeneration reproduced the documented SHA-256 hashes.

To inspect data, run `node apps/apple/scripts/update-emoji-data.mjs` after the pinned pnpm install and compare the resource README hashes. No visible app change is expected from this prerequisite.


mrosberghaus:
Independent review of this parser/emoji prerequisite (`acca42086`..`4fb7b356a`). No Critical or Important issues. I did not push a follow-up commit.

## What I checked

The PR matches the stated contract. SwiftSoup 2.13.9 and swift-markdown 0.8.0 are exact-pinned. Products sit on `SokosumiChatTests` only. The library target only gained `resources: [.copy("Resources/Emoji")]`. Resolved revisions match the tags (SwiftSoup `18b8032`, swift-markdown `3c6f952`, swift-cmark `924936d`). Committed `shortcodes.json` / `emoticons.json` hashes match the resource README. The export script walks web → remark-emoji 5.0.2 → node-emoji 2.2.0 → emojilib 2.4.0 / emoticon 4.1.0 and copies the MIT notices verbatim.

The Markdown tests hit the real Foundation gap: `AttributedString(markdown:options: .full)` drops empty fenced code and empty table rows. The unsafe-HTML tests assert SwiftSoup keeps `<script>`, `<style>`, and `javascript:` hrefs, so sanitization stays a later policy rather than a parser side effect. Apple CI is green on this head (Chat, Auth, CoreAPI, Realtime, Xcode build/app tests, SwiftLint/SwiftFormat).

## Why nothing landed

The Important bucket was empty. The items below are docs and assertion tightness. They do not break a pin, a resource load, or any production renderer, because this PR still does not wire rendering. Expanding a dependency PR for those would be scope, not a required fix.

## Left for the author (Minor)

1. **Design doc still leads with Foundation.** `apps/apple/docs/rich-message-rendering.md` still opens on `AttributedString(markdown:options:)` and says emoji data still needs selection. The new section at the bottom says replace Foundation after merge. Slice 10 will read the first heading. Rewrite "Rendering approach" so swift-markdown is the document tree, SwiftSoup is HTML fragments, and the bundled JSON is shortcodes/emoticons. Delete the stale "data source still needs selection" sentence.

2. **Empty-table assertions are weaker than the docs.** `RichTextDependencyTests.preservesEmptyTableRowsAndCells` locks body row counts and that the last row's cells are empty. It does not assert the leading empty body row, or empty header cells on `| | |`. Docs claim "empty table headers/middle/trailing rows/cells."

3. **Emoji tests lock counts and two keys, not the README hashes.** A hand-edit that kept 1,570 keys, `:smile:`, and `:+1:` would pass. The SHA-256s in `Resources/Emoji/README.md` are the real lock.

4. **`Document(parsing:)` is unqualified.** SwiftSoup also exports `Document`. It compiles because `init(parsing:)` is unique today. `Markdown.Document(parsing:)` would survive a later overload.

5. **PARITY checkpoint says "based on current main"** instead of pinning `acca42086`. PR #4370 recorded the SHA.

6. **`.copy("Resources/Emoji")` ships `README.md`** into the Chat bundle. Licenses belong there. The generator README does not.

## Slice 10, not this PR

Web's composer picker uses gemoji (`apps/web/src/lib/utils/emoji-shortcodes.ts`). This map is remark-emoji rendering (`+1` → 👍, no `:thumbsup:`). Keep that split. Do not treat `SwiftSoup.parse*` as sanitization; port `sanitizeMarkdown.ts` as a separate policy with fixtures. When the library target links these products, include the swift-markdown / swift-cmark / SwiftSoup notices in app distribution.

PR #4385, opened 2026-09-10
refactor(apple): organize app sources and share workspace coordination

mrosberghaus:
The Apple app's flat source tree mixed navigation, sidebar, composing and message rendering in large view files. This refactor groups sources by feature, names views for their roles, and moves cross-package lifecycle coordination into the UI-free `SokosumiWorkspace` package.

- Extract `ChatRootView`, `ConversationSidebarView`, `SignInView`, `ChatComposerView`, `MessageRowView` and `ParticipantAvatar`; rename transcript/thread/input/configuration files for their roles. Xcode's synchronized groups follow the real folders.
- Move `WorkspaceState` and thread coordination into `SokosumiWorkspace`, with the existing integration tests. The app injects its authenticated client provider and retains scene ownership and native adapters. Add a provider-boundary test and a package CI job. This local package extraction was explicitly approved; it introduces no new third-party library or API contract.
- Add an Apple README and update AGENTS/PARITY with ownership, dependency direction and contributor guidance.

Scope is behavior-preserving maintenance. Chat behavior, rendering, shared selection across windows, task cancellation/generation guards, API contracts and dependency versions remain unchanged. Web files are untouched.

Verification: Xcode app tests passed before and after extraction; all five package suites pass (54 workspace integration tests, 236 chat, 34 auth, plus CoreAPI/realtime). SwiftFormat and strict SwiftLint pass. The attempted iOS 17 deployment-target build was blocked because this Xcode installation lacks the iOS platform; iOS compilation is not claimed.

Manual check: build/run the Sokosumi scheme; confirm the navigator's feature folders, sign in, switch workspace/room, scroll history, hover Reply, open/close a thread, and focus the composer. Open Settings and another window to check existing shared state. No intentional visual changes.


mrosberghaus:
## Review follow-up: cleanup commit pushed (`2950ddc`)

I applied the review nits as a new commit on this branch. One important correction first: **only 4 of the 6 proposed fixes landed** — the other two turned out to be wrong, and this comment explains why so nobody re-attempts them.

### What changed (4 files, +6/−7)

1. **`apps/apple/Sokosumi/Chat/Timeline/TimelineStatusRows.swift`** — removed `import CoreAPI`, `import SokosumiAuth`, `import SokosumiChat`, keeping only `import SwiftUI`. `DaySeparatorRow`/`MembershipStatusRow` are pure SwiftUI over `String`, so the three carried-over imports were dead weight. Verified by full-file read plus a clean build.
2. **`apps/apple/.swiftlint.yml`** — the length-caps comment still cited `ContentView` as the example long view; it now cites `ChatRootView` / `ConversationSidebarView` (the types `ContentView` was split into). Sentence reflowed, meaning unchanged.
3. **`apps/apple/PARITY.md:448`** — the 09b follow-up note said "Auto-open in `ContentView`…"; fixed to ``ChatRootView``. (The `ContentView` mention in `AGENTS.md` is intentional generic guidance — "use role-specific names rather than generic `ContentView`" — so I left it.)
4. **`apps/apple/Packages/SokosumiWorkspace/Sources/SokosumiWorkspace/WorkspaceState.swift`** — added a doc comment on `public init` noting the host app must inject its authenticated client provider because the default (`{ _ in nil }`) resolves no client. Before the extraction, default construction fell back to `auth.coreClient()`; the default parameter now silently changes that, so the comment guards future callers.

### What I deliberately did NOT change (2 proposed fixes reverted)

The review also proposed removing `import CoreAPI` from `ChatComposerView.swift` and `ReplyThreadView.swift` as "unused". I tried both, and **the build failed**. Root cause, worth knowing:

- `Message` (and the room/thread schema types) are `Components.Schemas.*` typealiases from CoreAPI, re-exported through `SokosumiChat` (e.g. `ThreadSession.swift:9`, `RoomOutbox.swift:9`).
- Naming the type works through the intermediate import, but Swift 6.2 enforces *missing import of the defining module* on **member access**: `ChatComposerView` accesses `.id` (room, thread parent); `ReplyThreadView` accesses `.id`, `.content`, `.createdAt`, `.roomId`, `.threadReplyCount`.
- So `import CoreAPI` in both view files is load-bearing, not leftover. Both files are byte-identical to your version in the pushed commit.

Takeaway for future import cleanup in this codebase: `grep` for `Components.`/`OpenAPI` is not sufficient to prove a CoreAPI import unused — member access on typealiased schema types also requires it. Only a build proves it.

### Verification

- `xcodebuild … -scheme Sokosumi … build` (the exact CI invocation with ad-hoc signing): **BUILD SUCCEEDED**
- `mint run swiftformat --lint .`: 0/131 files require formatting
- `mint run swiftlint lint --strict`: 0 violations in 131 files

No behavior, API, or dependency changes — comments, docs, and one genuinely dead import block only.

PR #4386, opened 2026-09-10
feat(apple): add native emoji composing

mrosberghaus:
Adds emoji entry to room and reply composers: a native Emoji & Symbols button, partial-shortcode autocomplete (`:sm`), exact `:smile:` conversion, boundary-triggered emoticons, and trailing emoticon conversion before sending. Reuses existing bundled data and transport; no dependencies or API changes. Code literals remain unchanged.

PARITY splits independent emoji entry (12a) from rich formatting (12b). Includes renderer verification notes and an existing Xcode ordering-only edit.

Web references: [composer editor](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/composer-wysiwyg-editor.tsx), [emoticon rules](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/utils/composer-emoticons.ts), [shortcodes](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/utils/emoji-shortcodes.ts).

Validation: 241 Chat package tests passed; Xcode build and app tests passed, including UTF-16 selection, marked-text commit and undo/redo. Strict SwiftLint and SwiftFormat passed. Apple Development signed build passed including autocomplete.

Manual test: in a room and reply composer, type `:smile:` or `:D `, undo/redo, and verify draft restoration. Open Emoji & Symbols, search and insert an emoji over selected text. Check Enter submits and modified Enter inserts a newline; IME Return must only commit marked text.

Draft verification gap: button activation focused the editor, but computer control could not confirm the separate system picker panel. Picker presentation/insertion still needs a human check. No production test message was sent. Formatting toolbar, mentions and attachments are outside this slice. Autocomplete now belongs to this PR at the user’s request. Type `:sm`, navigate the native completion menu and accept with Tab/Return or click; Escape should preserve the draft. Visual completion-menu interaction also needs a manual check because computer control did not enter the test draft reliably.


mrosberghaus (commented):
## Review: no Critical issues, one Important (fixed and pushed)

I reviewed `e7117387..ffa29c2f` (10 files) against slice 12a plus the web parity references (`composer-wysiwyg-editor.tsx`, `composer-emoticons.ts`, `emoji-shortcodes.ts`). Note: this review covers the branch up to `ffa29c2f`; the shortcode-autocomplete commits you pushed while I was working (`4dae41d`, `bfd2dbf`) were not reviewed.

**Verified and cleared:** code-literal protection (swift-markdown `SourceLocation.column` is UTF-8 bytes, matching the `utf8.count + 1` math; `SourceRange.contains` is half-open), web parity (longest-first + first-wins dedupe, boundary punctuation set, whitespace equivalence with JS `trim()`, trailing-only flush, lowercase lookup, append-space rule), IME/marked-text guards, Enter/modified-Enter handling, draft save/restore and failed-send paths, and room + reply both flowing through `sendDraft`. The whole composer stack is `#if os(macOS)`, so no iOS gap. The acknowledged manual-picker-verification gap stands and correctly keeps 12a out of Done.

### What I changed and why (pushed as `9437ef8` on this branch)

**Important — accepted send left typing/conversion undo groups on the stack.** This PR newly enables composer undo (`allowsUndo = true`), but both send paths cleared the text with a plain `string = ""` assignment, which registers no undo itself — so the pre-send typing groups (including conversion edits with ranges into the old content) survived the clear. Outcome: Cmd+Z right after send would resurrect the just-sent text (confusion + duplicate-send hazard).

Fix: new `InputView.clearAfterSend()` (`MacComposerTextInput.swift`) that clears the text and calls `undoManager?.removeAllActions()`. It is used by the Enter path in `keyDown` (only when `submit()` returns true, so rejected sends keep text and undo) and by the Send-button path in `updateNSView` when the binding clears to empty. This is safe for manual select-all+delete: that flows through `textDidChange`, so the strings already match and the branch is never entered — user-cleared text stays undoable. Failed-send restoration is untouched (draft stays non-empty).

Tests: `acceptedSendClearsUndoHistory` (type `:D ` → converts → Return with accepted submit → asserts empty, `!canUndo`, `undo()` is a no-op) and `externalClearAfterSendDiscardsUndoHistory` (pins the helper used by the Send-button path). `MacComposerTextInputTests`: 38 passed, 0 failed; `swiftformat --lint` and `swiftlint lint --strict` clean. (One environment note: `xcodebuild test` with code coverage fails to link `cmark-gfm-extensions` with `Undefined symbol: ___llvm_profile_runtime` — pre-existing toolchain issue, unrelated to this change; tests were run with `-enableCodeCoverage NO`.)

### Remaining findings (minor, for you)

1. **Test gap: no shortcode-in-code case.** `preservesCodeLiterals` covers only emoticons (plus one `preparingToSend` fence case). The "code literals unchanged" requirement covers `:shortcode:` too; the shared `isCode` gate makes failure unlikely, but consider adding `` `x :smile: ` `` with caret before the closing backtick → nil, plus a non-ASCII-prefix code case (emoticon inside backticks on a line starting with emoji) to lock the UTF-8 column math.
2. **Undo test doesn't pin live behavior.** `emojiConversionSupportsUndoAndRedo` wraps typing in explicit `begin/endUndoGrouping`, but real typing has no explicit group — with the `breakUndoCoalescing` pair, live first-undo should restore the typed `":D "` (conversion reverted, space kept). Consider an ungrouped test asserting that.
3. **`canSend`/counter length-source split.** `ChatComposerView.swift:38` checks `canSend` on converted text while the counter/limit use raw `draft`. Only matters within a few UTF-16 units of the 10k limit with a trailing emoticon; consider computing `preparingToSend(draft)` once and sharing it.
4. **Readability nits in `ComposerEmoji.swift`:** `== 58` magic number for `:`; type name `ComposerEmoji` reads as a model rather than the caret-local edit it is; `preparingToSend` doesn't document that flush is emoticon-only (shortcodes intentionally excluded, matching web).
5. **`project.pbxproj` one-line reorder hunk** — ordering-only Xcode noise; consider reverting to keep the diff clean.
6. **Emoji button polish:** no screenshot for the new in-field button placement; picker presentation silently no-ops when `input.window == nil` (consider a log); consider an accessibility identifier for future UI tests.
7. **Shortcode scan relies on data invariants instead of web's `EMOJI_QUERY_PATTERN`.** `shortcode()` has no `[a-z0-9_+-]` gate, unlike `matchExactEmojiShortcodeClosed`. Behaviorally equivalent today (all 1570 `shortcodes.json` keys lowercase incl. `+1`/`-1`, no colons, so invalid names fail lookup; `::smile:`/`:x:smile:` correctly yield nil on both sides) — fragile only if the data changes. Optional hardening, not a blocker.

### Assessment

Ready to merge: **with fixes** — the Important item is fixed above; the rest are minor suggestions. Still needed: your planned human picker-panel check.

PR #4387, opened 2026-09-10
feat(web): open the New Task wizard in place from the sidebar

mrosberghaus:
## Summary

Clicking **New Task** in the sidebar navigated to the Task Manager with `?create=true`, loading the whole board behind the wizard. Now:

- **Seated members** get the New Task wizard as a modal over whatever page they are on. No navigation; cancel returns them to where they were. After creation they land on the task detail page as before.
- **Unseated organization members** keep the link to the Task Manager, which explains the Seat requirement. The item also stays a plain link while the instant-nav shell is loading.
- The wizard opens instantly: its chunk is warmed once the app shell is idle (off the shell's critical path), and everything it needs (assignees, projects, agent names, DESIGN.md) arrives in **one** server request behind a skeleton. Lists are fetched fresh on every open so a workspace switch never shows stale members.
- The `/tasks?create=true` deep link keeps working. Dismissing the modal now rewrites the URL only when that deep link opened it, so a wizard opened in place leaves the page's own URL alone (previously it would have stripped `create` from, for example, `/projects?create=true`).
- The unlinked `/tasks/new` page is removed together with the form's page variant (back link, details heading, coworker grid), its two grid-only components, and the five `NewTask` translation keys only that page used. The modal is now the only host of `TaskForm`, so the `variant` prop is gone.
- Extracts `listTaskAssigneeOptions` (owner Soko Bot, members, coworkers) and uses it from the new server action and both task edit pages instead of adding another copy of that composition.
- `CONTEXT.md` gains **Task Manager** and **New Task wizard**.

## Verification

- `pnpm --filter web test` (685 files; one unrelated admin Soko Bot test flaked under load and passes alone)
- `pnpm check`
- `pnpm --filter web build`
- Not clicked through in a browser; the worktree stack was not run.

No schema changes. No screenshots (no browser run).


mrosberghaus:
## Grok review

### Fixed
- Important: unmount the New Task wizard on dismiss (and load error) so `loadNewTaskWizardOptions` does not stay subscribed and refetch on window focus; query also uses `staleTime: 0` and `refetchOnWindowFocus: false` (`35dfd3efa` — `apps/web/src/app/(app)/components/new-task-wizard.tsx`, `apps/web/src/app/(app)/components/new-task-wizard-provider.tsx`)

### Remaining
- Minor: `apps/web/src/app/(app)/components/new-task-wizard-provider.tsx`: `requestIdleCallback` has no `{ timeout: 2000 }`; the 2s fallback only runs when the API is missing. `dynamic()` also has no `loading` UI, so a first click before warmup can look like a no-op.
- Minor: `apps/web/src/app/(app)/tasks/components/create-task-modal.tsx` / `task-form.tsx`: leftover unused `TaskFormLabels` (`continueLabel`, `previousLabel`, `nextLabel`, `searchPlaceholder`, `askPrompt`, `promptHint`, `allCompanies`) and unread `unassignedDescription` on edit pages still keep dead `App.Tasks.NewTask` keys looking used.
- Minor: `apps/web/src/app/(app)/components/user-avatar/workspace-switcher.tsx`: `getTaskDetailBasePath` still special-cases `/tasks/new` after that route is gone.
- Minor: `apps/web/src/app/(app)/tasks/components/create-task-modal.tsx`: wizard skeleton `aria-busy="true"` on a generic `div` with no `role="status"` or accessible loading text.
- Skipped / disagreed: no Critical findings. Minors left as listed; they are leftover page-variant polish, not the closed-modal refetch.

### Assessment
In-place wizard, seat/shell fallbacks, one-request load, deep-link URL rewrite, and `/tasks/new` removal match the spec; the closed wizard no longer keeps a standing options query. Not clicked through in a browser in this worktree.

mrosberghaus:
Addressed in `6b7691410`:

- **Close path**: kept the unmount-on-close from `35dfd3efa`, but the wizard now reports closing through a new `onClose` on `CreateTaskModalProvider`, called inside `handleClose`, instead of an effect watching `open` (the web effects rule lists that under "notifying parent"). The provider keeps a monotonic instance number plus an open flag, so a reopen never reuses the previous query key. `staleTime: 0` dropped since it is the default; `refetchOnWindowFocus: false` kept.
- `requestIdleCallback` now passes `{ timeout: 2000 }`.
- Unread `TaskFormLabels` (`continueLabel`, `previousLabel`, `nextLabel`, `searchPlaceholder`, `askPrompt`, `promptHint`, `allCompanies`, `unassignedDescription`) removed from the form, the modal, and both edit pages, together with the seven orphaned `App.Tasks.NewTask` keys and `App.Tasks.EditTask.unassignedDescription` in en/de/es. Parity green.
- `getTaskDetailBasePath` no longer special-cases `/tasks/new`.
- The skeleton wrapper is a `role="status"` region labelled from `App.Tasks.Actions.loading`.

Not changing: a `loading` UI on `dynamic()`. The shell it would render is a second `TaskFormModal` (a second Radix Dialog) that gets swapped for the real one when the chunk lands, so the modal re-animates on every first open. With the idle warm-up now bounded to two seconds, the window for a click before the chunk exists is well under a second on a cold load, and such a click is not lost: the wizard mounts the moment the chunk arrives.

PR #4411, opened 2026-09-11
fix(core): keep non-uuid mention keys out of sokoBot lookup

mrosberghaus:
## Why

`POST /v1/chats/rooms/:id/messages` mention fan-out crashed in production when a message named both a human and a Soko Bot. Human mention keys can be 32-character Better Auth ids. `loadChatMentionNames` passed that mixed list into `chatRoomSokoBotMember.sokoBotId`, which is a Postgres uuid column.

This change keeps the full key list on the text user and coworker lookups, and sends only hyphenated uuid keys to the Soko Bot member read. Mention names still resolve. The Prisma uuid error does not.

Fixes SOKOSUMI-CORE-3R

## Scope

- `apps/core/src/helpers/chat-mention-names.ts` partitions mention keys before `chatRoomSokoBotMember.findMany`
- `apps/core/src/helpers/chat-mention-names.test.ts` reproduces a mixed opaque-id plus Soko Bot uuid message, locks user/coworker `in` lists to the full key list, and locks the opaque-only soko bot query skip

Out of scope: changing mention token syntax, migrating legacy user ids, or wrapping Prisma in try/catch.

## Tradeoffs

Mention tokens are untyped (`@<id>`), so this partitions by id shape rather than a member-kind tag. That matches the schema. User and coworker ids are text. SokoBot ids are uuid.

## Blast Radius

Chat mention notification previews after `POST /v1/chats/rooms/:id/messages`. Without the fix, mixed human plus Soko Bot mentions fail open as an unhandled rejection and skip name resolution for that message. With it, those messages complete fan-out.

## Verification

- Fail: `pnpm --filter @sokosumi/core test src/helpers/chat-mention-names.test.ts` on the first commit. The mixed-key case throws `invalid input syntax for type uuid`.
- Pass: the same file after the fix. Follow-up test commit locks user/coworker `in: [AUTH_USER_ID, BEN_ID]` and the opaque-only soko bot skip. 8/8 passed.
- Related fan-out and mention-notification tests passed on the fix commit (75 tests across 3 files).
- `pnpm --filter @sokosumi/core typecheck` passed. Repo `pnpm check` and `pnpm typecheck` passed on commit.

<div><a href="https://cursor.com/agents/bc-9538c831-7311-4595-8ee2-4b9b6182f24f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9538c831-7311-4595-8ee2-4b9b6182f24f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/src/helpers/chat-mention-names.test.ts`: mixed-key test does not lock the user/coworker `in` lists, so a regression that uuid-filtered those reads could still pass.
- Minor: `apps/core/src/helpers/chat-mention-names.test.ts`: no 32-character-only mention case (no Soko Bot token) for the empty `sokoBotIds` skip path.
- Skipped / disagreed: none. Both remaining items are test-coverage nits; production filter and user/coworker full-key lookups are correct.

### Assessment
Ready to merge: Yes
The mixed-key uuid crash is fixed in the shared helper every mention fan-out path already calls; 7/7 `chat-mention-names` tests pass.

PR #4412, opened 2026-09-11
fix(core): reject banned and deleted users on Core's bearer paths

schaier-io:
Closes [SOK-997](https://linear.app/masumi/issue/SOK-997/securitycore-reject-banned-and-deleted-users-on-every-bearer) (security review finding AUTH-03, High).

## User-facing impact

Banning a user now takes effect on Core's bearer-authenticated API traffic. Before this, a ban cleared the person's sessions but left every API key and OAuth access token working until it expired on its own, and Better Auth API keys have no default expiry. Deleting an account was worse: `Apikey.referenceId` is a plain string with no relation to `User`, so the key row outlived the account and still authenticated, at the default `user` role.

An OAuth access token that was revoked on its own now stops working too. Only the refresh token's `revoked` column was consulted, so withdrawing a single access token had no effect while its grant lived.

## What changed

One predicate, `isActiveUser`, run on both user bearer paths in `authMiddleware`. It rejects a missing user and an active ban, and it is folded into the user lookup each path already made, so there is no extra query.

A ban whose `banExpires` has passed is not an active ban. The boundary follows Better Auth's own `banExpires.getTime() < Date.now()` rather than SOK-997's looser "null or in the future" wording, so that one instant cannot be a ban in Core and a lifted ban at sign-in. The divergence is documented in the JSDoc.

## Deliberately out of scope

Five surfaces still give a banned user some authority. None is a regression, and the last two are narrowed by this PR rather than untouched. Each is recorded with evidence in the commit messages or here.

- **Better Auth's own `/auth/**` routes.** They do not run through `authMiddleware`. With `enableSessionForAPIKeys`, the api-key plugin builds an in-memory session instead of calling `internalAdapter.createSession`, so the admin plugin's ban hook never fires (`@better-auth/api-key` 1.7.3, `dist/index.mjs:2426-2441`). It does reject a missing user (`:2422-2425`), so this is a banned-user hole only. That is SOK-996's subject.
- **`verifyAgentApiKey`.** A Soko Bot key carries an owner `userId` that is never checked against that owner's state. It is an agent credential rather than a user one; closest ticket is SOK-1006.
- **`coworkerContextMiddleware`.** A vendor coworker key with `X-Context-User-Id` acts inside that user's workspace, and the context user is validated with `select: { id: true }`, rejecting only a missing row (`apps/core/src/middleware/coworker-context.ts:77-84`). A banned user's workspace stays reachable that way. Sibling of the `verifyAgentApiKey` gap, different file and different credential class.
- **The session cookie cache (AUTH-04), narrowed but not closed.** The second commit adds the same guard to `sessionMiddleware`, which catches a ban whenever `getSession` reaches the database. But `session.cookieCache` is enabled with a 5 minute default (`apps/core/src/lib/auth.ts:282-286`, `apps/core/src/config/env.ts:60-63`), so a signed cookie minted before the ban still wins for up to that window.
- **`authenticationMethod` on API-key sessions (SOK-996).** A session minted from an API key reports `authenticationMethod: "session"` (`apps/core/src/middleware/auth.ts:708`), so it satisfies `requireInteractiveUserAuthContext` and `requireInteractiveAdminAuthContext` — the two helpers whose stated purpose is to require a human at the keyboard. Pre-existing, and a bigger change than this PR.

Ticket steps 4 (revoke credentials on ban and delete) and 5 (foreign key on `Apikey.referenceId`) are deferred. Once "reject a missing user" lands, an orphaned key can no longer authenticate through Core, so those become cleanup rather than the fix.

## Second commit: the session path

The first commit claimed the session path needed no change. That was wrong, and the second commit fixes it. With `enableSessionForAPIKeys`, the api-key plugin matches an unscoped `before` hook on the `x-api-key` header and assigns `ctx.context.session` without `internalAdapter.createSession` (`@better-auth/api-key` 1.7.3, `dist/index.mjs:2394`, `:2426-2441`), so the admin plugin's ban hook never runs on it. Since `authMiddleware` only reaches `bearerMiddleware` when an `Authorization` header is present, the same key sent as `x-api-key` skipped the new guard entirely.

`sessionMiddleware` now runs the predicate too. `getSession` already returns `banned` and `banExpires`, so this adds no query.

## Verification

- `pnpm --filter core test` — `Test Files 497 passed | 3 skipped (500)`, `Tests 5242 passed | 11 skipped (5253)`
- `pnpm --filter core typecheck` — clean
- `pnpm check` — `Checked 3899 files. No fixes applied.`
- Mutation check: with the new guards stripped from the source, exactly four of the new tests fail while 35 others still pass, so they bind to the fix rather than passing vacuously.

No schema changes, no migrations, no data scripts. No UI, so no screenshots.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened authentication checks across API keys, OAuth tokens, and sessions.
  * Blocked access for missing, banned, archived, or deleted users.
  * Rejected OAuth tokens that have been individually revoked.
  * Correctly restored access when temporary bans expire.
* **Tests**
  * Expanded coverage for user status, ban expiration boundaries, authentication methods, and revoked credentials.


schaier-io:
@coderabbitai full review

PR #4406, opened 2026-09-11
feat(chat): use stable human mentions and email subtitles

mrosberghaus:
Human mentions now store `@userId` instead of a label derived from the user's name. Both web and Apple show email below the display name in the picker and support finding people by email. Existing `@userId:slug` messages still resolve by ID and retain their stored text when edited.

This is the explicitly approved follow-up to #4392. It updates the existing web composer, Apple chips, and shared notification preview reader; it adds no dependencies or endpoints. Coworker, Soko Bot, and @all writers remain unchanged.

Web references: [room composer](https://github.com/masumi-network/sokosumi/blob/codex/human-mention-identity/apps/web/src/app/(app)/chat/components/room-composer.tsx), [composer serialization](https://github.com/masumi-network/sokosumi/blob/codex/human-mention-identity/apps/web/src/components/ui/mention-textarea-utils.ts), [room roster](https://github.com/masumi-network/sokosumi/blob/codex/human-mention-identity/apps/web/src/app/(app)/chat/components/rooms-client.tsx). Scope and handoff are recorded in apps/apple/PARITY.md.

Verification: 291 Chat and 55 Workspace tests; Xcode app build/tests; 148 targeted web tests; 141 preview tests; web typecheck; repo check/typecheck; pinned SwiftFormat/SwiftLint. Full web suite has 28 failures in task-activity and time-ago; the same 28 reproduce on unchanged main.

Native interactive verification passed on the signed build: searching `@nmkr` matched email domains, rows showed names with email subtitles, and Down + Tab inserted the selected person as a chip only after confirmation. The composer was restored empty; no message was sent. CI Web and Core tests also pass.

Remaining manual checks:
- On both clients, type @ and search by name or email. Human rows should show name, avatar, and email.
- Select a person, copy the chip as text, and confirm it contains only @userId. Verify room and thread sending.
- Open an older mention after a rename and confirm it displays the current name. Edit/copy preserves the old token.
- Check a mention followed by a comma, a missing-name user, and an agent mention.

No stored-message migration is needed. Existing readers already resolve bare IDs; old tokens remain supported. A rollback of writers does not require rewriting messages.

PR #2851, opened 2026-04-15
refactor(database): uuid(7) defaults and Better Auth UUID id generation

mrgrauel:
## Summary

Aligns primary-key generation with UUID v7 in Prisma, configures Better Auth to emit UUID-shaped identifiers for adapter-managed tables, and standardizes a few app-generated identifiers (UUID v7 where we want time-sortable correlation/lease tokens, UUID v4 where we align with external or legacy expectations). New DB rows use consistent defaults where applicable, without requiring a migration solely for default-expression churn when the physical schema already matches.

### Prisma (`packages/database/prisma/schema.prisma`)

- Replace `@default(cuid())` and `@default(uuid())` with `@default(uuid(7))` on model primary keys where applicable.
- Add `@default(uuid(7))` on `@id` fields that previously had no default (e.g. Session, Account, Verification, Passkey, RateLimit, Apikey, Subscription).
- Keeps existing `Workspace` / `workspaceId` `@db.Uuid` mapping where already migrated in the database.

### Better Auth

- **`apps/web/src/lib/auth/auth.ts`** and **`apps/core/src/lib/auth.ts`**: `advanced.database.generateId: "uuid"` so Better Auth generates UUID ids for records inserted through the Prisma adapter.

### Subscription helper and tests (`packages/database`)

- **`src/helpers/subscription.ts`**: initial free `subscription.create` omits `id` so Prisma applies `@default(uuid(7))`.
- **`src/helpers/subscription.test.ts`**: expectations updated so create payloads no longer include `id`.

### Other explicit UUID generation

- **`apps/core/src/services/sync-lock.service.ts`**: lock `ownerToken` suffix uses **UUID v7** (with `INSTANCE_ID` prefix).
- **`apps/web/src/lib/api/utils.ts`**: `handleApiError` default `requestId` uses **UUID v7**.
- **`apps/core/src/routes/v1/conversations/post.ts`**: default `openaiId` uses **UUID v4** (`uuidv4()`) when not provided (OpenAI-facing id shape).

### Debug / cleanup

- **`apps/core/src/routes/debug/pkce/post.ts`**: removed commented-out route and schema definitions; active PKCE helper unchanged (`state` still uses `randomUUID()` from Node).

### Documentation

- **`packages/database/AGENTS.md`**, **`apps/web/AGENTS.md`**, **`apps/core/AGENTS.md`**: notes on primary keys / UUIDs and Better Auth ID generation.

## Technical notes

[...]

PR #4422, opened 2026-09-11
fix(core): keep legacy auth ids out of the uuid soko bot mention lookup

schaier-io:
Fixes SOK-1043.

## What broke

Chat notifications stopped arriving for several people on 2026-09-11. Mentions produced nothing, in the notification center or as an OS banner, while plain direct messages kept working.

`loadChatMentionNames` passes every mention key in a message to three member lookups. Two of them filter plain `String` columns. The third filters `ChatRoomSokoBotMember.sokoBotId`, which is a `uuid` column. A human mention key can be a legacy 32-character Better Auth id, and Postgres rejects that comparison outright rather than matching nothing.

That read sits in `fanOutChatNotifications` above the per-recipient loop, so the rejection escapes before any recipient has been written a row. The caller is a bare `waitUntil(...)`, so nothing caught it and the fan-out's own Sentry capture never ran.

Users created before 2026-04-15 carry the 32-character ids; `generateId: "uuid"` landed that day in #2851. A message with no mention token produces no keys and returns early, which is why direct messages were unaffected.

Introduced by #4406, which widened the mention token regex to accept those legacy ids. Before that, every key was a UUID or `all`, and `all` was already filtered out.

## The fix

Filter the soko bot lookup to the keys that column accepts, and skip the query when none are left.

The predicate is a regex rather than a schema validator. `z.uuid()` rejects hyphenless UUIDs and off-spec version or variant nibbles, all of which Postgres accepts, so it would have silently dropped mentions the column would have matched.

## Verification

Reproduced against a migrated local database through the real Prisma client:

```
chatRoomUserMember (userId String): OK, 0 rows
chatRoomCoworkerMember (coworkerId String): OK, 0 rows
chatRoomSokoBotMember (sokoBotId @db.Uuid): THREW PrismaClientKnownRequestError code=P2007
Invalid input value: invalid input syntax for type uuid: "Xk3mQp7RvT2yLb9Nc4Wd8Hf6Jg1Zs5Aq"
```

Every shape the predicate admits was then checked against live Postgres through the same adapter: canonical, hyphenless, uppercase hyphenless, and off-spec nibbles all query cleanly; only the legacy auth id throws.

- `pnpm --filter core test` — `5232 passed | 11 skipped (5243)`
- `pnpm typecheck` — clean
- `pnpm check` — clean

Mutation tested: with the filter removed, two of the new tests fail. With the pattern tightened to mandatory hyphens, an RFC version class, or an RFC variant class, the off-spec test fails in each case.

## Not in this PR

`fanOutChatNotifications` still awaits `loadChatMentionNames` outside the per-recipient `try/catch`. A pool error or a `P2028` on that same query reproduces this outage exactly, with no invalid id involved. The preview is decorative and should degrade to an empty name map instead of costing everyone their notification. Left alone to keep this PR to the regression.

Separately, push subscriptions are never re-activated on app open, so a subscription that dies for any reason stays dead until the reader opens Account and presses the cell. That is an acceptance criterion of SOK-876, which is marked Done, but the code is not there.

### DIFF
--- a/apps/core/src/helpers/chat-mention-names.test.ts
+++ b/apps/core/src/helpers/chat-mention-names.test.ts
@@ -23,6 +23,8 @@ import { loadChatMentionNames } from "./chat-mention-names";
 const ROOM_ID = "550e8400-e29b-41d4-a716-446655440000";
 const ADA_ID = "019fc7e4-e4bd-7005-900c-66e44d33f5e4";
 const BEN_ID = "019fc7e4-e4bd-7005-900c-66e44d33f5e5";
+/** A member who signed up before Better Auth issued uuid ids. */
+const LEGACY_AUTH_ID = "Xk3mQp7RvT2yLb9Nc4Wd8Hf6Jg1Zs5Aq";
 
 beforeEach(() => {
   vi.clearAllMocks();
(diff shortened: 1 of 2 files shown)


schaier-io:
Superseded by #4411, which landed the same fix at 11:48 today. I opened this without fetching `main` first; the outage was already resolved.

The remaining differences are narrow and move to a follow-up stack: a hyphenless uuid key is valid to Postgres but filtered out by the merged regex, and the mention-name tests can pin the case-insensitive flag and the off-spec nibbles by mutation. Neither is needed to end the outage.

Closing so nobody reviews a duplicate.

PR #4423, opened 2026-09-11
fix(chat): drop participant hover card from sidebar DM avatars

mrosberghaus:
## Summary

Sidebar direct message rows no longer show a participant hover card (profile + **Message** button) when hovering a face. The row link already opens that direct, so the card duplicated the row's own action and covered the neighbouring rows.

- Sidebar DM avatars are now plain marks with presence dot (and the 1:1 screen-reader availability text stays).
- The hover card remains on message rows and mentions.
- `DirectRoomAvatarStack` lost its `canOpenHumanDirect` / `selectedRoomId` props and the `"use client"` directive (no hooks left).

## Verification

- `pnpm --filter web test src/components/chat` (30 tests pass)
- `pnpm check` and `pnpm typecheck` green (pre-commit hook)


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/direct-room-avatar-stack.test.tsx`: the no-hover-card assertion never opens the card, and the suite dropped the HoverCard mock that always mounted `data-testid="chat-participant-hover-card"`. Real Radix content is absent until open, and `interactive={false}` already strips `role` / `tabIndex` / `aria-label`, so putting the old wrapper back would still pass. Mock `ChatParticipantHoverCard` (or hover with that mock) and assert the test id is absent.
- Minor: `apps/web/src/components/chat/direct-room-avatar-stack.tsx`: `getDirectParticipants` is documented as “Other humans + all coworkers”; the filter also keeps soko bots (the always-online test depends on that). Say humans, coworkers, and soko bots, excluding the current user.
- Skipped / disagreed: PR body says the hover card remains on the roster panel — roster never used `ChatParticipantHoverCard`; it is a dedicated member list with Message. Not a product gap; do not add a hover card there to match the sentence.

### Assessment
Sidebar DM faces are plain marks with 1:1 SR availability and no nested hover-card action; message rows and mentions still mount `ChatParticipantHoverCard`. The leftover notes are a weaker regression test and a comment, not merge blockers.

mrosberghaus:
Addressed the two Grok notes in the follow-up commit: the test now mocks `ChatParticipantHoverCard` so the wrapper leaves a trace (verified the suite fails against the previous component), and the helper comment now names soko bots. Dropped the roster-panel sentence from the PR body.

PR #4433, opened 2026-09-11
fix(chat): insert newline on Enter from touch keyboards

mrosberghaus:
## Summary

On phones the on-screen keyboard has no Shift+Enter, so there was no way to write a multi-line message. Plain Enter in the room composer now inserts a line break on touch devices (`hover: none`) and the purple Send button sends, matching WhatsApp and iMessage on iPhone.

- Desktop unchanged: Enter sends, Shift/Cmd/Ctrl+Enter inserts a newline.
- Inline message edit unchanged on all devices: Enter still saves (it has no Save button).
- Room composer no longer sets `enterkeyhint="send"`, so the iOS return key reads "return" instead of "Send". The inline editor keeps it.
- `devicePrefersHover()` moved from `room-message-row.tsx` to `@/lib/utils/device-prefers-hover` so the composer reuses the same `(hover: hover)` check.

Reverses point 2 of [SOK-815](https://linear.app/masumi/issue/SOK-815) ("iOS return key is Send") per Andreas, 2026-09-11. Point 1 (first tap sends) is untouched.

## Verification

- `pnpm --filter web test` on `src/app/(app)/chat`, `src/components/chat`, `src/lib/utils`: 204 files, 1968 tests pass
- Biome via husky precommit
- Manual on a phone: open a channel, type, tap return → new line; tap Send → message sent. Not verified in a simulator (no iOS runtime installed on this machine).

### DIFF
--- a/apps/web/src/app/(app)/chat/components/room-message-row.tsx
+++ b/apps/web/src/app/(app)/chat/components/room-message-row.tsx
@@ -118,6 +118,7 @@ import type {
   ChatRoomUserParticipant,
 } from "@/lib/clients/generated/core";
 import { cn } from "@/lib/utils";
+import { devicePrefersHover } from "@/lib/utils/device-prefers-hover";
 import { classifyFilePreview } from "@/lib/utils/file-preview";
 import { getInitials } from "@/lib/utils/text";
 import { ChatParticipantHoverCard } from "./chat-participant-hover-card";
@@ -780,13 +781,6 @@ const LONG_PRESS_DELAY_MS = 450;
 const LONG_PRESS_MOVE_TOLERANCE_PX = 12;
 const TOUCH_MESSAGE_SELECT_NONE_CLASS =
   "[@media(hover:none)]:select-none [@media(hover:none)]:[-webkit-touch-callout:none]";
-
-function devicePrefersHover(): boolean {
-  if (typeof window === "undefined") {
-    return true;
-  }
-  return window.matchMedia("(hover: hover)").matches;
-}
 
 function clearDomTextSelection() {
   window.getSelection()?.removeAllRanges();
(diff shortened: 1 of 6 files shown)


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/composer-wysiwyg-editor.test.tsx:660`: `innerWidth` 390 still expects Enter to send. Phones are now `hover: none`, not width. Rename or drop; it passes only because happy-dom has `maxTouchPoints === 0`.
- Minor: `apps/web/src/components/chat/composer-wysiwyg-editor.test.tsx:498`: touch `matchMedia` stub is `query !== "(hover: hover)"`, so every other query matches. Use `matches: false` like `room-message-row.test.tsx`.
- Minor: `apps/web/src/components/chat/composer-wysiwyg-editor.tsx:133`: `modifierEnterSubmits` JSDoc still only mentions Ctrl/Cmd+Enter; it also forces desktop Enter on touch (inline edit).
- Skipped / disagreed: iPad + Smart Keyboard (`hover: none` so Enter never sends) — spec named iPhone / `hover: none`; Send still works. Not a miss.

### Assessment
Ready to merge: Yes
Room composer Return on `hover: none` inserts a newline and Send still sends; desktop Enter still sends; inline edit still commits with `enterkeyhint="send"`. SOK-815 first-tap-send is untouched.

mrosberghaus:
Addressed all three in the follow-up commit:

- Renamed the `innerWidth` 390 test to "submits on plain Enter in a narrow window that can still hover". The assertion stays: width no longer drives the policy, and a narrow hover-capable window still sends. It passes because happy-dom reports `(hover: hover)` as matching, not because of `maxTouchPoints`; the editor only checks the hover media query.
- Touch stub now returns `matches: false` for every query, same as `room-message-row.test.tsx`.
- `modifierEnterSubmits` JSDoc now states it also keeps plain Enter = save on touch.

iPad + hardware keyboard: agreed, out of scope. Send still works there.

PR #4434, opened 2026-09-11
fix(core): let a failed mention-name read cost only the preview

schaier-io:
`fanOutChatNotifications` read the mentioned members above the per-recipient
loop and outside its try/catch, so a throw there skipped every
`createNotification` and cost all of them both the notification center row and
the OS banner. That is how a mention key the soko bot column could not parse
took chat notifications down (#4411).

Read them the way the direct-room name below is already read, so a failure
reports and costs the reader the preview rather than costing everyone the
notification.

What the preview costs depends on the body. A mention nothing can name is
taken out of the sentence, and what is left reads as a whole sentence saying
something else: `ping @Bob not @Carl` becomes `ping not`. So a body carrying
one gets no preview at all, which leaves the reader the line that names the
author and the room, as a deleted message already does. A body whose mentions
all carry a slug says who without the lookup, and is shown as it is.

That is asked of the names the call actually holds rather than of whether the
read worked, because the read is not the only way to arrive without a name. A
member who has left the room, and a member whose display name is empty, both
come back missing from a read that worked, and the reader gets the same
inverted sentence with nothing to report.

The tokens are read from the message the way the preview reads it, which they
were not. `readChatMentionKeys` scanned the body as written while the preview
scanned it cleaned, so `_@<uuid>_` was no mention to the lookup and a mention
to the preview: the lookup never asked for that member's name and the preview
dropped the token for want of one. Both now read the one cleaned body, so the
names that are looked up are exactly the tokens that get rendered.

The trade is that a bare id nothing can name costs the whole preview wherever
it is written: inside a code span, inside a quote, or in the path of a pasted
link, where the builder used to drop the token and hand back the rest of the
sentence. A message is named or it is not named; there is no reading of a body
that tells a pasted id from an id the sender meant as a mention. The body is
read whole for this, past the length a preview is cut to, so an id pasted at
the end of a long message costs a line it would never have appeared on.

A reader whose row is left without a preview keeps it that way. An edit that
takes the mention out does not bring one back, because a rewrite never puts
text on a row carrying none.

The sweep that rewrites a preview after an edit does not ask this question. It
cannot yet: declining to write would leave an edit made to redact unredacted,
and writing nothing would empty a row that can never be filled again, by that
same rule. So a member who has left the room still loses their name there.
Fixing it means letting a rewrite refill a row it emptied, which is a rule of
its own and not this change. What the sweep does gain is the mentions it was
never looking up: an italicised `_@<uuid>_` is a mention to the key reader now,
so a rewrite names that member instead of dropping the token.

A row whose stored JSON will not read is reported too. Both columns are
written by `JSON.stringify` on every path that writes them, so a row that will
not parse is one something else wrote, or one damaged after it was written.
Every reader of it carries on with an answer that is safe rather than right: a
count that undercounts, a row left holding a deleted message's text. The row
stays that way for as long as it exists, and nothing else would ever say so.

What is reported says why and nothing of what the row held. A parse error
carries the opening characters of what it could not read, and those characters
are a message preview, an author's name, a room's name.

A column that parses into something that is not an object is the same damage
said more quietly, and is said the same way. `null` and a bare number both
parse, and neither carries a field any reader here asks for, so each leaves
its reader on the fallback a throw leaves them on. A column that is an object
without the field asked for is not damage: the first row of a room carries no
count, and a row written before a field existed carries none either.

One reader answers all three questions now, so these columns can only be read
the one way, and a fourth reader of them has one place to reach for.

A row is named once and then not again. It is not healed by being read, and
the paths that read it run again on every message, so said per read one
damaged row is rows times messages times recipients of the same row saying the
same thing. The memory of what has been said is per process, which is what
keeps it findable: an issue that fell off the page a month ago comes back
after a deploy rather than being suppressed for good by an instance that no
longer exists. Five hundred rows in, a process stops naming them; by then the
first five hundred have said everything the next one would. What is remembered
is the column and not just the row, so a row carrying two damaged columns is
named for both rather than having the second ride silently on the first.

The key that threw is filtered now; the shape that let one read speak for
every recipient was not.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat notification previews when mentions are incomplete, unnamed, or contain unusual formatting.
  - Preserved message slugs in generated previews.
  - Continued sending notifications when mention lookups or individual preview operations fail.
  - Added safer handling for malformed notification data, preventing invalid content from appearing in previews.
  - Improved handling of formatted messages, code spans, block quotes, and room-wide mentions.

- **Reliability**
  - Added sanitized error reporting and duplicate suppression for damaged notification records.


schaier-io:
@coderabbitai review

schaier-io:
@coderabbitai full review

schaier-io:
@coderabbitai review

PR #4435, opened 2026-09-11
fix(core): report a failed chat notification emit instead of losing it

schaier-io:
Every caller schedules these emits through waitUntil, after the reader has
been answered, so a rejection has no caller left to reach and arrives as an
unhandled one that names nothing. The fan-out does report, but only from
inside its per-recipient loop, which a throw above the loop never enters.

That is why #4411 ran for over half an hour with nothing in Sentry: the
notifications were not arriving and the failure left no trace anywhere.

Report at the emit seam, which every caller passes through, rather than at the
six call sites. The room-message effects already caught its own rejections and
wrote them to a console nobody reads until they already suspect the failure;
it reports them now too. Its roster read sat above that settled pair and so
outside it, which is the same shape again: it reports and stops.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of chat notifications when recipient, roster, or reader data cannot be loaded.
  * Prevented failed notification and reader-update operations from disrupting message processing.
  * Improved reliability when publishing chat room updates, allowing processing to continue when individual updates fail.
  * Added error reporting with relevant chat and room context for faster issue diagnosis.
* **Tests**
  * Added regression coverage for direct-message, mention, room-message, and chat-room update failures.


schaier-io:
@coderabbitai review

schaier-io:
@coderabbitai full review

PR #4436, opened 2026-09-11
fix(chat): name a member mentioned without the uuid hyphens

schaier-io:
A mention key keeps the spelling it was written in, and the lookup map is
keyed by the id the database stores, which is always hyphenated. So
`@019fc7e4e4bd7005900c66e44d33f5e4` named the same member as the hyphenated
spelling and found nobody, and the preview fell back to the slug or dropped
the mention.

Read the key as written first, then its canonical spelling. The order matters
when both spellings name somebody: a legacy 32-character auth id is hyphenless
too, and an all-hex one rewrites into a uuid that may belong to a different
member. The composer writes an auth id verbatim, so the key as written is the
one the author meant.

The soko bot filter takes the hyphenless form with it. Postgres parses a uuid
written either way, so a key the column would have matched must not be dropped
before it gets there.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed chat mentions using hyphenless UUID-based keys so they are correctly recognized and included in bot membership lookups.
  - Improved mention display-name resolution for hyphenless and uppercase UUID formats.
  - Preserved exact written-key matching when both hyphenated and hyphenless forms are available.
  - Maintained existing support for legacy hexadecimal authentication IDs.


schaier-io:
@coderabbitai review

PR #4437, opened 2026-09-11
feat(web): restore a push subscription that died on its own

schaier-io:
A Web Push subscription dies without anybody deciding it should: the push
service expires an endpoint, a browser drops it with its site data. Nothing
brought it back. Every activation sits behind a click on the account page, so
a reader whose browser went quiet had to find that page and press the cell
again, and nothing told them they had to. SOK-876 lists this and shipped
without it.

Repair it where a signed-in reader arrives, under three conditions read from
the browser alone, so an app open costs no request: Ably still holds a
registration for this browser, the notification permission is already granted,
and no subscription is live.

The registration is what makes this a repair rather than an opt-in. A browser
that never turned push on does not have one, and neither does one that turned
it off, so this reaches neither. The granted permission is why nothing is ever
asked: prompting belongs on a page the reader went to.

That reading of the registration only holds if turning push off always clears
it, and it did not. `deactivate()` clears the token only when it lands, so a
disable that failed halfway, or a sign-out the release cap cut short, left the
token beside a browser with no subscription: the exact shape this repairs. On
a shared browser that meant subscribing whoever signed in next. Signing out now
forgets the registration locally whatever Ably answered, and turning push off
forgets it once the browser subscription is gone. That cannot fail on the
network. A browser whose subscription survived a failed disable usually keeps
its token, because it is still being delivered to and a later sign-out needs
the token to deregister it. Not when the note cannot be written at all: a full
storage refuses the note and still answers every read, so the token would sit
there with nothing to say it is on its way out, and the token goes instead.
Ably may then be left holding a device whose endpoint is dead, which its own
delivery prunes.

Changing this browser's push state also happens one thing at a time, in the
order the reader asked. Turning push on deactivates the shared Ably client
halfway through, so two runs beside each other leave the browser subscribed
with no device registered: push reads as on and nothing is delivered. Turning
it off beside a run that is turning it on is worse, because the run finishes
afterwards and a reader who signed out is subscribed again. The repair makes
that overlap ordinary, because it starts an activation nobody pressed for. A
second reader waits rather than joins, since signing in again never reloads
the page, and a reader who turns push back on after asking for it off gets a
run of their own rather than the one their teardown is about to undo.

Three things guard the repair against undoing a reader's own decision. The
queue says whether anything is mid-flight, so a sign-out that reads an
untouched browser halfway through an activation queues its teardown anyway
rather than leaving the run to resubscribe the browser behind the reader. The
repair reads a count of teardowns before it reads anything else and again
after the chunk it loads, because until it queues something there is nothing
for a sign-out to order itself against, and every read it makes on the way
there is a window one can land in. And the Ably half of a teardown gets a
deadline, so a hang there cannot keep the token the repair reads as an
invitation; the reader is told that teardown failed, because an abandoned one
has not happened.

A teardown also leaves a note in storage while it runs, because its steps can
be cut short by a reload or a closed tab, and what they leave behind then is a
token beside a browser with no subscription: the shape the repair reads as a
subscription that died by itself. The note outlives the page that wrote it, so
the repair can tell an interrupted disable from an expired endpoint. Only a
deliberate activation takes it back, because an interrupted disable should go
on reading as off.

Whether the browser still holds a subscription is read from the browser rather
than inferred from whether the unsubscribe threw. A throw says the step failed,
and the common failure is a worker lookup that found nothing to unsubscribe: a
browser with no subscription would have kept the token, which is the shape the
repair turns back on.

Account deletion drops the token too (#4462). That path leaves the browser
unsubscribed and deliberately kept the token, which was safe while only a
reader doing something deliberate read it. The repair reads it as well, and
reads a token beside a browser with no subscription as one to bring back, so
keeping it would hand the browser to whoever signs in next, on a device record
belonging to an account that no longer exists. Said before the reads rather
than after them: a browser whose subscription had already died is not a
browser with nothing to do, it is that shape exactly.

What that path does not do is wait its turn. Every other change to this
browser's push state is queued; a deletion queued behind a hung
`client.push.activate()` would not unsubscribe at all, for the life of the
page, and signing in again never reloads it.

Running straight through means it can overtake an activation, and the ending
session does not stop one. `client.push.activate()` registers on the Ably
token the client already holds, minted with no ttl and so honoured for Ably's
default hour, and `ably@2.28.0` writes the identity token back from its own
state machine (`build/push.js:840`) rather than from the promise the caller
awaited. So the run lands after the deletion unsubscribed and forgot the
token, and leaves the browser subscribed, a device on the deleted reader's
channel, and the token back in storage. The next reader to sign in here reads
the Push cell as on, over a device that was never theirs.

So an activation asks, after every step that leaves the machine, whether a
teardown landed while it was away, and undoes itself when one did: the
subscription it just made and the token the SDK wrote back, in that order.
The count it reads is the one the deletion now takes. The Ably device record
stays, because deregistering it needs the session the deletion ended, and Ably
prunes a device whose endpoint stops answering.

The same question covers the sign-out, which is queued and did not need it: a
teardown waiting behind an activation unsubscribes a browser the activation
has already unsubscribed itself, which is a second call that finds nothing.

---

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery for missing push-notification subscriptions when supported and permitted.
  * Improved push-notification activation and deactivation reliability during concurrent operations, failures, and timeouts.
  * Push subscriptions now respond correctly when the signed-in account changes.

* **Bug Fixes**
  * Improved cleanup of stale push registrations and browser subscriptions.
  * Prevented interrupted sign-out or account deletion workflows from restoring push notifications.

* **Tests**
  * Expanded coverage for recovery, cleanup, failures, account changes, and concurrent operations.


schaier-io:
@coderabbitai review

schaier-io:
@coderabbitai full review

schaier-io:
@coderabbitai review

PR #4442, opened 2026-09-11
docs(web): forbid raw palette colors, hex literals, and opacity modifiers

schaier-io:
Components in apps/web painted the same meanings twice: 41 raw Tailwind
palette classes ran beside the semantic ramps, ~30 hex literals duplicated
brand and chart tokens, and 781 opacity modifiers invented tints at the call
site. Two adjacent status badges rendered "succeeded" in emerald and green
because each reached for the palette independently.

Adds .cursor/rules/color-tokens.mdc and links it from both AGENTS.md files.
Components name a role; the alpha-baked ramp steps (-tertiary / -quaternary
/ -quinary) replace opacity modifiers, and a new token is the answer when no
step fits.

Documentation only. The guard test the rule names lands with the cleanup
layers above this one.

---


schaier-io:
Folded into a smaller stack. Its commits now sit in #4443. The seventeen-PR stack is now four: #4443 token foundation, #4448 status roles, #4451 the opacity campaign, #4457 the token guard. No commit was dropped.

PR #4449, opened 2026-09-11
chore(web): delete the five status hues with no readers

schaier-io:
The badges moved onto the status role palette in the previous change, so
status-ready, status-running, status-awaiting, status-topped-up and
status-auth have no readers left in either theme block or the theme
bridge.

---


schaier-io:
Folded into a smaller stack. Its commits now sit in #4448. The seventeen-PR stack is now four: #4443 token foundation, #4448 status roles, #4451 the opacity campaign, #4457 the token guard. No commit was dropped.

PR #4450, opened 2026-09-11
fix(web): move the button focus ring outside the fill

schaier-io:
The focus core was invisible on three of the six variants, not merely
weak. --ring and --primary are the same purple, so focus-visible:
border-ring painted a purple border on a purple fill and measured
1.00:1 in both themes. On the destructive fill it measured 1.39. Only
outline looked right, because its fill is the page ground.

Put the mark outside the box instead. A 2px ring with a 2px
background-coloured offset measures against the page, so every variant
now reads the same: 7.04 in light and 5.07 in dark.

Point the destructive fill at the solid red and give it the near-white
label token. text-destructive-foreground had no token behind it at all,
so Tailwind emitted nothing and the label fell back to --foreground,
which is dark in light mode. The hover step had to darken with the
fill: the old dark hover carried white at 4.08, under the 4.5 floor,
and the new one carries it at 6.30.

---


schaier-io:
Folded into a smaller stack. Its commit now sits in #4443, moved down beside the control-contrast work it belongs with. The seventeen-PR stack is now four: #4443 token foundation, #4448 status roles, #4451 the opacity campaign, #4457 the token guard. No commit was dropped.

PR #4452, opened 2026-09-11
refactor(web): move the neutral fill tints onto solid steps

schaier-io:
275 call sites faded a neutral token by hand. bg-muted alone appeared at
20, 30, 40, 45, 50, 60, 70 and 80 percent, which is eight different
greys for what the screens treat as three surfaces: a faint wash, a
resting surface and a stronger one.

Each site now names the step whose role it was reaching for. The steps
are the base composited once against the page ground and frozen, so the
same class renders the same shade on a card as on the page. A modifier
could not do that, because it composites against whatever sits behind
the element.

The mapping is measured, not guessed. Every rewrite lands within OKLab
dE 3.0 of what the modifier rendered, and most within 1.0, which is
under the 2.0 just-noticeable difference. The 20 and 30 percent steps
collapse into the same surface as 40 through 60, because measured
against the tokens that actually generate a utility they sit 0.6 apart
in light and 1.9 in dark, both under that threshold.

The loading skeleton pins itself to the list view by class name, so its
selector moves with the fill class it mirrors.

Two smaller fills join them. The notification delete control landed on
main while this stack was in flight with a hand-faded foreground as its
focus fill, and bg-quinary is the resting neutral fill this sweep maps
that weight to. The dark select trigger hovered onto --border, the
divider role, which sits 1.12:1 from the --quinary rest state, so
hovering changed nothing a reader could see; --quaternary is the named
hover step and measures 1.54:1 against the same rest state.

The notification circle keeps --quinary, and the comment above it now
records the trade instead of telling the next reader to restore the
alpha the guard rejects: the disc edge is faint on a hovered row, and
the glyph inside it is what holds contrast.

---


schaier-io:

PR #4454, opened 2026-09-11
refactor(web): stop fading text by hand

schaier-io:
130 call sites dimmed a text token with a modifier.
text-muted-foreground alone ran at 40, 45, 50, 60, 70 and 80 percent.

These are not design steps. --muted-foreground already sits at the
lowest lightness that still clears 4.5 against the page, so every
fraction of it fails, and the faded value also changes with whatever
sits behind the element. The nearest solid match for
text-muted-foreground/50 is --tertiary, a border token, and borrowing a
border token as a text colour is exactly what the token rules forbid.

Each site now names the text role it meant. A genuinely quieter tier is
a token, not arithmetic at the call site.

The faint watermark in the upload dropzone is the one real step, so it
takes the primary tertiary tint instead.

---


schaier-io:

PR #4456, opened 2026-09-11
refactor(web): clear the last raw palette colours from the product UI

schaier-io:
Nineteen sites still named black or white directly. They survived the
earlier passes because no token described what they were doing, and a
literal is wrong in one of the two themes by construction.

Three groups, three answers. A check mark inside a filled primary circle
takes the primary label token. A label riding on a scrim takes the media
label token added with the scrims. The image viewer canvas and the CTA
that sits on a light photograph take --media-ground, which is
deliberately theme-independent: the ground is the photograph, not the
page, so following the theme would be the bug.

The destructive badge also moves onto the solid fill and its near-white
label, matching the button. Its dark override goes away, because the
solid token already carries a value for each theme.

---


schaier-io:

PR #4458, opened 2026-09-11
fix(web): give the two payment statuses their own labels

schaier-io:
STARTED and PAYMENT_PENDING both returned the paymentPending key, so
both printed "Hiring". Two badges read the same word with no way to tell
the states apart, and the history status filter offered two separate
options that also both read "Hiring". STARTED now uses the started key,
which already exists in all three locales.

Reword the payment stage in English and German. "Hiring" named the
outcome the user wants rather than what the job is waiting on, which is
the payment settling. Spanish already read "Pago pendiente" and does not
change, so this aligns the other two locales with it. PAYMENT_FAILED
moves with it, because it is the same stage failing.

The two statuses keep one role and one glyph. They are a single stage
seen twice, and the labels are what tell them apart.

---


mrosberghaus:
No! Don't merge this. That's on purpose because the user doesn't care about it. For him it's hiring.

schaier-io:
Folded into a smaller stack. Its three commits moved down nine layers into #4448, so the label split now sits in the same PR as the status roles whose comments pointed at it. The seventeen-PR stack is now four: #4443 token foundation, #4448 status roles, #4451 the opacity campaign, #4457 the token guard. No commit was dropped.

PR #4464, opened 2026-09-11
fix(core): report a row the banner count could not read

schaier-io:
`arrivalsOn` is the second reader of `messageParams` and was the silent one.
It is a byte-for-byte twin of the fan-out's `countOn`: both parse the column,
both take `count` off it, and both answer one when they cannot. The fan-out
now says when that happens and this did not, so the same damaged row reported
itself on the path that writes and said nothing on the path that counts.

What it costs is a banner that undercounts. A row standing for nine messages
reads as one, and the reader is told four arrived when twelve did. Nothing
about that is visible from outside: a banner saying four looks exactly like a
banner saying four, whatever it was counted from.

Read through `readNotificationRowJson`, so a column that will not parse and a
column that parses into something that is not an object are each said with the
row's id and nothing of what the row held. Said once per row, which this path
needs more than the one below it does: the count re-reads every unread row in
the room on every publish, and a damaged row is not healed by being read, so
per read it would be rows times messages times recipients of one row saying
one thing. The count itself is
unchanged: a row nobody can read the params of still stands for the message
that wrote it, and one is still the answer that undercounts.

The hidden row's own metadata is read the same way, two lines above. Parsed
in the loop, a column that would not read threw out of it into the catch
below, which names no row and fires on every publish of every message. What it
answers is unchanged: a hidden row that cannot say whether it was banner-only
takes the whole count down with it, parsed either way.

The row's id has to be selected for both, which the count read was not asking
for. It reads two columns of the row already; this adds the one that names it.

---

Stacked on #4437.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved notification processing when row data contains unreadable or malformed values.
  - Prevented notifications with invalid required data from being published.
  - Preserved fallback behavior when notification group counts are uncertain.
  - Enhanced error reporting with row-specific, sanitized diagnostics while preventing duplicate reports.
- **Tests**
  - Expanded coverage for malformed message parameters, hidden-row metadata, and unread notification queries.


schaier-io:
@coderabbitai review

schaier-io:
@coderabbitai full review

schaier-io:
@coderabbitai review

PR #4465, opened 2026-09-11
chore(core): say that core sends every error

schaier-io:
`sampleRate` was unset, which means one: every error event is sent. The two
rates beside it are 0.005, so a reader scanning the three together has every
reason to read the missing one as sampled as hard as its neighbours, and to
answer "why is this not in Sentry" with the sampler instead of with the code
that should have reported it.

Nothing changes at runtime. What changes is that the answer is in the file
rather than in the SDK's defaults, and a test holds it there, so raising or
lowering it becomes a decision somebody makes rather than a default somebody
inherits.

It matters more now than it did. The reporting added below this in the stack
fires once per damaged row and once per failed emit, and each one is the only
thing that will ever say so. A rate applied to those is not a trimmed sample,
it is the defect going unreported.

---

Stacked on #4464.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error monitoring reliability by ensuring all error events are captured, while performance traces and profiles continue using lower sampling rates.

- **Tests**
  - Added coverage verifying the updated error sampling behavior.


schaier-io:
@coderabbitai review

PR #4467, opened 2026-09-11
fix(core): move avatar top-up off GET onto a POST

schaier-io:
Closes SOK-1021.

## User-facing impact

`GET /v1/soko-bots/avatars?topUp=true` generated mascot images before it answered. That writes rows and bills FAL, and a GET reaches none of the protections a mutation needs: it is cacheable by intermediaries, and a cross-site top-level navigation carries the Core session cookie because Better Auth issues it `SameSite=Lax`. An attacker page could make a signed-in user's browser refill the shared avatar pool and spend the platform FAL budget.

The GET is now read-only and the `topUp` query parameter is gone. Generation moved to `POST /v1/soko-bots/avatars/top-up`.

The only caller that asked for a top-up is the avatar picker, through a Next Server Action. Those are POST and carry Next's own origin check, so the picker still works and previews still fill an empty grid rather than showing nothing. That preview behaviour is why `topUp` existed: Vercel runs crons on production deployments only.

## Scope of the fix, stated plainly

This closes the cross-site half of the finding. It does **not** stop an authenticated user from spending FAL budget in a loop: `POST /avatars/top-up` has no rate limit, and because the availability count applies the same `excludeIds` filter, harvested ids force generation even when the pool is full. That was equally true before this change. VERIFIED: `grep -rn "rateLimit" apps/core/src --exclude="*.test.ts"` matches only Better Auth's own limiter in `lib/auth.ts`. Tracked as SOK-1050.

## Review findings fixed in this branch

Two adversarial passes ran. Both found real defects.

1. The new POST route was missing `required: true` on its body. `@hono/zod-openapi` skips body validation **entirely** when a request carries no JSON content-type unless that flag is set, so `take` reached the service as `undefined`, and Prisma reads `take: undefined` as "no limit". A body-less POST would have returned the whole unclaimed pool instead of one capped page. The repo already pins this same trap on the admin refund route.
2. The Core `excludeIds` cap was 50 while the web action that feeds it allows 60, so the picker would have started erroring on the ninth request for a fresh set.
3. The second pass demonstrated my first test file passing while the vulnerability was back in the GET. It asserted against the route file's source text, and the slice it inspected stopped at the first `});`, which falls inside the `listAvailableAvatars(take, { excludeIds })` call. Replaced with tests that drive the mounted router.

## Correction to the ticket

SOK-1021 predicted a body-less POST would 422. It returns 200. With `required: true` the validator runs, parses `{}`, and applies the schema defaults, so `take` is 6 rather than `undefined`. The fix is effective; the expected status in the ticket was wrong. The test asserts what actually matters: the service is called with 6, not with `undefined`.

## Verification

- `pnpm --filter @sokosumi/core test`: `Test Files 500 passed | 3 skipped (503)`, `Tests 5278 passed | 11 skipped (5289)`
- `pnpm --filter web test src/lib/actions/soko-bot`: `Tests 5 passed (5)`
- `pnpm typecheck`: `Tasks: 19 successful, 19 total`
- `pnpm exec biome check .`: exit 0
- Mutation check, `required: true` removed with the tests kept: `Tests 1 failed | 3 passed (4)`
- Mutation check, both source files reverted with the tests kept: `Tests 3 failed | 14 passed (17)`

## Out of scope, found while here

- The picker accumulates seen ids without ever resetting them, so after 11 clicks of "show others" it exceeds the 60 cap and shows an error toast. Pre-existing, and this change improves it: the old query-string transport capped `exclude` at 1000 characters, about 26 ids, so it broke sooner.
- `apps/core/src/routes/v1/soko-bots/index.ts` is 1781 lines, over the 750-line ceiling. It was already over before this change.
- The regenerated client sweeps in two unrelated operations, because the checked-in client on `main` was behind Core. The OpenAPI snapshot is gitignored, so it could not be narrowed without hand-editing generated files.

## Summary by CodeRabbit

* **New Features**
  * Added explicit avatar top-up support to generate missing Soko Bot avatars.
  * Avatar requests can specify the number of results and IDs to exclude.
  * Avatar browsing now separates read-only listing from pool replenishment.

* **Bug Fixes**
  * Prevented regular avatar listing from unexpectedly generating new avatars.
  * Added validation for invalid quantities and excessive exclusion lists.


schaier-io:
@coderabbitai review

PR #4440, opened 2026-09-11
feat(apple): attach existing Drive files in chat

mrosberghaus:
Adds From Drive to the native chat attachment menu. Browse folders, search the current folder, and select an existing file into the same removable, persisted attachment chips used by local uploads. Rooms and reply threads share this flow, scoped to the active personal or organization workspace.

Uses the existing Core Drive endpoint. Prerequisite #4432 is merged; the regenerated client accepts the real pagination response without total. Networking/state remain in portable packages; presentation uses a native SwiftUI sheet and list. No dependencies or web changes.

Web references:
- [Drive picker](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/drive/drive-file-picker.tsx)
- [Drive listing](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/utils/drive-file-list.client.ts)
- [Room composer](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/room-composer.tsx)

Scope: attachment browse/search/select, duplicate prevention, draft persistence, loading/errors/retry. Standalone Drive management is excluded.

Validation:
- 313 SokosumiChat tests and 55 SokosumiWorkspace tests passed, including uncounted file/folder pagination, workspace query parameters, repeated cursor rejection, stale request handling and saved attachments.
- Xcode app tests and signed macOS build passed; Workspace cross-build for iOS 17 passed.
- Pinned SwiftFormat/SwiftLint and commit hook (repository check/typecheck) passed.
- Live checks passed: menu/sheet opening, listing, full-row folder clicks, nested browsing, breadcrumbs, prefix search and empty/no-results feedback. Initial live check revealed #4432; corrected client loads successfully. Sampled folders had no selectable files, so live selection/send remains pending. Selection persistence/deduplication is covered by tests. Composer unchanged; no message sent. Apple CI passed at bbe7cd2d6; full-row click refinement passed signed build and pinned lint, new CI pending.

Test: paperclip → From Drive → open folder/search → select file → check chip → remove/reopen. Repeat in a thread and personal workspace. Confirm Cancel preserves the draft and duplicate selection adds no second chip. Deliberately send a test attachment only in a suitable test conversation.

PR #4469, opened 2026-09-11
feat(apple): preview PDF attachments from compact file tiles

mrosberghaus:
Document attachments now appear as compact 64-point tiles with native macOS file-type artwork. Attachment-boundary newlines are removed to avoid excess top spacing, and tiles have consistent bottom padding. Clicking a PDF opens a native PDFKit viewer with the filename, Open in Browser, and Save actions in its toolbar. This is the user-approved viewing follow-up to #4440.

Web references:
- [File thumbnail tiles](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/ui/file-chip-mini-preview.tsx)
- [Document viewer](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/ui/document-viewer.tsx)

Scope: document tiles, native PDF preview, unavailable/error fallback, existing download/export reuse. Other document types retain Open/Save through the fallback. Office/text preview, audio/video changes, and external dependencies are out of scope. No web or Core changes. PDFKit adapters are isolated for macOS and iOS; shared parsing trims only boundary newlines around attachments, preserving internal paragraphs and inline spaces.

Validation: all 316 Chat tests, signed macOS build, full Xcode app tests including a two-page native PDF hosting/resize test, pinned SwiftFormat/SwiftLint, and repository commit checks passed. Live opening of the user's existing PDF displayed its pages and toolbar correctly. Live PDF scrolling and the native Save dialog were also verified; export was canceled without writing. Actual saved-file verification, light-mode acceptance, and visual acceptance of the latest tile spacing remain pending. PARITY.md records the remaining scope.

How to test:
1. Open a conversation containing a PDF and click its compact tile.
2. Scroll the PDF, resize the window, select/copy PDF text, and close with Escape.
3. Reopen it and test Save and Open in Browser from the toolbar.
4. Open an unsupported document and confirm the fallback still offers Open/Save.

Apple CI passed at f9233f11e (run 34636804041): Xcode build, app tests, all five Swift package suites, and Swift lint/format.


mrosberghaus:
## Muse review

### Fixed

### Remaining
- Minor: `apps/apple/Sokosumi/Chat/Rendering/DocumentAttachmentPreview.swift`: PDF detection is extension-only (URL + filename). An extensionless PDF (e.g. a `/deliverables/` id URL with a bare label) lands in the unavailable fallback. Web behaves the same (viewer kind from url/fileName/mediaType, no content sniffing), so this is parity — a future option is `%PDF-` magic-byte sniffing or plumbing MIME.
- Minor: `apps/apple/Sokosumi/Chat/Rendering/AttachmentFileIcon.swift`: the iOS tile always shows the generic `doc.richtext` glyph for every file type; only macOS gets per-type artwork.
- Minor: `apps/apple/Sokosumi/Chat/Rendering/MessageAttachmentView.swift`: the 8pt tile bottom padding stacks with the 8pt segment VStack spacing (16pt below tiles followed by text); the PR already flags spacing acceptance as pending.
- Minor: `apps/apple/Sokosumi/Chat/Rendering/AttachmentFileIcon.swift`: `UniformTypeIdentifiers` is imported unconditionally but used only on macOS. No warning emitted; trivial.
- Skipped / disagreed: requiring magic-byte PDF sniffing as a fix — rejected because web parity does not do it and the PR scope excludes it; kept as a Minor note above.

### Assessment
The slice matches its stated scope and web parity with package tests (10/10) and lint (0 violations on the 4 changed Swift files) green in this checkout; full Xcode app tests and CI were not re-run here.

mrosberghaus:
Verified the four notes against the current code. No Critical/Important fix was pushed by the reviewer.

Fixed in 0ae4918:
- Removed per-tile bottom padding. The segment container now adds bottom padding only when its final visible segment is a file, so a file followed by text uses the existing 8pt gap rather than 16pt.
- Scoped UniformTypeIdentifiers to the macOS import block.

Deferred:
- Extensionless PDFs: confirmed the native preview needs a PDF extension. Web can additionally use supplied MIME metadata, but neither implementation sniffs bytes. With a bare label and no MIME metadata, the fallback is consistent. Content sniffing/MIME plumbing is not required for this PDF slice.
- iOS file artwork: the generic glyph plus extension remains the portable fallback. Per-type iOS artwork belongs with the future iOS UI; macOS currently uses native file-type artwork.

Validation: Xcode build, SwiftFormat, SwiftLint and repository commit checks pass. The first commit attempt exposed stale installed dependencies after the main merge; installing the existing frozen lockfile resolved the web typecheck failure. No dependency manifest or lockfile changes. Previous-head full Apple CI passed; new-head CI is pending.

PR #4471, opened 2026-09-11
feat(web): grow the chat room search icon into the header field

mrosberghaus:
## Summary

The chat room search field no longer sits permanently in the room header. On desktop it rests as an icon-sized ghost field and grows into the full field (search icon, placeholder, ⌘F / Ctrl+F hint) with a width transition when it gains focus, the hotkey fires, or a query is present. It shrinks back on blur once the query is empty; a second Escape on an empty field collapses it. Mobile is unchanged: icon trigger, field inside the popover.

- Collapsed state is the same element as the field, so the expand is a real CSS width transition, no remount
- Collapsed look matches the sibling ghost icon buttons (transparent border, accent hover)
- Placeholder and shortcut hint fade with the width
- Query survives a close and re-runs on reopen, as before

## User-facing impact

The room header is calmer: one icon instead of a 224–256px input with a keyboard hint. Click or ⌘F opens the same search as before.

## Verification

- `pnpm --filter web test src/app/(app)/chat/components/room-search-panel.test.tsx` — 30 passed
- `pnpm --filter web test src/app/(app)/chat/components/__tests__/rooms-client-room-header.test.tsx` — 14 passed
- `pnpm check`, `pnpm typecheck` — clean
- Not yet verified in a running browser


mrosberghaus:
## Grok review

### Fixed
- Important: emptying the query closed the popover, which unmounted the mobile field and left previous hits in state so the next keystroke could reopen on the old fetch (`a5aef8ba8` — `apps/web/src/app/(app)/chat/components/room-search-panel.tsx`)

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/room-search-panel.tsx`: width expand uses `transition-[width]` with no `prefers-reduced-motion` opt-out
- Minor: `apps/web/src/app/(app)/chat/components/room-search-panel.tsx`: after a failed search, `showLoading` and the error paragraph can both render during the 250ms debounce of the next keystroke
- Skipped / disagreed: none

### Assessment
Desktop icon-to-field expansion matches the spec; the two Important clear-query bugs are fixed and covered by tests.

mrosberghaus:
Re: Grok review

- **Fixed (mobile unmount / stale hits)**: verified the bug and kept `a5aef8ba8` as-is.
- **Reduced motion**: `motion-reduce:transition-none` on the width transition — `35194dc5e`.
- **Error and loading overlap during debounce**: a stale error now yields to the loading state while the next keystroke is pending; covered by a new test — `35194dc5e`.

Panel tests: 35 passed. Biome and typecheck clean.

PR #4472, opened 2026-09-11
fix(chat): close the pinned panel on jump and memoize the markdown parse

mrosberghaus:
## Summary

Tapping a pinned message took about 3 s before the transcript moved on the Vercel preview, and on a phone the wait was invisible.

- **Loading state on the tapped pinned row.** The panel stays up while the window loads, with a spinner on that row, and closes once the message is on screen. A jump that gave up leaves the panel open so the row is still there to tap again.
- **Markdown parse is memoized.** `Markdown` ran react-markdown inside its own render on every parent update, so every row on screen re-parsed on every room render. The preprocessing chain and the `ReactMarkdown` element are now built in `useMemo`, keyed on the source text, the highlight term, and the components map.
- **Room jump no longer re-renders the transcript twice.** The jump toggled a page-level hold flag on and off, and each toggle re-rendered every row, around 200 after the window merges. The flag exists for the thread panel, which a room jump never touches, so the jump now holds the scroller through the stick-to-bottom hook's refs alone and the transcript renders once, on the merge.
- **A closed emoji picker is now just a button.** Each row renders one. While closed it rebuilt the emoji catalog sections on every render, held three viewport listeners, and read the root font size through `getComputedStyle` from its `useSyncExternalStore` snapshot, which forces a style recalc per row per render. The body now lives in its own component inside `PopoverContent`, so Radix mounts it only while open.
- **Two overlay roots fewer per row.** Every row mounted a touch action sheet and a delete confirm dialog that almost never open. A closed Radix dialog still costs a root, a portal, and its contexts per row. Both mount on first open and stay mounted after, so exit animations still run. This also trims the initial room render.

## Where the time goes

Sentry, last 14 days, production:

| Segment | p50 | p95 | max |
| --- | --- | --- | --- |
| Core `GET /v1/chats/rooms/:id/messages` | 92 ms | 165 ms | 1.25 s |
| Web → Core call for that route | 118 ms | 308 ms | 8.7 s |
| Web server actions on `/chat/rooms/[roomId]` | 94 ms | 379 ms | 1.8 s |

The server is not the cost. The 3 s recordings that motivated this PR were taken from a network whose traffic enters Vercel at the Sydney edge (`x-vercel-id: syd1::fra1`) and is forwarded to the Frankfurt function. From there a trivial unauthenticated request to production has a time to first byte of 0.65 to 1.5 s, and every server action pays that. The jump makes one such round trip, so it cannot land in under about 1.5 s on that path no matter what the client does. The same recordings showed the pinned list itself, a tiny fetch, taking 2 s to load.

The client-side changes in this PR still cut work that scales with transcript size and are worth keeping, but they do not move the number on that network.

Not changed: the Core `around` request and its 100-row window, and the transcript is still not virtualized.

## Verification

- `pnpm --filter web test "src/app/(app)/chat" src/components/chat src/components/markdown` — 1307 passed, 149 files
- `pnpm check` and `pnpm typecheck` — clean (precommit hook)

New tests: `emoji-picker.test.tsx` proves a closed picker adds no resize listener and reads no computed style; `markdown.test.tsx` proves the parse runs once across a same-source rerender; `pinned-messages-panel.test.tsx` covers the busy row, close on landing, and stay-open on a failed jump; `room-message-jump.test.ts` covers the landed answer; `use-room-message-jumps.test.ts` asserts the room jump leaves the page-level hold flag alone.

## Manual check

Open a Channel with pinned messages, open the pinned panel, tap a pinned message older than the loaded page. The row shows a spinner, then the panel closes and the transcript is on the message, marked. Long-press a message on touch and confirm the action sheet still opens; delete a message and confirm the dialog still opens.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `use-room-message-jumps.test.ts` never asserts `handleJumpToMessage`’s boolean (merge → true, invalidate → false).
- Minor: `emoji-picker.tsx`: reopen during Presence exit can reuse the still-mounted panel (stale query, no refocus).
- Minor: `room-message-row.tsx`: Sheet/AlertDialog stay mounted after first open (intentional for exit animation).
- Minor: `room-notification-jump.ts:25` still types `jumpInRoom` as `Promise<void>` (`Promise<boolean>` is assignable).
- Skipped / disagreed: panel `onClose` not keyed to the active jump — `performRoomMessageJump` only returns true when `isNewestJump()` still holds (`highlight` is `isNewestJump() && highlightRoomTranscriptMessage`). A later tap bumps generation before the first jump can return true, so the first call cannot close the panel. The leftover window is a microtask after resolve; a click cannot land there.

### Assessment
Ready to merge: Yes
Close-on-land, markdown memo, room-jump hold via scroller refs, closed emoji picker, and lazy overlay mount match the spec; the one Important close-race does not survive the jump generation check.

mrosberghaus:
Re the Grok review, addressed in b10d43504:

- **Hook test boolean:** fixed. The invalidated jump asserts `false`, the merged jump asserts `true`.
- **Reopen during the popover exit animation:** fixed. Each opening now keys a fresh `EmojiPickerPanel`, so a reopen inside the exit window gets a reset query and a refocused search instead of the still-mounted panel.
- **Sheet and dialog staying mounted after first open:** intentional, kept. Unmounting on close would cut the exit animation.
- **`jumpInRoom` typed `Promise<void>` in `room-notification-jump.ts`:** kept. That utility awaits the jump and never reads the answer, and the hook wraps it in an `async` block that returns nothing. The type says exactly that.

PR #4474, opened 2026-09-11
perf(chat): memoize transcript rows with stable callbacks

mrosberghaus:
## Summary

Opening a busy room loads 100 to 200 transcript rows, and every room state change re-rendered all of them: a realtime message from another member, a reaction, an edit session opening, a refresh merge after a coworker stream settled. On a phone the transcript stuttered after each of those events even though one row changed.

A change that touches one message now re-renders that row only. The transcript looks and behaves as before.

- **`ChatMessageRow` is memoized** with the default shallow comparison. Its DOM, props contract, and callback signatures are unchanged; the existing row tests pass without edits.
- **Row callbacks from the room screen are referentially stable.** `RoomsClient` keeps its handlers as they are and exposes one stable wrapper per callback. Each wrapper reads the handler from the latest render through a ref, so no wrapper closes over stale room state. This is a "latest handlers" ref rather than the "latest values" ref the spec sketched; it reaches the same goal with no handler rewrite and no dependency-array audit.
- **Message identity survives a refresh merge.** `mergeRoomMessages` keeps the existing object when a re-sent message is structurally equal, so a refresh page after a stream settles re-renders only rows that actually changed.
- **`handleRetryOutbound`** was already a `useCallback` keyed on the room object; it now goes through the same wrapper so its guard is unchanged.

Per-row lookups already reached the row as primitives (`isPinned`, `showOutboundSentTick`, `isEditing`, `editDraft`), so that stays.

Closes [SOK-1054](https://linear.app/masumi/issue/SOK-1054/memoize-transcript-rows-with-stable-callbacks). Builds on #4472.

## Tests

- New `rooms-client-row-render-isolation.test.tsx`: the room-screen harness with a memoized row stub that counts renders per message. Realtime create, reaction toggle, edit start, and draft typing each re-render only the touched row. Two stale-closure guards: pin after a realtime pin calls unpin, and save after a realtime update sends the current draft.
- Row test: re-render with an identical prop set does not re-run the row body.
- Merge helper: unchanged incoming messages keep their object identity; changed ones get the new object.

## Verification

- `pnpm --filter web exec vitest run` — 5486 passed, 1 skipped
- `pnpm check` — clean

Not changed: virtualization, the thread panel's own per-row prop construction, and Core.

### DIFF
--- a/apps/web/src/app/(app)/chat/components/__tests__/rooms-client-row-render-isolation.test.tsx
+++ b/apps/web/src/app/(app)/chat/components/__tests__/rooms-client-row-render-isolation.test.tsx
@@ -25,8 +25,9 @@ import type {
  * prop with a new identity or value.
  */
 
-const { rowRenders } = vi.hoisted(() => ({
+const { rowRenders, fetchRoomMessagesMock } = vi.hoisted(() => ({
   rowRenders: new Map<string, number>(),
+  fetchRoomMessagesMock: vi.fn(),
 }));
 
 vi.mock("next/navigation", () => ({
@@ -139,8 +140,9 @@ vi.mock("@/app/chat/actions", () => ({
   unpinRoomMessageAction: vi.fn(),
 }));
 
+// Scheduled room recovery reads go over GET; the poll is the refresh merge.
 vi.mock("@/components/chat/fetch-room-messages", () => ({
-  fetchRoomMessages: vi.fn(async () => ({ messages: [], nextCursor: null })),
+  fetchRoomMessages: fetchRoomMessagesMock,
 }));
 
 vi.mock("@/components/chat/organization-chat-list.actions", () => ({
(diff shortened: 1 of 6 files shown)


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/rooms-client.tsx` — `openingDirectParticipantKey` is still passed to every row, so starting a DM re-renders the whole transcript (not one of the events this PR claimed to isolate).
- Minor: `apps/web/src/app/(app)/chat/utils/merge-room-messages.ts` — `isSameMessage` is `JSON.stringify`; a key-order mismatch (Ably vs Core page) re-renders that row. Comment already says that is extra render only, never wrong data.
- Minor: `mergeRoomMessages` always returns a new array even when every object is reused, so a no-op refresh still re-renders `RoomsClient` (memoized rows still skip).
- Minor: isolation harness does not cover refresh-merge at the room screen; that path is only locked in `merge-room-messages.test.ts`.
- Skipped / disagreed: field-by-field equality instead of stringify — YAGNI; the spec already accepted key-order extra renders. `handleRemoveOutbound` is already `useCallback([])`. Thread-panel per-row props are out of scope.

### Assessment
Ready to merge: Yes
`ChatMessageRow` is memoized, room-screen callbacks are stable latest-handler wrappers (stale pin/save guards included), and merge keeps object identity for unchanged rows — a one-message change re-renders that row only.

mrosberghaus:
Went through the four remaining items.

- **Refresh merge at the room screen.** Added in 31fec8555. Two cases drive the scheduled recovery read with fake timers: a page that re-sends the same messages renders no row, a page that changes one message renders that row only.
- **`openingDirectParticipantKey` re-rendering every row.** Not changed. Opening a direct navigates to that room, so the transcript is left after the two renders. Isolating it would need the row to decide which of its participants the key belongs to, which changes the row contract and is out of scope here.
- **`mergeRoomMessages` returning a new array on a no-op refresh.** Not changed. The goal of this PR is row work; the parent render on a no-op poll is cheap now that every row bails out.
- **`JSON.stringify` key order.** Agreed, already noted in the code comment. A mismatch costs one extra row render and never wrong data.

PR #4484, opened 2026-09-12
feat(chat): load a 30-row jump window and mark gaps with a boundary row

mrosberghaus:
Implements SOK-1057 (https://linear.app/masumi/issue/SOK-1057).

A jump into room history loaded 100 rows on top of the 100 on screen and merged them into one flat list, so the transcript read as continuous where it was not, and Load older at the top could never fill the hole. A search hit avoided the hole by swapping the timeline and freezing the room.

- A jump (pinned message, notification, search hit) loads a 30-row window and merges it as its own loaded range. The newest page stays on screen and the room stays live.
- A boundary row marks every stretch of missing history. Above the oldest range it reads "Load older messages" and replaces the old top button; between two ranges it reads "Messages are missing here". The whole row is the hit area.
- The row loads on tap and on scroll-into-view, shows an inline spinner, and keeps a failed load on the row with "Try again" (no auto-load after a failure).
- Scroll position is held while rows land above the viewport. Done by hand because Safari has no `overflow-anchor`.
- The room-level historical freeze flag is removed; all room jumps go through one merge path. The thread panel is unchanged.

## How

- New pure model `room-transcript-ranges.ts`: the flat message list plus the older edge of every contiguous range, with `mergeRoomHeadPage`, `mergeRoomJumpWindow`, `mergeRoomOlderPage`, and `buildRoomTranscriptRows` (the flat row list SOK-1058 virtualization will consume).
- New `TranscriptBoundaryRow` component and `transcript-scroll-anchor.ts` helper.
- `listRoomMessagesAction` accepts `limit`; Core is unchanged.
- Translations added to `en`, `de`, `es` (`App.Channels.Boundary.*`).

## Verification

- `pnpm --filter web test`: 711 files, 5522 tests passed
- `pnpm check`, `pnpm typecheck`: clean
- No live browser run


mrosberghaus:
## Grok review

### Fixed
- Important: gap auto-load restored against the cursor below the hole and could shove a just-landed jump target off screen (`624e1efb9` — `apps/web/src/app/(app)/chat/utils/transcript-scroll-anchor.ts`, `apps/web/src/app/(app)/chat/components/rooms-client.tsx`)
- Important: continuation chrome still grouped the first head row with the jump window across "Messages are missing here" (`624e1efb9` — `apps/web/src/app/(app)/chat/utils/room-transcript-ranges.ts`, `apps/web/src/app/(app)/chat/components/rooms-client.tsx`)
- Important: in-flight boundary loads were not bound to a room generation, so A→B→A could merge a stale page into a fresh head (`624e1efb9` — `apps/web/src/app/(app)/chat/components/rooms-client.tsx`)
- Important: ranges suite did not pin outbound shells after the head (`624e1efb9` — `apps/web/src/app/(app)/chat/utils/room-transcript-ranges.test.ts`)

### Remaining
- Minor: `room-transcript-ranges.ts`: adjacent `around` windows that share no id do not join (14+1+15), so jump-to-50 then jump-to-80 can flash a 1-id hole until auto-load closes it. Matches the written join rule; fights "never see a missing row where nothing is missing."
- Minor: `transcript-boundary-row.tsx`: `IntersectionObserver` uses the viewport, not the transcript scroller.
- Minor: `transcript-boundary-row.tsx`: failed alert sits outside the button, so a tap on "Couldn't load messages." does not retry.
- Minor: `room-search-jump.test.ts`: stale comment still says a parent window swaps the timeline and marks it historical.
- Skipped / disagreed: none of the Important findings. Other nits (constant location, identity already locked in the overlap test, one-frame stale transcript on prop-loaded room switch) left as-is.

### Assessment
Jump landing, gap chrome, and room-switch loads now match SOK-1057; leftover notes are observer root, retry hit area, and adjacent-window join flash.

mrosberghaus:
Reply to the Grok review remaining notes:

- **Failed alert outside the button**: fixed in 7dddea16e. The error text is now inside the same full-width button as the retry, so a tap on "Couldn't load messages." retries.
- **Stale comment in `room-search-jump.test.ts`**: fixed in the same commit.
- **`IntersectionObserver` uses the viewport, not the transcript scroller**: leaving as is. With `root: null` the observer clips the intersection rectangle through every scrollable ancestor (Intersection Observer spec, "compute the intersection"), so a boundary row scrolled out of the transcript scroller does not report as intersecting. Setting `root` to the scroller would change nothing here.
- **Adjacent `around` windows that share no id do not join**: leaving as is. Two windows that touch without overlapping give the client no signal that nothing is missing between them; the only way to learn that is a fetch. The gap row above the newer window is that fetch: it auto-loads once visible, the page overlaps the older window, and the ranges join. The row shows for one round trip in the exact case where it cannot be avoided without a fetch.

PR #4485, opened 2026-09-12
perf(chat): mount the row action pill and hover cards on first hover or focus

mrosberghaus:
## Summary

Implements [SOK-1056](https://linear.app/masumi/issue/SOK-1056/mount-row-action-bar-and-hover-cards-on-hover-only). Every transcript row mounted six action buttons, an emoji popover root, an overflow dropdown root, and two participant hover-card roots on first paint, although a reader only ever hovers or tabs into a few rows. A busy room renders 100 to 200 rows and a jump merges another window on top.

- **One row-level latch.** The row flips an `interacted` flag on the first pointer enter or the first focus inside the article and never resets it. Before the latch it renders no action pill and bare hover-card triggers; after it, the full chrome, kept for the life of the row. The right gutter padding stays, so mounting never shifts layout.
- **Tab order unchanged.** The participant hover card gains an `active` flag. Inactive, it renders the same trigger element with the same role, tab stop, and accessible name and no Radix root. Callers that do not pass the flag keep today's behaviour, so mention hover cards in message bodies are untouched.
- **Keyboard focus survives activation.** When focus is on a bare trigger at the moment the row activates, the mounted Radix trigger takes focus on attach and the card opens on focus as before. The decision is made in the bare trigger's ref cleanup, before the node leaves the DOM, because Chrome fires blur when a focused node is removed.
- **Screen readers keep the always-mounted "more actions" button**, which still opens the touch action sheet. The sheet and delete dialog were already lazy from #4472.

## Verification

- `pnpm --filter web test "src/app/(app)/chat" src/components/chat` — 145 files, 1313 passed
- `pnpm --filter @sokosumi/core test` — 5171 passed
- `pnpm check` and `pnpm typecheck` — clean (precommit hook)
- Full `pnpm test`: one unrelated admin Soko Bot actions test hit its timeout under load and passes on its own

New row tests cover the pill and hover cards before and after hover, keyboard focus retention on tab, the latch surviving a re-render with a fresh message object, continuation rows, tabbing onto the "more actions" button, and deleted, thinking, editing, and outbound rows still rendering no pill after hover. Existing pill tests gained one hover step each.

## Manual check

Open a busy Channel in Chrome. Tab into a row: focus should land on the avatar, the participant card should open, and Tab should continue through the body into the action pill. Hover a row with the mouse and confirm the pill fades in as before.


mrosberghaus:
## Grok review

### Fixed
- Important: first hover on a row's avatar or name left the participant card closed (`a6cc8c731` — `apps/web/src/app/(app)/chat/components/chat-participant-hover-card.tsx`). Swapping the bare trigger for the Radix trigger under the pointer skipped `pointerenter`; attach now replays `pointerover` so Radix opens as usual.

### Remaining
- Minor: `cloneElement` overwrites the child's `ref` / pointer handlers. Mention triggers currently have neither, so this is latent.
- Minor: on a continuation row, tabbing onto the always-mounted "more actions" control mounts the pill *behind* that stop; Shift+Tab reaches it. The sheet still opens from "more actions".
- Minor: the `interacted` latch never resets, so a long hover-scroll still accumulates chrome. That matches the PR's one-way latch.

### Assessment
First-paint skip of pill and hover-card roots matches the spec; keyboard focus restore was already covered, and first-hover on the trigger now opens the card too.

mrosberghaus:
Follow-up on the Grok review.

**Fixed item, a6cc8c731 (replay pointerover on attach).** Checked and kept. Only the bare trigger's own pointer enter arms the replay, so hovering the row body alone does not open the card, and no ancestor in the transcript listens for pointer enter, so the bubbled synthetic event has no other effect. One correction to the description: without the fix the card opened on the next pointer movement, because browsers dispatch a fresh pointerover for the node that appeared under a stationary pointer on the next move; it did not need a leave and re-enter. The fix still matters for a pointer that stops dead on the avatar.

**Continuation-row tab order.** Fixed in 0de115853. The always-mounted "more actions" button now precedes the pill in DOM order, so on a row with no other tab stop the Tab after it walks into the pill's first control instead of leaving the row. The button is screen-reader only, so nothing visible moves. The tab test asserts the next stop is the "add reaction" button.

**`cloneElement` overwriting a child's ref and pointer handlers.** Left as is. No caller passes either today, so composing them would be code for a need that does not exist yet.

**Latch never resets.** By design, per the spec's one-way latch; a row that was hovered once keeps its chrome for the life of the row.

PR #4497, opened 2026-09-12
feat(chat): virtualize the room transcript with react-virtuoso

mrosberghaus:
Closes [SOK-1058](https://linear.app/masumi/issue/SOK-1058/virtualize-the-room-transcript). Parent: SOK-1053.

## User-facing impact

The room transcript mounts only the rows near the viewport (about one screen plus a buffer each side) instead of every loaded message. Opening a room, jumping to an old message, and receiving a message cost the same on a room with thousands of messages as on one with fifty. Everything else looks and behaves as before: open on the newest message, stay pinned while new messages and late images arrive, own sends reveal the bubble, older history loads above without the reader's row moving, search, pinned, notification and quote jumps land and mark the row, boundary rows still auto-load.

## What changed

- `react-virtuoso` 4.18.13 (MIT base `Virtuoso` only), attached to the existing room scroller via `customScrollParent`, so the room shell and Instant skeleton chrome are untouched.
- New `TranscriptViewport` owns the live-edge pin, jump landings, the quote scroll, and the scroll anchor. `RoomsClient` reaches it through a ref with the same handle names the stick-to-bottom hook exposed, so the jump hook, composer, and existing tests change minimally.
- Pure decision logic lives in `transcript-viewport-model.ts`: row keys, first-item-index shift for prepends, growth follow rule.
- Jumps and highlights answer from the row model instead of a document query; the quote click goes through a handler instead of the DOM helper (deleted).
- Thread panel unchanged; it keeps `useStickToBottom` and the DOM highlight. The viewport is written list-agnostic so threads and pins can adopt it later.
- Adds the `react-virtuoso` skill app-scoped under `apps/web/.agents/skills` (only the base-component skill; the sibling skills document commercial packages).

## Deviations from the spec, found in the browser

- Virtuoso's own follow-output covers row-count changes only. Late growth (unfurls, images, measured rows replacing estimates) is handled by a growth rule that re-pins the live edge and, off the live edge and at rest, by hand-rolled scroll anchoring: the first fully visible row is recorded on scroll and put back whenever the list height changes. While scrolling, Virtuoso compensates itself and the hold stays out of the way (both at once showed as jumps).
- Virtuoso answers a `firstItemIndex` prepend with a deferred scroll by its own estimate of the new rows, on top of the anchor restore. The restore is issued again 100ms later so the retrying scroll-to-index has the last word.
- A row that leaves the list and comes back re-renders its images from zero height, then grows by the image a frame later. `useRememberedImageSize` keeps natural sizes per URL after the first load and hands them back as width/height attributes (unfurl images and markdown images), so a remount reserves the box. Unfurl images additionally reserve their `h-48` cap until the first load, since wide link previews land there.
- Overscan is 2400px above and 600px below, so rows measure while the reader is still reading (where the hold corrects before paint) rather than while scrolling toward them. Rows that have never been measured can still shift on a fast first scroll through old history; that is Virtuoso measuring dynamic heights, and the row moves at most once.
- Collapsed `mt-2` article margins escaped the measured item box (list 48px taller than Virtuoso thought, live edge clipped). Both row wrappers are `flow-root`, which keeps the margins inside the box rather than rewriting row spacing to padding.
- `alignToBottom` is not used; the existing `min-h-full justify-end` wrapper bottom-aligns short transcripts (verified on a 4-message Direct).
- The scroller reaches the viewport as state, not a ref: on a same-commit mount the shell's ref attaches after the child's effects run.

## Verification

- `pnpm check`, `pnpm typecheck`, `pnpm --filter web test` (712 files pass).
- Browser, Everyone channel (thousands of messages) and Directs:
  - Open: 10 rows mounted for a 7.4k px list, view exactly on the live edge after unfurls load.
  - Scroll up through the boundary row: older page loads, anchored row offset 56 → 56, no late correction frames.
  - Search hit two days back: row marked and centered.
  - Pinned message 15 days back (outside loaded ranges): window loads, row marked and centered, gap row appears below the window on scroll.
  - Quote click: smooth scroll to the quoted message, no mark.
  - Short Direct (4 rows): rows sit at the bottom, no scroll.
  - Scroll up 4.8k px in wheel-like steps: second pass over measured rows shifts once at most (was 9 shifts); boundary loads hold the anchored row in 3 of 3 runs; first pass through never-measured rows still shifts on tall rows (10, was 14 with repeated ±192px flickers).
  - Not exercised: sending a message (write to the shared database); covered by the scroll-on-send tests and Virtuoso follow-output.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/thread-panel.tsx`: quote taps now call the room viewport. A quote of a thread reply (not in the room transcript) is a no-op; previously `scrollToRoomMessageElement` found the panel copy. Spec story 23 only requires a scroll when the room holds the message.
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.test.tsx`: pin, own-send, hold, restore-anchor, and quote-without-mark are not driven through real Virtuoso (happy-dom). Covered at the pure layer, RoomsClient stub spies, and the PR's browser steps — matches the spec's testing decision.
- Minor: `apps/web/src/app/(app)/chat/components/room-shell-layout.tsx`: comment still mentions pixel `minHeight` after that prop was removed from the room path.
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`: `TranscriptViewportAnchor.index` is stored and never read; restore looks the row up by id. `landOnMessage` rAF poll is not cancelled on unmount.
- Skipped / disagreed: firstItemIndex + restoreAnchor "double correction" — Virtuoso `scrollToIndex` takes the 0-based data index (not `firstItemIndex` + offset); the PR's browser check (offset 56 → 56, no late frames) matches that. Empty/failed still mounting an empty Virtuoso is harmless. Skill reference dump is the installed `react-virtuoso` skill, not product code.

### Assessment
The viewport owns pin, land, quote, and prepend anchoring; `react-virtuoso` is pinned at 4.18.13; CI is green; remaining notes are nits, not blockers.

mrosberghaus:
Addressed the Grok review notes in 63e439145:

- Thread-only quote: a reply quoting another reply is now scrolled in the open panel when the transcript does not hold it (`onJumpToQuotedMessage` falls back to the thread list's copy). Verified the old helper found the panel copy via a document-wide query.
- `TranscriptViewportAnchor.index` was never read; the type is gone and the handle uses `TranscriptScrollAnchor` directly.
- `landOnMessage` poll: unmount bumps the landing generation, which ends any poll still waiting for its row.
- Room shell comment no longer mentions the pixel `minHeight`.

No change on the test-coverage note; it matches the spec's fallback (pure layer, stub spies, browser steps). On `firstItemIndex`: the double correction the earlier review flagged was real but for a different reason than index math. Virtuoso answers a prepend with a deferred `scrollBy` by its estimate of the new rows, on top of the restore. Fixed in 534a14786 by re-issuing the restore after that window; the PR body describes it.

mrosberghaus:
Correction: the commits above are `4dd5af0b8` (scroll jumps) and `ce7f1eda8` (review notes) after rebasing onto the merge of main; the SHAs in the previous comment were pre-rebase.

mrosberghaus:
## Grok review

### Fixed
- Important: a second prepend (or a jump) inside the 100ms restore window could fire the first retry against a stale data index. The pending timeout is now cleared on the next restore, a jump, a live-edge scroll, and unmount; the row is looked up when the restore actually runs (`ca18f11d2` — `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`)

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`: 100ms prepend retry is still a guess against Virtuoso's 150ms list-refresh window. Last-wins is enough; do not add a third pass.
- Minor: `apps/web/src/app/(app)/chat/components/rooms-client.tsx`: empty and load-failed chrome still mount an empty `TranscriptViewport`. Harmless with zero rows.
- Minor: `apps/web/src/components/ui/file-chip-mini-preview.tsx`: large attachment images are not in `useRememberedImageSize`; a recycled file-chip row can still grow on remount.
- Minor: `apps/web/src/app/(app)/chat/components/__tests__/transcript-viewport-stub.tsx`: `scrollToMessage` always returns true, so the thread-only quote fallback is not exercised in RoomsClient tests.
- Skipped / disagreed: missing happy-dom coverage for land-on-unmounted-row — spec testing decision allows proving that at the pure layer plus stub, with the real-browser check in the PR body. Same note as the last review; not a merge blocker.

### Assessment
Viewport owns pin, land, quote, and prepend restore; `react-virtuoso` is pinned at 4.18.13; the stale prepend-retry race is fixed; leftover notes are nits.

mrosberghaus:
On the second Grok review, in 8a86e6609:

- Large attachment previews (`file-chip-mini-preview`, large variant) now go through `useRememberedImageSize` with `h-auto`, so a recycled row with a big image reserves its box on remount.
- The RoomsClient stub's `scrollToMessage` now answers from the DOM like the real viewport (true only when the room list holds the row), so the thread-only quote fallback is reachable there.

Not changed:
- Empty and load-failed rooms still mount an empty viewport on purpose. The handle has to exist before the first row commits: an empty room's first own send calls `pinToBottomAfterOwnSend` in the same tick that appends the pending shell, and the scroll-on-send test renders an empty room for exactly that.
- The 100ms prepend retry stays a single retry, per the note.

`ca18f11d2` (stale retry superseded) is pulled and verified: the retry is cleared on a later restore, a jump, a live-edge scroll, and unmount, and the index is resolved when it runs.

PR #4525, opened 2026-09-14
fix(core): retire moved occurrences on legacy schedule PUT

mrosberghaus:
## Problem

Legacy `PUT /tasks/{id}/schedule` can orphan a moved occurrence on a v2 series.

Verified path:

1. A v2 series has a moved `PLANNED` row (human exception on the current epoch).
2. Public OpenAPI `PUT /tasks/{id}/schedule` with a **changed** rule calls `buildUpdatedTaskScheduleMetadataV2`, which starts a **new `epochId`** and recomputes `nextRunAt` from the new rule.
3. The handler then called `replaceTaskSchedulePlannedOccurrences`, which is built to **keep durable exceptions**. The moved row stays `PLANNED` on the old epoch.
4. Sync releases via `findNextReleaseableOccurrence(..., current epochId)`, so the old-epoch row is never released. Calendar still lists any `PLANNED` row, so the occurrence stays visible.

Web live-series edits already go through `editCalendarSeries` → `PUT /calendar-schedule`, which retires the future half. Public create/update schedule does not.

Staff Engineer nightly hunt (2026-09-14, `bc-a45ccd9f`) reported this as a hypothesis. The hole is real; the fix is the existing calendar-schedule pair, not a new primitive.

## Smallest design

On legacy PUT, only when a v2 series actually starts a new epoch:

1. `retireTaskScheduleFutureOccurrences` (cancels moved/skipped future exceptions; deletes ordinary future projections)
2. `createTaskSchedulePlannedOccurrences` (projects the new epoch)

Same-rule / no-op PUT (`epochId` unchanged) still uses `replaceTaskSchedulePlannedOccurrences` and keeps `nextRunAt`. No schema or migration.

## What this mirrors

`PUT /tasks/{id}/calendar-schedule` / `editCalendarSeries`:

- retire the old epoch’s future half
- then project the new epoch

Legacy PUT does **not** copy calendar-schedule’s “always new epoch even when the rule is unchanged” contract. Same-rule PUTs remain a no-op for exceptions.

## Proof

In `apps/core/src/routes/v1/tasks/[id]/schedule/put.test.ts`:

- v2 series → seed a moved `PLANNED` row → legacy PUT with a changed rule → that row is `CANCELED`, no live `PLANNED` rows remain on the old epoch, `nextRunAt` matches the new rule, and the next live row is on the new epoch.
- Same-rule PUT keeps the moved `PLANNED` row, the old `epochId`, and the existing `nextRunAt`. Wiring tests assert replace vs retire/create.
- Changed-rule wiring also pins retire-before-create via `invocationCallOrder` (same pattern as calendar-schedule). A swap of the two calls fails.

Follow-up (test-only, Andreas Grok-review minor):

```
$ pnpm --filter core test src/routes/v1/tasks/[id]/schedule/put.test.ts

 RUN  v5.0.0 /workspace/apps/core

 Test Files  1 passed (1)
      Tests  17 passed (17)
   Start at  07:29:27
   Duration  3.08s (import 55%, transform 43%, tests 2%)
```

Original proof:

```
$ pnpm --filter core test src/routes/v1/tasks/[id]/schedule/put.test.ts src/helpers/task-schedule-occurrence-index.test.ts src/routes/v1/tasks/[id]/calendar-schedule/put.test.ts

 RUN  v5.0.0 /workspace/apps/core

 Test Files  3 passed (3)
      Tests  45 passed (45)
   Start at  00:31:32
   Duration  3.55s (import 51%, transform 39%, tests 9%, setup 1%)
```

Also: `pnpm --filter core typecheck` (tsc --noEmit) passed.

<div><a href="https://cursor.com/agents/bc-2d5ed5a3-a788-4c4f-8a23-f8e3c1b7fb0e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2d5ed5a3-a788-4c4f-8a23-f8e3c1b7fb0e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/src/routes/v1/tasks/[id]/schedule/put.test.ts:579-596`: changed-rule wiring test asserts both helpers were called, not retire-before-create order (`invocationCallOrder`). Calendar-schedule’s test does. Ledger test would still catch a swap.
- Minor: `apps/core/src/routes/v1/tasks/[id]/schedule/put.test.ts:253-267`: `useRealOccurrenceIndex` installs real helpers and `beforeEach` only `clearAllMocks()`, so `mockImplementation` leaks. Fine while those tests stay last and Core does not shuffle.
- Minor: `apps/core/src/routes/v1/tasks/[id]/schedule/put.ts:79,147-152,194-199`: `scheduledAt` is captured before the tx; `computeScheduleNextRun(metadata)` uses a later `new Date()`. Calendar-schedule has the same split. Drift, not a new orphan; out of scope for this hole.
- Skipped / disagreed: none. Pre-existing orphans on a non-current epoch stay until a later *changed-rule* PUT; no backfill was asked.

### Assessment
Ready to merge: Yes
Legacy PUT now retires the old epoch’s future half only when a v2 `epochId` actually changes, then projects the new epoch; same-rule PUT still replaces and keeps the moved `PLANNED` row.

PR #4533, opened 2026-09-14
refactor(chat): move TranscriptViewport to TanStack Virtual

mrosberghaus:
## What

Swaps `react-virtuoso` for `@tanstack/react-virtual` (3.14.12, core 3.17.10) behind the unchanged `TranscriptViewportHandle`. Room and thread stay on the one `TranscriptViewport`.

Spike and plan: [SOK-1063](https://linear.app/masumi/issue/SOK-1063/spike-migrate-chat-transcriptviewport-from-virtuoso-to-tanstack). Pulled forward from the mid-October checkpoint because of a jump-then-blank-fill while scrolling up with momentum on iOS Safari, which TanStack's momentum-scroll deferral (core 3.15) targets.

## User-facing impact

- Same behaviour by design: open on the newest message, live-edge follow only when already there, own send always revealed, search and notification jumps land and mark, history loads above without moving what the reader was looking at.
- On iOS, scroll corrections during momentum are deferred by the virtualizer until the finger and momentum are done, instead of being written mid-scroll.

## What TanStack takes over

| Was hand-built on Virtuoso | Now |
| --- | --- |
| `firstItemIndex = 1_000_000` offset trick, `firstItemIndexAfterRowsChange` | `anchorTo: "end"`: the row under the top edge is held by key across prepends |
| `followOutput` + `held` state + `totalListHeightChanged` follow | `followOnAppend` + `scrollEndThreshold`; a hold sets the threshold to 0 |
| `captureAnchor` / `restoreAnchor` round trip through `RoomsClient`, 100 ms retry | removed; core anchors synchronously before commit, so running both would double-correct |
| hand `holdVisibleAnchor` on height growth | core compensates first measurements above the fold and streaming growth at the end |
| `VirtuosoMockContext` in tests | offset-height and rect shims in the test harness |

One residual hold stays ours: when the row under the top edge is a boundary row that the loaded page replaces or removes, core has no key to hold. `rowHoldAfterRowsChange` then holds the first message row below it. Unit-tested.

## Not covered by TanStack (known)

- A row that straddles the top edge and grows (an unfurl landing inside it) is not compensated; core skips that case on purpose for streaming rows ([#1218](https://github.com/TanStack/virtual/issues/1218)). Virtuoso's hand anchor did put it back.
- Open upstream: [#1281](https://github.com/TanStack/virtual/issues/1281) (React 19 `flushSync` warning when measuring from a ref callback; not seen in tests) and [#1287](https://github.com/TanStack/virtual/issues/1287) (desktop Safari drops the prepend anchor write during top rubber-band).

## Verification

- `pnpm --filter web test "src/app/(app)/chat"`: 117 files, 1049 tests green
- `pnpm check`, `pnpm typecheck` green (husky pre-commit)
- Local browser pass on Team Soko (desktop Chrome via the Browser pane, dev server):
  - opens at the bottom, 31 of 101 rows mounted, container height = scroller minus the 24 px padding
  - jump to top, older page loads: the row that sat under the boundary is at exactly 0 px afterwards, no blank frame
  - search jump to a hit far up the room: centered and marked, rest dimmed
  - thread panel: rows pushed to the panel bottom, container height correct
  - no `flushSync` warning during scrolling with `useFlushSync: false`
- Not yet: iPhone. Please check the preview deploy on the same room as the recording: scroll up fast through history, jump to a search hit and a thread deep link, own send while scrolled up, streaming Thought at the live edge.

### Found and fixed during the browser pass

- The hold for a replaced boundary row scrolled after commit, so rows mounted by that scroll were measured against a stale offset. It now sets the virtualizer offset during render, as the built-in key anchor does.
- The virtualizer's "scrolling" flag was still set from a scroll 2 s earlier, so nothing was measured on mount, and its debounced reset then re-applied a stale offset. `useScrollendEvent: true` lets the browser end the scroll with a fresh offset (Safari < 26 falls back to the debounce).
- The hold is clamped at the survivor's start: a new, estimated row straddling the edge is not compensated when measured.

## Also

- Removes the `react-virtuoso` agent skill under `apps/web/.agents/skills` and its `skills-lock.json` entry with the dependency.


mrosberghaus:
## Grok review

### Fixed
- Important: prepend at the live edge now pins to the newest row instead of holding the top boundary (`60baa92c1` — `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`)
- Important: overscan restored to 30/8 rows (old 2400/600 px at the 80px estimate) (`60baa92c1` — same file)
- Important: `scrollMargin` rounded to an integer so float jitter cannot loop the layout effect (`60baa92c1` — same file)

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`: inline `containerRef` callback reattaches every parent render
- Minor: `apps/web/src/app/(app)/chat/components/rooms-client.tsx`: comment still says the viewport owns “the scroll anchor”
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.test.tsx`: `offsetHeight` / `getBoundingClientRect` / `ResizeObserver` shims are never restored
- Skipped / disagreed: boundary-hold integration test is not the proof the PR claimed — `rowHoldAfterRowsChange` is unit-tested; the integration case is an honest happy-dom smoke check, not worth a layout rewrite
- Note: no real-browser pass yet (iOS momentum, search/thread jump, own send, unfurl growth) — still on the author

### Assessment
Live-edge prepend, overscan, and margin rounding are in; confirm on the preview iPhone before treating behaviour as unchanged.

mrosberghaus:
Addressed the three remaining minors in the latest commit:

- `containerRef` callback is memoized with `useCallback`, so it is no longer detached and reattached every render.
- `rooms-client.tsx` comment no longer names the scroll anchor the viewport gave up.
- The viewport test restores the `offsetHeight` / `clientHeight` / `scrollHeight`, `getBoundingClientRect`, and `ResizeObserver` shims in `afterAll`.

On the two "remaining" items that this review predates: 7b0aff2a9 moved the boundary hold to render time and the integration test now asserts the held row at exactly 0 px; a desktop Chrome pass (open at bottom, older-page load hold, search jump, thread panel) is in the PR body. iPhone is still open.

Clean cycle on the final version: `pnpm check`, turbo typecheck, full web suite (721 files, 5642 tests) green.

PR #4534, opened 2026-09-14
fix(core): count the real avatar pool before generating

schaier-io:
The availability count applied the caller's own `excludeIds` filter, the same
one the read uses. A caller could list the pool through the GET, hand the ids
straight back as `excludeIds`, and drive the count toward zero, so generation
ran even when the unclaimed pool was full. FAL bills per image, so that was
spend available on demand to anyone inside the beta.

The count now looks at the real unclaimed pool. The read still honours
`excludeIds`, so the picker keeps showing a fresh set.

Pre-existing: the same narrowed filter fed the count before the top-up moved
off the GET. Part of SOK-1050.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Avatar availability checks now count all unclaimed avatars before generating additional avatars.
  - Excluded avatars remain filtered from returned listings.


schaier-io:
@coderabbitai review

PR #4535, opened 2026-09-14
feat(core): attribute generated avatars to the user who asked

schaier-io:
`SokoBotAvatar` recorded nothing about who caused a row, so there was no way
to count generation per user and no way to cap it. The chat-invitation limiter
counts `inviterId` in a window; this column gives the avatar pool the same
thing to count.

`requestedByUserId` is TEXT with an `ON DELETE SET NULL` foreign key, matching
`soko_bot_turn.requestedByUserId`. It is not `@db.Uuid`: `user.id` is TEXT, and
rows created before Better Auth started emitting UUID ids do not parse as one,
so a UUID column would make Postgres reject those users with 22P02.

The foreign key is added `NOT VALID` and validated in a second statement. Both
forms lock `user` against writes, but a plain `ADD CONSTRAINT` holds that lock
across its validation scan, and `NOT VALID` releases it at once. The scan then
runs without blocking a single user write.

The column is nullable because the pool cron generates on its own schedule and
no user asked for that. Charging it to somebody would spend an allowance on
work they never triggered. The top-up route stamps the session user, never a
body field, so a caller cannot bill a different account.

`reservedAt` marks a row claimed as a generation slot before the paid image
call, so an attempt is countable the moment it starts rather than only once it
succeeds. Nothing writes it yet; the layer that reserves slots does.

From here on the top-up route stamps every generated row with its caller.
Nothing reads that column yet, so no user-visible behaviour changes; the
layer that caps generation is what starts counting it.

Part of SOK-1050.

---

## Summary by CodeRabbit

- **New Features**
  - Avatar top-ups now record which authenticated user requested the additional avatars.
  - Avatars generated through scheduled pool stocking remain unassigned to an individual requester.
  - Avatar generation supports tracking in-progress reservations to improve top-up processing and accountability.

- **Bug Fixes**
  - Improved consistency between authenticated requests and the user recorded for generated avatars across supported request formats.


schaier-io:
@coderabbitai review

PR #4536, opened 2026-09-14
fix(core): cap avatar generation per user per hour

schaier-io:
The pool check is not a rate limit. It asks whether the unclaimed pool is
short, so a caller who empties the pool keeps it short and keeps buying paid
FAL images. Nothing counted what one account had already spent.

This counts the rows the user caused in the last hour and refuses past the cap
with a 429, matching the guest-invitation limiter. It returns the remaining
allowance and shrinks the batch to fit, rather than only asking "are you under
the cap": that check would pass a user sitting at 23 and then generate a full
page of 6, finishing at 29 against a cap documented as 24.

Reads are untouched and the check runs only once the pool is short, so a user
at their cap can still browse a full pool. The pool cron writes a null
`requestedByUserId`, so background refills consume nobody's allowance.

`tooManyRequests` now takes the same optional metadata as the other error
helpers, so the 429 carries a `kind` rather than only a message. No client
reads the `kind` yet; the web action matches on the status, as the turn action
already does for its 409.

The picker gets its own sentence for this case. It collapsed every failure
into "Could not load pictures.", which reads as breakage and invites an
immediate retry. A spent allowance is a wait, so the action maps the 429 to
its own error code and the toast says to try again later.

The picker also stops its skeletons on a failed load. It only cleared them on
success, so a refusal left six tiles pulsing under the toast, which reads as
"still loading". That was survivable while a failure meant Core was broken; it
is not, now that a refusal is an expected answer.

Known gap, not closed here: the count runs before the paid call and the row is
written after it, so concurrent requests can each pass the check, and a run
that bills FAL but fails to store the image is never counted. Bounding that
needs allowance reserved before the spend. Tracked on SOK-1050.

Part of SOK-1050.

---

## Summary by CodeRabbit

* **New Features**
  * Added a limit of 24 Soko Bot avatar generations per user per hour.
  * Requests exceeding the remaining allowance are automatically reduced or rejected.
  * Avatar browsing remains available without consuming the generation allowance.
  * Added localized rate-limit messages in English, German, and Spanish.

* **Bug Fixes**
  * Rate-limit errors now display a dedicated message and stop avatar loading indicators.
  * Improved handling of rate-limit responses while preserving generic error messages for other failures.

* **Documentation**
  * Documented the avatar top-up rate-limit response.


schaier-io:
@coderabbitai review

PR #4537, opened 2026-09-14
fix(core): reserve the generation slot before spending on it

schaier-io:
The hourly cap counted rows that only existed after the paid call returned.
Two holes followed. Concurrent top-ups all read the same count and all passed,
so the cap bounded a serial loop and not a burst. And a run that billed FAL but
failed to store the image wrote no row at all, so a blob outage held the count
at zero while every request kept spending.

Generation now claims its slots first and fills the images in afterwards.
`reserveDraws` counts the caller's recent rows and inserts the reservations
inside one `serializableTransaction`, so two concurrent top-ups cannot both
read the same allowance and both spend it. Postgres aborts the loser with a
serialization failure, which the helper retries and finally surfaces as a 409.
The FAL call stays outside the transaction, so no paid call holds a lock.

An attempt is countable from the moment it starts, which is what the cap needs.

`reservedAt` keeps an unfilled row out of every read that serves it, so nobody
is shown a blank avatar and no bot can claim one. Draw selection still sees
those rows on purpose, so a reservation holds its triple while it is in
flight. A run that fails hands its unspent
slots back. The pool cron sweeps reservations whose run died, but ages them
off `createdAt`, the same column the cap reads, so a swept row still counts
against the hour it was spent in.

A caller who cannot generate still gets the pool it asked for. A spent
allowance and a lost serialization race both say nothing about the rows
already sitting there, so both fall back to a plain read and only surface the
failure when that read comes back empty. Any other error still propagates: a
broken pool must not read as a merely empty one.

Releasing unspent slots cannot fail the request. It runs while handling
a failure the caller is meant to survive, and the blip that broke the run can
break the delete too, so it logs and leaves the rows to the sweep.

The sweep also runs on the top-up path. Vercel runs crons on production only,
and this route exists because preview has none, so a reservation whose run
died there would otherwise hold its draw and the user's allowance until
somebody shipped to production.

The blob upload gets the same timeout the two fetches already carry. It was
the one unbounded step of a fill, so an upload that hung past the sweep cutoff
would have had its own reservation deleted out from under it.

Closes SOK-1050.

## Verification

All VERIFIED in this session unless tagged otherwise.

- `pnpm --filter web test` → `Test Files 722 passed (722)`, `Tests 5657 passed | 1 skipped (5658)`
- `pnpm typecheck` → `Tasks: 19 successful, 19 total`; `biome check .` → `No fixes applied`
- `pnpm --filter web messages:parity` → `de: ok`, `es: ok`
- `pnpm --filter @sokosumi/core test` → `Tests 8 failed | 5280 passed | 20 skipped (5308)`. The 8 failures are in 6 files this stack does not touch (`stripe.client`, `routes/sync/index`, `routes/v1/index`, `workspaces/index`, `well-known/oauth-issuer-metadata`, `task-schedules-sync`), all with ~5000 ms timeout durations. Re-run alone: `Test Files 6 passed (6)`, `Tests 85 passed (85)`. Read as load flakes, not regressions.
- Migration applied to a scratch database: `All migrations have been successfully applied.` Catalog afterwards: `requestedByUserId|text`, `reservedAt|timestamp without time zone`, and the FK `convalidated|t`.
- The generated Core client is byte-exact generator output: re-running `generate:core:snapshot` at HEAD leaves no diff.

### Mutation testing

Each guard was removed or inverted and the suite re-run; every one is caught by the test that names it. Covered: the Serializable wrapper, the cap throw, the allowance clamp, clearing `reservedAt` on fill, releasing unspent slots, the `reservedAt` read filter, the top-up sweep, the recoverable-failure predicate in both directions, the empty-fallback guard, the 429 action mapping in both directions, and the blob `abortSignal`.

### Concurrency, measured rather than argued

A review raised that two concurrent `reserveDraws` calls would collide on `@@unique([subject, background, seed])` and surface as Prisma `P2002` (a 500) rather than a retryable `P2034`. Measured directly with two `SERIALIZABLE` psql sessions on PostgreSQL 18.6. Both interleavings abort with `40001`:

- loser inserts against the winner's uncommitted key → `Canceled on identification as a pivot, during conflict in checking`
- winner commits first, loser inserts after → `Canceled on identification as a pivot, during write`

The full-table read in `nextAvatarDraws` puts a relation-level predicate lock in the read set, so the write is always a serialization failure, which `serializableTransaction` retries. INFERRED caveat: this holds while the planner picks a seq scan. The pool is bounded small by design (floor of 24), so that is the realistic plan; a much larger table read through an index could take finer-grained predicate locks.

## Rollback cost

This build writes rows with `reservedAt` set and `imageUrl` empty. The previous build has no `reservedAt` filter on its reads and no guard in `claimAvatar`, so a rollback while reservations exist would let the old picker render a blank tile and write an empty `avatarImageUrl` onto a bot. Reservations clear within the hour on their own. To clear them immediately after a rollback: `DELETE FROM soko_bot_avatar WHERE "reservedAt" IS NOT NULL;`

## Found during review, deliberately not fixed here

All pre-existing on `main` and outside this PR's own changes.

- `avatar-picker.client.tsx` grows `seen` by one page per "show others" click and never trims it, while `excludeIds` is capped at 60. Repeated clicks eventually fail validation and keep failing.
- The picker's start-over retry uses `.then((r) => (r.ok ? r.value : []))`, so a failure on that second call is swallowed with no toast.
- `createMySokoBot` does not declare a 404 although `claimAvatar` can raise one. Unreachable in practice, since reserved ids never reach a client.
- The regenerated Core client carries drift unrelated to avatars, because the committed snapshot on `main` was stale. Correct per the generated-files rule, but it widens this diff.
- The commit `test(core): drive the avatar routes instead of reading their source` also renames a web server action, so its `test` type understates it. Left as is: rewriting it means rewriting a branch that carries merge commits under an open PR.

---

## Summary by CodeRabbit

* **New Features**
  * Improved Soko Bot avatar generation with safeguards against concurrent requests.
  * Added hourly generation limits and automatic cleanup of stale generation attempts.
  * Improved recovery when generation fails, including fallback to available existing avatars.
  * Avatars remain unavailable until generation is complete, preventing incomplete results from being shown.
* **Documentation**
  * Documented the `409 Conflict` response returned when a generation slot cannot be reserved due to concurrent activity.


schaier-io:
@coderabbitai review

PR #4552, opened 2026-09-14
feat(web): honor chat history read budget backoff (SOK-1065)

mrosberghaus:
## User impact

When Core throttles sustained chat history polling, the web app now honors the asked-for wait: all chat background readers (open room transcript, sidebar collections, unread indicator) pause quietly for ~N seconds and resume on their own. No retry storm burns the shared per-user budget, displayed content never blanks, and realtime updates keep flowing throughout.

## Links

- Linear: SOK-1065 (consumes the SOK-1060 throttle contract; no Core changes)

## Verification

- `pnpm --filter web test`: full suite green (40 new tests: shared clock, carrier, both route families, fetch seam, scheduler seam incl. sibling-quieting proof)
- `pnpm --filter web typecheck` and `pnpm --filter web check`: clean
- Two-axis code review (standards + spec): no blocking findings
- Review follow-ups: jitter-tail resume timer consumed on read start; throttle clock input hardened (non-finite/sub-1s delays fall back); header-over-body precedence pinned at fetch seam


mrosberghaus:
## Grok review

### Fixed
- Important: a jittered resume timer could outlive Core's throttle clock and fire a second catch-up read (`8757d5126` — `apps/web/src/components/chat/use-chat-refresh-scheduler.ts`)

### Remaining
- Minor: `apps/web/src/lib/clients/core.request.ts`: generic Core client imports the chat throttle module only for `parseRetryDelaySeconds`
- Minor: `apps/web/src/components/chat/fetch-background-json.ts`: header-over-body delay is pinned on the Core carrier, not at the fetch arming seam
- Minor: `apps/web/src/lib/chat/chat-read-throttle.ts`: `noteChatReadThrottled` trusts a raw `0`/`NaN` argument; current callers only pass parse output

### Assessment
Shared clock, proxy 429 forwarding, and scheduler quieting match SOK-1065; the leftover resume-timer catch-up is gone, three layering/test nits are unfixed.

mrosberghaus:
Review follow-up (commit `9e05908c4`):

**Verified the fixed item.** Traced the jitter-tail scenario: deferred read with jittered resume at 37.5s, clock clears at 30s, an explicit request at 32s reads immediately, and the leftover timer would have fired a stale second read at 37.5s. Consuming the resume timer on read start is correct and mirrors the existing main-timer handling.

**Minor 3 — implemented.** `noteChatReadThrottled` now routes any unusable delay (missing, non-finite, < 1s) to the bounded fallback. This was worse than described: `NaN`/`Infinity` permanently poison the clock via `Math.max`, and `0` silently un-throttles. 5 new unit tests, watched fail then green.

**Minor 2 — implemented.** Added a fetch-seam test with conflicting header 30 / body 45 asserting the catch-up lands at 30s. Proved non-vacuous by temporarily inverting precedence (exactly that test failed), then reverted.

**Minor 1 — declined.** The `retryAfterSeconds` carrier field has exactly one reader (the chat proxy), so the parser serves this feature exclusively. Moving it to a neutral home anticipates a second consumer that doesn't exist; cohesion + YAGNI. 10-minute move if you disagree.

Verification: typecheck clean, 85 tests across the 3 touched files green, Biome clean.

PR #4558, opened 2026-09-14
fix(chat): preserve message expansion while scrolling

mrosberghaus:
Expanding a long message with **Show more**, scrolling it out of view, and returning caused the virtualized row to remount collapsed. Its changed height made the transcript jump. The overflow control also disappeared for the initial render of a returning row.

Remember expansion and overflow state within each open room or thread transcript, restoring both before the virtualizer measures the row. Message bodies and individual quotes have separate stable keys, with only the latest content version kept for each. Editing text back to an earlier version does not revive its old expansion state. Collapsed content is remeasured after resizing; measurements update overflow without changing expansion.

Verification:
- `pnpm --filter web test`: 723 files passed; 5,708 tests passed, 1 skipped.
- `pnpm precommit`: repository Biome checks and all workspace typechecks passed.
- Browser component reproduction with the real message body, Markdown, and transcript viewport: the report previously shrank from 1,180px to 888px on return. Four fixed-state revisit checks now preserve 888px when collapsed and 1,180px when expanded.
- Authenticated end-to-end chat was not tested; the local app stack was unavailable.

Browser component screenshot after scrolling away and back (sample content):

expanded-after-revisit


mrosberghaus:
## Grok review

### Fixed
- Important: remount did not assert the first commit's measured height (`53647f204` — `apps/web/src/app/(app)/chat/hooks/use-clamped-overflow.test.tsx`)
- Important: body vs quote keys shared one Map with no independence test (`53647f204` — same file)

### Remaining
- Minor: `use-clamped-overflow.tsx`: `measureOverflow` always writes `expanded: false` (queued ResizeObserver could snap an expanded row shut)
- Minor: `use-clamped-overflow.tsx`: Map keys keep the full message body until the transcript unmounts
- Minor: `use-clamped-overflow.test.tsx`: "separate transcripts" is unmount-A then mount-B, not two live providers
- Minor: editing back to the original text can restore the old expansion (content lives in the key)
- Skipped / disagreed: a `TranscriptViewport` scroll-out/in remount test — happy-dom shims every `data-index` node to 40px and does not honor virtualizer scrolls, so it would test the shim. The new measure-callback-ref test is the same commit timing as `measureElement`. Existing first-render `onRender` already fails if restore moves into `useLayoutEffect`.

### Assessment
Restore-during-render, per-viewport provider, and prefixed body/quote keys match the bug; leftover notes are map-key lifetime and a possible observer write of `expanded: false`.

mrosberghaus:
Verified the linked review against head `53647f204c`.

- The two important test fixes are valid. The remount test asserts the first callback-ref measurement, and the body/quote test checks restoration independently in both directions. These are hook/commit-timing tests; they do not claim real browser layout coverage. All 6 committed hook tests pass.
- I disagree with treating the queued ResizeObserver callback as a demonstrated defect. The layout-effect cleanup disconnects the old observer on expansion. The [ResizeObserver specification](https://drafts.csswg.org/resize-observer/#dom-resizeobserver-disconnect) clears both observation targets and active targets on disconnect. Manually invoking a disconnected mock callback would not establish the reported browser race. No defensive production change recommended without a reproducer.
- The two-live-providers coverage note is fair. An isolated probe with both providers mounted simultaneously passes: expanding one leaves the other collapsed. The provider implementation already isolates state correctly.
- Full content in keys does retain visited content versions until the transcript unmounts. That lifetime is explicit, and I found no evidence here requiring an eviction/cache redesign.
- Editing A → B → A really does restore A’s earlier expansion; verified with a second probe. This is a minor disclosure-state quirk, not a scroll-remount regression. A stable identity plus content revision would address both old-version retention and this behavior if pursued separately.
- The virtualizer-test limitation is accurate: the existing happy-dom harness assigns every `data-index` row 40px. Keep the previously recorded real-browser reproduction as layout evidence rather than presenting that shim as real scrolling validation.

No new blocking finding from this review. I would keep this PR focused and merge once its required checks pass. Verification used the exact reviewed source in an isolated checkout with existing installed dependencies; the package manager warned that those dependencies were out of sync with the lockfile, so this is targeted test evidence, not a fresh-install CI claim. No production changes made.

PR #4559, opened 2026-09-14
feat(web): show a pinned note on pinned channel messages (SOK-1066)

mrosberghaus:
## Summary

Pinned messages in a Channel now show a small, always-visible **Pinned** note (pin icon + label) on their transcript row, so readers can spot them while scrolling. The note appears and disappears live on pin/unpin.

Resolves [SOK-1066](https://linear.app/masumi/issue/SOK-1066/show-a-pinned-note-on-pinned-messages-in-a-channel).

### Why a Core change
The web transcript only learned which messages were pinned after the pinned panel loaded a page (paginated, no pin cap). A row-only change would have hidden the note until someone opened the panel, and never shown it for older pins.

### Changes
- **Core:** `ChatRoomMessage` gains nullable `pinnedAt`, read from the existing pin relation in the shared message include/mapper. Null for unpinned, deleted, Direct and thread messages. Additive; no migration.
- **Web client:** regenerated from the Core snapshot (only `pinnedAt` added). The Ably full-message schema and hydration carry `pinnedAt`.
- **Web transcript:** rows derive `isPinned` from `message.pinnedAt`. Own pin/unpin, pinned-panel unpin and realtime pin events patch `pinnedAt` on the loaded message. The separate `pinnedMessageIds` set and the panel's `onIdsLoaded` seeding are removed. The leftover room-change effect is replaced with a render-time guard.
- **Row:** new `MessagePinnedLabel`, which sits next to the timestamp/edited cue on group-leading rows and on its own line above the body on continuations. Hidden on deleted rows; never shown in the thread panel. The icon is `aria-hidden`.
- **i18n:** `App.Channels.PinnedMessages.pinned` in en/de/es.
- **Apple:** follow-up queued in `apps/apple/PARITY.md` (derive native pinned state from `pinnedAt`, localize the label). No Swift changes.

## Verification
- `pnpm check`, `pnpm typecheck`: pass
- `pnpm test`: web 722/722 files pass. Core 503/504 files pass. The one failure is a 5s timeout under parallel load in `src/routes/sync/index.test.ts` (cron auth, unrelated); it passes 38/38 when run alone.
- New tests:
  - Core messages GET (pinned / unpinned / deleted-but-pinned) and a thread-reply `pinnedAt: null` assertion.
  - `RoomMessageRow` note placement: header, continuation, pinned + edited, deleted, unpinned, hidden icon.
  - `RoomsClient`: first load without opening the panel, realtime unpin, other-room pin ignored, only the touched row re-renders.

## Screenshots
Not captured yet: the UI has not been checked in a browser. Screenshots (light/dark, header + continuation) will follow before marking the PR ready.

## Schema changes
None (no migration). The API contract gains one additive nullable field, and the web Core client is regenerated.


mrosberghaus:
## Grok review

### Fixed
- Important: Ably full-message schema required `pinnedAt`, so a web-first deploy dropped every live create/update/delete from a Core that omitted the field (`8708a7815` — `apps/web/src/lib/ably/schema.ts`). Optional now; omitted updates keep the loaded pin; explicit `null` still clears it.
- Important: no test that a full Ably update keeps the note on an already-pinned row (`8708a7815` — `apps/web/src/app/(app)/chat/components/__tests__/rooms-client-row-render-isolation.test.tsx`)

### Remaining
- Minor: `apps/web/src/lib/ably/apply-chat-room-message-id-envelope.ts`: id-envelope tombstones keep `pinnedAt`; the note is already hidden via `!isDeleted`
- Minor: `apps/core/src/schemas/chat-room.schema.ts`: OpenAPI says Direct/thread `pinnedAt` is always null; the mapper only reads `pins[0]` and clears on delete (write-path rejection makes it true today)
- Minor: `apps/web/src/app/(app)/chat/components/thread-panel.tsx`: thread-panel hiding is default `isPinned=false`, untested
- Minor: `apps/core/src/routes/v1/chats/rooms/helpers.test.ts`: no direct `mapChatRoomMessage` unit cases for pinned / unpinned / deleted-but-pinned (HTTP GET covers them)

### Assessment
Channel rows now derive pin state from Core `pinnedAt` without the panel, and live events survive a web-first deploy; remaining notes are nits plus draft screenshots.

mrosberghaus:
Follow-up on the Grok review.

### Fixed items
- **Ably `pinnedAt` optional + fallback (`8708a7815`)**: reverted in `4f4c0ae9b`. Core and web ship together, so `pinnedAt` stays required on the Ably full-message schema, like `unfurls` and `membership`. A missing field is a contract bug and should fail loudly, not be tolerated silently. The `omittedPinnedAt` flag and `messageWithPreservedPinnedAt` were a permanent shim for a deploy window we don't have (AGENTS.md: no silent dual-read layers). Kept the two realtime tests: a full update carrying `pinnedAt` keeps the note, and one carrying `null` clears it.

### Remaining items
- **Id-envelope tombstones keep `pinnedAt`**: no change. The row hides the note via `!isDeleted`, Core REST and full events null it on delete, and the next read replaces the tombstone.
- **OpenAPI "always null" for Direct/thread**: accurate. The pin route rejects Directs (`room.kind !== "channel"`) and thread replies (`parentMessageId != null`), so no pin row can exist for them.
- **Thread panel hiding untested**: skipped. All four thread-panel tests stub the row, so a test would only assert a missing prop. The row test covers `isPinned` → visible note.
- **No `mapChatRoomMessage` unit cases**: skipped. They are covered at the agreed seam: messages GET (pinned / unpinned / deleted-but-pinned) and thread GET (`pinnedAt: null`).

Verification: web chat + Ably tests 142/142 files, `pnpm check`, `pnpm typecheck` pass.

PR #4568, opened 2026-09-14
fix(chat): keep the live edge in view when the scroller resizes and stop snapping near-end readers on refresh

mrosberghaus:
## What

Two stick-to-bottom fixes in the web chat transcript (`TranscriptViewport`):

1. **Last message hidden under the composer.** The scroller kept its top edge fixed whenever its height changed, so any composer growth (more lines, attachments, format strip, mobile keyboard) pushed the live edge under the composer. Only the format strip and quote chip had a resize hook. The viewport now watches the scroller's height with a `ResizeObserver` and keeps a near-end reader's distance from the live edge. The `onChromeResize` / `scrollToBottomIfPinned` plumbing through the composer, session composer, thread panel and rooms client is removed.
2. **"Stick to bottom" jumping after opening a room.** Every rows refresh (the 3s fallback poll builds a new array even when nothing changed) re-pinned any reader within 200px of the end to the exact end. Now only a view exactly at the end is put back there; the virtualizer's `followOnAppend` still pulls a near-end reader down for genuinely new messages.

3. **Visual gap between messages and composer.** (see also 5) The composer's top padding (8px / 12px) sat between the scroller and the composer box, so messages were clipped above the box. That spacing now lives at the bottom of the list content (room, thread, Instant loading view): the last message keeps its room above the box and scrolled content runs flush to it.
4. **Window resize left the view far above the end.** The resize follow wrote `scrollTop` without telling the virtualizer, which learns of a scroll from the event a frame later; rows it measured in between were compensated against the old offset and undid the write. The offset is now handed to the virtualizer first. Verified on the preview deployment: distance to the end stays 0 through 1000×900 → 820×700 → 1400×1000 (was 478 / 750 / 286 before).

5. **Edge.** The scroller's bottom edge meets the composer box directly and cuts content hard there. An intermediate attempt that scrolled the list under the composer box was reverted: it let the newest message slide beneath the box as soon as the reader scrolled up. The avatar-at-the-corner look is deliberately left as is.
6. **Clamped virtualizer scroll.** After a row grows at the end, the virtualizer's scroll can be clamped by the browser before the rows below are moved; the viewport now finishes that scroll on the next virtualizer change and re-reads the edge flags there.

## Verification

- `pnpm --filter web typecheck`, Biome, and the chat test folder (118 files / 1071 tests) pass.
- Three new regression tests in `transcript-viewport.test.tsx` fail on `main` and pass here.
- Reproduced and re-verified both symptoms in a real browser (headless Chromium at 60fps) using a throwaway harness rendering the real viewport with a growable composer: composer growth keeps distance 0 / 120 as appropriate, poll with the reader 120px up no longer snaps, exact-end poll stays at the end, far-up reader stays top-anchored.


mrosberghaus:
## Grok review

### Fixed
- Important: `onChange` finished a clamped virtualizer scroll whenever the DOM was short by any amount, which could snap a reader back from history if `scrollOffset` was still at the end. The write now requires a growth-sized gap (`30ce8706b` — `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`).
- Important: later physics was untested. Added scroller-grow (clamped composer-shrink), near-end grow, exact-end poll, near-end append follow, and `settle()` after resize (`30ce8706b` — `apps/web/src/app/(app)/chat/components/transcript-viewport.test.tsx`).

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`: hold path still uses `virtualizer.isAtEnd` while everything else uses DOM distance — jump-button state during a search hold can disagree with `distanceFromEndRef`.
- Minor: `atEndRef` duplicates `distanceFromEndRef <= STICK_TO_BOTTOM_NEAR_PX`.
- Skipped: no test that drives `onChange` finishing a clamped row-growth scroll — happy-dom does not reproduce the virtualizer-scrolls-before-rows-move clamp. The write is gated instead.
- Stale “ResizeObserver is made inert” comment was corrected in the same test file.

### Assessment
Live-edge resize, poll-vs-append, and the clamped-scroll snap are covered; leftover notes are hold-path flag drift and an `onChange` path the harness cannot drive.

mrosberghaus:
Follow-up to the review in `97cf65ef8`:

- Hold path: the `virtualizer.isAtEnd` effect is removed. The distance is recorded on every virtualizer change (an append under a hold mounts and measures the new row, which notifies), so the jump control and `distanceFromEndRef` now come from the same DOM reading.
- `atEndRef` is gone; the near-end check is derived from the recorded distance where it is needed.

The growth-sized gate on the clamped-scroll finish in `30ce8706b` is kept. One known residual: a size change committed between a wheel input and its scroll event can still finish a scroll against the stale offset, but that window is one frame at the very start of a gesture from the exact end and moves the reader by at most that tick; the gate does not change it and a fix would need the virtualizer's private intended-offset field.

PR #4569, opened 2026-09-14
feat(calendar): erase calendar data with accounts and workspaces

enjojoy:
## Summary

Account and organization erasure now removes retained Calendar records in the same transaction as the parent deletion. This covers schedules, occurrences, links, quarantine records, project history, and personal-workspace Calendar data, while blocking erasure when task payments remain unresolved or payment authorizations are still live.

- Coordinate Calendar writers and erasure with user, workspace, project, and task locks. Lock organization membership rows for deletion before notification cleanup so concurrent email dispatch cannot deadlock with the organization cascade.
- Preserve account payment-ownership checks and post-commit task-file cleanup.
- Remove workspace Calendar outbox records during erasure, including legacy parent-delete cascades. Account cleanup preserves pending revocations in surviving organizations so the existing worker can cancel scheduled emails and deliver access revocation.

Project deletion remains unavailable, as shipped in #5083. This PR contains no Delete UI, deletion eligibility fields, deletion replay table, or additional ordinary-project-delete guard. No Web or public API schema changes remain.

## Migration

`20260922140000_calendar_workspace_erasure/migration.sql` adds an `AFTER DELETE` workspace trigger to remove that workspace's Calendar outbox records. It replaces this branch's unmerged deletion-protection migration; no tombstone table or Prisma model is added. No data-migration script is required.

## Verification

- `pnpm check` and `pnpm typecheck`: passed after integrating current `main` (17 typecheck tasks).
- Focused erasure, organization/account deletion, and notification-email tests: 83 passed.
- CI configuration tests: 76 passed.
- Full migration history applied successfully to a disposable PostgreSQL 17 database.
- Updated CI PostgreSQL step: 7 tests passed across Projects API, project activity, and Calendar erasure.
- Concurrent deletion/email regression reproduced PostgreSQL `40P01` / Prisma `P2034` before the fix and passed afterward. Both transactions now finish without a deadlock.
- CI now applies migrations and runs the Calendar erasure suite, including the concurrency regression, on every applicable Core test run.
- Final independent review: no remaining actionable findings. No browser verification; no Web changes remain.

## Tracking

- [SOK-890](https://linear.app/masumi/issue/SOK-890)
- Follows #5083.

## Summary by CodeRabbit

* **New Features**
  * Organization and account deletion now removes associated Calendar data and invalidation records.
  * Deletion is blocked when unresolved payments or active payment authorizations could be affected.
  * Calendar changes and task operations now use the requesting user’s context for safer conflict handling.

* **Bug Fixes**
  * Improved cleanup of pending and published Calendar notifications during workspace and account deletion.
  * Added safeguards to prevent deleting Calendar data belonging to another workspace.


enjojoy:
@coderabbitai review

mrosberghaus:
## Project deletion review

Focused on ordinary project deletion: eligibility, the delete route, the database guard, idempotency, and the close-vs-delete dialog. Account, organization, and personal-workspace erasure are out of this pass.

The trigger, the occurrence foreign key, and the tombstone hold for the normal path. An open project with a live schedule, a `SCHEDULE` link, or an occurrence is not deleted. A replay stays on the same workspace, project, and actor. Close does not become a delete.

Five gaps still show the wrong action.

### 1. Eligibility misses archived quarantine

`hasProjectCalendarHistory` requires `archivedAt: null` for metadata, `nextRunAt`, and quarantine (`apps/core/src/helpers/project-calendar-history.ts:13-18`). The trigger's quarantine check does not (`packages/database/prisma/migrations/20260914130000_protect_calendar_deletion_erasure/migration.sql:60-66`).

`GET` can return `canDelete: true`. Confirm then returns 409. The route tests mock `findFirst`, so they never see this filter. `project-calendar-delete-guard-db.test.ts` runs only with `RUN_DATABASE_INTEGRATION_TESTS=true`, which CI does not set.

Use one predicate for `GET` and the trigger. Add an archived-quarantine case and run that database test in CI.

### 2. Every skipped delete is labeled calendar history

After the helper reports no history, `deleteMany` with `count !== 1` still returns `project_has_calendar_history` (`apps/core/src/routes/v1/projects/[id]/delete.ts:115-126` and `143-149`). A closing, closed, or close-operation project hits that branch because of the `WHERE`, not because of history. The route test locks this in (`project-id.routes.test.ts:550-591`).

The dialog then explains history and offers Close. Close of an already-closing project fails. Return `project_has_calendar_history` only from the history predicate. Use a different conflict when the project is closing or closed.

### 3. An archived schedule can leave the project

The helper and the trigger both ignore archived tasks that still have `metadata` or `nextRunAt` (`migration.sql:54-59`). `Task.projectId` is `ON DELETE SET NULL` (`packages/database/prisma/schema.prisma:2501`). Archive only sets `archivedAt` (`apps/core/src/routes/v1/tasks/[id]/delete.ts:70-78`). Sync already treats "archived plus `nextRunAt` still set" as a real interleaving (`apps/core/src/services/task-schedules-sync.ts:92`).

Delete then succeeds, writes a tombstone, and that row is no longer on the project. Links and occurrences stay blocked. Count archived `metadata` / `nextRunAt` in the guard, or clear those columns in the archive update.

### 4. A 409 sticks to the next project

`historyRejected` is cleared only when the component goes away. Closing the dialog clears the typed name and the idempotency key, not `historyRejected` (`apps/web/src/app/(app)/projects/components/project-detail-actions.tsx:292-299`). The page does not set `key={project.id}` (`apps/web/src/app/(app)/projects/[projectId]/page.tsx:102`).

After a 409, client navigation keeps the history hint and hides Delete even when `canDelete` is true. Set `key={project.id}` on `ProjectDetailActions`, and reset `historyRejected` when `projectId` changes.

### 5. Core and web skew the action

New Core requires a UUID `Idempotency-Key` before the transaction (`delete.ts:36-45`). Web that does not send one gets 422 on every delete. New web treats a missing `canDelete` as false (`project-detail-actions.tsx:96`). Old Core omits the field, so every project hides Delete and the hint invites Close. Close does not make the project deletable again.

Ship Core and web together, or accept a missing key for one release and treat a missing `canDelete` as unknown.

### Minor

- A missing user or workspace lock is returned as 404 "Project not found" (`delete.ts:77-82`).
- The history hint is a `<p>` inside the menu (`project-detail-actions.tsx:213-217`).
- `en`, `de`, and `es` delete keys match. The menu hint does not say that Close leaves the project undeletable. The post-409 `historyError` does.

mrosberghaus:
## Recommendation: remove project deletion

Remove the ability to delete a project. Close should be the only way a person ends one.

`DELETE /v1/projects/{id}` is already shipped, so this is an API removal, not a small cut on this PR. Account and workspace erasure can still delete the row. That path is not a user action on one project.

Delete and Close are different promises, and this PR's machinery exists to stop Delete from breaking the Close promise.

- Delete removes the project, detaches linked tasks and jobs, and deletes project files. The dialog says those tasks and jobs stay available.
- Close keeps the project, stops future scheduled runs, and retains calendar history. The dialog says the project cannot be reopened. A second close is rejected. Already-released work is not canceled.
- A project with a schedule, a `SCHEDULE` link, or a retained occurrence cannot be deleted, so the safe action is already Close. Ordinary tasks and jobs do not block Delete: their `projectId` is cleared and the project still goes. The trigger, `canDelete`, the 409 race, and `project_deletion_tombstone` exist only so a project with no calendar history can still be hard-deleted, and so a retry can return success after the row is gone.

Removing Delete removes that machinery. The earlier review findings on eligibility, the mislabeled 409, archived-schedule detach, the sticky dialog, and Core/web skew apply only if Delete stays.

Close is not yet a full substitute. These are follow-up improvements, not a reason to keep Delete:

1. A closed project stays in the project list. Close cannot be undone, and Core has no reopen route. The list needs a clear place for closed projects, or an empty project created by mistake stays visible forever.
2. Delete is the only action that detaches every task and job and removes project files. A task can already be moved off one by one. If "unwrap this project and delete its files" is still required, put that on Close or on an explicit move.
3. Close can stop in `CLOSE_FAILED` and then needs retry or "cancel owed work." That recovery becomes the only exit, so it needs to be the solid path.

mrosberghaus:
https://github.com/masumi-network/sokosumi/pull/5083

mrosberghaus (changes requested):
As discussed in the chat

enjojoy:
@coderabbitai review

mrosberghaus (changes requested):
The lock design and delete ordering look good. Two cleanup gaps should be fixed before this ships, because both leave data behind after an erasure that reports success.

## Must fix

**1. Queued notification emails are still sent after erasure.**
[`calendar-erasure.ts:168`](https://github.com/masumi-network/sokosumi/blob/8a67ba626efd20efc98a7211842a8cae21cfd961/apps/core/src/helpers/calendar-erasure.ts#L168) runs `tx.notification.deleteMany({ where: { workspaceId } })`. Some of those rows have an `emailScheduledAt` and a Resend `emailId`, and nothing cancels those sends. The member-delete trigger doesn't help: the rows are already gone, and the new workspace trigger removes the revocation outbox row for that workspace.
- Scenario: a user has an unread notification with an email queued for an hour from now, then deletes their account. The email still reaches the erased address. The same happens to the remaining member of a deleted organization.
- Fix: follow the pattern at `notifications.ts:580-591`. Select the scheduled rows inside the transaction, return them from `eraseWorkspaceCalendarData`, and call `cancelNotificationEmails` after commit in both callers.

**2. Deleting an organization leaves its task files publicly reachable.**
[`organization-deletion.ts:69`](https://github.com/masumi-network/sokosumi/blob/8a67ba626efd20efc98a7211842a8cae21cfd961/apps/core/src/helpers/organization-deletion.ts#L69) erases every task in the organization's workspace, and the `TaskFile` rows are deleted with them. The stored blobs are never deleted, which the account path does do (`user-deletion-tasks.ts`, `deleteTaskFileIfOwned` after commit). Before this PR, an organization with tasks could not be deleted at all (`task_workspaceId_fkey` is RESTRICT), so this leak is new.
- Fix: collect `taskFile` rows for `task: { workspaceId }` inside the transaction and delete the blobs after commit.

Both fixes fit one seam: `eraseWorkspaceCalendarData` returns the post-commit cleanup (emails to cancel, blobs to delete) to both callers. A mocked test for the organization path should assert that both cleanups run.

## Should fix

**3. Organization erasure deletes other users' payment records unchecked.**
[`calendar-erasure.ts:151-156`](https://github.com/masumi-network/sokosumi/blob/8a67ba626efd20efc98a7211842a8cae21cfd961/apps/core/src/helpers/calendar-erasure.ts#L151-L156) deletes every sweepable `TaskX402Payment` in the workspace, whoever was charged. Tasks owned by former members stay in the organization's workspace, so their VERIFIED or REFUNDED records belong to live users. The account path blocks this exact case with `TASK_X402_PAYMENT_BILLING_OWNER_MISMATCH`. Please either add the same foreign-charge check or record that organization erasure may delete these records on purpose.

**4. Payment blockers no longer page anyone, and the organization-path message can't be acted on.**
- In the account path, a payment with an unknown status now maps to `TASK_X402_PAYMENT_UNRESOLVED` without the Sentry page the old branch sent.
- In the organization path ([`organization-deletion.ts:71-80`](https://github.com/masumi-network/sokosumi/blob/8a67ba626efd20efc98a7211842a8cae21cfd961/apps/core/src/helpers/organization-deletion.ts#L71-L80)), no payment blocker pages anyone, and the message "Resolve it and retry." can't be followed, because only support can resolve a PENDING payment.
- Suggest paging the way `throwPendingX402PaymentDeletionBlocker` does, and telling the user to contact support.

**5. The Postgres erasure test never runs in CI.**
`calendar-erasure.postgres.test.ts` needs `RUN_DATABASE_INTEGRATION_TESTS`, and `.github/workflows/test.yml:138` runs only two named files. CI also builds the schema with `prisma db push`, which doesn't create the new trigger, so the legacy-cascade case would fail there. Please either run it in CI against a database built with `migrate deploy`, or cover the trigger with a SQL assertion test like `calendar-collaboration-migration.test.ts`.

## Minor

- **Lock queries may scan whole tables while holding locks.** The occurrence and link lock queries in `calendar-erasure.ts`, and the affected-workspace and project queries in `user-deletion-tasks.ts`, OR across joined relations. `user-deletion-tasks.ts` already uses UNION arms to avoid this.
- **Deadlocks in the organization path return a 500.** `prepareOrganizationForDeletion` doesn't map P2034 to the retryable `ORGANIZATION_DELETION_CONCURRENT_CHANGE`. Heavier locking makes deadlocks more likely.
- **Web shows a generic error for the new codes.** It doesn't handle `ORGANIZATION_DELETION_TASK_PAYMENT_BLOCKED` or `ORGANIZATION_DELETION_CONCURRENT_CHANGE`, and the deletion preflight can't report the payment blocker. Fine as a follow-up, given the no-schema-change scope.
- **Project-close sync now reads the next task before locking it.** This is at `project-close-sync.service.ts:538-552`. `processSeries` locks again and reloads, so the risk is low.

## Out of scope, worth a follow-up

An organization with any `Job` still can't be deleted, because `Job_workspaceId_fkey` is RESTRICT and the delete returns a raw FK 500. This predates the PR.

## What's solid

- **Lock order is consistent.** Writers and erasure take locks in the same order (actor, then workspace, projects and tasks), and a writer acting as the user being erased blocks at the user row.
- **Organization deletion is now atomic.** Evaluation runs under the organization row lock, and the deletes run sequentially instead of in `Promise.all`.
- **Delete order works with the foreign keys.** Occurrences are deleted before projects, which matters because of the RESTRICT on `sourceProjectId`. Outbox rows go last, so they include the rows the triggers add.
- **The migration can be re-run safely.** It uses `CREATE OR REPLACE` and `DROP TRIGGER IF EXISTS`.

PR #4572, opened 2026-09-14
fix(auth): block email spam with Turnstile

mrosberghaus:
Registration spam can currently trigger paid verification emails without a bot check. This change validates a fresh Cloudflare Turnstile token through Better Auth before signup, email sign-in, verification resends, email changes, magic links, or password-reset requests can send mail. Resend remains the delivery service, and the existing database-backed rate limiter remains active.

A shared, translated verification dialog waits for the check before each request. Cancelling sends no request; expired/reused tokens and provider failures are rejected by Core. CORS and the legacy auth proxy forward the captcha header.

### Deployment

- Configure `NEXT_PUBLIC_TURNSTILE_SITE_KEY` on Web and sensitive `TURNSTILE_SECRET_KEY` on Core for Production and Preview. Both must belong to the same Managed widget, allowing `sokosumi.com` and its subdomains.
- Deploy Web and Core together. Core's secret is optional: omitting it disables verification in any environment. Web requires its site key in deployed environments; local development can omit both keys. CI uses Cloudflare's always-fail test site key only.
- Setup and test-key behavior are documented in `apps/core/README.md`.

### Verification

- `pnpm check` and `pnpm typecheck` passed.
- `pnpm ci:test`: 65 passed. Turborepo dry runs confirm the public site key is automatically inferred and changing it invalidates the Web build cache.
- Focused Core auth/config regression tests: 100 passed, including startup without the secret and signup without a challenge when disabled.
- `pnpm --filter core test`: 5,295 passed; the subsequent deployment-env regression run also passed.
- `pnpm --filter web test --maxWorkers=4`: 5,733 passed, 1 skipped.
- `pnpm build` passed with Cloudflare test keys.
- Real local HTTP: all six protected endpoints return 400 without a token; rejected-token signup returns 403. In-memory Better Auth regression tests verify blocked requests invoke no email callback and valid tokens allow legitimate email actions, including authenticated email changes.
- Browser: signup challenge, failure/cancellation, keyboard focus restoration, light/dark rendering, and 320px layout verified. No paid emails were sent during verification.
- Standards and Spec reviews passed after closing the email-change bypass.

### Screenshot

Failure state using Cloudflare's always-fail test key; the red test-only badge is absent with real widget keys.

Turnstile signup protection


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(auth)/signin/components/form.test.tsx` — sign-in tests mock captcha but do not assert cancel or `x-captcha-response`; magic-link covers the header on success only.
- Minor: `apps/web/src/components/auth-captcha-provider.tsx` — 30s widget timeout sets `failed` but leaves the pending Promise unresolved until Cancel.
- Minor: `apps/core/src/lib/auth-captcha.ts` — `{ id: "captcha" }` no-op when the secret is omitted could be mistaken for an active plugin.
- Skipped / disagreed: add `NEXT_PUBLIC_TURNSTILE_SITE_KEY` to turbo `globalEnv` — `scripts/ci/__tests__/turbo-remote-cache.test.mjs` forbids `NEXT_PUBLIC_*` in hashed/passthrough lists; Next.js inference is the intended cache contract.

### Assessment
Ready to merge: Yes
Server-side Turnstile gates the six email-sending Better Auth routes fail-closed; Web dialog, CORS, proxy, and env split match the PR.

mrosberghaus:
## Code review (three axes) + fix for the one Important finding

Reviewed `1eabd8e...446f17a` along three axes — **Standards** (repo conventions), **Spec** (PR description as spec; no linked Linear issue or spec file found), and a general deep review that ran the suites and audited every email callback against the pinned better-auth 1.7.4 source. The fix below was re-verified on the latest head `9b7a7cdf8` after it landed (the delta itself — comment clarifications plus new edge-case tests — was not put through the three axes). I see `9b7a7cdf8` already resolves one Minor below — nice.

**Verdict: ready to merge once the startup-warning fix lands** (committed locally as `7008792`, push pending — see "Landing the fix").

### Standards — no hard violations

Compliant throughout: pinned `@marsidev/react-turnstile` version; real de/es translations with no message-bag change needed; no direct `process.env` in web code; correct `@/` aliases; legitimate `use client` with `useMountEffect`; colocated tests including failure paths; no fixed-px type; no DB-boundary breach. Three non-blocking judgement calls: the `requestCaptcha` guard repeated in 5 handlers, the identical `vi.mock` block in 4 test files, and the `"x-captcha-response"` / `"auth"` literals in ~8 files (a shared constant in `@sokosumi/utils` would single-source that contract).

### Spec — faithful, no gaps

All six anonymous email flows are gated at both layers; every cancel path aborts before any request; CORS and the legacy proxy forward the header; deployment terms match. One behavior-neutral scope note: the diff renames `isDeployedSokoBotEnvironment` → `isDeployedEnvironment`, which the spec doesn't ask for — but the fix below now uses that helper, so the rename earns its keep. Two confirmed non-issues: a concurrent `requestCaptcha()` resolving `null` (sensible anti-double-submit) and ungated org invitations (authenticated-admin-only, outside the six anonymous flows).

### General deep review — 0 Critical, 1 Important, 3 Minor

- Confirmed against the pinned upstream source that custom `endpoints` replace better-auth's captcha defaults (nothing dropped); audited every email callback (`sendResetPassword`, `sendVerificationEmail` including sign-up/sign-in variants, `sendMagicLink`, `changeEmail`) — all reachable only through gated endpoints, no emailOTP/phone plugins, no bypass at any web call site. Suites, typecheck, and Biome all green at review time.
- **Important → fixed as `7008792`: silent fail-open in production.** Omitting `TURNSTILE_SECRET_KEY` disabled verification in *any* environment with zero log output — a typo'd var looks identical to a protected deploy while spam flows. `validateEnv()` now logs a startup warning when the secret is unset in a deployed environment. Behavior unchanged; 3 new tests (warn / silence-with-secret / silence-in-local-dev). Verified on latest head: applies cleanly, `env` + `auth-captcha` + `auth` + `routes/auth` suites 4 files / 106 tests pass, `biome check` and `tsc --noEmit` clean, red-green confirmed for the new test.
- Minor: captcha rejections surface untranslated server English in the auth forms (legit users hit this mainly during a Cloudflare outage) — consider mapping to translated `AuthErrorCode` copy. Pre-existing pattern, optional.
- Minor (already resolved by `9b7a7cdf8`): the disabled-state stub no longer occupies the real plugin id (`captcha-disabled`). Nothing to do.

### Landing the fix

The fix commit lives in a scratch worktree on this machine, as a direct child of `9b7a7cdf8`:

```
git -C /tmp/pr4572-fix push origin HEAD:codex/auth-turnstile-protection
```

That fast-forwards this PR branch with `fix(auth): warn when Turnstile secret is unset in deployed environments`.

PR #4574, opened 2026-09-14
feat(auth): run the security check inline instead of in a dialog

mrosberghaus:
## Summary

Follow-up to #4572. The Turnstile security check now renders inline in every gated form instead of opening a dialog on submit.

- `interaction-only` widget, pre-rendered on mount, so most visitors never see it and the token is ready before they click. Tokens reset after each use, so a wrong-password retry gets a fresh one silently.
- The hidden widget takes no room in the layout; when Cloudflare asks for interaction the checkbox appears directly above the submit button, in the app's resolved theme and locale.
- Submit waits up to 8 s for a token, then shows an inline load-failure message.
- One `Security Check` analytics event with `entry` and `step` (`interactive`, `solved`, `failed`, `load_error`).
- Dialog, root provider, and the `title` / `description` / `cancel` strings are removed in en/de/es. `CONTEXT.md` gains **Security check**.
- Core endpoint gating is unchanged.

**Behaviour change to review:** the email-verification notice in the sonner toast and the notification dropdown now navigates to `/notifications` (dropdown item labelled "View") instead of sending the email, because neither surface can host the widget. The card on that page keeps the resend button with the inline check.

## Verification

- `pnpm check`, `pnpm --filter web typecheck`, `pnpm --filter web test` (727 files, 5763 tests, message parity ok)
- Browser check with Cloudflare test keys on sign-in, sign-up, forgot-password, and the magic-link row: pass-through key adds no space and issues a token; forced-interactive key shows the checkbox above the button.

## Screenshots

Attached in the conversation: sign-in with the pass-through key (nothing visible), with the forced-interactive key, and the open magic-link row.


mrosberghaus:
## Grok review

### Fixed
- Important: widget `onError` / unsolved 8s timeout showed the load-failure / content-blocker copy. Script errors now report `load_error` once; an interactive timeout uses `missingResponse`; widget errors reset and stay retryable (`6f7015c57` — `apps/web/src/components/auth-captcha.tsx`)
- Important: next-themes hydrating `resolvedTheme` remounted Turnstile (`theme: "auto"` → `light`/`dark`) and dropped the pre-fetched token. The widget mounts only after the theme is `light` or `dark` (`6f7015c57` — same file)
- Important: no `onExpire` handler; marsidev leaves `widgetSolved` true, so a stale tab could reject immediately. Expire now `reset()`s (`6f7015c57` — same file)

### Remaining
- Minor: `apps/web/src/app/(app)/components/account-notice-action.test.ts` — path is asserted; the menu "View" label and toast no longer sending mail are not.
- Minor: `apps/web/src/app/(app)/components/verify-email-button.test.tsx:75` — case is still named "cancelled"; it now covers timeout/load failure.
- Minor: `apps/web/src/app/(auth)/components/social-buttons.tsx` — `useAuthCaptcha("magic-link")` runs even when `showMagicLink` is false.
- Note: marsidev `size: "flexible"` is 65px tall; the `absolute size-0 overflow-hidden` clip is load-bearing. Pass-through test keys still issue a token while clipped; a production managed key can be stricter about visibility.

### Assessment
Inline `interaction-only` check, gated forms, and notice navigation match the PR; leftover notes are test naming and a production-visibility caveat on the clipped widget.

mrosberghaus:
Reply to the Grok review:

**Kept from `6f7015c57`:** split failure copy, polling for the widget ref before the deadline, and `onExpire` resetting the widget.

**Changed in `a9d33bf6e`:** resetting inside `onError` restarted the widget immediately and looped on a persistent error. Measured with Cloudflare's always-block test key: 13 `failed` analytics events in 22 s, then stable after the change. Turnstile already retries on its own (`retry: "auto"`), so the callback now only reports the first failure of a streak.

**On the theme gating:** next-themes computes `resolvedTheme` in its initial client state (`useState(() => a === "system" ? getSystemTheme() : …)`), so the widget never mounted with `theme: "auto"` and no remount dropped a token. The extra `widgetTheme` gate is harmless and stays.

**Remaining notes:** verify-button test case renamed. `useAuthCaptcha("magic-link")` must run unconditionally because it is a hook; the widget only mounts inside the open row. The `size-0` clip matches the library's own invisible-mode box, and Cloudflare's docs have no visibility requirement for `interaction-only`; the widget un-clips before any interactive challenge.

PR #4592, opened 2026-09-15
feat(credits): add spendable wallets and fix the monthly bar

mrosberghaus:
## Summary

Admin, the account chip, and `/you` could show different credit figures for the same organization. Personal leftover, the org pool, extras, and the enterprise contract were all labeled “credits”, and the monthly bar could count empty leftover period buckets.

**User-facing**

- Monthly bar uses the live current-period grant only (no leftover `migrated-member-period` tombstones).
- Chip and `/you` say personal vs organization credits from payload `scope`.
- Admin Users column is **Personal credits**.
- Admin org **Organization credits** is spendable remaining: extras plus the enterprise pool when present (ASK-style leftover FREE is included). Billing still lists the enterprise pool on its own line.

**API (additive `/v1`)**

- New: `scope` (`personal` | `organization`), `spendable` (same number as `credits.total`), top-level `enterprise` (contract + pool, or `null`).
- Kept, OpenAPI-deprecated: `credits` (`total` / `buffer` / `subscription`) and `extra.enterprise`.
- Web and Soko Bot read only the new fields.

Spend rules are unchanged: personal workspace spends personal credits; organization workspace spends the shared org pool for seated members.

## Verification

- `pnpm --filter core test src/helpers/subscription.test.ts src/schemas/user.schema.test.ts src/lib/soko-bot/context-packet.test.ts src/helpers/admin-organization-overview.test.ts`
- `pnpm --filter web test` for chip, `/you`, billing, transformers, credit-usage mapping
- `pnpm check && pnpm typecheck` via husky on each commit
- `pnpm --filter web generate:core:snapshot` after schema changes


mrosberghaus (commented):
## Code review

**Ready to merge? Yes** — all five PR requirements verified against the diff. No Critical or Important issues; a few Minor nits (3 inline, notes below).

### Strengths

- **Additive payload done right.** `spendable: total` is assigned from the same variable (`subscription.ts:441`), so the "same number as `credits.total`" claim is structural. The deprecated `credits` mirror and `extra.enterprise` are built by untouched code paths — old keys keep working byte-for-byte.
- **Monthly-bar fix is mathematically sound.** Traced `currentPeriodBucketContribution` + `summarizeCurrentPeriodSubscriptionBuckets` through the cases: live grant untouched, fully pre-consumed tombstones contribute exactly 0, partially pre-consumed buckets contribute only their true remaining. New remaining/total are weakly ≤ old values, so over-counting is impossible; the `min(periodUsed, lifetimeUsed)` cap handles the inter-query race gracefully.
- **Complete caller migration, verified by grep.** No product-code reads of `credits.total`, `credits.buffer`, or `extra.enterprise` remain outside schema description strings and unrelated domains. Web and Soko Bot all read the new fields.
- **Labels verified against data sources.** Admin Users passes `organizationId: null` (personal — “Personal credits” correct); admin org `totalCredits` is pool-only on enterprise contracts and org balances on self-serve. All three locales updated consistently.
- **Regenerated client consistent** with the schema source by inspection — descriptions match verbatim, transformer additions follow the existing pattern, no hand-edit smell.
- **Good test updates** — tombstone tests encode the exact motivating scenario, self-serve-with-pool-buckets pins `enterprise: null`, chip/`/you`/transformer coverage for the new fields.

### Notes (lines outside the diff — can't anchor inline)

1. **Stale doc pointer** (`apps/core/src/schemas/user.schema.ts:101`): the deprecated `credits.buffer` description still says “(see extra.enterprise for pool)”. Since `extra.enterprise` is now itself deprecated in favor of top-level `enterprise`, point readers at the new field. One-line doc edit + regen.
2. **Hot-path query growth**: every org credits read now adds `resolveOrganizationBillingPlan` (plus `getEnterpriseContractBillingSummary` for enterprise orgs), and `getCurrentSubscriptionCredits` does `findMany` + 2× `groupBy` instead of 2 aggregates. Fine at current scale, but `buildCreditsPayload` runs per Soko Bot turn — worth keeping an eye on.
3. **Expiry filter has no behavioral test** (`subscription.ts:125-129`): the `expiresAt > now` clause is asserted only structurally via `toHaveBeenCalledWith`, because the test `tx` mock ignores `where`. A `where`-respecting fake (or one integration test) would prove expired buckets are actually excluded. Same pre-existing limitation applies to `activatesAt`. Related: buckets expiring mid-period take their this-period usage out of `used` with them — coherent with “live grant only”, just noting the bar can hide entirely at expiry (`total <= 0` → `resolveCreditUsage` returns null).

Note: the test suite was not executed as part of this review (base..head range reviewed read-only), so CI is the final word on green.

apps/core/src/helpers/subscription.ts
```diff
@@ -115,33 +119,72 @@ export async function getCurrentSubscriptionCredits(params: {
       };
 
   const currentPeriodBucketWhere = {
-    AND: [creditBucketActivatesAtOrBefore(now), currentPeriodBucketScope],
+    AND: [
+      creditBucketActivatesAtOrBefore(now),
+      currentPeriodBucketScope,
+      {
+        OR: [{ expiresAt: null }, { expiresAt: { gt: now } }],
```
mrosberghaus: **Nit (unreachable branch):** given the scope already requires `expiresAt: { gt: periodStart, lte: periodEnd }`, the `{ expiresAt: null }` alternative here can never match. Harmless as defensive posture, but a one-line comment would save the next reader the same confusion.

apps/core/src/helpers/subscription.ts
```diff
@@ -345,6 +428,48 @@ export async function buildCreditsPayload(params: {
...
+  if (!params.organizationId) {
+    return null;
+  }
+
+  const billingPlan = await resolveOrganizationBillingPlan(
+    params.organizationId,
+    params.tx,
+  );
+  if (billingPlan.mode !== "enterprise_contract") {
+    return null;
```
mrosberghaus, line 453: **Nit (confirm intent):** the enterprise wallet is now mode-driven — this returns `null` whenever billing mode isn't `enterprise_contract`, even if enterprise pool buckets still exist (e.g. right after a downgrade). Previously `extra.enterprise` (and therefore Soko Bot's `enterpriseRemainingCredits` and the `"enterprise"` plan fallback) was non-null whenever pool buckets existed. The self-serve-with-pool-buckets test pins the new behavior, so this looks deliberate — just confirming Eve ignoring lingering pool buckets in that window is the intent.

apps/web/src/app/(app)/components/sidebar/components/account-summary-types.ts
```diff
@@ -2,11 +2,14 @@ import type { SessionUser } from "@sokosumi/utils";
 import type { MemberWithOrganization } from "@/lib/clients/generated/core";
 import type { CreditUsage } from "@/lib/types/credit";
 
+export type CreditWalletScope = "organization" | "personal";
```
mrosberghaus: **Nit (drift risk):** this re-declares `CreditWalletScope = "organization" | "personal"` instead of deriving it from the generated client (e.g. `GetUsersByIdCreditsResponse["data"]["scope"]`, already imported in `sidebar/index.tsx`). Consider deriving it so a future Core enum extension fails loudly in web instead of silently diverging.

apps/core/src/helpers/subscription.ts
mrosberghaus: Dropped the `{ expiresAt: null }` arm. Period scope already requires `expiresAt` in `(periodStart, periodEnd]`; the remaining clause is only `{ expiresAt: { gt: now } }`.

apps/core/src/helpers/subscription.ts
mrosberghaus: Intent confirmed: top-level `enterprise` is contract-mode-driven. Leftover pool buckets after a downgrade still count in `spendable` (and deprecated `extra.enterprise`). Soko Bot `availableCredits` uses `spendable`; `enterpriseRemainingCredits` is “on a contract”, not “has leftover ENTERPRISE_* rows”.

apps/web/src/app/(app)/components/sidebar/components/account-summary-types.ts
mrosberghaus: `CreditWalletScope` is now `GetUsersByIdCreditsResponse["data"]["scope"]`.

mrosberghaus (commented):
Addressed the nits on this review (commit d3dbddf).

Inline:
- Unreachable `{ expiresAt: null }` — dropped. Period scope already requires `expiresAt` in `(periodStart, periodEnd]`; live filter is `{ expiresAt: { gt: now } }`.
- Mode-driven `enterprise` — confirmed. Top-level `enterprise` is null unless billing mode is `enterprise_contract`. Leftover pool rows after a downgrade still sit in `spendable` (and deprecated `extra.enterprise`). Soko Bot `availableCredits` uses `spendable`; `enterpriseRemainingCredits` means “on a contract”.
- `CreditWalletScope` — now `GetUsersByIdCreditsResponse["data"]["scope"]`.

Notes:
- `credits.buffer` description now points at top-level `enterprise` (regen included).
- Query growth on org credits reads — left as-is.
- Expiry `where` is still only asserted structurally — left as-is.

The review’s “admin org totalCredits is pool-only on enterprise” was true at c1c33f6. Later commit 8af0b58 makes that card extras + pool (spendable).

PR #4593, opened 2026-09-15
feat(web): add slack-style quick reactions to message toolbar

mrosberghaus:
## What changes for users

Hovering a chat message on desktop now shows three one-click reactions ahead of the emoji picker, like Slack. They start as 👍 ❤️ 😂 and then follow the emojis the reader uses most. The mobile long-press sheet shows five from the same list.

- **Stable order:** the three emojis stay in place until the pointer leaves the toolbar, so reacting never moves them under the cursor.
- **Already reacted:** an emoji the reader has already added to this message is tinted and marked `aria-pressed`, so it's clear a click removes it.
- **Tooltip:** shows the shortcode, e.g. `:+1:`, like Slack.
- **Size:** the emojis are one step smaller (`text-sm`) so they match the weight of the outline icons.
- **Hover:** the emoji grows to 1.15× over 100ms on top of the standard background highlight. Only the emoji grows, so the circle keeps its size. Off when reduced motion is on.

## How

- **Emoji history.** The per-browser history is now a log of the last 50 uses, still under the `sokosumi.emoji-picker.recent.v1` key. Existing data is already a valid log, so no migration is needed. `rankFrequentlyUsedEmojis` ranks it by use count, with ties going to the most recent. An emoji you stop using drops out on its own.
- **Shared hook.** `useFrequentlyUsedEmojis` / `recordEmojiUse` (`apps/web/src/hooks/use-frequently-used-emojis.ts`) is a `useSyncExternalStore` hook shared by the picker's "Frequently used" section and the quick reactions. It also picks up changes made in other tabs.
- **One save per use.** The picker only reads the history. Whoever uses the pick saves it once: the message row when a reaction is added (never when one is removed), and the composer. This keeps counts from doubling.
- **Quick reaction list.** `resolveQuickReactions` (`chat/utils/quick-reactions.ts`) takes the ranked list and pads it with the defaults.
- **Layout.** The space kept free for the toolbar on the right of each message grows from `pr-48` to `pr-64`, so the wider toolbar doesn't cover message text.

## Scope notes

- History is per browser. It doesn't sync across devices or to the Apple app.

## Verification

- `pnpm --filter web test`: all 737 files pass. New tests cover the defaults, ranking by use, the order holding under the pointer, the tint for reactions you've already added, shortcode tooltips, "removing doesn't count", the picker not saving picks, and the composer saving them.
- `pnpm check` and `pnpm typecheck` passed in the pre-commit hook.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/lib/utils/emoji-shortcodes.test.ts` — `rankFrequentlyUsedEmojis` case is 3/2/1 counts, not a tie; the test name overclaims recency-on-tie.
- Minor: `apps/web/src/app/(app)/chat/components/room-message-row.tsx` — a second click while the parent request is in-flight can log two uses for one add (ranking noise only).
- Minor: `apps/web/src/app/(app)/chat/utils/quick-reactions.ts` — comment still says “most recently used”; the list is already frequency-ranked and this helper only pads defaults.
- Minor: `apps/web/src/app/(app)/chat/components/thread-panel.tsx` — `reserveHoverActionGutter={false}` was already there; the wider pill can cover more text in the thread column.
- Minor: `room-message-row.tsx` — `pr-64` has ~4px slack on the busiest desktop pill; a `SokoBotChainBadge` on the same pill can still overlap.

Skipped / disagreed: none. All five are nits; none were one-line fixes in a file already being edited.

### Assessment
Ready to merge: Yes
Frequency log, shared hook, single writer, freeze-on-hover, pressed state, shortcode titles, reduced-motion scale, and gutter change match the PR and are covered by tests.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/4593#issuecomment-5678691160). Fixed in cd4b89434.

- **Ranking test name:** confirmed, the old case had counts 3/2/1 and no tie. It's now "ranks the most used first", plus a new case with an actual 2/2/1 tie.
- **Double count while a toggle is in flight:** confirmed. `rooms-client` ignores the second click, but the row saved the use before handing the click over. The row now counts at most one use per emoji until the message's reactions change. New test: two clicks during one pending toggle save one use.
- **`resolveQuickReactions` comment:** it now says the input is the already-ranked frequently used list.
- **Wider toolbar on Soko Bot chain messages:** confirmed and measured. The chain badge plus 7 buttons comes to about 276px, against the 256px (`pr-64`) kept free. Before this PR it was 186px against 192px, so this PR caused the overlap. Rows whose toolbar shows the chain badge now keep `pr-72` (288px) free. Other rows keep `pr-64`, where the busiest toolbar (8 buttons) is 252px. New test covers it.
- **Thread column overlap: no change.** The thread panel sets `reserveHoverActionGutter={false}` on purpose, so replies can use the full width of the narrow column. The toolbar already covered text there, and it only appears on hover. Reserving about 256px in that column would cost much more reading width than the brief overlap does.

Verification: `pnpm --filter web typecheck` is clean, and all affected chat, emoji, composer and hook tests pass (162 files, 1468 tests).

PR #4598, opened 2026-09-15
fix(web): stack week calendar tasks full width and default to week view

enjojoy:
## Summary

The Schedules week view squeezed concurrent tasks into narrow side-by-side lanes in FullCalendar's time grid, so tasks scheduled at the same time were unreadable (markers only, titles clipped).

- Switch the week view from `timeGridWeek` to `dayGridWeek`: every task is a full-width stacked row and the day column grows to fit them all.
- Show each week event's start time before the task name, since the hour axis is gone.
- Default the calendar to the week view when the URL has no `view` param, and offer the week switch on mobile as well so the active view always has a highlighted button.
- Remove the time-grid-only code: hover slot highlight and pointer tracking, the `timegrid` plugin, `allDaySlot` / `slotEventOverlap`, and the week-only transparent event CSS.

Behaviour note: clicking an empty week cell now creates a task at midnight of that day (same as month view) instead of at the hovered hour.

## Verification

- `pnpm --filter web test "src/app/(app)/calendar"` — 101 tests pass
- `pnpm check && pnpm typecheck` — clean (pre-commit hook)
- Visual check pending on the Vercel preview


enjojoy:
@coderabbitai review

PR #4610, opened 2026-09-15
fix(web): render dates in the viewer's time zone

mrosberghaus:
## What

Dates in Task detail (Created / Updated / Next run / Upcoming), project headers, `TimeAgo` fallbacks and schedule occurrence rows rendered in **UTC**. A viewer in Germany saw `Sep 15, 2:50 PM` for a run due at 16:50 CEST.

## Why

#3042 pinned every short date to UTC (`HYDRATION_STABLE_TIME_ZONE`) so server and client markup would match. That fixed the hydration mismatch by showing everyone the wrong clock.

## How

- The browser records its IANA zone in a `sokosumi.timezone` cookie on first visit (`TimeZoneSync`) and refreshes once.
- next-intl request config reads that cookie as `timeZone`, so `getFormatter` / `useFormatter` / `useTimeZone` all agree on the server and the client.
- `formatShortDate` / `formatShortDateTime` take an explicit `timeZone`. Server pages pass `getTimeZone()`, client code passes `useTimeZone()`. A missing value still falls back to UTC so hydration stays deterministic.
- Schedule occurrences keep the zone the rule was captured with; only legacy rows without one fall back to the viewer's zone.

First page load without the cookie still renders UTC, then refreshes to local time once the cookie exists.

## Follow-up consolidation (second commit)

- Admin x402 payments, the public share page, passkey settings, API key rows and authorized clients formatted dates with hand-rolled `Intl.DateTimeFormat` / `toLocaleDateString`. They now use the next-intl formatter, which carries the viewer zone.
- Browser zone detection collapses onto `getDefaultTimezone()` (TimeZoneSync, Soko Bot schedule form, task schedule section).
- Left alone on purpose: chat message times (already client-gated to the browser zone), admin quality day buckets (UTC by design), schedule cron/zoned helpers (explicit zone), and date-only pickers.

## Verification

- New regression test: `2026-09-15T14:50:00Z` → `Sep 15, 4:50 PM` in `Europe/Berlin` (red before, green after).
- `pnpm --filter web test`: 738 files / 5899 tests pass.
- Not verified in a live browser (no dev server for this worktree).


mrosberghaus:
## Grok review

### Fixed
- Important: `TimeZoneSync` refreshed the root layout whenever the browser zone disagreed with the request zone, without checking that `sokosumi.timezone` stuck. Android WebViews that report `Factory` (rejected by `Intl`) and browsers that ignore `document.cookie` left the server on UTC, so the mismatch never resolved. `getDefaultTimezone()` now returns `UTC` for invalid IANA ids; `TimeZoneSync` skips `router.refresh()` if the cookie is absent after the write (`55ea7a467` — `apps/web/src/i18n/time-zone-sync.tsx`, `apps/web/src/lib/schedules/timezones.ts`).
- Important: no coverage of that sync path. Added client tests for write+refresh-once, cookie-blocked, and already-matching zone, plus a `Factory` → `UTC` unit test (`55ea7a467` — `apps/web/src/i18n/time-zone-sync.test.tsx`, `apps/web/src/lib/schedules/timezones.test.ts`).
- Minor (while in the same files): dropped the local `isValidTimeZone` copy; request + browser paths both use `@sokosumi/utils` `isValidTimezone`.

### Remaining
- Minor: `apps/web/src/i18n/time-zone.ts`: timezone cookie omits `Secure` on HTTPS; locale/consent cookies add it. Wrong clock if rewritten, not session theft.
- Minor: `apps/web/src/lib/utils/datetime.ts`: `timeZone?` on `formatShortDate` / `formatShortDateTime` still falls back to UTC with no type error. Spec for first paint; a required arg would catch the next silent UTC regression.
- Minor: `shared-task-view.test.tsx` / `passkey-settings.test.tsx` mock UTC, so they do not prove the viewer zone. `datetime.test.ts` is the contract.
- Skipped / disagreed: first-visit UTC flash, then local after cookie — that is the spec, not a bug. Chat times, admin quality day buckets, and cron/zoned helpers were left on purpose.

### Assessment
Viewer-zone formatting matches the PR (Berlin `14:50Z` → `Sep 15, 4:50 PM`); the refresh loop is closed. Leftover notes are cookie `Secure` and optional formatter args.

mrosberghaus:
Follow-up to the Grok review, in `ae5d7c5df`:

- **Cookie `Secure` flag:** `sokosumi.timezone` now uses the same `browserCookieFlags()` as the locale cookie (exported from `i18n/locales.ts`), so it is `Secure` on real HTTPS hosts.
- **Required `timeZone`:** `formatShortDate` / `formatShortDateTime` take a required `timeZone: string`. Server callers pass `getTimeZone()`; the two hook consumers (`TimeAgo`, `useLocalizedDateTime`) spell out `useTimeZone() ?? DEFAULT_TIME_ZONE`. A forgotten zone is now a type error.
- **UTC mocks in `shared-task-view` / `passkey-settings` tests:** left as is. Those tests cover component behaviour; the viewer-zone contract is asserted in `datetime.test.ts` and `time-ago.test.tsx` (Berlin case).

Full web suite, typecheck and Biome are green.

PR #3300, opened 2026-07-14
feat(vendor-grants): workspace grants, GRANT_PENDING parking, and Serviceplan auto-grant

mrosberghaus:
## Summary

Introduces workspace-scoped vendor grants for personal and organization workspaces, a first-class **`GRANT_PENDING`** task status for delegated creates that wait on approval, and automatic **Serviceplan** workspace access for new and existing workspaces.

Each vendor has one **workspace** grant per workspace (read, comment, and create are all-or-nothing). Core enforces access; the web app surfaces grant management, approval banners, and kanban placement for blocked tasks.

### Vendor workspace grants

- New `vendor_grant` table: `permission: workspace`, statuses `PENDING` / `GRANTED` / `DENIED` / `REVOKED`, unique on `(vendorId, workspaceId)`.
- Org and personal APIs: list, proactive create (`{ vendorId }`), approve, deny, revoke.
- Coworker access beyond baseline upserts a pending grant and returns `403 grant_required` for read/comment when access is missing.
- Grant requests persist in a dedicated transaction so route rollbacks do not drop pending upserts or approver notifications.
- Gated Postgres integration tests cover concurrent grant create races and approve/unpark serialization (`RUN_DATABASE_INTEGRATION_TESTS=true`).

### GRANT_PENDING task parking

Parked delegated creates no longer rely on `pendingApproval` / exposed grant IDs on the task API.

| Concern | Behavior |
| --- | --- |
| Active status | `GRANT_PENDING` |
| Resume target | `grantResumeStatus` (`DRAFT` or `READY`) on task API while pending |
| Grant link | `pendingVendorGrantId` exposed on task API while `GRANT_PENDING` ([in-repo doc](https://github.com/masumi-network/sokosumi/blob/main/docs/coworker/vendor-workspace-grants-api.md)); cleared on unpark/cancel |
| On approve | Transition to `grantResumeStatus`; emit task event with resume status |
| On deny/revoke | `GRANT_PENDING` → `CANCELED`; emit cancel event |
| Coworker events | Create shows `GRANT_PENDING`; approve/deny visible via `/coworkers/me/events` polling |

List filtering uses `status=GRANT_PENDING` (removed `pendingApproval` and `includeParkedReady` query params).

### Serviceplan auto-grant

- On **new workspace create** (personal or org), Core grants the seeded Serviceplan vendor (`slug: serviceplan`) with status `GRANTED`.
- Create hook is idempotent and **does not overwrite** existing grant rows (preserves explicit deny/revoke).
[...]

PR #4611, opened 2026-09-15
feat(core): restore agent access to scheduling

enjojoy:
## Summary

Restores agent access to scheduling and removes the workspace-grant approval round-trip from scheduled creation, without changing scheduling behavior.

**Schedule mutations regain their former access model**
- `PUT/DELETE /v1/tasks/{id}/schedule`, `PUT /v1/tasks/{id}/calendar-schedule`, `PUT /v1/tasks/{id}/calendar-source`, and `PATCH /v1/tasks/{id}/schedule/occurrences` admit agent actors through the existing Task collaboration gate (`requireTaskCollaboration`) instead of the human-only owner guard.
- `GET /v1/tasks/{id}/schedule/occurrences` admits agents through the same Task read gate as `GET /tasks/{id}`; human owners keep ownership-scoped reads, and reads stay free of the parked-task guard.

**Scheduled creation loses the approval round-trip**
- `POST /v1/tasks/scheduled` no longer requires a GRANTED-only check and no longer commits a GRANT_PENDING request. Coworker contexts now go through the standard authorized user binding (`requireAuthorizedUserContext`): a GRANTED workspace grant **or** a baseline assignee/sibling task relationship with the contextual user, plus the `tasks` capability. The request-grant helper (`requireScheduledTaskCreatorOrRequestGrant`) and the gate type are deleted.
- The endpoint still never parks work and never creates `GRANT_PENDING` rows: the Task, v2 rule epoch, and occurrence index are created atomically.
- `GET /v1/workspaces/calendar/sources` reports scheduling availability from the same binding + capability check instead of grant state.

**Cross-cutting**
- Calendar beta and organization-seat checks follow the effective user whenever one exists (session, contextual coworker via `X-Context-*`, or Soko Bot). Standalone coworker keys have no user context and are scoped by Task collaboration only. New helper: `resolveUserContext`.
- Task event attribution now uses the shared `resolveTaskEventActorFields` helper (also reused by `POST /tasks/{id}/events`); occurrence `actorUserId` is set only for user actors.
- Docs updated: `docs/coworker/vendor-workspace-grants-api.md`.

## Post-mortem: how scheduling access evolved

| When | Change | Author |
| --- | --- | --- |
| Jun 24, 2026 | Task scheduling ships: `PUT/DELETE /tasks/{id}/schedule`, owner-scoped; a contextual coworker can schedule the owner's task (#3237). | Francis |
| Jul 13, 2026 | Schedule routes move to `requireTaskCollaboration`: an assigned coworker key can edit/remove its task's schedule (#3289). | Andreas |
| Jul 23, 2026 | Owner-user guard lands on schedule delete (#3394). | Andreas |
| Aug 28, 2026 | Owner-user guard plus organization seat land on schedule put (#3995). Agent schedule mutations end. | Andreas |
| Sep 8, 2026 | Calendar launch (#4126): atomic `POST /tasks/scheduled` replaces the web's create-then-schedule two-step; new Calendar routes are human-only per the SOK-871 spec; the coworker creation path is born with a hard `GRANTED` vendor-workspace-grant gate — chosen because a scheduled Task is an immediate workspace-wide `QUEUED` commitment that cannot park like `POST /tasks`. | Us |
| Sep 15, 2026 | #4588: a missing grant commits a PENDING request and returns 403 `grant_required`; the caller retries after human approval. | Us |
| **This PR** | Agents get schedule mutation access back (Task collaboration), and scheduled creation drops the approval round-trip in favor of the standard coworker-user binding. | Us |

Where the pieces came from: the vendor workspace grant mechanism itself (`GRANTED`/`GRANT_PENDING` parking, `vendor-grants.ts`, the coworker grants doc) is Andreas's #3300. The scheduled-creation gate and its request flow are ours (#4126, #4588). Scheduled creation never had an ungated past: before #4126 the endpoint did not exist, and the equivalent two-step flow was gated by regular task access plus owner-only schedule edits. So opening creation beyond GRANTED is an explicit product decision, not a revert.

## Decisions

1. **Mutations restore, creation opens.** Existing schedule routes return to the Task collaboration model agents had Jul 13 – Aug 28. Creation is opened beyond GRANTED-only: a vendor already bound to the workspace (GRANTED grant or baseline task) creates scheduled work without any approval round-trip.
2. **User-scoped gates follow the effective user.** Calendar beta and organization seats apply whenever a user context exists, so agents cannot bypass an organization's billing or rollout gate on behalf of a user. Standalone coworker keys have no user to gate.
3. **Delete, don't bypass.** The grant gate, its request helper, and its tests are removed rather than short-circuited, so there is one creation path.
4. **No new impersonation hole.** Removing the grant gate alone let any `tasks`-capable vendor key assert an arbitrary `X-Context-User-Id` and create work owned by that user. Coworker contexts therefore pass through `requireAuthorizedUserContext` (GRANTED grant or baseline task), the platform default for user-scoped coworker access. That is a binding on whose behalf the vendor may act, not an approval round-trip.

## Verification

- `pnpm --filter core test` — 5431 passed
- `pnpm typecheck`
- `pnpm check`

## Notes

- Human behavior is unchanged: session actors keep the same gates, seat, and attribution as before.
- Scheduled creation still requires a contextual workspace user; standalone coworker keys cannot create (same as `POST /tasks`).
- Coworker-user binding applies to creation. Schedule mutations stay assignment-scoped (`requireTaskCollaboration`: the Coworker must be the Task's assignee), which is the platform rule for Task collaboration, not a grant check.
- Review follow-ups: CodeRabbit's authorization-bypass finding is fixed in `e3372f9`; its docs note, the pre-merge two-axis review, and Andreas's helper questions are addressed (unreachable agent-only branch removed, occurrence actor fields resolved once, Calendar gate docs split per route, `resolveUserContext` documented in the Core actor menu, direct unit tests for `requireTaskScheduleReadAccess`). Branch is rebased on current `main` (`ee2b19688`).


enjojoy:
@coderabbitai review

apps/core/src/middleware/auth.ts
```diff
@@ -266,6 +267,26 @@ export function requireUserContext(
...
+export function resolveUserContext(
+  authContext: AuthenticationContext,
+): UserContext | null {
+  if (isCoworkerAuthContext(authContext) && !authContext.context) {
+    return null;
+  }
+
+  return requireUserContext(authContext);
+}
+
```
mrosberghaus, line 289: @enjojoy Are you sure this is needed? We’re doing similar things in many other pull requests.

apps/core/src/helpers/task-event-actor.ts
```diff
@@ -0,0 +1,39 @@
...
+      sokoBotId: authContext.sokoBotId,
+    };
+  }
+
+  return {
+    userId: null,
+    coworkerId: requireCoworkerAuthContext(authContext).coworkerId,
+    sokoBotId: null,
+  };
+}
```
mrosberghaus, line 39: @enjojoy Same question here... We already handle different actors in other parts of the application.

apps/core/src/middleware/auth.ts
enjojoy: It is needed for the restored access, but happy to move it if you prefer another home.

The schedule routes gate Task access through `requireTaskCollaboration`, which intentionally accepts a standalone coworker key (assigned task, no `X-Context-*`). On top of that they still apply user-scoped gates (Calendar beta, organization seat) — but only when the actor actually has an effective user. The alternatives I considered:

- `requireUserContext`: throws for a standalone coworker key, which would drop the assigned-key schedule access this PR restores.
- `requireAuthorizedUserContext`: belongs to user-scoped routes; these are Task-collaboration routes and it also rejects standalone keys.
- Branching on `authContext.actor` in each of the five handlers, which the handler actor menu explicitly asks us not to do.

So `resolveUserContext` is the single place that says "effective user when one exists, `null` for a standalone coworker key". If you'd rather it live somewhere else (or be named differently), tell me and I'll move it.

apps/core/src/helpers/task-event-actor.ts
enjojoy: Not a new parallel path — it is the consolidation. `POST /tasks/{id}/events` had three near-identical local mappers (`getStatusEventActorData`, `getCommentEventActorData`, `getAgentActorData`); they are now this one shared function, and the four schedule routes needed the same mapping, so sharing it avoided four more copies. The existing `eventActorFields` in `task-domain.service.ts` takes a `TaskDomainActor` (not an `AuthenticationContext`), so it cannot be reused here. Happy to rename or move it if you have a preferred home.

apps/core/src/services/task-schedule-create.service.ts
```diff
@@ -135,22 +129,20 @@ function coworkerScheduledTaskCreator(
...
- * workspace access, or request it for the contextual user when it is missing.
+ * Resolves the creator of a scheduled Task: the session or contextual user, or
+ * the Coworker acting for them. Coworkers need the `tasks` capability, exactly
+ * like any other Task they create; scheduled creation has no workspace-grant
+ * approval gate.
  */
-type ScheduledTaskCreatorGate = "require-granted" | "request-grant";
-
-async function resolveScheduledTaskCreator(
+export async function requireScheduledTaskCreator(
```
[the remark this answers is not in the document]
enjojoy: Fixed in e3372f9. `requireScheduledTaskCreator` now resolves coworker contexts through `requireAuthorizedUserContext` (GRANTED workspace grant or baseline assignee/sibling task) before the `tasks` capability check; no approval round-trip is reintroduced. Added a service test for the rejected unbound context.

docs/coworker/vendor-workspace-grants-api.md
```diff
@@ -172,10 +170,11 @@ returns the original Task without creating another occurrence ledger.
...
 | POST | `/v1/tasks` | Delegated create flow above. |
-| POST | `/v1/tasks/scheduled` | Atomic v2 scheduled creation; no grant → creates a PENDING request (approvers notified) and returns **403** `grant_required`; PENDING → **403** `grant_required`; DENIED/REVOKED stay blocked; never parks work. |
+| POST | `/v1/tasks/scheduled` | Atomic v2 scheduled creation; not workspace-grant gated. Coworker needs the `tasks` capability and a contextual user; Calendar beta / seat checks apply to that user; never parks work. |
 | POST | `/v1/tasks/{id}/events` | **`GRANT_PENDING`** → **403** `task_parked`. |
 | POST | `/v1/tasks/{id}/jobs` | Parent **`GRANT_PENDING`** → **403** `task_parked`. |
 | PATCH | `/v1/tasks/{id}` (+ schedule, etc.) | Collaborators cannot mutate parked tasks. |
+| PUT/DELETE/PATCH | `/v1/tasks/{id}/schedule`, `/calendar-schedule`, `/calendar-source`, `/schedule/occurrences*` | Coworker collaborators may edit, remove, or move schedules and occurrences. Calendar beta / organization-seat gates follow the effective user (session, contextual coworker, or Soko Bot); standalone coworker keys skip them. |
```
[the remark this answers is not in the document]
enjojoy: Fixed in e3372f9. The table now splits legacy `PUT /tasks/{id}/schedule` (no Calendar beta; seat applies to the effective user), `DELETE /tasks/{id}/schedule` (neither gate), and the Calendar-specific mutation routes, plus a note that standalone coworker keys skip user-scoped gates.

apps/core/src/middleware/auth.ts
mrosberghaus: Okay for now, but we need some consolidation. We have too many single-use-case helpers that mostly do the same thing.

PR #4614, opened 2026-09-15
fix(apple): honor shared chat read cooldown

mrosberghaus:
When Core returns a chat-read 429, Apple currently discards retry timing and other readers continue issuing requests. This change shares one cooldown across room, thread, search, message, pin and room-list reads. It honors Retry-After or the body delay, uses a five-second fallback for invalid timing, and adds one-sided jitter when resuming.

The cooldown uses the existing generated-client middleware stack and lives in the portable SokosumiChat package. OAuth's existing session generation prevents an old login's waits or responses from affecting the next login; token refresh preserves the cooldown. Stopping a recovery scheduler now cancels its active task. Writes and realtime/stream delivery bypass the gate. A throttled response still reaches the existing error handling; it is not replayed automatically.

Web reference: [shared cooldown](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/lib/chat/chat-read-throttle.ts), [response handling](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/fetch-background-json.ts), [refresh scheduling](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/use-chat-refresh-scheduler.ts). This completes PARITY.md slice 07b. No web files, generated sources, dependencies or Core contracts changed. Server limits, new chat features and UI redesign are outside scope.

Validation:
- All five Swift package suites passed: Chat 374, Auth 35, Workspace 73, CoreAPI 1, Realtime 27. After adding two further race cases, all 11 cooldown tests passed separately.
- Full Xcode app test suite and macOS build passed; generic iOS 17 Workspace build passed.
- Strict SwiftLint, SwiftFormat, git diff check and normal Biome/typecheck commit hooks passed.
- Tests cover actual generated-client room-to-thread sharing, header/body precedence, malformed fallback, extending an active wait, session changes, stale completions, cancellation before transport, and write/stream bypass.
- No production account was deliberately throttled. Timing tests use controlled clocks and fake HTTP transport. Test-host apps were closed.

To test: run `swift test --package-path apps/apple/Packages/SokosumiChat --filter ChatReadCooldownTests`, then `swift test --package-path apps/apple/Packages/SokosumiChat --filter ChatRefreshSchedulerTests`. Both exercise the cooldown and stopping an active delayed read without waiting real server cooldown intervals.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/ChatReadCooldown.swift:36-41` — per-chunk jitter on 60s slices can cancel itself on long delays; web jitters the full remaining window once. Dormant for Core’s current ~1s refill.
- Minor: `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/ChatReadCooldownMiddleware.swift:24-28` — operation-ID `Set` rebuilt on every intercept; hoist to `private static let`.
- Minor: `apps/apple/Packages/SokosumiChat/Tests/SokosumiChatTests/ChatReadCooldownTests.swift` — all clocks inject `jitter: { 0 }`; no assertion that production one-sided jitter lengthens the wait.
- Minor: `apps/apple/PARITY.md:6` vs `:11` — resume block says 07b is in review; “Next implementation” still points at 07b.
- Skipped / disagreed: none of the above are bugs in the cooldown path; left for a follow-up.

### Assessment
Shared middleware cooldown matches slice 07b (Retry-After then body, 5s fallback, session generation, scheduler cancel, writes/stream bypass); remaining items are jitter effectiveness on delays Core does not currently send, plus small cleanup.

mrosberghaus:
Verified the four minor findings against the implementation and web's `chatReadThrottleResumeInMs`. Addressed them in 292a7db83.

- Long waits now apply jitter once to the full remaining window, then sleep in bounded chunks. The shared deadline is checked again before the request resumes; cancellation and session checks remain between chunks. A 300-second delay with 25% jitter previously waited exactly 300 seconds; it now waits 375 seconds.
- Added deterministic nonzero-jitter regression coverage for 1-, 60-, and 300-second delays. The long-delay case failed before the fix and passes afterward.
- Hoisted the read operation allowlist to `private static let`.
- Corrected PARITY.md: 07b is the active PR; quick reactions (20b) is next after merge.

Validation: all 377 SokosumiChat tests pass, strict SwiftLint reports zero violations, SwiftFormat is clean, and the macOS Xcode build succeeds. Repository pre-commit checks also pass. CI is rerunning on the new commit.

No Critical/Important fixes were pushed by the reviewer, so there was no separate reviewer patch to validate.

PR #4618, opened 2026-09-15
feat(web): add 12-hour/24-hour time format preference

mrosberghaus:
## What

Users can pick how times are written: **Account → Preferences → Time format** offers **Auto**, **12-hour** and **24-hour**. An English-UI user in a 24-hour region now sees `Sep 15, 16:50` instead of `Sep 15, 4:50 PM`. Language still drives month names, order and separators.

Linear: https://linear.app/masumi/issue/SOK-1081/time-format-preference-12-hour-24-hour

## How

- **Storage.** The explicit choice lives in the `sokosumi.timeformat` cookie (`12h` / `24h`; absent = Auto). `TimeFormatSync` records the browser's own hour cycle in `sokosumi.timeformat.auto` and refreshes once, the same way `TimeZoneSync` handles the zone. Switching back to Auto just deletes the explicit cookie. Both cookies use `browserCookieFlags`.
- **Resolution.** `resolveRequestHourCycle` picks explicit → detected → none. None means Intl's language default, which is today's behaviour.
- **Named formats.** `createFormats(hourCycle)` in `src/i18n/time-format.ts` is the request's next-intl `formats`: `time`, `timeWithSeconds`, `dateTime`, `dateTimeWithYear`, `dateTimeShort`, `dateTimeMedium`, `dateTimeWithSeconds`. The root provider inherits them, so server and hydrated client render the same string.
- **Call sites.** About 60 labels that show a clock time now name a format instead of passing `{ hour, minute }` / `{ timeStyle }` options. That covers task and project headers, schedule next run and occurrences, `TimeAgo`, chat and Soko Bot timestamps, Drive, jobs, refunds, billing, the share page and admin tables. Date-only labels keep ad-hoc options.
- **Preferences UI.** A new card sits beside Language. The Auto option shows what it resolves to in this browser, e.g. "Auto (24-hour)". Copy is added in en, de and es.

## Deviations from the spec, on purpose

- **Guard is a lint rule, not a type error (story 23).** Format names type-check through next-intl's `AppConfig`, so `"notAFormat"` fails `tsc`. next-intl's public `dateTime(value, options)` overload can't be narrowed by types, though. Instead, a Biome GritQL plugin (`apps/web/biome-plugins/named-clock-formats.grit`) flags two things: clock options written inline in `.dateTime(…)`, and `const` option objects carrying `hour` / `timeStyle` / `hourCycle`. It does not see raw `Intl.DateTimeFormat` or `toLocale*String` calls.
- **`formatShortDateTime` is removed rather than given an hour-cycle argument.** Every caller moved to the `dateTime` format, and a client component can't read the resolved hour cycle outside next-intl's formats. Keeping the helper would have meant a second place deciding the hour cycle. The Berlin `14:50Z` contract now lives in the `TimeAgo` and share-page tests.

## Changes outside the core ask

- `TimeAgo` takes its locale from next-intl instead of a `locale` prop. Its absolute fallback already uses the provider's locale, and a second locale source would disagree. Side effect: the ratings list and the members table now show relative times in the UI language rather than always in English.
- Formats use a numeric hour. Drive, chat, Soko Bot activity and personal-assistant timestamps used to zero-pad it. 12-hour labels there now read `4:05 PM` instead of `04:05 PM`.
- The schedule editor's next-runs preview used `toLocaleString(undefined, …)` in the browser locale. It now uses `dateTimeMedium`.

## Left alone, per the spec's out-of-scope list

- Schedule **rule** labels: recurring ("Daily at 9:00 AM", from `src/lib/schedules/cron.ts`), the one-time footer label, and the editor presets. They keep their own formatting for now, and the one-time label carries an explicit `biome-ignore` saying so. Next to a 24-hour "Next run", a recurring label can still read 12-hour. That calls for a follow-up that moves all rule labels together.
- Admin analytics day buckets (UTC by design), date-only inputs and pickers, and FullCalendar's built-in labels.

## Verification

- TDD seams per the spec:
  - resolver unit tests
  - `TimeFormatSync`: write + refresh, stale rewrite, cookie didn't stick, already in sync, explicit choice outranks Auto
  - Preferences: 24-hour writes the cookie and reloads; Auto deletes it and names its resolved value
  - provider seam: `TimeAgo` and schedule occurrences render `16:50` vs `4:50 PM` for the same instant
  - server path: the share page, built with the real `createFormatter`
- `pnpm --filter web test`: 744 files / 5930 tests pass.
- `pnpm check` and `pnpm typecheck` (11 packages): clean. The new Grit rule is clean on the tree.
- **Not verified in a live browser** (no dev stack in this worktree). Screenshots of the Preferences card are still needed.


mrosberghaus:
## Grok review

### Fixed
- Important: Preferences select stayed on Auto after a real reload because SSR `useState` could not read `document.cookie`. The account page now passes the cookie into the select (`156f01782` — `apps/web/src/app/(app)/account/page.tsx`, `preferences-section.tsx`).
- Important: test now asserts a server-passed `24h` value shows as selected (`156f01782` — `apps/web/src/app/(app)/account/components/preferences-section.test.tsx`).

### Remaining
- Minor: `apps/web/src/app/(app)/account/components/passkey-settings.test.tsx` and some calendar tests still mock `useFormatter` as if the second argument were Intl options. Named format strings then spread into `DateTimeFormat` and those tests skip the clock they think they cover.
- Minor: `apps/web/src/i18n/time-format-sync.tsx` only runs on mount, so Auto does not follow an OS clock change in a long SPA session. Same limit as `TimeZoneSync`; a reload is enough.
- Minor: first visit can fire two `router.refresh()` calls (`TimeZoneSync` + `TimeFormatSync`). Next may batch them.
- Minor: `detectBrowserTimeFormat` maps any non-`h23`/`h24` `hourCycle` to `12h`, including a missing cycle when `hour12 === false`.
- Skipped / disagreed: none of the Critical/Important findings. Schedule rule labels staying 12-hour is the PR's stated out-of-scope follow-up.

### Assessment
Hour-cycle path (cookies → request formats → named labels) matches SOK-1081; the Preferences control now restores the stored choice after reload. Leftover formatter mocks and Auto-on-OS-change are polish, not blockers. Live-browser screenshots of the Preferences card are still missing, as the PR said.

mrosberghaus:
Follow-up on the [Grok review](https://github.com/masumi-network/sokosumi/pull/4618#issuecomment-5683040543):

- **Formatter mocks.** Fixed in 0cdd3239b. `passkey-settings.test.tsx` and the three `workspace-calendar` tests now build `useFormatter` from `createTestFormatter`, so a named format renders its clock. Constant stubs (`() => ""`, `toISOString()`) never read the second argument, so I left them alone.
- **Select after reload (156f01782).** Kept. The server passes the stored choice, so SSR and hydration agree.
- **Auto re-detects only on mount.** Unchanged on purpose. `TimeZoneSync` behaves the same way, and a reload picks up a new browser language or OS clock setting (story 21). Browsers fire no event for an OS clock change.
- **Two refreshes on first visit.** Unchanged. It happens only on a first visit without cookies, costs one extra RSC request, and the last refresh carries both cookies.
- **Missing `hourCycle`.** Unchanged. ECMA-402 reports `hourCycle` whenever the format has an hour field, and every browser in Next 16's default targets (Chrome, Edge and Firefox 111, Safari 16.4) implements it, so the `hour12` fallback would be dead code.

Verification after merging main: `pnpm --filter web test` passes 745 files / 5934 tests, and `pnpm check` and `pnpm typecheck` are clean.

PR #4620, opened 2026-09-15
feat(apple): add quick reactions to message toolbar

mrosberghaus:
Adds three frequent emoji shortcuts before React in the native message toolbar. Empty history uses 👍, ❤️ and 😂; existing history fills the slots without duplicates. Choices stay fixed while interacting, and own reactions remain highlighted and toggle off on another click.

Matches [web shortcut selection](https://github.com/masumi-network/sokosumi/blob/2a7a20bf4/apps/web/src/app/%28app%29/chat/utils/quick-reactions.ts) and [web message actions](https://github.com/masumi-network/sokosumi/blob/2a7a20bf4/apps/web/src/app/%28app%29/chat/components/room-message-row.tsx). Reuses the existing native history, reaction coordinator, picker and error handling. Narrow thread panes use icon-only controls with accessibility labels. No web, API, generated-source or dependency changes. Account-synced preferences, custom emoji and mobile toolbar design are out of scope.

Validation:
- All five Swift package suites pass, including 378 Chat tests and new history/default/deduplication coverage.
- Full macOS app tests, macOS compilation, iOS 17 Workspace build, strict SwiftLint, SwiftFormat and repository pre-commit checks pass.
- Local fixture using the production row verified add/remove, selected state, stable choices, history on re-entry, and wide/compact picker presentation. Task-launched apps closed.
- Physical Tab/Space navigation and pointer hover animation remain unverified: automation could activate accessibility buttons but could not deliver those inputs to the fixture.

How to test: hover a room/thread message, click a shortcut twice, choose another emoji in React, leave and re-enter the row, and narrow the thread pane. Check keyboard focus, hover feedback and light/dark appearance.

Light toolbar
Compact dark toolbar


mrosberghaus:
## Grok review

### Fixed
- Important: record emoji history only when adding a reaction, matching web `handleToggleReaction` (`be223e947` — `apps/apple/Sokosumi/Chat/Timeline/MessageRowView.swift`)

### Remaining
- Minor: `MessageRowView.swift`: native uses `accessibilityValue` for own-reaction state; web uses `aria-pressed`. `.accessibilityAddTraits(.isSelected)` would match VoiceOver toggle semantics.
- Minor: `MessageRowView.swift`: reaction popover is attached to the whole toolbar pill, not the React control; arrow may look off-center.
- Minor: `ReactionEmojiHistory.swift`: comment says “same order as web” while web still lists five defaults (`🎉`, `👀` unused at toolbar count 3).
- Minor: physical Tab/Space and pointer hover scale remain unverified (automation limit, already noted in the PR).
- Skipped / disagreed: ViewThatFits duplicate `@FocusState` bindings — `ViewThatFits` keeps the fitting child in the interactive hierarchy; the skill’s ambiguous-binding example is two visible fields, not two layout candidates. Geometry-driven compact would be a larger, flicker-prone rewrite with no observed keyboard bug.
- Skipped / disagreed: extra package test for “remove does not record” — the guard is one line in the view; Apple’s required suite does not cover SwiftUI layout; ranking/dedupe tests already lock history selection.

### Assessment
Shortcut ranking now matches web on remove; compact layout and unverified keyboard/hover stay as the PR already documented.

mrosberghaus:
Verified `be223e947` against web's `handleToggleReaction`: deciding add/remove from the current user's reaction before the async action correctly prevents removal from teaching emoji history. The native path retains its existing behavior of recording after the action succeeds. The fix is valid.

Addressed two minors in `69655da53`:
- Selected shortcut buttons now add SwiftUI's `.isSelected` accessibility trait, alongside the existing spoken value.
- Clarified that the native defaults are the first three web defaults, matching the desktop slot count. With three output slots, the extra web defaults do not change selection results.

Keeping one popover attached outside `ViewThatFits` deliberately. During the original compact-layout fixture check, attaching a presenter inside both candidate toolbars caused the compact picker not to appear. The single presenter restored both layouts. The arrow is anchored to the toolbar rather than precisely to React; changing that now would trade a working presentation for an unverified alternative.

No evidence supports replacing `ViewThatFits` or adding a geometry-driven layout. Physical Tab/Space navigation and pointer hover animation remain explicitly unverified in the PR; accessibility button activation, stable shortcuts, selected state, and wide/compact picker presentation were verified. I am not claiming those remaining manual checks passed.

Validation after these follow-ups: macOS build succeeds, all 378 Chat tests pass, pinned strict SwiftLint and SwiftFormat pass, and repository pre-commit checks pass. Updated CI is pending on the new commit.

PR #4626, opened 2026-09-15
feat(apple): add room thread overview

mrosberghaus:
Adds a native Threads inspector for the selected room. Readers can browse unread-first threads, load older pages, open a thread and return with native Back, or mark all room threads read. Opening the overview itself does not mark anything read.

The toolbar shows a dot or an unread count (capped at 99+) according to the existing server preference. Portable state handles retry, stale responses, rapid reopen and mark-all failures. Previews reuse Markdown parsing off the main actor; generated bindings come from existing Core endpoints. No web code, Core contract or dependencies change.

Web parity sources: [thread list](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/thread-list-panel.tsx), [header attention](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/unread-threads-panel.tsx), [preview rules](https://github.com/masumi-network/sokosumi/blob/main/packages/utils/src/chat-message-preview.ts), [realtime refresh rules](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/utils/thread-unread-event.ts).

Scope: PARITY slice 24, room-only as approved. Workspace-wide unread threads, preference editing and notification deep links remain outside this slice.

Verification:
- All five package suites pass: Chat 387, Workspace 73, Auth 35, Realtime 27, CoreAPI 1.
- macOS build and full app tests pass; strict SwiftLint/SwiftFormat and repository check/typecheck pass.
- Workspace package builds for iOS 17.
- Minimum-width inspector rendered in light/dark. Native fixture confirmed overview → thread → Back; its text-only destination establishes navigation state, not actual reply-view sizing. Existing reply-view tests cover layout. Fixture removed; test apps closed.

How to test: open Threads in a room with replies, check previews and unread attention, open a row and go Back, load older threads, then use Mark all read. Check empty/error/retry states and resize the inspector.

Light inspector
Dark inspector


mrosberghaus:
## Grok review

### Fixed
- Important: Threads VoiceOver now says `Threads` when unread is 0 and `more than 99 unread` past the visible `99+` cap (`b7a85e1ca` — `apps/apple/Sokosumi/Chat/Search/RoomToolsModifier.swift`)
- Important: Inspector hides while a thread is open without dropping the overview request; lasting tests cover that predicate (`b7a85e1ca` — `apps/apple/SokosumiTests/Chat/Search/RoomToolsModifierTests.swift`)
- Important: Slice 24 now records that native previews flatten `MessageMarkdown` plus URL strip/128 cap, not the banner helper (`b7a85e1ca` — `apps/apple/PARITY.md`)

### Remaining
- Minor: `RoomThreadOverviewView.swift`: Mark all is also disabled while `isLoading`, so a reload after Back briefly blocks it (web only disables while marking)
- Minor: `RoomToolsModifier.swift`: Threads trigger has no open-state equivalent of web `aria-expanded`
- Minor: `ChatReadCooldownMiddleware.swift`: list reads are on the cooldown allowlist; `threads/unread-count` is not
- Minor: `RoomThreadOverviewView.swift`: copy is `Mark all read` (web: `Mark all as read`); error/retry row lacks the padding used by empty/loading states
- Skipped / disagreed: full port of `packages/utils/src/chat-message-preview.ts` — native rows go through `MessageMarkdown` (same parser as the timeline). The util is the lock-screen/banner sanitizer; a Swift port of HTML-comment / empty-mention / name-seam rules is out of this slice. Documented the gap instead of copying 650 lines.

### Assessment
Room-only overview, GET-only open, explicit Mark all, stale/race guards, and Core binding regen are solid; leftover notes are copy, cooldown on the count GET, and expanded-state a11y.

mrosberghaus:
Verified the review against the current web/Core implementations and pushed follow-ups in `699bfe56e`.

- The VoiceOver count fix is valid. The inspector helper extraction preserves the existing behavior; its new tests provide regression coverage rather than fixing a previously different predicate.
- Added Expanded/Collapsed accessibility state, matched “Mark all as read”, and padded the error/retry group. Targeted app tests, strict SwiftLint and SwiftFormat pass; the updated 280-point light/dark renders fit the longer label.
- Retained the loading guard for Mark all as read. The native model serializes list replacement and the mutation; removing only the UI guard would expose an action the model ignores while loading.
- Kept unread-count outside the history-read cooldown. Its Core route does not call `assertChatMessageReadBudget`; the thread-list route does.
- I disagree that the preview sanitizer is banner-only or outside thread parity. Web `formatUnreadThreadsPreview` directly delegates to `buildChatMessagePreview`. The documented HTML/comment, empty-mention and name-seam differences remain a real parity gap, so I am not treating documentation alone as resolution or recommending merge yet.

The first local test attempt failed on TreeSitterYAML coverage linking; the retry with `-enableCodeCoverage NO` passed. The test host exited. CI will verify the new head.

PR #4631, opened 2026-09-15
fix(web): anchor message action bar on the row's top edge

mrosberghaus:
## Summary

- The message hover action bar now sits centred on the hovered message's top edge, like Slack and the Apple client (`apps/apple/Sokosumi/Chat/Timeline/MessageRowView.swift`), instead of inside the row. It no longer hangs below one-line messages or drops its tooltips onto the composer.
- The hidden bar (opacity 0, still mounted after the first hover) no longer catches the pointer, so it cannot steal hover from the message above. While the More menu is open the bar stays inert, so the click that closes the menu does not also trigger a bar action.
- The right hover gutter (`pr-64` / `pr-72`) is unchanged.

## Known trade-offs

- The message at the very top of the scroll area gets the top half of its bar clipped by the scroller, as in the Apple client.
- A message with the Soko Bot chain badge has a wider bar (about 17rem). Over a message with the normal 16rem gutter, it can cover about 1rem at the end of that message's last full-width line while hovered.
- In the thread panel (no gutter), the bar's upper half now covers the end of the previous reply instead of the hovered message's header.

## Verification

- `pnpm --filter web test src/app/(app)/chat/components/room-message-row.test.tsx`: 126 passed, including two new pointer-guard tests.
- Full `pnpm --filter web test` before the final review fixes: 748 files, 5955 passed, 1 skipped.
- `pnpm --filter web typecheck` and Biome clean (pre-commit hook passed).
- Headless Chrome on a throwaway page with real `ChatMessageRow`s (not committed): bar centre equals the row's top edge; the pointer over the overlap zone keeps the upper message hovered; moving from a message up onto its own bar keeps the bar reachable.


mrosberghaus:
## Grok review

Picked up the review trigger from @mrosberghaus.

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/room-message-row.test.tsx` — new tests lock pointer-event class tokens only; nothing asserts `top-0` / `-translate-y-1/2`, so a later restore of `top-1.5` would still pass CI.
- Minor: same test (`keeps the hidden hover pill from taking the pointer over the row above`) overclaims two-row hover stealing; it checks the class combination on one hovered row.
- Minor: `MESSAGE_ACTIONS_PILL_CLASS` dropped `flex`; layout comes from `[@media(hover:hover)]:flex` at the call site. Fine with one consumer.
- Skipped / disagreed: none — the three Minors are real but out of this pass (no Critical/Important to piggyback a one-line fix onto).

### Assessment
Bar is centred on the row top edge like the Apple overlay, the latched opacity-0 pill no longer steals hover from the row above, and More-open inertness matches Radix modal pointer-events. Remaining notes are test coverage/naming, not product bugs.

mrosberghaus:
Replies to the three Minors.

**Test name overclaimed** — fixed in 6650111e. The test is now `gives the hover pill the pointer only while it shows`, with a comment saying that hit testing across rows needs real CSS and that the behaviour itself was checked in a browser.

**No position assertion** — deliberate, and the repo owner's call. The position is a visual result that `top-0` / `-translate-y-1/2` only stand in for, so an assertion on those tokens would break on an equivalent rewrite while proving nothing about the rendered position. It was measured in headless Chrome instead (bar centre equals the row's top edge) and is visible on the preview deployment. A revert to `top-1.5` passing CI is a real gap; the screenshots in this PR are the guard.

**Dropped `flex`** — no change. There is one consumer, and `cn` (tailwind-merge) dropped that `flex` anyway, because the `hidden` in the call site's class string wins on display.

PR #4639, opened 2026-09-16
fix(chat): align unfurl preview image with its text column

mrosberghaus:
## Problem

A link preview's image was sized by its **height** budget, so anything taller than the box's own ratio drew narrower than the title above it and left dead space on the right.

Web compounded it: the card was `w-fit max-w-full`, so a long title stretched it across the whole message column while the image stayed at the 192px height cap. That is the "text much wider than the image" in the report.

## Change

Both apps now fill the text column with the image and crop what does not fit, and the web card is capped at 400px to match the Apple card.

- **apple**: `MessageImageView` gains `fills`, which clamps the layout ratio to the box's ratio and switches the overlay to `scaledToFill().clipped()`. Attachments keep fit — cropping a reader's own photo would be wrong.
- **web**: the preview image goes `w-full object-cover`; the card goes `max-w-100`.

Measured on macOS at a 1160pt window, image width against a 380pt text column:

| image | before | after |
| --- | --- | --- |
| 2:1 (typical OG card) | 380 | 380 |
| 1:1 | 192 | **380** |
| 3:4 | 144 | **380** |

## Reviewer note

`room-message-row.test.tsx` previously asserted the opposite — *"Keep intrinsic aspect ratio (do not force w-full + max-h + object-cover)"*. That assertion is flipped here and its comment rewritten to record the new intent. If the original reason still holds, say so on this PR.

## Verification

- `xcodebuild … build` (macOS, Debug) — green
- `xcodebuild test -only-testing:SokosumiTests` — 126 passed, 0 failed. Needs `-enableCodeCoverage NO`: with coverage on, `TreeSitterTypeScript` fails to link with `Undefined symbol: ___llvm_profile_runtime`. Pre-existing, unrelated to this change.
- `mint run swiftformat --lint .` / `mint run swiftlint lint --strict` — clean
- `pnpm --filter web test src/app/(app)/chat/components/room-message-row.test.tsx` — 126/126
- `pnpm check` + `pnpm typecheck` via precommit — green

An earlier revision of this section claimed `-only-testing:SokosumiTests/MessageImageLoadingTests` passed. That filter matches nothing — the suite is nested in `NativeWindowTests` — so the run was vacuous. The whole-target result above replaces it.


mrosberghaus:
## Grok review

### Fixed
- Important: web unfurl card used `max-w-100` only, so a column under 400px could size the card to 400px and clip under `overflow-x-clip`. Both wrapper and link are now `max-w-[min(100%,25rem)]` (`d98bfd0ae` — `apps/web/src/app/(app)/chat/components/room-message-row.tsx`).
- Important: Apple `MessageImageLoadingTests` only asserted height on a 1.5:1 PNG, so reverting `fills: true` still passed. Added a portrait SVG unfurl width assertion (`d98bfd0ae` — `apps/apple/SokosumiTests/Chat/Rendering/MessageImageLoadingTests.swift`).

### Remaining
- Minor: `MessageImageView.swift` `layoutRatio` clamps against the 380×200 budget, not the laid-out column, so a 1:1 unfurl in a 300pt pane is cropped harder than a 300×200 box would need. Alignment is still correct.
- Minor: typical OG 1.91:1 at 380px wants ~199px height; web `max-h-48` (192) now crops a sliver. Apple’s 200pt budget does not. Pre-existing caps.
- Skipped / agreed: the old web test that forbade `w-full` + `object-cover` — that pair is the fill-and-crop the PR wants. Attachments still fit.

### Assessment
Fill-and-crop matches the plan on both apps; the 400px cap now also respects a narrower parent.

mrosberghaus:
Worked through the Grok review. One finding held, one did not, and the Apple fix needed a follow-up because it left CI red.

### `max-w-100` → `max-w-[min(100%,25rem)]` — could not reproduce

Built the real nesting from `room-message-row.tsx`: `flex min-w-0 max-w-full overflow-x-clip` row → `flex-1 min-w-0 overflow-x-clip` column, 340px pane, 36px avatar + 14px gap, so a 290px column. Long title, `w-full` image with a 1200×600 intrinsic source. Measured in Chromium:

```
pane=340  column=290
max-w-100               card=290  img=268  ok
max-w-[min(100%,25rem)] card=290  img=268  ok
```

`width: fit-content` already resolves against the available space, and `min-w-0` on both flex ancestors keeps that space definite, so the 400px cap never binds below 400px. No clip either way.

Keeping the change — it is equivalent here and strictly more defensive if an ancestor ever becomes intrinsically sized — but the severity looks overstated. If you saw a real clip, what pane width and which ancestor?

### `unfurlPortraitImageFillsTheTextColumn` — right intent, broken as written

The concern was correct: nothing caught a revert of `fills: true`. The test did not either. It asserted `host.fittingSize.width >= 360`, and `fittingSize` reports the **ideal** width. `Color.clear` contributes no ideal width, so the card's ideal was just the title:

```
host.fittingSize.width : 67.0     # with fills: true
host.fittingSize.width : 67.0     # with fills: false
```

Failed both ways, discriminated nothing, and made this PR red.

Fixed in `0b52d48dd` by making the view declare what it actually wants — a filling image asks for the whole width budget:

```swift
.frame(idealWidth: fills ? maxSize.width : nil, maxWidth: maxSize.width, maxHeight: maxSize.height, alignment: .leading)
```

`nil` when not filling, so fit mode is untouched. The assertion now reads 400 with `fills: true` and 67 with `fills: false` — verified by mutating the call site and running it both ways. `-only-testing:SokosumiTests` is 126 passed / 0 failed.

Note for future runs: `-only-testing:SokosumiTests/MessageImageLoadingTests` matches nothing. The suite is nested in `NativeWindowTests`, so the path is `SokosumiTests/NativeWindowTests/MessageImageLoadingTests`. The short form exits `TEST SUCCEEDED` having run zero tests.

### Remaining notes

- **`layoutRatio` clamps against the 380×200 budget, not the laid-out column** — agreed, leaving it. In a 300pt pane the box lays out 300×158 instead of 300×200, so a tall image crops a little harder than it needs to. Alignment, which is what this PR is about, is unaffected. Reading the real column needs a `GeometryReader` on the hot path of every previewed message; not worth it for a narrow-pane fidelity detail.
- **1.91:1 at 380px wants ~199px, web caps at 192** — real, ~3.5% off the top and bottom of a typical OG card. Apple's budget is 200. Happy to move web to `max-h-50` / `h-50` so the two platforms use the same cap; say the word and it is one value plus the test.

mrosberghaus:
Closing out the second remaining note — web preview height cap raised to 200px in `15344abc1`.

`max-h-48` was 12rem because that is the nearest spacing token, not because anything about link previews wanted 192px. The number with a reason is the smallest cap that leaves a standard card whole:

```
1.91:1 (Open Graph recommendation, 1200×630)
  web:   378 / 1.91 = 198
  apple: 380 / 1.91 = 199
```

So 200 leaves both 1.91:1 and 2:1 uncropped and 192 clips both. Web now uses `h-50` / `max-h-50`; Apple's budget was already 200pt, so both platforms land on the same number.

Verified the utility actually compiles rather than silently no-opping — built the real Tailwind output against `apps/web`:

```
h-50      .h-50     { height: calc(var(--spacing) * 50); }
max-h-50  .max-h-50 { max-height: calc(var(--spacing) * 50); }
```

`--spacing` is not overridden in `globals.css`, so that is the 0.25rem default → 12.5rem → 200px.

`pnpm --filter web test room-message-row.test.tsx` 126/126, biome clean.

The other remaining note (`layoutRatio` clamping against the budget rather than the laid-out column) stays as-is, for the reason in the previous comment.

PR #4651, opened 2026-09-16
feat(notifications): make the notification center one list where reading is the only done state

mrosberghaus:
Implements [SOK-1085](https://linear.app/masumi/issue/SOK-1085/notification-center-as-one-list-where-reading-is-the-only-done-state). The Unread-only view follows in [SOK-1084](https://linear.app/masumi/issue/SOK-1084/unread-filter-in-the-notification-center), which is blocked by this.

## What changes

The Notification Center was two lists that behaved differently, with two ways to be done with a notification: read it or delete it. It is now one list where reading is the only change a reader makes.

- **One list:** the header panel and `/notifications` render one list component over one provider state. Newest first, 20 rows a page, and older rows load on their own as the reader scrolls toward the end, until nothing older is left. The page's own fetching hook, the list merge and the deletion event bus are gone.
- **Row control:** each row carries one control, shown on hover and on keyboard focus. A double check marks an unread row read, and a filled dot marks a read row unread, with the action named in its tooltip and accessible label. A mouse click does not leave focus on the control, so it hides once the pointer leaves. A failed read or unread write puts the row back and shows an error.
- **Panel:** a Radix Popover instead of a DropdownMenu, because a scroll-loaded list with per-row buttons fought the menu's roving focus. Opening it focuses the panel rather than its first row, so no row shows its control before the reader touches it. On mobile the bell is a link to `/notifications`.
- **Mark all read:** relabelled from "Mark all as read" in English, in both frames. The panel header keeps one height with or without the button, and the page keeps its action row while the list has rows, so reading the last unread row no longer moves the list.
- **Removed:** per-row delete, Clear all and its confirmation dialog in both frames, with their translation keys in `en`, `de`, `es`.
- **Core:** removes `DELETE /v1/notifications/{id}` and `DELETE /v1/notifications` and their tests, and regenerates the OpenAPI snapshot and the web client. Web was the only caller. The Apple app has no Notification Center, and its OpenAPI subset lists no notification routes.
- No Prisma schema change and no data migration. Notifications are kept forever; there is no purge job.

## User-facing impact

A reader can no longer lose a notification to a stray click or a Clear all. Everything they received stays reachable by scrolling, in the panel as well as on the page. Reading a row settles it, and a read can be undone per row.

## Verification

VERIFIED, run in this worktree on Node 24.14.1 at head `471342f2e`, after merging `main` including #4663:

- `pnpm check` exit 0
- `pnpm typecheck` exit 0, `Tasks: 20 successful, 20 total`
- `pnpm --filter web test` exit 0, `Test Files 753 passed (753)` / `Tests 6014 passed | 1 skipped`, including message catalogue parity for `de` and `es`
- `pnpm --filter @sokosumi/core test` exit 0, `Test Files 521 passed | 3 skipped` / `Tests 5559 passed | 11 skipped`
- `pnpm --filter web generate:core:snapshot` at the head produces no diff, so the generated client merged from `main` matches the Core source.

Tried on the Vercel preview during review: opening the panel, marking rows read and unread, the header height, the control hiding after a click, and the icon choice. No screenshots are attached.

## Notes for the reviewer

- This reverses SOK-983, which added delete and Clear all. Newest-first order plus SOK-1084's Unread-only view covers the problem it solved.
- **Paging guards, each with a test:**
  - When a page adds no new rows, the next request follows Core's `nextCursor`, so it never asks for the same page twice.
  - When Core refuses a cursor with a 400 because that row left the feed, the row is dropped and paging continues from the row above, once per attempt.
  - A failed page waits for the reader's retry, and a refresh does not restart it.
  - An older page still loading when a refresh starts the list over is dropped when it lands (`5edfbfa5b`), and the loading state holds until then, so scroll loading re-arms (`9a56a93ad`).
- A refresh keeps the pages the reader loaded, unless the new first page no longer reaches them because more rows arrived than a page holds. Then the list starts over from that page. A realtime unread event for a row older than everything loaded is ignored until paging reaches it.
- The IntersectionObserver effect from the chat transcript boundary row moved into `useLoadWhenVisible`, which that row and the new notification boundary row now share.
- Rows and their loading placeholders use one padding in both frames, so page rows are slightly denser than before.
- The stale read-on-close sentence in the unread route's OpenAPI description, and the client regen it needed, shipped in #4663.


mrosberghaus:
## Grok review

### Fixed
- Important: in-flight `loadOlder` could append under a start-over refresh and leave a hole in the list (`5edfbfa5b` — `apps/web/src/contexts/notification-provider.tsx`)

### Remaining
- Minor: `apps/core/src/routes/v1/notifications/[id]/unread/patch.ts`: unread route copy still says the center marks rows read on close
- Minor: `apps/web/src/app/(app)/notifications/components/notifications-loading-view.tsx`: skeleton rows still use `p-4` while live rows use `p-3`
- Skipped / disagreed: both Minors — not one-line fixes in a file this pass already edited; the Core copy also needs a client regen

### Assessment
One list, reading as the only done state, delete/clear removed through Core and the generated client; sequential paging guards hold, and the concurrent start-over hole is closed.

mrosberghaus:
Follow-up on the Grok review above.

- **In-flight older page after a start-over refresh** (`5edfbfa5b`): the stale-page drop stays. The branch that set the older status to idle as soon as the refresh landed is removed. It re-armed the boundary while the stale request still blocked `loadOlder`, so the boundary fired once, was ignored, and never armed again, leaving scroll loading stuck. The status now stays `loading` until the stale page lands and is dropped, which re-arms the boundary. New island test, which fails on `5edfbfa5b`: 9a56a93ad
- **Loading rows `p-4` vs live rows `p-3`**: fixed in c94dff674.
- **Unread route copy still describing read-on-close**: confirmed stale, not changed in this PR. The copy change needs a web client regen, and regenerating on current `main` also rewrites unrelated Soko Bot and `X-Organization-Slug` header types. A clean `origin/main` regenerates to the same drift, so its committed client already differs from its Core source. That regen and this copy fix belong in their own change.

Verified on c94dff674: `pnpm check`, `pnpm --filter web typecheck`, `pnpm --filter web test` (753 files passed, locale parity ok).

mrosberghaus:
The unread route copy fix and the Core web client regen it needed are in #4663.

PR #4658, opened 2026-09-16
fix(core): expose direct recipient schemas to generated clients

mrosberghaus:
Swift OpenAPI Generator currently omits the nullable `SokoBotState.sokoBot` field, and generated recipient requests cannot specify the workspace header. This blocks the Apple Start Direct recipient picker.

Use the existing nullable-object pattern for the state field while keeping the shared `SokoBot` schema non-nullable. Reuse `withOrganizationSlugHeaderParameter` on GET `/coworkers` and GET `/soko-bots/me` to document the header their middleware already supports. Runtime payloads and authorization are unchanged.

Separate, approved Core prerequisite for Apple Start Direct. No database migration or web product changes. Client generation was verified in temporary output directories; the Apple snapshot will be regenerated in its feature PR after this merges.

Validation:
- 63 tests passed across schema, OpenAPI headers, coworker, and Soko Bot route tests.
- `pnpm check`, `pnpm typecheck`, and pre-commit checks passed.
- Core OpenAPI export and web client generation succeeded (web generator reports unsupported-transformer warnings).
- Swift client generation with the Apple configuration produced the nullable bot field and both typed workspace headers, with zero diagnostics.
- Regression coverage checks nullability in both schema registration orders and header declarations on both routes.


mrosberghaus:
## Grok review

Picked up a grok review request from @mrosberghaus.

### Fixed

### Remaining
- Minor: `apps/core/src/routes/v1/coworkers/get.ts`, `apps/core/src/routes/v1/soko-bots/index.ts`: header wrapper applied at `app.openapi(...)` instead of the route const (other callers wrap at the `createRoute` assignment)
- Minor: `apps/core/src/schemas/soko-bot.schema.ts`: comment does not name the Swift vs hey-api split; a later change could “fix” this back to named-schema `z.union`
- Minor: `apps/core/src/schemas/soko-bot.schema.test.ts`: nullability test does not assert `anyOf` is absent on `sokoBot`

### Assessment
Ready to merge: Yes
The OpenAPI source emits a locally nullable bot field without poisoning `SokoBot`, and both recipient routes document the optional workspace header their middleware already honors.

mrosberghaus:
Verified the review against the current implementation and addressed all three minors in 7b21afe98.

- Moved both organization-header wrappers to the route declarations, matching the existing route convention. This is a consistency improvement; registration-time wrapping was functionally equivalent.
- Clarified that hey-api supports the named-schema null union, while Swift codegen drops the field, and why the nullable object must stay local.
- Added an explicit assertion that the bot property has no `anyOf`, in both schema registration orders. The existing type assertion already rejects the original union regression; this also guards against accidentally adding `anyOf` alongside the nullable object.

Validation: all 63 targeted tests pass; repository Biome and typechecks pass through the pre-commit hook. Re-exported the full Core OpenAPI document and confirmed it is structurally identical to the pre-follow-up export. No runtime or API behavior change from these follow-ups.

No verified Critical/Important findings remain. CI was green on the reviewed commit; checks are rerunning for this follow-up.

PR #4666, opened 2026-09-16
fix(web): size app shells with dvh so the iOS home-screen app stops scrolling the room

mrosberghaus:
## Summary

On the installed (home-screen) iOS web app the whole chat room, composer included, could be dragged up and down under the fixed header. Screen recording from Andreas on 2026-09-16.

**Cause.** In standalone mode WebKit reports `svh` off by the status-bar height. Measured on an iPhone 17 simulator (iOS 26): `100svh` = 874 while the layout viewport (`innerHeight`, `visualViewport`, `-webkit-fill-available`, `100dvh`) = 812 with the default status-bar style; with `black-translucent` it flips (812 vs 874). Every shell sized with `svh` (body, sidebar wrapper, app `main`, chat room shell, auth/flows shells, full-screen modals) was 62px taller than the screen, so the document scrolled and the composer sat partly below the bottom edge. Safari and headless Chromium report `svh` = `dvh`, which is why this never showed outside the PWA.

**Fix.** Use `dvh` for viewport-height sizing throughout the web app. `dvh` matched the real layout viewport in every mode measured (Safari, both standalone status-bar styles).

## User-facing impact

- Installed iOS app: the room no longer scrolls or rubber-bands as a whole; the composer sits fully on screen above the home indicator.
- Sign-in, flows and full-screen mobile modals get the same correction (they used the same unit).
- Desktop and Android: no change (`svh` = `dvh` there for our non-scrolling shells).

## Verification

- Throwaway auth-exempt probe page replicating the app frame around `RoomOpenLoadingView`, installed as a web clip in the iPhone 17 simulator: main `clientHeight` 874 vs `innerHeight` 812 before (composer cut off), 812 / 812 after (composer fully visible). Probe deleted before commit.
- `pnpm check`, `pnpm --filter web typecheck`, full `pnpm --filter web test` (all green).
- The shell class guard test (renamed to `shell-height-class-guards.test.ts`) now also bans `svh` units so this cannot regress silently.

## Follow-up (not in this PR)

The app sets no `apple-mobile-web-app-status-bar-style`, so the PWA never paints under the status bar although the header carries safe-area handling for it. Switching to `black-translucent` is a design decision; it works with this change (verified in the same probe) but is out of scope here.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/tasks/components/tasks-view.tsx` and `tasks-loading-view.tsx` still use `max-h-[calc(100vh-150px)]` (same files already switched kanban `svh` → `dvh`). Pre-existing; not the room/composer rubber-band.
- Minor: leftover `100vh` / `min-h-screen` / `h-screen` in `global-error.tsx`, auth callback/signup, `maintenance/page.tsx`, calendar + task-schedule dialogs, `image-viewer.tsx`. Guard only bans `svh`, so these can return without failing the new test.
- Skipped / disagreed: converting those `vh`/`screen` utilities in this PR — the measured standalone bug was `svh` off by status-bar height; `dvh` matched `innerHeight`. Extra unit sweep is out of scope.

### Assessment
All product `svh` shells are now `dvh`, the chat room height token matches the app shell, and the class guard will catch `svh` regressions; leftover `100vh`/`min-h-screen` is a follow-up, not a blocker for the iOS home-screen room scroll.

mrosberghaus:
Follow-up to the Grok review: the leftover `100vh` / `min-h-screen` / `h-screen` sites are the same defect, not a separate sweep. In the installed app with the default status-bar style `vh` and `lvh` report the full screen height (874) while the layout viewport is 812, the same 62px overflow as `svh`. Commit 6eedb719e converts the remaining 25 sites (global error, maintenance, auth callback and social signup shells, the tasks board max-height, dialogs, image viewer) to `dvh` and widens `shell-height-class-guards.test.ts` to ban `vh`, `lvh` and `h-screen` heights too. Width utilities (`w-screen`, `svw`, `dvw`) are unchanged. Biome, web typecheck and the full web suite pass.

PR #4670, opened 2026-09-16
feat(web): add unread filter to the Notification Center

mrosberghaus:
## What readers get

The Notification Center gains a view strip, drawn once and shown in both frames: the header panel and the notifications page. Two tabs, All and Unread, with the live unread count on Unread and the active tab underlined in the unread colour, the same purple as the rail on an unread row. Unread lists only rows that are still unread, newest first, with the same scroll loading. Switching is a lens — it never marks anything read and never removes anything. The default is All, the choice is shared between panel and page for the session, and it resets to All on reload. An empty Unread view says “You’re all caught up” (translated in de/es) instead of “No notifications yet”.

A strip of tabs rather than a chip or a segmented pill: each view is a different list, so the control is a single choice, and a later view (rows that need the reader, SOK follow-up) joins as a third tab.

### Panel layout

- **Header**: the title on the left, Mark all read on the right while something is unread. Nothing else shares the row, so the button can leave without moving anything.
- **Strip**: All · Unread, directly above the rows. Its bottom border is the line between header and list.
- **Footer**: the link to the full page, right-aligned and muted.
- The header, the strip, and the browser-permission primer sit on the popover's 16px inset.
- On the page the strip is the card's top edge, so it reads as the card's own control and not as page navigation. Mark all read keeps the action row above the card.

## Notes

- The view lives in the Notification Center provider (`view`/`setView`); switching clears loaded rows, drops the older-page cursor, and refetches with the Core list's existing `isRead=false` query.
- The strip is the Radix `Tabs` primitive with the underline drawn as the active trigger's bottom border in `border-primary`; the count pill uses the primary ramp (`-quinary` fill), no opacity modifiers.
- No Core change, no generated-client regeneration, no new endpoint. Badge counting, read semantics, and chat room read sync are untouched.
- In the All view a read row stays where it is. In the Unread view it leaves once the pointer and focus are off it, folding up over 200ms, so nothing slides under the cursor that just clicked its control and the way back stays until then. A tapped row folds right after the tap. Hiding is a lens over the shared list, not a delete: the row stays in state so a failed mark-read can roll back, and emptiness and paging follow the unread badge. Mark all read empties the Unread view at once. Putting a row back to unread before leaving it cancels the departure.

## Links

- Linear: SOK-1084

## Verification

- `pnpm --filter web test` on the notification island, reducer, header, and page suites: 13 files, 149 passed
- `pnpm --filter web typecheck`: clean
- `pnpm check` (Biome): clean
- `pnpm --filter web messages:parity`: de/es ok
- Tailwind 4.3 compile check: `aria-pressed:*` and `group-aria-pressed/filter:*` emit CSS
- Preview checked in dark mode: chip, header spacing, footer, and Mark all read leaving without moving the chip

## Schema changes

None.


mrosberghaus:
## Grok review

### Fixed
- Important: Unread view no longer deletes feed rows after the 200ms fold. Hiding is a list lens, empty/paging follow `unreadCount`, and `loadOlder` no-ops when the badge is already 0. That stops “You’re all caught up” while more unread pages remain, a mark-all refill under a 0 badge, and a failed mark-read that cannot roll back (`3b14ca3b8` — `apps/web/src/components/notifications/notification-center-list.tsx`, `apps/web/src/contexts/notification-provider.tsx`).
- Minor (same files): reduced-motion skips the 200ms hide delay; the leave test now asserts Mark all read is gone instead of that the count endpoint was called.

### Remaining
- Minor: `apps/web/src/components/notifications/notification-center-list.tsx`: touch has no durable “pointer inside”, so a tap can fold the row before undo or leave it until the next tap.
- Minor: `apps/web/src/app/(app)/notifications/page-content.tsx`: empty Unread still mounts a blank `min-h-8` action row above the card.
- Minor: PR Notes still describe the `aria-pressed` chip; head is the tab strip.
- Skipped: Linear SOK-1084 still says stay-until-fetch and a segmented control in the header — the PR body is the newer contract; not rewriting the issue from this review.

### Assessment
The All/Unread lens, shared session state, and Core `isRead=false` query are sound; the leave-and-remove coupling that could lie about “caught up”, refill after mark-all, and lose a failed mark-read is fixed. Touch-undo and a blank page action row can wait.

mrosberghaus:
Re the Grok review (issuecomment-5703065324): the lens fix in `3b14ca3b8` holds up against the reducer on all three counts, and it stays. Remaining items handled in `86d293f4b`:

- **Touch**: only a hovering pointer holds a read row now (`pointerType !== "touch"`). A tapped row folds right after the tap, since the finger is gone and nothing can slide under it. The undo window on touch is the trade-off; Mark unread remains in the All view.
- **Blank action row on the page**: the row is rendered only while there are rows to hold space for. An Unread view that is empty from the start gets no row.
- **PR notes**: rewritten for the tab strip and the hide-as-lens behaviour.
- **Removed**: the effect clearing `hiddenIds` on view change. The All view never reads the set and an Unread fetch never returns a read row, so a stale id can only match a row that is unread again, which the lens shows regardless.
- **Linear SOK-1084**: left as is; the PR body is the contract.

Checks: 216 tests across the notification, header, page, and reducer suites; Biome and web typecheck clean.

PR #4671, opened 2026-09-16
style(web): put layout and border lengths on whole pixels

schaier-io:
Stacked on #4673.

## Why

A CSS pixel is not a device pixel. On a 2x Mac display `1.5px` is exactly three device pixels, so it looks deliberate to whoever wrote it. On a 1x display, which is most Windows hardware, there is no half pixel: the browser rounds, and not consistently. A `1.5px` border lands at 1px on one edge of the box and 2px on the opposite edge. A `0.2px` border rounds away to nothing on some surfaces and to a hairline on others.

The result is an element that is visibly lopsided on the hardware most users have, and correct only on the hardware the author has.

## The four sites

| File | Was | Now |
| --- | --- | --- |
| `globals.css` `.chat-input-border-anchor` | `padding: 1.5px` / `border: 1.5px` | `2px` / `2px` |
| `agents/agent-card.tsx` | `border: "0.2px solid transparent"` | `1px` |
| `agents/agent-detail/header.tsx` | `border: "0.2px solid transparent"` | `1px` |
| `agents/coworker-gallery-section.tsx` | `p-[1.5px]` | `p-0.5` (2px, back on the 4pt scale) |

## The guard

`apps/web/src/lib/utils/__tests__/whole-pixels.test.ts` scans every `.ts`, `.tsx` and `.css` file under `apps/web/src` and fails on a fractional px in three forms: a CSS declaration, a React style object, and a Tailwind arbitrary value. Generated clients and test files are skipped.

It covers only lengths that resolve to a painted edge or a box dimension. Blur radii, keyframe translations and unitless scale factors are interpolated rather than snapped, so they stay fractional and are deliberately out of scope. Three sites keep a fractional px for that reason: `--chat-jump-dim-blur`, the `sokosumi-loader` keyframe translations, and a `presence-dot.tsx` comment that already states the principle.

**There is deliberately no line-level exemption list.** A first draft carried one, and it made the whole check vacuous: the exemption matched `shadow-lg` elsewhere on the same line and disarmed every utility beside it.

## Docs

The rule is written up in `.cursor/rules/whole-pixels.mdc` and linked from root `AGENTS.md`, `apps/web/AGENTS.md`, and the Shapes section of `DESIGN.md`, beside the existing rule against fixed px type.

## Verification

VERIFIED by mutation, not by a green run alone. Reintroducing each of the three violation shapes in turn, one at a time, makes the guard fail:

```
TW probe  (p-[1.5px] beside shadow-lg)     -> CAUGHT
CSS probe (padding: 1.5px)                 -> CAUGHT
OBJ probe (border: "0.2px solid …")        -> CAUGHT
clean tree                                 -> Tests  1 passed (1)
```

Full suite on the rebased branch:

```
pnpm --filter web test
 Test Files  763 passed (763)
      Tests  6122 passed | 1 skipped (6123)

pnpm typecheck
 Tasks:    20 successful, 20 total
```

Blind spot: the guard is three regexes over source text. It cannot see a px length assembled at runtime by string concatenation or computed from a variable, and it does not read `.mdx` or any inline `<style>` outside the three extensions it walks.


schaier-io:

PR #4672, opened 2026-09-16
fix(auth): remove the redirect chain from sign-in and sign-up

mrosberghaus:
## Why

Submitting the login form showed four screens before the app: the form vanished and the marketing hero photo swapped, a spinner sat beside the panel for ≥200 ms, then a fully blank page, then the sidebar skeleton. Only the last one is needed.

Cause: credential sign-in hard-navigated through `/auth/callback/signin` (a full document load under the `(auth)` layout, which picks a new random hero per request) purely to fire a GTM event, then navigated to `/`, where the workspace gate's chrome-free Suspense fallback rendered a blank page before the shell skeleton.

## What changes for the user

| Before | After |
| --- | --- |
| Form vanishes, hero photo swaps | Form keeps its spinner until the app loads |
| Spinner beside the marketing panel | Gone for credential and passkey |
| Blank page | Sidebar + header skeleton |

Google, Microsoft and magic link still land on `/auth/callback/signin` (Core must 302 somewhere), but that page now renders outside the marketing layout: plain background + spinner, no second hero.

## How

- Credential sign-in, credential sign-up and passkey pass no `callbackURL`. They wait for the session, count the conversion only when one exists, then leave for the destination. Shared as `finishAuthInPlace` in `lib/auth/finish-auth.client.ts`, which takes an `eventType` mirroring `SocialAuthCallback`.
- The leave is a **full document load** (`window.location.replace`), not `router.replace`. The Next client router cache still holds the pre-login middleware result for the destination (anonymous `/` → `/signin`), so a soft nav returns that cached redirect and bounces back to the form. Same pattern, and the same reasoning, as the workspace-gate leave in `identity-onboarding-form.client.tsx`. It lands directly on the app, so there is still no hero swap and no interstitial.
- `/auth/callback/{signin,signup}` move out of the `(auth)` route group, guarded by a route contract test.
- `(app)/layout.tsx` keeps one gate and one Suspense boundary. `AuthenticatedAppFrame` already made the identical session + workspace-access decision, so `WorkspaceAccessGate` and `AppAccessCheckingFallback` are deleted.
- Dead code removed with the branches that used it: `getAuthOAuthRedirect` (its `redirect` payload only ever appeared because `callbackURL` was set; the OAuth consent destination rides on `returnUrl`), both success toasts and their message keys in en/de/es.

## Analytics bug fixed on the way

`TRACKING.md` records that placing `fireGTMEvent.*` after a Better Auth call with a `callbackURL` makes it dead code, because the redirect plugin navigates inside its own fetch hook. That is why credential `login` showed zero in GA4 against 145 `login_area_form_start`.

The signup form still had exactly that shape. Its `fireGTMEvent.signUp("credential")`, its success toast **and** its `handleUtmConversion()` call all sat after the awaited `signUp.email` that had already started the navigation. So credential signups and their UTM attribution were very likely under-reporting the same way. Both now run before the leave, gated on a real session.

## Why this PR touches Core

`signUp.email`'s `callbackURL` does double duty. Better Auth also embeds it in the verification email as the post-verify destination, and Core has `sendOnSignUp` enabled. Dropping it client-side would have left that link pointing at `/` **relative to Core's own origin**, landing verified users on the API host instead of the product.

`apps/core/src/lib/verification-email-callback.ts` anchors a relative verification callback to the web app base URL and leaves an absolute one alone for Better Auth's own `originCheck` to validate against `trustedOrigins`. The destination no longer depends on what the browser sent. Five unit tests cover it.

Sign-in is unaffected by this: `requireEmailVerification` is false, so the sign-in branch that sends a verification email never runs.

**Behaviour note:** a signup carrying a `returnUrl` used to put that deep link in the verification email. Verification now always lands on the web app root. The common case is unchanged; carrying a deep link through an email opened days later was not worth keeping.

## Fixed in review

- The first push regressed sign-in on the preview: the form bounced back to `/signin` with the fields re-enabled. That was the `router.replace` soft nav described above. Passkey already carried the same latent bug on `main` and is fixed here too.
- Comments in the sign-in form, the callback page, `TRACKING.md`, `auth.utils.ts` and the verify-sokosumi docs still described the leave as a soft navigation. All corrected.
- The verify-sokosumi harness gave the credential leave only a fixed 1 s, because its `/auth/callback/` poll now exits immediately and credential never goes there. A slow session settle would have false-failed a good login as "stayed on auth page". It polls for the real condition now, bounded at 12 s.

## Trade-off to be aware of

SOK-797's acceptance criterion says a not-ready user "sees no sidebar/header". They now see the shell **skeleton** for the duration of the workspace-access round trip before being redirected to `/setup`. Accepted deliberately: ready users are the overwhelming majority and the blank frame hit every one of them on every login.

## Verification

Run at the merged head:

| Check | Result |
| --- | --- |
| `pnpm --filter web test` | 760 files, 6147 passed |
| `pnpm --filter core test` | 525 files, 5700 passed |
| `pnpm typecheck` | 20 tasks successful |
| `pnpm check` | clean |

Credential sign-in confirmed working in a browser on the preview deployment after the soft-nav fix.

Reviewer, on the preview: sign up with a fresh email and confirm you land in the app (or `/setup`) with no callback hop, then click the verification email and confirm it returns you to the **app**, not the API host. Sign in and confirm the form does not bounce.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(auth)/signin/components/form.tsx:107`, `apps/web/TRACKING.md:110-112`, `apps/web/src/app/(auth)/components/social-auth-callback.tsx:33`, `.cursor/skills/verify-sokosumi/features/sign-in.md` — still say “soft-navigate” / `router.replace`. The leave is `window.location.replace` (`finish-sign-in.client.ts`). Tests already assert that; comments can re-teach the bounce.
- Minor: `.cursor/skills/verify-sokosumi/bin/verify-sokosumi` (`assert_authed_url`, `wait_leave_auth_callback`) still describes a credential hop through `/auth/callback/signin`. Credential now stays on `/signin` until the full-document leave. Happy path (~200ms + getSession) fits the 1s wait; a slow session settle could false-fail as “stayed on auth page”.
- Minor: `apps/web/src/lib/auth/auth.utils.ts:73-76` (and Ably singleton comments) still say no sign-in replaces the document. Credential/passkey now do. Discard still runs before unload; harmless.
- Skipped / disagreed: OAuth/magic-link `SocialAuthCallback` still `router.replace`s after the callback full load — that is a new document, so the Next client cache that bounced credential/passkey does not apply. Signup still hard-redirects through `callbackURL` — called out as out of scope.

### Assessment

Credential/passkey finish in place, fire `login` only when a session exists, then `window.location.replace` so middleware re-runs with the cookie. Callback pages sit outside `(auth)` (contract-tested). Duplicate workspace gate is gone; one Suspense + shell skeleton. Off-origin `returnUrl` is sanitized. Remaining notes are comment/harness drift, not a bounce or blank-frame regression.

mrosberghaus:
Addressed in 754b3f9cd. The review ran against an earlier commit, so a couple of its notes were already stale.

**Stale comments — fixed.** All four sites still called the leave a soft navigation: `signin/components/form.tsx`, `components/social-auth-callback.tsx`, `TRACKING.md` and `auth.utils.ts`, plus the verify-sokosumi feature doc. Worth flagging that two earlier `TRACKING.md` edits of mine had silently failed to match, so that file was further out of date than the review could see; the whole "why events go missing" bullet is rewritten.

**Harness — fixed, and this one had teeth.** `wait_leave_auth_callback` only polled while the URL contained `/auth/callback/`. Credential no longer goes there, so it returned on the first iteration and the entire grace period for the in-place leave was the fixed `wait 1000`. That leave is at least 200 ms of session settle, plus a `getSession` round trip, plus a full document load. A slow settle would have reported "stayed on auth page" on a perfectly good login. Replaced with `wait_leave_auth_page`, which polls until the URL is off both the auth entry pages and the callback path, bounded at 12 s so a genuinely rejected login still reports instead of hanging.

**Ably comment — fixed.** `auth.utils.ts` claimed no sign-in replaces the document. Credential and passkey now do, where the discard is a harmless no-op because it runs before unload. The callback page and the re-authentication dialog still need it, so the code is unchanged and only the doc moved.

**Agreed on the disagreement.** `SocialAuthCallback` keeping `router.replace` is correct: it runs in a fresh document, so there is no pre-login router cache to be served. I put that reasoning in the comment so it does not get "fixed" later.

**Signup note is stale — now done.** It also surfaced something the review could not have seen. `signUp.email`'s `callbackURL` does double duty: Better Auth embeds it in the verification email as the post-verify destination, and Core has `sendOnSignUp` enabled. Dropping it would have left that link pointing at `/` relative to Core's origin, landing verified users on the API host. Core now anchors a relative verification callback to the web app and leaves an absolute one for its own `originCheck`, so the destination no longer depends on what the browser sent.

Verified locally: web 758 files / 6099 tests, core 524 files / 5600 tests, `pnpm typecheck` 20 tasks, `pnpm check` clean.

mrosberghaus:
## Grok review

### Fixed
- Important: `anchorVerificationCallbackToWebApp` treated `//host` (and `/\host`) as a relative path, so `new URL` promoted it to `https://host` inside the verification email. Off-origin resolutions now fall back to the web app root (`43769e85c` — `apps/core/src/lib/verification-email-callback.ts`).

### Remaining
- Minor: `apps/web/src/app/(auth)/signin/components/form.test.tsx`: test title still says "soft-navigates"; the leave is `window.location.replace`.
- Skipped / disagreed: SOK-797 "no sidebar/header" vs shell skeleton — documented trade-off in the PR body, not a regression of the gate itself (`AuthenticatedAppFrame` still fail-closes to `/setup`).
- Skipped / disagreed: GTM `login`/`sign_up` immediately before `location.replace` — same sendBeacon race as `begin_checkout`; the actual GA4-zero bug was firing *after* a `callbackURL` hard redirect.

### Assessment
Credential/passkey in-place finish, callback pages outside `(auth)`, and Core verification anchoring match the PR; remaining items are naming and accepted UX/analytics trade-offs.

mrosberghaus:
Addressed in 8f5788417.

**The protocol-relative finding was a real bug and mine.** `callbackUrl.startsWith("/")` accepted `//evil.example`, and `new URL("//evil.example", webAppBaseUrl)` promotes that to `https://evil.example/`, so the verification email could have carried an off-origin link. I verified the pushed fix against the hostile inputs rather than assuming it covered them:

| `callbackURL` | resolved destination |
| --- | --- |
| `//evil.example` | `https://app…/` |
| `/\evil.example` | `https://app…/` |
| `//evil.example/x` | `https://app…/` |
| `/\/evil.example` | `https://app…/` |
| `/chat` | `https://app…/chat` |

Resolving first and then comparing `origin` is the right shape: it rejects by outcome instead of trying to enumerate the syntax that escapes. Two tests cover it.

**Stale test title — fixed.** `form.test.tsx:293` now reads "leaves for returnUrl" instead of "soft-navigates". Assertions were already correct; only the name lied. Grepped the rest of the tree: the other `soft-nav` mentions are pre-existing chat and Instant Nav comments, unrelated to this branch.

**Agreed on both skipped items.**

The SOK-797 note is a documented trade-off, not a gate regression. `AuthenticatedAppFrame` still fail-closes to `/setup` on a not-ready user or a workspace-access error; what changed is only which fallback paints during the round trip.

The GTM point is correct and worth stating precisely, because it is easy to conflate the two. The GA4-zero bug was firing *after* Better Auth had already navigated, which is unreachable code, not a race. Firing before `location.replace` is the ordinary `sendBeacon` race that `begin_checkout` already lives with, and `TRACKING.md` documents it as "push before navigating, never after". Same class as the existing code, not a new hazard.

Verified: web `form.test.tsx` 12 passed, core `verification-email-callback.test.ts` 7 passed, `pnpm check` clean.

PR #4675, opened 2026-09-16
feat(apple): create channels from the native sidebar

mrosberghaus:
## Create channel

Organization members can create a channel from the native sidebar toolbar. The two-step sheet validates the handle, display name, topic and visibility, then adds all organization members or selected people, coworkers and the personal assistant. External visibility is available to owners/admins. Conflicts return to details; other failures retain the draft for retry.

Reuses the Direct recipient loader and room-create transport. Workspace changes invalidate pending requests, duplicate submissions are blocked, and completion cannot steal navigation from a newer room selection. Shared models/networking remain UI-free and iOS 17-compatible.

### Web sources and scope
- [Create channel dialog](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/create-channel-dialog.tsx)
- [Wizard validation](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/create-channel-wizard.ts)
- [Participant selection](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/app/%28app%29/chat/components/participant-checkboxes.tsx)

PARITY row 29 only. Channel browsing, management and invitations remain later slices. No web, Core contract or dependency changes. The Apple OpenAPI snapshot selects two existing read operations; existing paths and schemas remain unchanged.

### Validation
- macOS build and full app test suite pass; four native rendering cases pass after capture synchronization.
- All five package suites pass: Chat 423, Workspace 84, Auth 35, Realtime 27, CoreAPI 1.
- iOS 17 Workspace package build passes.
- Strict uncached SwiftLint, uncached SwiftFormat, Biome and repository typecheck pass.
- Inspected both sheet steps in light/dark appearance. Test hosts exited.

To try: select an organization, choose Create channel in the toolbar, enter a free handle, choose visibility, advance, select everyone or specific participants, and create. Verify the new channel opens. Try a taken handle and switch workspace while a request is pending. Live Core creation, physical keyboard navigation and VoiceOver acceptance remain unverified; model/transport tests cover the state and request paths.

### Native captures
Details
Participants


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/Sokosumi/Chat/Sidebar/CreateChannelView.swift`: slug focus uses `.onAppear` instead of `.defaultFocus` / `.task` (Start Direct already focuses inside `.task`; `onAppear` can miss).
- Minor: `apps/apple/Sokosumi/Chat/Sidebar/ConversationSidebarView.swift`: two `.sheet(item:)` modifiers. macOS window-modal sheets block the other toolbar button in practice; an enum sheet would still match the SwiftUI sheet skill.
- Minor: `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/ChannelCreation.swift`: load/create fallbacks skip `friendlyMessage(for:)` so URLErrors stay generic instead of the Direct picker's network copy.
- Minor: `CreateChannelView.swift`: fixed `480×320` / list `240` frames will clip at large Dynamic Type.
- Skipped / disagreed: migrating new `ChannelCreation` to `@Observable` — matches `DirectRecipientPicker` / `WorkspaceState` and the iOS 17 package baseline.

### Assessment
Ready to merge: Yes
Slice 29 matches the web wizard (slug rules, debounce, external role gate, all-vs-specific members, slug-taken bounce, duplicate/nav guards) without Core/web contract changes; leftover notes are polish.

mrosberghaus:
Checked the four minors against the implementation. Two fixes are in 058b4b1db:

- Handle focus now runs in the field’s `.task`, when the field is mounted after loading, rather than the sheet’s initial `.onAppear`. It also runs when returning to details.
- Load/create transport failures now use the existing `friendlyMessage(for:)`. Added coverage for offline loading and timed-out creation.

Keeping the two sheet presentations: native window-modal presentation blocks the other toolbar action, and both are dismissed by the same workspace-context invalidation. An enum would be a structural preference rather than a verified behavior fix here.

The Form and List have scrollable viewports, so fixed viewport height alone does not establish clipped content. The existing light/dark fixtures render correctly, but do not prove large Dynamic Type behavior. That case remains explicitly unverified; I have not claimed it fixed or replaced the native controls. The portable model remains consistent with the existing ObservableObject architecture.

Validation: 424 Chat tests, all four native sheet rendering cases, strict uncached SwiftLint, uncached SwiftFormat, Biome and repository typecheck pass. Test-host app exited. CI was green on the previous head and is rerunning for this follow-up.

PR #4678, opened 2026-09-16
revert(core): restore delegated organization access

PatrickTobler:
## Problem

PR #4621 added workspace-context binding to organization and workspace reads made with coworker and Soko Bot API keys.

Before #4621, an authorized coworker could use its Bearer key with `X-Context-User-Id: <user-id>` to act for that user. The integration could:

- fetch the user with `GET /v1/users/{id}`
- fetch the user credits with `GET /v1/users/{id}/credits`
- fetch every organization membership with `GET /v1/users/{id}/organizations`
- use those memberships to read organization and workspace data available to that user

Our external customer console depends on this behavior. It is workspace-first: after a user signs in, it lists the Sokosumi organizations for that user so the user can select a workspace.

After #4621, `GET /v1/users/{id}/organizations` returned only the organization in `X-Context-Organization-Id`. With personal context, where the organization id is null, it returned `data: []`. The same change also rejected organization and workspace reads outside the current workspace context of the key.

The delegated user was valid, the coworker grant was valid, and the organization memberships still existed. However, the console could no longer discover or access them. Users saw only "Personal" and could not open their organization workspaces.

This was not caused by a header rename. `X-Context-User-Id` and `X-Context-Organization-Id` remain the canonical headers, while the existing legacy delegation headers remain supported.

## Solution

Fully revert #4621, not only the organization-list filter.

This restores the pre-#4621 contract for coworker and Soko Bot keys:

- `GET /v1/users/{id}/organizations` returns every organization membership of the delegated user.
- Organization metadata and slug lookups again use delegated-user membership without restricting the target to the current workspace context of the key.
- Organization credits, member data, and `DESIGN.md` access again use delegated-user membership.
- Workspace and calendar reads again use delegated-user access without additional context-workspace binding.
- Existing coworker user-binding and grant checks remain unchanged.
- Session-user behavior remains unchanged.

The implementation reverses all 22 files changed by #4621. It also refreshes the Apple Core API snapshot added after #4621, for 23 changed files in this PR.

## Security impact

This intentionally reopens SOK-1020 / PERM-N02.

A coworker or Soko Bot key scoped to Workspace A can again act through a delegated user to discover and read Workspace B resources when that user belongs to both organizations. This is the previous production contract required by current external integrations.

## Verification

- Core tests: 520 files passed, 3 skipped; 5,562 tests passed, 11 skipped
- Core typecheck passed
- Web typecheck passed
- Core Biome check passed
- Web and Apple Core API artifacts regenerated


PatrickTobler:
@coderabbitai review

PatrickTobler:
@coderabbitai review

Scope changed to a full reversal of #4621. Please review the new head commit.

PatrickTobler:
@schaier-io @mrosberghaus please discuss if this was the correct security fix or not in the first place

PR #4679, opened 2026-09-16
fix(web): stop the running glyph spinning where nothing is running

schaier-io:
Stacked on #4671.

## The problem

The spin is the one part of a status marker that makes a claim about time: it says the work is in flight as you read it. Two callers rendered that claim where it is false.

**The task activity feed shows history.** "Hannah changed status to Running, 5h ago" is a value someone set hours ago. The badge spun as though the task were running now.

**The status picker's dropdown lists options.** The Running row spun there too, so a menu of statuses you could choose animated as though one of them were underway.

## The change

`StatusMarker` takes `live`, defaulting to `true`, and spins only when the caller says the glyph stands for something happening right now. The two callers above pass `false`. They keep the same glyph, the same box, the same colour and the same stroke; only the motion goes.

`TaskStatusBadge` forwards the same flag, with the same default, so every existing caller is unchanged.

## What still spins

The picker's **trigger** is untouched. It spins while a status change is pending, and it spins when the task really is running. Both are claims about now, and both hold.

`motion-reduce:animate-none` is unchanged: the spin was already off for readers who ask for less motion.

## Verification

VERIFIED by mutation, not by a green run alone. Dropping the `live` term from the spin condition fails exactly one test:

```
× holds the running glyph still when the badge is not live
  Tests  1 failed | 4 passed (5)
```

Restoring it passes 5 of 5.

Full suite on the rebased branch:

```
pnpm --filter web test      6085 passed | 9 failed | 1 skipped
the 9, re-run alone         Tests  22 passed (22)   Test Files  4 passed (4)

pnpm typecheck              Tasks: 20 successful, 20 total
```

The 9 are the repo's known load flake: every one is pinned near 5000ms, they sit in the share page, project page, sidebar cache and admin coworker suites, none of them renders a status marker, and all pass in isolation.

Blind spot: the tests assert the class on the rendered glyph. They prove `animate-spin` is absent, not that the browser paints no motion from some other rule.

A second commit covers one surface the first pass missed. The share page
renders the same status history as the activity feed, so a status someone
set hours ago was still spinning there. It now passes `live={false}` like
the other history rows, and a guard matches every `<TaskStatusBadge>`
element: if it reads `event.status`, it must carry `live={false}`.


schaier-io:

PR #4681, opened 2026-09-16
refactor(web): give the app one horizontal gutter so rules reach the edge

schaier-io:
A rule that spans the view escapes the page's horizontal padding with a
negative margin. That only works when it escapes every level of padding
above it, and the app had two.

`main[data-app-main]` sets `p-4`, so every authenticated page carries a
16px gutter it cannot see. Some pages then added a second one. The Team
page set `px-4 lg:px-6` and its rule escaped exactly that much, so the
rule stopped 16px short of the edge on both sides. That is the gap you
can see under the Team heading.

The arithmetic ran the other way too. `-mx-6` is 24px and was correct on
the two pages that added `px-2` on top of the shell's 16px. On every
other page, which adds nothing, the same class overshot the gutter by
8px and was saved only by `overflow-x-hidden` clipping it.

So the gutter is now singular. `main` keeps its `p-4` and owns the only
horizontal gutter in the app. Five pages that added a second one drop
it: soko-bots, history, tasks, and the chat room page. Every bleeding
rule becomes `-mx-4`, which is the one value that now matches.

`main` also gains `scrollbar-gutter: stable`. It is the scroll
container, so without that the scrollbar takes its width out of the
content box on the right only, and every full-bleed rule stopped
further from the right edge than the left. That asymmetry was visible
next to the 16px shortfall and is a separate cause.

A guard test reads every .tsx under apps/web/src and fails on a
negative horizontal margin that is not -mx-4 on an element that also
paints a rule. It checks the value, not whether a given element should
bleed: a centred column inside a max-w wrapper keeps its own padding and
has no rule crossing the view.

Two tests asserted the old 24px and now assert 16px. They are converted
here, in the layer that changed the value.

Verified by mutation: putting -mx-6 back on the history list fails the
guard with the file, the line and the offending class. Restoring passes.

VERIFIED, re-measured on this branch after the rebase onto main 8459b24:

  pnpm --filter web test   762 files, 6119 passed, 1 skipped, exit 0
  pnpm typecheck           20 successful, 20 total
  pnpm check               4136 files, no fixes applied

Blind spot: the guard reads class strings, so it cannot see a gutter that
is composed at runtime or applied from a CSS file.

---


schaier-io:

PR #4682, opened 2026-09-16
fix(web): make the wizard divider span the panel

schaier-io:
The vertical rule between the coworker rail and the detail column is a
border on the detail column, so it runs that column's height. The modal
body held the vertical padding, which left the rule 12px short at the top
and 12px short at the bottom.

Move that padding from the body to the two columns at md, where the
divider shows. Measured against these exact class strings: before, the
divider was 228px inside a 252px panel; after, it is 252px with both gaps
at 0, and the content keeps its 16px inset.

Add a guard that pins the mechanism for this panel. Whether an ancestor
holds vertical padding is not something a regex can decide in general, so
the guard does not try. It pins the one panel that had the bug: the body
gives its vertical padding up at md, and the column that paints the
divider takes it.

The two other vertical dividers in the app were checked and need no
change. `coworker-gallery-section.tsx:318` sits in a grid with no
vertical padding, and its own `md:pt-6 md:pb-10` is inside its border, so
it does not shorten it. `offer-card.tsx:632` sits in a grid inside a
`DialogContent` that is `p-0`.

VERIFIED on this branch after the rebase onto main 8459b24:

  pnpm --filter web test   762 files, 6119 passed, 1 skipped, exit 0
  pnpm typecheck           20 successful, 20 total
  pnpm check               4136 files, no fixes applied

Verified by mutation: putting `py-3` back on the wizard body and
`md:pt-1` back on both dividers fails the guard, 2 of 2. Restoring
passes, 2 of 2.

Blind spot: the divider geometry was measured against these exact class
strings in a browser, not inside the signed-in modal. Local auth is not
available in this checkout (VERIFY_SOKOSUMI_* unset, no fixture users).

---


schaier-io:

PR #4685, opened 2026-09-16
refactor(core): fold read-for-reference into notifications read route

mrosberghaus:
## Summary

- `PATCH /v1/notifications/read` now takes a union body: `{ ids }` (1..25) **or** `{ kind: "TASK" | "JOB", referenceId }` (non-empty). Response stays `{ count }`.
- The ids form still schedules `publishClearedNotifications`; the reference form does not republish (chat stays excluded, same reasoning as before).
- Both variants are strict objects, so a body carrying ids *and* a reference gets a 422 instead of matching whichever variant comes first.
- `PATCH /v1/notifications/read-for-reference` is removed (route, tests, mount). Its tests now live in `read/patch.test.ts`.
- Web: regenerated Core client; `notificationsBrowserClient.patchNotificationsReadForReference` → `patchNotificationsRead`; `MarkNotificationsRead` and its tests updated.

No user-facing change: the task and job pages still mark their notifications read when opened. Apple calls neither route (checked with grep).

## Verification

- `pnpm --filter core test src/routes/v1/notifications src/services/notification-follow-up-sync` passes
- `pnpm --filter web test` for the browser client and `mark-notifications-read` tests passes
- `pnpm --filter web typecheck`, `pnpm typecheck`, `pnpm check` pass


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/src/routes/v1/notifications/read/patch.test.ts`: the “publishes no clear” case mocks zero rows, so a later `waitUntil` gated on `clearedRoomIds.length` would still pass. Return a TASK/JOB row and assert no publish.
- Minor: `apps/web/src/lib/clients/core.notifications.browser.client.test.ts`: `patchNotificationsRead` now takes the union body; only `{ kind, referenceId }` is asserted, not `{ ids }`.
- Skipped: no-body / extra-key 422 cases on a single variant — schema + mixed-body test already cover the exclusive union.

### Assessment
Ready to merge: Yes

Union is exclusive (`strictObject` + 422 on ids+reference), chat stays refused, ids still `waitUntil(publishClearedNotifications)`, reference does not republish, web call sites and generated client point at `PATCH /notifications/read`, old `/read-for-reference` is gone.

mrosberghaus:
Addressed both Minors from the Grok review in 6f062f36d:
- `read/patch.test.ts`: the reference "publishes no clear" case now writes a TASK row, so a republish gated on cleared rows would fail it.
- `core.notifications.browser.client.test.ts`: `patchNotificationsRead` is asserted with `{ ids }` as well as `{ kind, referenceId }`.

Agree with skipping the single-variant no-body / extra-key cases.

PR #4687, opened 2026-09-16
feat(web): add a Needs you view to the Notification Center

mrosberghaus:

A third view, **Needs you**, beside All and Unread in the Notification Center (header panel and full page). It lists only the rows whose request is still open: a task or job paused on input, a pending vendor grant or coworker access request. Reading a row leaves it there; answering the request removes it on the next fetch. The tab shows its own live count, and an empty view says nothing is waiting.

Same lens rules as Unread: switching never marks anything read, the choice is shared between panel and page for the session, and it resets to All on reload.

## How

- **Core** `GET /v1/notifications` takes `needsAction=true`. "Still waiting" is decided from the record each row points at (task status, computed job status, grant / access status), not from stored metadata, one row per record (the newest).
- **Core** `GET /v1/notifications/counts` returns `{ unread, needsAction }` and replaces `GET /v1/notifications/unread-count`. Web was its only caller (Apple does not call it); one request now feeds the bell and both tabs.
- **`@sokosumi/utils`** owns the four actionable message keys (`NEEDS_ACTION_MESSAGE_KEYS`); Core filters with them, web uses them to keep rows that never asked out of the view when they arrive live.
- **Web** one more `TabsTrigger`, a third `view` value, `needsActionCount` next to `unreadCount`. No new list component. Copy in en / de / es. `CONTEXT.md` gains the **Needs you** term.

Spec: the latest comment on [SOK-1097](https://linear.app/masumi/issue/SOK-1097/needs-you-view-in-the-notification-center).

Closes SOK-1097.

## Verification

- `pnpm --filter @sokosumi/core test src/routes/v1/notifications src/helpers` ✅
- `pnpm --filter web test src/components/notifications src/contexts src/lib/clients` ✅
- `pnpm --filter @sokosumi/utils test` ✅
- `pnpm typecheck` ✅, `pnpm check` ✅, `pnpm --filter web messages:parity` ✅
- Full `pnpm test`: running locally at the time of opening; CI covers it.

## Out of scope (per the issue)

Filters by kind, combining views, other task pauses (approval / sign-in / credits), push or email for actionable rows, a needs-action badge on the bell.


mrosberghaus:
## Grok review

### Fixed
- Important: accepting a grant or access request removed the row but left the Needs you tab count high (`5c89a5269` — `apps/web/src/contexts/notification-state.ts`)
- Important: an empty loaded page then said nothing was waiting while that count was still > 0 (`5c89a5269` — `apps/web/src/components/notifications/notification-center-list.tsx`)

### Remaining
- Minor: `apps/core/src/helpers/notification-feed.ts`: `findNeedsActionNotificationIds` loads every historical row with those four keys on each list/counts request. Same shape as the stale grant lookups; fine until this route shows up in traces.
- Minor: `apps/web/src/contexts/notification-state.ts`: a Needs you refresh keeps older loaded rows by message key only, so an answered request on a later page can stay until the view is switched.
- Skipped / disagreed: pick the newest *feed-visible* row per record — a later silenced (`inApp: false`) event would then keep nagging from an older in-app row. Newest-then-exclude matches “the newest row per request” and the reader’s current in-app preference.
- Skipped: other task pauses (approval / sign-in / credits) — out of scope per the issue.

### Assessment
Core’s `needsAction` list/counts match the spec (record still waiting, one newest row, read does not dismiss). The two live-count bugs on the tab are fixed.

mrosberghaus:
Checked `5c89a5269` against the code:

- **Count on remove**: correct. `removeNotification` is called only from the vendor-grant and coworker-access action rows after accept/deny, so the row's record has stopped waiting at that point; a decrement there matches what the next `/counts` would return. A grant or access row is written once per record, so the removed row is the one the count holds.
- **Empty page with a count > 0**: correct, and the `narrowedCount` generalisation is the same rule Unread already applied.

Web tests (147) and typecheck pass on the merged head.

Remaining notes, left as they are:
- Historical rows per request: bounded by the reader's actionable rows and the same shape as the stale-grant lookups; the spec's follow-up if it shows in traces is a resolved-at on the row.
- Later-page rows kept by key after a refresh: the same range rule Unread uses; a refetch or view switch reconciles.

PR #4688, opened 2026-09-16
refactor(web): fold the info role into working and move it off the brand hue

schaier-io:
The brand moved from hue 263.5 to 201.8, which left three roles inside 15
degrees of one another: --primary at 201.8, --status-working at 205, and
--semantic-info at 193. Anything that close reads as one colour, so the
Soko Bot badge could not tell "running" from "delegate".

--semantic-info is retired rather than moved. All five of its consumers
meant "in flight right now" and none meant "here is information": an
active turn, a processing outcome twice, a RUNNING count, and the badge's
RUNNING and STARTING. It was the working role under a second name, so the
duplicate goes and every consumer reads --status-working.

Working then moves from hue 205 to 300. Sweeping all 360 hues against the
task tints and the badge tones at once, 300 is the only hue whose nearest
neighbour in either set stays above 7, at 7.08 in each. Violet 272 scores
better against the badge alone and fails against the task set, landing
3.76 from --status-external at 240.

The badge fills move from -quinary to -quaternary in the same change. The
quinary steps sit at 93 to 95 percent lightness, where all 15 pairs
measured under dE 6.0 and the worst under 1.0. At -quaternary that is 1
pair of 15, and the pair that stays close is neutral against accent, which
never competed as a category.

Magenta carries less luminance than blue at the same HSL lightness, so
--status-working-label drops from 35.4 to 33.4 percent. At 35.4 it
measured 4.39:1 on its own fill and missed the 4.5:1 that 12px badge text
needs. It now measures 4.80:1.

Every running job badge in the app turns magenta, not only the Soko Bot
ones, because --status-working also paints agent-job-status-badge and the
task-created celebration. That is the intended consequence of one role
having one colour.

The guard pins the separation, not the hue: move either role and it says
so, move one far enough and it passes whatever value you pick.

---


schaier-io:

PR #4689, opened 2026-09-16
fix(core): verify current organization membership on every request

schaier-io:
Closes SOK-1005 (security finding PERM-02, High).

## What this changes

Core took the organization from the session and never rechecked it. `apps/core/src/middleware/auth.ts` copies `session.activeOrganizationId` into the auth context, and the organization middleware only looked up membership when that value was empty. Better Auth clears the field from the remover's own session only, so removing someone else left their other sessions still naming the organization. Any route without a seat gate served them, and only five routers carry `requireOrganizationProductSeat`.

Two commits:

1. **Verify membership at request time.** `organizationContextMiddleware` (renamed from `organizationHeaderMiddleware`, which no longer described it) reads the caller's current `Member` row whenever the request carries an organization. The lookup uses the `@@unique([userId, organizationId])` index, the same shape `coworker-context.ts` already uses.
2. **Move the caller instead of locking them out.** A removed member's session is switched to their personal workspace through Better Auth `setActiveOrganization`, and the request continues. Rejecting instead would have locked them out of all of `/v1`, the organization list and `/v1/admin/*` included, for as long as the stale field survived.

The `X-Organization-Slug` path still answers 403. Naming an organization you do not belong to is an error, not stale state.

Soko Bot and coworker paths are deliberately untouched. The branch sits inside `isUserAuthContext`, and coworker keys already do their own request-time membership read in `coworker-context.ts`.

## Why the switch goes through Better Auth

A direct `session` row update would be invisible for the lifetime of the session cookie cache (`apps/core/src/lib/auth.ts:187-191`, 300s default). Going through `setActiveOrganization` writes the row and re-signs the cookie. The write is guarded by a session read with `disableCookieCache: true`, so it happens once rather than on every request, and a request that started before a deliberate switch to another organization cannot write `null` over that choice.

## Verified before choosing this shape

- A null organization resolves to the personal workspace at `workspace.repository.ts:137-148`, never to an unfiltered query.
- No Prisma `where` turns a session-derived null into `undefined`. The one real filter-drop, `enterprise/contracts/get.ts:57`, reads its value from a query parameter, not the session.
- Credit scope is explicit: `credit-bucket-scope.ts` emits `organizationId IS NULL` for personal.

## Known limits

- The refreshed cookie reaches the browser only on a browser-direct call. Web calls Core server side and does not relay `Set-Cookie` (`apps/web/src/lib/clients/core.client.ts:24-27`), and a handler returning a raw `Response`, such as the chat SSE stream, drops it. The row is written once either way; the cookie lags, and the middleware drops the organization on every request until the cache expires.
- A user with no personal workspace still gets a 403, now reading "Workspace is missing" (`workspace.ts:31-33`).
- The organization seat gate is skipped once the organization is gone (`organization-product-seat.ts:29-31`). Every seat-gated route is workspace-scoped, so this grants nothing beyond the caller's own workspace.
- One extra indexed `member.findUnique` per authenticated session request.

## Out of scope, reported not fixed

- SOK-1005 item 2: `resolveWorkspaceForContext` still ignores the user on the organization branch.
- SOK-1005 item 3: member removal still does not clear and revoke the removed user's other sessions. That is what ends the stale-cookie window this PR works around.
- Soko Bot keys are still not re-checked; `auth.ts` puts `apiKey.sokoBot.workspace.organizationId` on the context. SOK-1006 owns that.
- 28 route files under `apps/core/src/routes/v1/` declare no `403` in their OpenAPI contract. That gap pre-dates this change: the header path already threw `forbidden`. Worth its own issue.

## Verification

- `pnpm --filter core test` on the rebased tree: `Test Files 530 passed | 6 skipped (536)`, `Tests 5838 passed | 37 skipped (5875)`.
- `pnpm typecheck`: 20 of 20 tasks successful. `biome check .`: clean.
- Rebased onto main (158 commits) with no conflicts. `pnpm prisma:generate` and a `@sokosumi/database` rebuild were needed afterwards before the suite reflected reality.
- Mutation checks, not just green runs: reverting the whole feature fails four of the new tests; removing only the guard's early return fails exactly the two guard tests (`Tests 2 failed | 8 passed`).

## Tests

- `organization.test.ts`: membership verified on the session path; the session switched to personal when the membership is gone; the organization still dropped when Better Auth is down; no write when the session already moved to another organization or is already clear; the header ignored when the session carries an organization.
- `jobs/[id]/get.test.ts`: route-level proof on a route with no seat gate. It asserts the organization the route observed was `null` and that the job query ran against the personal workspace id.
- Ten suites that model a member now answer the membership read, eight through the shared `test-fixtures/organization-membership.ts` stub.


mrosberghaus (commented):
PERM-02 is closed for user sessions. Ready to merge. SOK-1005 items 2 (`resolveWorkspaceForContext` still ignores the user on the org branch) and 3 (revoke every session at removal) stay residual, as the PR says.

Four nits from a local pass. Three applied in f342ceee0; one declined. Inline notes below.

apps/core/src/routes/v1/jobs/[id]/get.test.ts
```diff
@@ -272,6 +330,49 @@ describe("GET /jobs/{id}", () => {
...
+    getSessionMock.mockResolvedValue({
+      session: { activeOrganizationId: "org_123" },
+      user: { id: "user_123" },
+    });
+    observedOrganizationIds.current = [];
+  });
+
+  it("reads the personal workspace when the organization membership was removed", async () => {
+    // Production jobs router also sets requireOrganizationProductSeat.
+    // This harness does not, so membership is the only control here.
```
Comment (mrosberghaus) [inline apps/core/src/routes/v1/jobs/[id]/get.test.ts:347]: Nit (applied in f342ceee0): the previous comment said this route has no seat gate. Production `jobs/index.ts` sets `requireOrganizationProductSeat: true`. The harness is `new OpenAPIHonoWithAuth()` without that option, so membership is the only control on this request, not on the production route.

Real no-seat org-scoped routers remain `agents/[id]/jobs` (SOK-1004), tools, workspaces, chats. The personal-workspace `where` assertion still holds.

apps/core/src/routes/v1/jobs/[id]/get.test.ts
```diff
@@ -272,6 +330,49 @@ describe("GET /jobs/{id}", () => {
...
+    });
+    expect(observedOrganizationIds.current).toEqual([null]);
+    expect(jobFindFirstMock).toHaveBeenCalledWith({
+      where: {
+        id: "job_123",
+        workspaceId: PERSONAL_WORKSPACE_ID,
+      },
+      include: expect.any(Object),
+    });
+    expect(response.status).toBe(200);
```
Comment (mrosberghaus) [inline apps/core/src/routes/v1/jobs/[id]/get.test.ts:375]: Nit declined: do not change this to `404`.

The test is the continue-to-personal path from the second commit, not a reject. Production would 404 an org job after the workspace strip, but this mock always returns a job, so `200` is a harness artifact.

The security proof is `workspaceId: PERSONAL_WORKSPACE_ID` and `observedOrganizationIds === [null]`. Asserting `404` would recast the test as rejection, which is the behaviour that commit walked away from.

apps/core/src/lib/hono.ts
```diff
@@ -51,12 +51,12 @@ export type EnvVariables = {
  *
  * Auth middleware is automatically applied - all routes are protected
  * Coworker context middleware runs after auth to attach optional workspace scope from headers.
- * Organization header middleware is also applied to set organizationId from X-Organization-Slug header
+ * Organization context middleware verifies session membership, or sets organizationId from X-Organization-Slug
```
mrosberghaus, line 54: Nit (applied in f342ceee0): leftover "Organization header middleware" after the rename to `organizationContextMiddleware`. It now also verifies session membership, not only `X-Organization-Slug`.

apps/core/src/middleware/organization.test.ts
```diff
@@ -141,9 +173,97 @@ describe("organizationHeaderMiddleware", () => {
...
+    const app = createUserApp("org_removed");
+    const response = await app.request("http://localhost/");
+
+    expect(response.status).toBe(200);
+    expect(await response.json()).toEqual({
+      actor: "user",
+      userId: "user_123",
+      organizationId: null,
+      role: "user",
+    });
```
mrosberghaus, line 227: Nit (applied in f342ceee0): these two guards skipped the session write but never checked that *this request* still has `organizationId: null`.

`switchSessionToPersonalWorkspace` can return early while `setAuthContext` still nulls the request. A refactor that skipped the context drop when it skipped the write would keep org A on the request and stay green.

Same assertion added on "does not write again once the session is already clear".

PR #4711, opened 2026-09-17
feat(chat): add private self-direct notes

mrosberghaus:
Adds **Message yourself** to New Direct so each user can keep private work notes in one conversation across their workspaces, including accounts without a personal workspace. The picker remains usable when recipient loading fails; the sidebar and header show **You** with the user's avatar, and the empty state explains private notes in English, German, and Spanish.

Core creates or reuses a personal Direct containing only the caller, preserves its identity across concurrent opens and archive restoration, and exposes a generated `isSelfDirect` DTO flag. Existing membership authorization and message flows keep notes private without self-notifications or AI dispatch. No database migration or dependency changes.

Implements [SOK-1099](https://linear.app/masumi/issue/SOK-1099/message-yourself-for-private-work-notes-and-to-dos).

## Validation

- `pnpm check` and `pnpm typecheck` passed, including commit hooks.
- Full Core suite passed: 5,754 tests, including nine real Postgres integration tests against a disposable migrated database. Final focused Core rerun: 66 passed.
- Final focused Web tests passed: 50 tests covering the picker, action boundary, header, avatar, and display name. Translation parity passed.
- `pnpm test` was run. Web reported 6,209 passes and one admin action test failure; that unchanged test passed on an isolated rerun. The unchanged CLI TUI test `test/tui/status-app.test.ts` remained pending when rerun separately. The interrupted Core suite was subsequently run to completion as noted above.
- Separate standards and spec code reviews found no actionable issues. Browser smoke testing is pending with the author, as requested.


mrosberghaus:
## Grok review

Reviewed `d329d293b..ace082264` against SOK-1099 (authorization, visibility, other users cannot read notes).

### Fixed

### Remaining
- Minor: browser smoke from the spec (open Message yourself, send, reload, switch workspaces) is still pending with the author.
- Minor: `apps/core/src/routes/v1/chats/rooms/helpers.test.ts` never asserts `buildDirectParticipantRoomKey` for `memberUserIds: [currentUserId]` → `direct:self:…` (covered by route + Postgres tests).
- Minor: `isPrismaUniqueViolationOnField` matches `directKey` as a substring of the index name. Fine for this index; a looser field name would false-positive.
- Skipped: Apple/CLI “You” labeling — spec out of scope (no new SwiftUI). Extra `isSelfDirect` on ChatRoom is ignored by the snapshot client.
- Skipped: coworker/Soko Bot create in the integration test uses IDs that are not in the database, so 400/403 can come from missing actor rather than Self Direct rejection. User-actor routing still cannot open that shape (`memberUserIds` is emptied and the coworker/bot is the other participant).

### Assessment

Membership stays the auth boundary; identity is `direct:self:${userId}` (not leftover member count); lists/reads carry derived `isSelfDirect`; another user, org admin, coworker, and Soko Bot cannot see the room; send skips self notifications and AI dispatch. CI is green. Browser proof is the only spec item still outstanding.

mrosberghaus:
Follow-up to [the Grok review](https://github.com/masumi-network/sokosumi/pull/4711#issuecomment-5712581330): addressed the test gaps in c0b8aaa433c2a52b48a3f1c02d41498139c09909.

- Added a direct assertion that `buildDirectParticipantRoomKey` returns `direct:self:user_self` when the caller is the only participant.
- Replaced nonexistent coworker/Soko Bot IDs with persisted, valid actors. The integration test now proves that their create requests succeed with separate AI Directs (`isSelfDirect: false`) while access to the owner's Self Direct remains denied. Before adding the fixtures, the stronger assertion reproduced the misleading 400 responses.
- Left the substring matcher unchanged after checking current index names and callers: no current collision was found. Apple/CLI labeling remains outside this issue's scope.
- Validation: 86 helper tests and all nine Postgres integration tests passed; repository formatting checks and typecheck passed. Browser smoke remains with the author, as requested.

PR #4713, opened 2026-09-17
fix(web): keep collapsed sidebar rail scrollable

mrosberghaus:
## What

The collapsed (icon) sidebar could not scroll, so channels and DMs below the fold were unreachable. The rail now scrolls vertically like the expanded sidebar, with the scrollbar hidden so the 3.5rem rail keeps its full width for icons.

## Cause

`SidebarContent` carried shadcn's default `group-data-[collapsible=icon]:overflow-hidden`. Replaced with `group-data-[collapsible=icon]:[scrollbar-width:none]` while keeping `overflow-y-auto`.

## Verification

- New regression test `apps/web/src/components/ui/__tests__/sidebar-collapsed-scroll.test.tsx` (red before, green after)
- `pnpm check`, `pnpm typecheck` (pre-commit hook), `pnpm --filter web test` on sidebar tests


mrosberghaus:
## Grok review

Reviewed `e11a66cb4bb6001bbb8fc340616be01fe12ff328..edad9bc6f1250551cfba06e4effa2b02035b7ef6` (`SidebarContent` overflow + regression test).

### Fixed

### Remaining
- Minor: `apps/web/src/components/ui/sidebar.tsx`: icon-mode scrollbar hide is only `group-data-[collapsible=icon]:[scrollbar-width:none]`. Elsewhere this repo uses the triple `[-ms-overflow-style:none] [scrollbar-width:none] [&::-webkit-scrollbar]:hidden` (e.g. `apps/web/src/app/(app)/agents/components/categories.tsx`). Safari 18.2+ honors `scrollbar-width`; older WebKit can still paint a bar that eats the 3.5rem rail. Overlay scrollbars (default macOS/iOS) do not, so this is polish, not the original clip bug.
- Minor: `apps/web/src/components/ui/__tests__/sidebar-collapsed-scroll.test.tsx`: className string check (`overflow-y-auto`, no `collapsible=icon]:overflow-hidden`). It does not assert `[scrollbar-width:none]`, and the class string is static so `defaultOpen={false}` is unused. Fine as a revert-guard; it does not prove computed overflow at overflow heights.
- Skipped / disagreed: adding WebKit/ms hide in this PR — the bug is `overflow-hidden` clipping, which is gone. Extra hide classes are YAGNI until a classic-scrollbar Safari report.

### Assessment
Dropping `group-data-[collapsible=icon]:overflow-hidden` while keeping `overflow-y-auto` + `min-h-0 flex-1` (and the app sidebar’s non-shrinking inner column) is the right root-cause fix; header/footer stay pinned, channels/DMs below the fold can scroll on the desktop icon rail.

mrosberghaus:
Addressed both notes in edad9bc → HEAD:

- Scrollbar hide now uses the repo's existing triple (`[-ms-overflow-style:none] [scrollbar-width:none] [&::-webkit-scrollbar]:hidden`, same as `categories.tsx`) under the icon-mode variant. Kept for pattern consistency; it costs two classes.
- Test now also asserts the icon-mode `[scrollbar-width:none]` class so a revert of the hide is caught. `defaultOpen={false}` stays: the class is static, but the test documents the collapsed scenario it guards.

Agree the test is a revert-guard only; happy-dom has no Tailwind, so computed overflow is verified in the browser, not here.

PR #4714, opened 2026-09-17
fix(web): hide room row trailing cluster in collapsed sidebar

mrosberghaus:
## Summary

In the collapsed (icon-mode) chat sidebar, each room row still showed its trailing cluster: the pin/mute glyph and the ellipsis menu. They overlapped the room icon. The cluster now hides in icon mode, matching the unread count and mention badge.

## Changes

- `TRAILING_CLUSTER_CLASS` gains `group-data-[collapsible=icon]:hidden`.
- Cluster div tagged `data-slot="room-trailing"` (mirrors the existing `room-leading` slot) so tests can address it.
- Regression test in `chat-room-sidebar-row.test.tsx`: renders a pinned room, asserts the glyph and menu button live in the cluster, and that the cluster carries the icon-mode hidden class. Fails without the fix.

## Verification

- `pnpm --filter web test src/components/chat/chat-room-sidebar-row.test.tsx` — 32 passed
- `pnpm check` — clean
- `pnpm typecheck` — clean


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/chat-room-sidebar-row.test.tsx`: new test asserts the `group-data-[collapsible=icon]:hidden` class string (and that pin + menu live in `data-slot="room-trailing"`), not a collapsed `group` ancestor. Same pattern as the unread/badge tests; does not prove CSS actually applies.
- Skipped / disagreed: mute glyph is not asserted separately — it shares the cluster wrapper, so the class covers it. `svg.lucide-pin` is a Lucide class detail; existing expanded-row menu tests already guard that an unconditional `hidden` would break actions.

### Assessment
Ready to merge: Yes
The cluster is a `z-10` sibling of the size-8 button, so `overflow-hidden` cannot clip it; `group-data-[collapsible=icon]:hidden` is the same hide used by unread, mentions, and archived trailing controls.

mrosberghaus:
Re the Minor on the class-string assertion: kept as is. The web test environment is happy-dom without compiled Tailwind, so a collapsed `group` ancestor would not exercise the CSS either. The assertion follows the existing collapse tests in this file (unread count, mention badge). No changes pushed.

PR #4715, opened 2026-09-17
feat(chat): show emoji reactions before confirm with idempotent reaction routes

mrosberghaus:
## Summary

Emoji reactions now show the moment you tap them and only move again on the server's answer. Design and rejected options: [ADR 0032](docs/adr/0032-reactions-are-idempotent-and-shown-before-confirm.md).

**Core**
- `POST …/reactions` (blind toggle) is replaced by `PUT` and `DELETE …/messages/{messageId}/reactions/{emoji}`. Repeating either returns `200` with the message and publishes no realtime event, so a retry can never flip a Reaction back.
- The `FOR UPDATE` row lock is gone: set semantics make delete-then-create interleaving harmless.
- Route tests cover 👍, ❤️ (variation selector) and 👨‍👩‍👧 (ZWJ sequence) through the path in both directions.

**Web**
- A tap shows the Pending reaction at once: count, own highlight, your name at the end of the reactor list.
- Requests per message+emoji run one at a time; only the newest intent is sent (on/off/on ≤ 2 requests).
- Realtime patches apply first, the pending change is re-applied on top.
- Success takes only that emoji's entry from the response; failure rolls back only that entry and shows the existing toast.
- Timeline and thread panel share one pending state. Leaving the room hides the overlay; the request still finishes.

**Apple**
- Calls the new routes based on the current own-reaction flag; still waits for the server (instant version is PARITY row 20c).
- `scripts/update-core-api.py` now drops paths Core no longer serves, so the removed toggle left the snapshot.

**Shared**
- `MAX_LISTED_CHAT_REACTION_REACTORS` moves to `@sokosumi/utils` (web and Core both use it).
- `CONTEXT.md` defines **Reaction** and **Pending reaction**.

## Verification

- `pnpm test` (all workspaces), `pnpm typecheck`, `pnpm check`
- Apple: `swift test` for SokosumiChat (437) and SokosumiWorkspace (85), `mint run swiftlint lint --strict`, `mint run swiftformat --lint .`
- Web client regenerated via `pnpm --filter web generate:core:snapshot`; Apple snapshot via `scripts/update-core-api.py`

## Notes for review

- API contract change: the toggle route is removed rather than deprecated. No Apple build has shipped.
- A tap queued behind a request that then fails is dropped with the rollback (decision: snap back + toast).


mrosberghaus:
## Grok review

Range: `d1405ae93ba44400ea42e359367d2f0b82e88146` → `9c610973f304ae7fa20d151936101393a9408327`.

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/room-message-row.tsx` — comment still says the parent drops in-flight taps. It no longer does; `recordedUsesRef` keys off `target.reactions` identity, so on/off/on can record the same emoji twice in quick-reaction history.
- Minor: `mergeConfirmedReaction` can replace a richer live same-emoji row if someone else reacted during the request. Apple refetches in that case; web follows ADR 0032 (“success takes the server’s entry”). Next Ably event or room reload repairs it.
- Minor: `applyPendingReaction` + Ably personalize — if the viewer is outside the 20 named reactors, a pending add can show count+1 until HTTP confirm (wire DTO has no full reactor id list).
- Minor: Core DELETE tests omit deleted-message / membership / 404; the shared handler is covered on PUT.
- Skipped / disagreed: 3-way merge or Apple-style refetch on web — ADR 0032 explicitly takes that emoji’s server entry. Queued tap dropped on failure is the recorded decision. Apple still waits for Core (PARITY 20c).

### Assessment

Idempotent `PUT`/`DELETE`, last-intent serial requests, overlay identity, confirm-only-that-emoji, and failure rollback all match the PR; leftover notes are display edges, not contract bugs.

mrosberghaus:
Addressed in `3388b9f33`:

- **Stale comment on `recordedUsesRef`**: reworded. The behaviour itself did not change: the "off" overlay is the confirmed row object (`applyPendingReaction` returns the same message when the intent already holds), so on/off/on within one request hits the same `reactions` identity and records once. A second record only happens once a confirm has replaced the array, which was also the case before this PR.
- **DELETE guard tests**: the deleted-message, membership-status, 404, 422 and coworker cases now run for both `PUT` and `DELETE`.

Left as-is, per ADR 0032:

- **Richer live row replaced on confirm**: success takes that emoji's server entry; the next patch or reload repairs a count that moved during the request.
- **Viewer outside the 20 named reactors**: pre-existing Ably-path edge (`personalize-chat-room-message-event.ts`). The tap sends an idempotent `PUT`, and the personalized REST response corrects the count on confirm.

PR #4718, opened 2026-09-17
feat(web): channel tile in collapsed sidebar rail

mrosberghaus:
## What changed

Collapsed to icons, every public channel in the sidebar was the same `#` glyph, so ten channels were ten identical marks. Channel rows now show a **Channel tile** when the rail is collapsed: a rounded square of the name's initials on the same muted fill DM avatars fall back to, with a small lock or globe corner mark for private and external/matched channels. Public channels carry no mark. Hovering the collapsed row shows the channel name.

The expanded list is unchanged and keeps the `#` / lock / globe glyph beside the name.

- New `ChannelRoomMark` renders the glyph expanded and the tile collapsed; CSS picks one, the same way the row already hides its trailing controls.
- `channelKindIcon` is now the one place that maps discoverability to lock/globe; the glyph and the tile both use it.
- `ChatRoomSidebarRow` passes the room name as the sidebar button tooltip. The sidebar only shows it when collapsed on desktop, so DM rows gain the same hover name.
- `getInitials` returns two letters for single-word names (`general` → `GE`), which its docstring already promised. Applies to all callers, including people avatars.
- Glossary: "Channel tile" added to `CONTEXT.md`.

Closes SOK-1100. Follow-ups: SOK-1101 (unread signal on the tile), SOK-1102 (Apple sidebar).

## Screenshots

Pending: to be attached from the Vercel preview.

## Verification

- `pnpm check`
- `pnpm --filter web test` (764 files pass)


mrosberghaus:
## Grok review

Range: `732297690` (`origin/main`) → `886aabc29`. Reviewed against SOK-1100 and the PR body (collapsed rail tile, expanded glyph unchanged, tooltip, `getInitials`).

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/channel-room-mark.tsx`: corner mark is `size-3.5` with `-right-1 -bottom-1` on a `size-5` tile, so it covers the second initial. DM presence on the same row uses `size-2` / `-right-0.5`.
- Minor: `channel-room-mark.test.tsx` only asserts className tokens. It never mounts under `group` + `data-collapsible=icon`, so it does not prove CSS actually swaps glyph/tile.
- Minor: `CONTEXT.md` Channel tile text says private/external; implementation also marks `matched` (correct per SOK-1100, glossary omits it).
- Minor: hyphenated single-token names (`eng-oncall`, `eng-platform`) both become `EN`. Follows the chosen `getInitials` rule; hover name is the disambiguator.
- Skipped / disagreed: `SidebarMenuButton` `overflow-hidden` clipping the hanging corner mark — not verified. Collapsed button is `min-w-10!` (40px) with `p-3!`, which leaves room for `-right-1` / `-bottom-1`.

### Assessment
Matches SOK-1100 (tile in icon rail, expanded `#`/lock/globe kept, tooltip, single-word initials). CI is green. Leftover is polish, not a missed requirement.

mrosberghaus:
Addressed in c87e96fb5.

- **Corner mark over the second initial**: fixed. The mark is now `size-3` at `-right-1.5 -bottom-1.5`, so it hangs off the tile's corner instead of sitting on the lower half of the second letter. Still inside the collapsed button's 40px box.
- **Glossary omits matched**: fixed. The `CONTEXT.md` entry now reads "private, external, or matched".
- **Test asserts class names only**: leaving as is. happy-dom applies no Tailwind, so mounting under `data-collapsible=icon` would not prove the swap either. The row's trailing-cluster tests take the same approach.
- **Hyphenated single-token names**: by design, and rarer than the example suggests. Core derives the display name from the slug as spaced title-case words, so `eng-oncall` becomes "Eng Oncall" and the tile reads `EO`. Only a name typed with a hyphen collapses to two letters of one word.

PR #4720, opened 2026-09-17
fix(web): allow attachments in coworker direct rooms

mrosberghaus:
## What

Users could not upload pictures (or any file) in a direct chat with a coworker agent. The attach button and the drag-and-drop zone were hidden for coworker DMs only; human DMs and channels worked. Same gate existed on Apple.

## Why

`rooms-client.tsx` passed `allowAttachments={!isCoworkerStreamRoom}` (and the same gate on `RoomFileDropZone`). Apple's `WorkspaceState.canAttachFiles` excluded the active direct-stream room. Both gates dated from when the coworker stream path could not persist uploads. Today both composers append attachments to the message text as markdown, and Core persists that text verbatim on `POST /v1/chats/rooms/{id}/stream`, so the gates only blocked the user.

## Change

- Web: remove the coworker gate on the room composer, thread composer and drop zone.
- Apple: `canAttachFiles` no longer excludes the direct-stream room; `PARITY.md` slice 14 note updated.
- Regression test: `rooms-client-coworker-attachments.test.tsx` renders a coworker-only direct room and asserts the composer, thread composer and drop zone allow attachments (red before, green after).

## Verification

- `pnpm --filter web test src/app/(app)/chat/components` (69 files pass)
- `pnpm check`, `pnpm typecheck` (pre-commit hook)
- `swift test` in `Packages/SokosumiWorkspace` (87 pass), `mint run swiftformat --lint`, `mint run swiftlint lint --strict`

## Not in scope

The coworker model still receives the attachment as a markdown link, not as a multimodal image part. That remains the deferred item noted in Core `stream/post.ts`.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/__tests__/rooms-client-coworker-attachments.test.tsx` clones the full RoomsClient mock harness (~300 lines) for two data-attribute asserts. ThreadPanel is stubbed, so the thread-composer ungating in `rooms-client.tsx` is untested.
- Minor: Apple still blocks attach in coworker DMs (`WorkspaceState.canAttachFiles` requires `directStream.roomId != roomId`). `apps/apple/PARITY.md` slice 14 still says coworker stream rooms retain web's restriction.
- Skipped / disagreed: none. The web gate was UI-only; `POST /v1/chats/rooms/{id}/files` still uses `requireChatRoomUserMembership`, and room stream still uses `requireChatRoomUserWriteAccess` plus the 1:1 coworker-only check. Drop-zone `enabled` does not widen grants.

### Assessment
Ready to merge: Yes
Composer and drop zone no longer special-case coworker DMs; attachments still fold into markdown and persist on the existing stream path, with membership/write checks unchanged.

mrosberghaus:
Addressed both Minor items in c68616424.

- **Thread composer untested:** the regression test now opens a thread in the coworker DM and asserts the `ThreadPanel` receives attachments enabled. Verified red with the old gate restored, green with it removed. The harness clone matches the sibling `rooms-client-*.test.tsx` files, which each carry their own mock set.
- **Apple gate:** `WorkspaceState.canAttachFiles` dropped the `directStream.roomId != roomId` clause. Apple already folds attachment links into the draft text (`ComposeAttachment.message`) and `sendMessage` passes that text to `directStream.send`, so the stream path persists them the same way as web. `PARITY.md` slice 14 updated. SokosumiWorkspace package tests pass (87), pinned SwiftFormat and SwiftLint clean.

PR #4721, opened 2026-09-17
feat(chat): add copy message link

mrosberghaus:
## Summary

Readers can copy a link to a chat message from the action overflow.

- **Web hover More:** Copy link (not Copy text). Desktop selection still works in the transcript.
- **Web long-press sheet:** Copy link, then Copy text. Copy text stays here because selection is not possible.
- **Apple More, context menu, and VoiceOver:** Copy link only.

The copied URL is the existing notification Message link (`/chat/rooms/{roomId}?message={messageId}`). Opening it jumps to that message (or its thread) for anyone who can already read the room. It is not a public share.

## Verification

- `pnpm --filter web test src/lib/utils/notification-href.test.ts src/app/(app)/chat/components/room-message-row.test.tsx`
- Husky `pnpm check` and `pnpm typecheck` on commit
- `swift test --package-path apps/apple/Packages/SokosumiChat --filter ChatLinkTests`

Not clicked in a live browser or the Mac app.


mrosberghaus:
## Grok review

### Fixed
- Nothing pushed — Important `ChatLinkTests` compile break (`link?.roomId` / `link?.messageId` on an enum with no those members) already landed in `9a2b66719` (`apps/apple/Packages/SokosumiChat/Tests/SokosumiChatTests/ChatLinkTests.swift`). `swift test --package-path apps/apple/Packages/SokosumiChat --filter ChatLinkTests` passes on that commit.

### Remaining
- Minor: `ChatLink.href` interpolates `roomId` into the path; web `chatRoomMessageHref` uses `encodeURIComponent`. Current room ids are opaque CUIDs, so the copied URLs match.
- Minor: Apple copy discards `NSPasteboard.setString` success and has no toast; web uses `copyTextWithToast`. Matches other Mac chat actions.
- Minor: `chatRoomMessageHref` tests live inside `describe("getNotificationHref")`.
- Minor: No hover-overflow test that an empty body still shows Copy link (the long-press sheet test covers that case).
- Skipped / disagreed: iOS Copy link — `MessageRowView` is `#if os(macOS)`; there is no iOS timeline to hang the action on.
- Skipped / disagreed: extracting a clipboard helper from `CopyTextButton` — that type is a labeled View, not a pasteboard util.

### Assessment
Overflow Copy link on web (hover More without Copy text; long-press sheet with both) and Mac (More, context menu, VoiceOver) copies the existing `/chat/rooms/{roomId}?message={messageId}` URL; jump stays membership-gated (403/404 → no message); i18n keys exist in en/de/es.

mrosberghaus:
Took the remaining notes from https://github.com/masumi-network/sokosumi/pull/4721#issuecomment-5714368240.

Done in `688b51365`:
- `chatRoomMessageHref` cases live in `describe("chatRoomMessageHref")`.
- Hover overflow still offers Copy link when the body is empty (and still hides Copy text).

Left as-is:
- `ChatLink.href` does not percent-encode `roomId`. The parser splits the path and does not decode, so encoding here would break jump for any encoded id. CUIDs already match web.
- Mac pasteboard stays silent. Same as other Mac chat actions; Copy text is web-sheet-only by spec.
- No iOS Copy link (`MessageRowView` is macOS-only).
- No clipboard helper extracted from `CopyTextButton` (that type is a labeled View).

PR #4725, opened 2026-09-17
feat(chat): send coworker stream attachments as file parts

mrosberghaus:
Follows #4720 (merged); rebased onto `main`.

## What

Uploaded files in a coworker DM now reach the coworker model as image/file input, not only as a markdown link in the text. A vision model can look at an uploaded picture.

## Why

#4720 unblocked the attach button, but the stream request carried a single text part. Core already accepted `file` parts on `POST /v1/chats/rooms/{id}/stream` and the Sokosumi provider already maps them to Responses `input_image` / `input_file`. The gap was the clients and the thread path.

## Change

- **Web:** the composer passes its attachment chips through the send request; the stream hook turns each typed attachment into an AI SDK `file` part next to the text. Drive picks carry no media type and stay link-only.
- **Core:** the stream route collects `file` parts from the last user message and passes them to the thread builder, which re-attaches them to the newest user turn (persisted thread rows are text-only, so they would otherwise be dropped). The top-level path already forwarded them via `convertToModelMessages`.
- **Apple:** `startDirectStream` sends the same `file` parts; attachments are threaded from the composer through `sendMessage` / `sendThreadReply` and `DirectStreamSession.send`. No OpenAPI regeneration: the snapshot already carried the part.

The persisted room message is unchanged: the text part still contains the markdown links.

## Verification

- `pnpm --filter web test src/app/(app)/chat` (70 files pass), `pnpm --filter @sokosumi/core test src/helpers/room-stream-thread.test.ts src/routes/v1/chats/rooms/[id]/stream` (54 pass)
- `swift test` in `SokosumiChat` (442) and `SokosumiWorkspace` (87); `mint run swiftformat --lint .`; `mint run swiftlint lint --strict`
- CI's macOS `xcodebuild … build` succeeds

## Not covered

Live check with a real coworker and an image is manual: send a picture in a coworker DM and ask what it shows.


mrosberghaus:
## Grok review

### Fixed
- Important: Core 400s file parts whose URL is not this user's chat upload in this room, so the coworker model cannot fetch another room's blob or an arbitrary URL (`0fd96b971` — `apps/core/src/routes/v1/chats/rooms/[id]/stream/post.ts`)
- Important: Apple omits Drive picks with an empty media type so they stay link-only, matching web (`0fd96b971` — `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/ChatService+Streaming.swift`)

### Remaining
- Minor: `apps/core/src/helpers/room-stream-thread.test.ts`: subsequent-turn path does not assert `appendFilePartsToLastUserMessage` after `convertToModelMessages`
- Minor: `apps/core/src/routes/v1/chats/rooms/[id]/stream/post.ts`: `lastUserFileParts` type predicate still keys only on `type === "file"` (request schema already requires `url` + `mediaType`)
- Skipped: extra web `rooms-client` send-path test for `files: request.attachments` — `toCoworkerStreamFileParts` already drops untyped Drive picks

### Assessment
Membership/auth were already in place; blob ownership and Drive link-only are now enforced. Leftovers are test nits.

mrosberghaus:
Both remaining items addressed in fe879f550.

- **Subsequent-turn coverage:** `room-stream-thread.test.ts` now has a case where an assistant reply already exists under the root, `convertToModelMessages` runs, and the request's file part is re-attached to the converted newest user turn as `[text, file]`.
- **Predicate:** `lastUserFileParts` narrows with `Extract<typeof part, { type: "file" }>` from the request schema union instead of a hand-written intersection, so `url` and `mediaType` come from the schema, not a claim.

Verified the two fixes in 0fd96b971 before building on them: the ownership check reuses `isOwnedUserChatRoomFileUrl` (same `users/{userId}/chats/{roomId}/` prefix the upload grant mints) and Apple's Drive picker sets `mediaType: ""`, so the empty-string guard is the right mirror of web's `null` drop. Core stream tests 47, thread tests 10, Core typecheck clean.

PR #4728, opened 2026-09-17
fix(web): make Turnstile site key optional

mrosberghaus:
## Summary
Web exited at startup without `NEXT_PUBLIC_TURNSTILE_SITE_KEY` whenever `NODE_ENV=production` or Vercel reported preview/production, so a local `next build` / `next start` needed a key. Core already treated `TURNSTILE_SECRET_KEY` as optional. Now both keys are optional everywhere; with no keys the auth security check is skipped.

- Remove the site-key requirement from `apps/web/src/config/env.public.ts`
- Test covers development, production, preview and Vercel production without a key
- Drop the placeholder site key from the CI build env
- Document both keys as optional (`apps/web/.env.example`, `apps/core/README.md`)

**Note:** deployed Web no longer fails without the site key. Core still warns when its secret is unset in a deployed environment. Set both keys or neither — a Core secret without a Web site key rejects every protected auth request.

## Verification
- `pnpm --filter web exec vitest run src/config/env.public.test.ts src/components/auth-captcha.test.tsx` (17 passed)
- `pnpm --filter core exec vitest run src/config/env.test.ts src/lib/auth.test.ts` (92 passed)
- `pnpm check && pnpm typecheck` (pre-commit hook)


mrosberghaus:
## Grok review

Range: `c90bcf3e88`..`f944380c34` (5 files).

### Fixed

### Remaining
- Minor: `apps/web/src/config/env.public.test.ts` — covers omitted key in development/production/preview, but does not assert a configured site key is still parsed (Core has `preserves a configured secret`).
- Minor: Web has no deployed-env warning when the site key is unset. Core warns; Web now boots and `useAuthCaptcha` skips the widget. Footgun only if Vercel env is dropped.
- Minor: `apps/web/src/components/auth-captcha.tsx` still says the skip is “during local development”; that path now also runs in production when the key is omitted (pre-existing comment, not in this diff).
- Skipped / disagreed: restore Vercel production/preview fail-closed for the site key — PR body makes optional-everywhere the requirement; Core already matches. Smallest local-`next start` fix would have been dropping only the `NODE_ENV === "production"` check; that would fight this PR.

### Assessment
Ready to merge: Yes

Production still enforces CAPTCHA when both keys are set (`useAuthCaptcha` sends the header; Core `createAuthCaptchaPlugin` still verifies). Omitting keys is allowed in every environment by design, not local/dev-only. Build, Typecheck, Biome, Test Core, and Test Web are green on this head.

mrosberghaus:
Addressed the Grok review in 3f849b2b2:

- **Configured key test:** added `preserves a configured site key` to `env.public.test.ts`.
- **Stale comment:** `auth-captcha.tsx` now says omitting the site key skips the check in any environment.
- **Web deployed-env warning: skipped.** The public env is parsed in the browser too, so a warning there would print on every page load for every visitor. The dangerous misconfiguration (Core secret set, Web key missing) already fails loudly, because every protected auth request is rejected. Dropping both keys is covered by Core's existing deployed-env warning.

PR #4731, opened 2026-09-17
ci(apple): drop duplicate Xcode build job and cache SwiftPM builds

mrosberghaus:
## Why

The Apple workflow compiled the app twice per PR, and every Swift package leg recompiled its third-party dependencies from scratch.

Last full run ([35222141094](https://github.com/masumi-network/sokosumi/actions/runs/35222141094)): `Xcode build` 279s duplicated the build inside `Xcode app tests` (364s). The slowest leg, `Swift tests (SokosumiWorkspace)`, took 389s.

## What

- **Remove the `Xcode build` job.** `xcodebuild test` in the app leg builds the same scheme, config and destination. Saves about 4.5 macOS runner-minutes per PR; doesn't change how long the run takes.
- **Cache SwiftPM `.build` per package.** Key: package + toolchain hash + `Package.resolved` hash + SHA. PRs only restore, by prefix. Main pushes and manual dispatches save.
- **Run `test-suite` on path-filtered pushes to main** so the cache is saved on the default branch (PR caches are ref-scoped, same reason lint already runs there).

## Before merge: ruleset change

`Xcode build` is a required status check in the **Default Branch** ruleset. It has to be removed there, or this PR (and every PR after) waits on a check that no longer runs. `Xcode test` stays required and covers the app build.

## Verification

- [x] `actionlint` passes; its only warnings are about the `xcode-27` runner label, which it doesn't recognize and which the workflow already used
- [x] Dispatch run 1 (cold, saves cache) vs dispatch run 2 (warm) on this branch: package leg timings compared in a comment
- [x] PR run is green except the removed `Xcode build` check


mrosberghaus:
## Cache timings (3 dispatch runs, same commit)

| Package job | Run 1: from scratch ([35224906576](https://github.com/masumi-network/sokosumi/actions/runs/35224906576)) | Run 3: cache hit ([35227907254](https://github.com/masumi-network/sokosumi/actions/runs/35227907254)) | `swift build` scratch → cached |
| --- | --- | --- | --- |
| SokosumiRealtime | 335s | 265s | 174s → 82s |
| SokosumiWorkspace | 295s | 136s | 153s → 43s |
| SokosumiChat | 273s | 120s | 144s → 44s |
| CoreAPI | 166s | 84s | 102s → 22s |
| SokosumiAuth | 66s | 58s | 24s → 12s |
| **Sum** | **1135s** | **663s** | |

- Restoring the cache costs 14–42s per job. Saving costs 6–33s, but only on `main` and manual runs; PRs skip it.
- Run 2 ([35226984541](https://github.com/masumi-network/sokosumi/actions/runs/35226984541)) mostly missed: `xcode-27` runners currently come with Xcode `27A5252f` or `27A266a`, and the key includes the toolchain on purpose. Each package now has a cache for both builds, except SokosumiRealtime (only `27A5252f`).
- SokosumiRealtime stays slow because its tests take about 85s to run, not because of compiling.
- **Package jobs use about 8 fewer runner-minutes per PR.** With the removed `Xcode build` job, that's about 12.5 minutes saved per PR.
- **Total run time barely changes:** `Xcode app tests` (~365s) is now the slowest job. The next thing to speed up would be caching that job's build (`-derivedDataPath`).

mrosberghaus:
## Grok review

### Fixed

### Remaining
- Merge blocker (ops, not this diff): Default Branch ruleset still requires `Xcode build`. Until that context is removed, this PR and later Apple PRs wait on a check that no longer runs. `Xcode test` stays required and still aggregates the matrix.
- Minor: `apps/apple/AGENTS.md:79` still says CI “runs build + tests in parallel” and only mentions the Mint cache on main.
- Minor: restore-keys line 2 (`…-spm-$package-$toolchain-`) can restore a `.build` from a different `Package.resolved`. SwiftPM should invalidate; dispatch run [35227907254](https://github.com/masumi-network/sokosumi/actions/runs/35227907254) still executed Chat 450 / Workspace 92 / Auth 35 / Realtime 27 / CoreAPI 1 tests after a cache hit.
- Skipped / disagreed: dummy `Xcode build` job to satisfy the ruleset — that would reintroduce a fake required check. Remove it from the ruleset instead.
- Skipped / disagreed: skip `cache/save` on exact `cache-hit` — dispatch re-save warns (`Cache already exists`) and does not fail the job.

### Assessment
Dropping `Xcode build` is safe (`xcodebuild test` builds the same `Sokosumi.app` scheme); the SwiftPM cache restores third-party `.build` artifacts and still runs tests. Do not merge until `Xcode build` is removed from the Default Branch required checks.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/4731#issuecomment-5716725298)

- **Ruleset still requires `Xcode build`:** agreed that this blocks merge. Removing it from the Default Branch ruleset is a repo-settings change, so it's waiting on @mrosberghaus. No dummy job.
- **`apps/apple/AGENTS.md:79` out of date:** fixed in 5c18bbfe1. It now says there's no separate build job, that tests also run on pushes to `main`, and that those runs save the SwiftPM cache (keyed by toolchain) while PRs only restore it. The `PARITY.md` entries that mention "Xcode build" are dated verification logs, so they stay as they are.
- **The last `restore-keys` fallback ignores `Package.resolved`:** keeping it on purpose. After a dependency bump, PRs still reuse every dependency that didn't change. SwiftPM re-resolves and rebuilds whatever changed, and the toolchain stays in every key, so a `.build` from a different compiler is never restored. The next save on `main` writes a cache under the new `Package.resolved` hash.
- **Don't add a job just to satisfy the check, and don't skip saving when the key already exists:** agreed with both.

PR #4732, opened 2026-09-17
fix(web): show hover hints for all collapsed sidebar items

mrosberghaus:
## What
In the collapsed sidebar, only **Search** showed its name on hover. New Task, Agents, Projects, Tasks, Calendar, Files and History showed nothing, only the browser's link preview.

## Why
`menu-items.tsx` passed a `tooltip` to `SidebarMenuButton` only when the item had a shortcut label, and only Search has one. Link items never got one.

## Change
- Every item now passes a tooltip: the plain label, or the label plus shortcut for Search.
- `SidebarMenuButton` already hides the tooltip unless the sidebar is collapsed on desktop, so the expanded sidebar and mobile look the same as before.
- Added a test checking every menu item has a hover hint.

## Verification
- `pnpm --filter web exec vitest run "src/app/(app)/components/sidebar/components/menu-items.test.tsx"`: 12 passed (new test failed before the fix)
- `pnpm check && pnpm typecheck`: green
- Not checked by hand in the browser yet


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/components/sidebar/components/menu-items.test.tsx`: the new test only asserts `data-has-tooltip="true"`. It does not check that the hint is the item label, or that Search still includes the shortcut.
- Minor: `apps/web/src/app/(app)/components/sidebar/components/menu-items.tsx`: Search shortcut in the tooltip still uses `text-muted-foreground` on `TooltipContent`'s `bg-primary-solid`. Pre-existing (moved, not introduced). Skip.
- Skipped / disagreed: no Critical/Important to push back on.

### Assessment
Ready to merge: Yes
Every rail item now passes `tooltip` through the existing `SidebarMenuButton` gate (`hidden` unless desktop collapsed), so expanded and mobile stay label-only.

mrosberghaus:
Addressed the Grok review:

- **Test only checked tooltip presence**: fixed in 664d2cc56. The test now asserts the exact hint for each item in order: the label for every link and New Task, and label plus shortcut (`searchCtrl+K`) for Search.
- **Search shortcut uses `text-muted-foreground` on the tooltip's `bg-primary-solid`**: agreed it's pre-existing and only moved in this PR, so it stays out of scope here.

PR #4733, opened 2026-09-17
feat(apple): show live participant presence

mrosberghaus:
## Summary

Apple chat parity row 36 — participant presence. The Mac app now enters Ably org presence for the active organization and mirrors teammates' online/away/offline marks, matching web's `OrgPresenceProvider` (ADR 0003).

Web sources audited at `132376a62`: `apps/web/src/components/chat/presence-dot.tsx`, `live-member-presence-dot.tsx`, `apps/web/src/contexts/org-presence-provider.tsx`, `apps/web/src/lib/ably/use-org-presence-map.ts`, `use-org-presence-publisher.ts`, `should-publish-org-presence.ts`, `apps/web/src/hooks/use-self-presence.ts`, `packages/utils/src/chat-presence.ts`, and the surfaces in `room-roster-panel.tsx`, `chat-participant-hover-card.tsx`, `direct-room-avatar-stack.tsx`, `sidebar-account-chip.client.tsx`.

## Included

- **Publishing** (`SokosumiRealtime/OrgPresenceChannel`, `OrgPresenceScope`): authorize, read the token capability, enter `presence:org_{orgId}` only when `presence` is granted, `update` with `enter` fallback, re-sync on every `connected`, 15 s retry after a failed mint, leave-then-enter on workspace switch, leave/unsubscribe/detach on sign-out and app termination. Personal workspaces never enter.
- **Map** (`SokosumiChat/ChatPresence`, `OrgPresence`): `presence.get` hydration plus live events, `{userId}:{instanceId}` aggregation (any device online → online, connected → away, absent → DTO fallback), malformed ids/payloads rejected, 30 s local reclassification, roster dropped on organization change.
- **Publisher bookkeeping** (`OrgPresencePublisherState`): `lastActiveAt`/`visible`, forced publishes on visibility/enter/reconnect, activity throttled to 4 min inside the 5 min window, idle/unchanged presence never publishes.
- **Coordinator** (`WorkspaceState+Presence`): organization sync on socket start and switch, activity/visibility hooks, tick, `presence(for:)` with AI members pinned online, `prepareForTermination`.
- **UI**: `PresenceDot` (filled disc / crescent / hollow ring with a ground halo) on roster avatars (with web's Coworker / Personal assistant badges beside the name), participant details (visible label for humans), sidebar Direct faces (none on self Directs) and the Me section's local self-approximation; accessibility values on the avatar button, 1:1 Direct rows and the account menu. macOS `NSEvent` local monitor feeds activity; `willTerminate` leaves presence.

## Excluded / native adaptations

- No typing indicator (none on web). No room-header avatar stack (the native toolbar Members button stays).
- The self dot reads offline from Ably connection health (after the socket was up once) rather than `navigator.onLine`; window activation counts as activity through the existing `scenePhase` path.
- No Core contract, OpenAPI snapshot, dependency or web changes: `POST /v1/realtime/ably-token` already grants `presence` + `subscribe` on every membership organization.

## How to test

1. Sign in on two clients (web and the Mac app, or two Mac installs) in the same organization workspace; open a channel's Members inspector, a Direct row and a participant popover.
2. Hide/unfocus one client: its dot turns to the away crescent on the other within seconds; foreground it again: green. Idle for 5 min: away. Quit or sign out: hollow ring (Ably disconnect grace applies to a crash).
3. Switch to the personal workspace: dots fall back to Core's snapshot values; switch back: live again.

- `swift test` — SokosumiChat 469, SokosumiWorkspace 95, SokosumiRealtime 30, SokosumiAuth 35, CoreAPI 1 passed.
- `mint run swiftformat --lint --cache ignore .` — 0/335 need formatting; `mint run swiftlint lint --strict --no-cache` — 0 violations.
- `xcodebuild … build` (macOS arm64, ad-hoc) succeeded; `xcodebuild … test -only-testing:SokosumiTests -enableCodeCoverage NO` — 163 tests, 0 failures.
- `swift build --package-path Packages/SokosumiWorkspace --triple arm64-apple-ios17.0 --sdk "$(xcrun --sdk iphoneos --show-sdk-path)"` succeeded.
- Light/dark render fixtures inspected: `apps/apple/docs/screenshots/presence-dots-*.png`, `sidebar-presence-*.png`, `room-details-*.png`.

Unverified: live Ably presence against production, the `NSEvent` monitor and `willTerminate` leave from a signed build, VoiceOver and large Dynamic Type. Details in `apps/apple/PARITY.md` (slice 36).


mrosberghaus:
## Grok review

### Fixed
- Important: `OrgPresenceChannel.stop()` was `queue.async`, so `ARTRealtime.close()` could run before `presence.leave` was issued. Sign-out and `willTerminate` would then wait on Ably’s disconnect grace instead of going offline immediately. `stop()` is now `queue.sync`; `disconnect()` leaves, then closes (`f8ca2d678` — `apps/apple/Packages/SokosumiRealtime/Sources/SokosumiRealtime/OrgPresenceChannel.swift`, `AblyRealtimeConnection.swift`).
- Important: a failed `presence.enter` returned silently after the coordinator had already `markPublished`, so the 30s tick would not retry. Enter failures now retry after 15s (same cadence as a failed mint) (`f8ca2d678` — `OrgPresenceChannel.swift`).

### Remaining
- Minor: `apps/apple/Sokosumi/Shared/PresenceDot.swift` — marks use SwiftUI `.green` / `.orange` rather than web’s semantic success/warning tokens. Silhouettes still distinguish the three states; screenshots already cover light/dark.
- Minor: sidebar dots paint the halo in `.background`, not the sidebar material (already called out as unverified in `PARITY.md`).
- Minor: `OrgPresenceChannel.attach` ignores `presence.get` errors (same as web’s `hydrate` catch). First paint then waits on live events or the next `connected`.
- Minor: `PresenceLifecycleModifier` is a no-op on iOS (no activity monitor, no terminate leave). Packages still build for iOS 17; there is no iOS app target yet.
- Skipped / disagreed: coordinator force-publish on `connectionHealth(true)` — Ably `connected` already re-authorizes, re-hydrates, and resends `latestData`. That keeps teammates inside the 5 min online window; an extra force would only duplicate web’s mount+connected publish.

### Assessment

Capability-gated org channel, no client API key, leave-then-enter on switch, DTO fallback in personal workspaces, and AI members pinned online all match ADR 0003. Live Ably against production, the `NSEvent` monitor, and a signed `willTerminate` leave remain unverified, as the PR already recorded.

mrosberghaus:
Response to the Grok review.

**`stop()` async vs `close()`; failed `enter` never retried** — confirmed fixed in f8ca2d6 (leave-then-close is synchronous, enter failures retry on the 15 s cadence). Realtime and Workspace suites pass locally on that head.

**`.green` / `.orange` instead of web's semantic tokens** — leaving as is. The app has no semantic color assets (only `AccentColor`); every status color in the app is a system color (`.red` for errors in the composer, sheets and root view). `Color.green` / `.orange` are Apple's adaptive semantic colors, so they track light/dark and Increase Contrast the way web's tokens track its theme, and the silhouettes carry the state without color. Introducing an asset catalog palette is a separate, app-wide change.

**Sidebar halo painted in `.background`, not the sidebar material** — leaving as is, still recorded as unverified. `DirectRoomAvatarStack` has drawn its separator ring in `.background` since slice 03 and that rendering was accepted; the dot halo follows the same convention. Swapping both to a cutout mask over the material is the right fix if the halo proves visible on the translucent sidebar, and it belongs to both marks at once, not the dot alone.

**`presence.get` errors ignored** — matches web's `hydrate` catch, as noted; live events and the next `connected` re-sync fill the map.

**`PresenceLifecycleModifier` no-op on iOS** — intentional; there is no iOS app target, and the packages keep the portable parts.

Force-publish on `connectionHealth(true)`: agreed with the skip — `connected` already re-authorizes, re-hydrates and resends `latestData`.

PR #4734, opened 2026-09-17
fix(web): keep mention badge still on sidebar row hover

mrosberghaus:
## Summary

Hovering a chat room row in the sidebar shifted the mention count badge left.

**Cause:** the badge sits against the row's trailing spacer. At rest the spacer is a 16px glyph hole (pinned/muted) or 0px; on hover it grows to 28px for the `…` menu, pushing the badge.

**Fix:** rows with a mention badge hold one 12px spacer in every state on hover-capable devices, so the badge never moves and ends where the `…` button's box begins (~6px to the visible dots). Rows without a badge are unchanged.

## Verification

- New test in `chat-room-sidebar-row.test.tsx` (red before, green after)
- `pnpm --filter web exec vitest run src/components/chat/`: 334 passed
- Biome clean
- Not yet checked in a browser


mrosberghaus:
## Grok review

Range: `78977ea96` (`origin/main`) → `4c7cc672a` (`fix-mention-badge-hover-jump`). Fix commit `65cb81d9f`.

### Fixed

### Remaining
- Minor: `chat-room-sidebar-row.test.tsx` — fixture is pinned + mentions only. Unpinned mention rows take the same `badgeCount > 0` branch (the larger rest `size-0` → hover `size-7` jump). Same assertion would cover it; not worth a follow-up commit.
- Minor: that test uses `toContain("[@media(hover:hover)]:size-7")`. It is not a substring of the `group-hover/…:size-7` classes (checked), but this file already splits tokens for `md:size-7`. `className.split(" ").includes(...)` would match the existing seam.
- Skipped: nested rest-size ternary — readable enough; a helper would be extra.
- Skipped: PR body Claude footer — convention, not the diff.
- Not verified in a browser (author already said so). CSS-only reservation; collapsed still hides badge + spacer via `group-data-[collapsible=icon]:hidden`.

### Assessment
Ready to merge: Yes

Badged rows reserve `[@media(hover:hover)]:size-7` at rest so the mention badge does not move when the menu slot appears; unbadged rows, touch (`hover:none`), and the collapsed icon rail are unchanged. Web tests / Biome / typecheck are green on CI.

mrosberghaus:
Re [Grok review](https://github.com/masumi-network/sokosumi/pull/4734#issuecomment-5717384164) (reviewed `4c7cc672a`; since then `f64e94836` changed badged rows from a 28px to a fixed 12px spacer in every state so the badge sits beside the `…` button):

- **Unpinned badged rows:** fixed in `1d0ac20e2`. The test runs for pinned and unpinned rows via `it.each`.
- **Exact token match:** fixed in `1d0ac20e2`. The test now splits `className` and asserts on whole tokens, including that the three hover/focus/open `size-7` classes are absent.
- Nested ternary and PR footer: left as is, agreed.

38/38 row tests pass; pre-commit typecheck and Biome green.

PR #4735, opened 2026-09-17
feat(core): add send to yourself endpoint for chat messages

mrosberghaus:
Closes [SOK-1104](https://linear.app/masumi/issue/SOK-1104/send-a-chat-message-to-yourself-from-the-message-menu)

## What changes for users

A new **Send to yourself** message action (web and Apple) puts a quote of the message into your Self Direct ("You"). If the Self Direct doesn't exist yet, it is created.

- The saved message is only a quote card: author, snippet and attachment, with no body text. Clicking the card opens the Message link to the original.
- You stay in the room. Web shows a "Sent to yourself" toast with **Open**, and Apple shows a confirmation with **Open**.
- The action appears wherever Copy link does, except on messages inside the Self Direct.
- If you later lose access to the source room, the saved quote stays readable. Only the link stops working.
- Normal quoting still works only within the same room. Cross-room quotes are allowed only into your own Self Direct.

## Implementation

- **Core:** new `POST /v1/chats/rooms/{id}/messages/{messageId}/send-to-self`, which returns 201 with the saved `ChatRoomMessage`.
  - It returns 404 for rooms you can't read and for missing or deleted messages. It returns 400 for non-content messages and for messages already in your Self Direct.
  - `ChatRoomMessageQuote` gains an optional `roomId`. Existing quotes don't have it and behave as before.
  - I extracted `buildRoomQuoteSnapshot` / `roomQuoteSourceSelect` and `isSelfDirectRoom` from the existing helpers.
- **Web:** added a service method and a server action, plus a menu item in the hover overflow and the touch sheet, including the thread panel. A quote card from another room follows `chatRoomMessageHref`, and a message that is only a quote renders no empty body. New i18n keys in en/de/es.
- **Apple:** `ChatService.sendMessageToSelf`, `WorkspaceState.canSendToSelf` / `sendMessageToSelf`, and a context menu, More menu and accessibility action in the timeline and threads. `quoteSourceURL` routes cross-room quote taps through the existing chat link handler.
- Regenerated the web Core client and the Apple `openapi.json`.

## Verification

- `CHAT_SELF_INTEGRATION_DATABASE_URL=… pnpm --filter @sokosumi/core exec vitest run src/routes/v1/chats/rooms/self-direct.integration.test.ts`: 13/13 against a local Postgres. This covers success, creating the Self Direct on demand, access denied, deleted or missing messages, rejection inside the Self Direct, and the snapshot surviving a room exit.
- Core chat route tests: 601 passed. `pnpm --filter web exec vitest run src/app/(app)/chat src/lib/services`: 1380 passed.
- `pnpm typecheck` (core, web) and `messages:parity` pass.
- Apple: `swift test` passes for SokosumiChat (452) and SokosumiWorkspace (92). The `xcodebuild` build passes, and so do `swiftlint --strict` and `swiftformat`.
- Not verified in a running app: I haven't clicked through the UI on web or macOS, so there are no screenshots yet.


mrosberghaus:
## Grok review

### Fixed
- Important: Apple **Open** after the first send-to-self (Self Direct created on demand) — `openRoomLink` only selects rooms already in the sidebar, and Apple ignores `chat_rooms_changed`, so Open failed until the 15s recovery refresh (`c778bd4de` — `apps/apple/Packages/SokosumiWorkspace/Sources/SokosumiWorkspace/WorkspaceState+SendToSelf.swift`)

### Remaining
- Minor: `apps/core/src/routes/v1/chats/rooms/[id]/messages/[messageId]/send-to-self/post.ts`: first `prisma.$transaction` is read-only (access + snapshot). A plain `prisma` client is enough.
- Minor: no Core test that membership-status messages return 400 (`assertChatRoomContentMessage`). Web/Apple already hide those rows from the action.
- Minor: Apple `MessageQuoteView` still only makes the author name tappable; spec says clicking the card. Same as existing same-room quotes.
- Minor: web `handleSendMessageToSelf` uses `selectedRoom?.id` instead of `message.roomId`. Fine while the transcript is that room.
- Skipped / disagreed: none of the Critical/Important findings were wrong. Auth, cross-user isolation, and blob ownership checked out.

### Assessment

Core always writes into the caller's Self Direct after a membership read on the source room; other users cannot copy into your notes, and quote attachments keep the original blob URLs (no ownership transfer).

mrosberghaus:
Fixed three of the four remaining notes in `bf0e130d8`, and kept one as-is.

- **Read-only transaction:** the access check and the snapshot read now run on the plain client. The snapshot is durable, so the two reads don't need to agree on one point in time.
- **Membership-status 400:** covered in `self-direct.integration.test.ts` (14 pass against a local Postgres).
- **Web room id:** `handleSendMessageToSelf` now sends `message.roomId` instead of `selectedRoom?.id`.

Not changed: Apple `MessageQuoteView` making only the author name tappable. That view is shared by the composer preview and every same-room quote, so widening the tap target would change all quote cards, not just the ones sent to yourself. Worth doing, but as its own change. Say the word and I'll file it.

Your `c778bd4de` is right about Open: `openRoomLink` only selects listed rooms and Apple ignores `chat_rooms_changed`, so the first send left Open dead until the recovery refresh.

mrosberghaus:
Dropped the database-backed cases in `e8050d0f0`.

`self-direct.integration.test.ts` is back to its state on main, and the endpoint is now covered by `send-to-self/post.test.ts` beside the route, in the same mocked-Prisma style as the sibling `pin/post.test.ts`. Six cases: the quote written into the Self Direct (room id, empty body, snapshot with the source `roomId`, room bump, read state, realtime), the sidebar refresh firing only when the Self Direct was created, 404 for an unreadable room, 404 for a missing or deleted message, 400 inside the Self Direct, and 400 for a membership status message.

Only `createOrGetDirectRoom` is stubbed, since it runs its own transaction. Room access and the quote snapshot still execute for real against the mocked client.

Core chat suite: 609 passed, and it no longer needs a database.

mrosberghaus:
Diagnosed the missing message properly rather than trusting the first fix. Root cause confirmed, with a test that reproduces your exact symptom (`99a199c30`).

**Root cause:** the transcript's visibility rule was content-centric. `shouldKeepPersistedMessage` in `merge-room-messages.ts` kept only membership rows, coworker shells and messages with a non-empty body, so it dropped the saved quote before rendering. The room opened and the transcript was empty. It was never a routing or jump problem.

**The loop:** `rooms-client-saved-quote.test.tsx` renders `RoomsClient` for a Self Direct with a quote-only message, both as server-rendered history and as hydrated history. With the fix removed, both cases fail with `Unable to find an element by: [data-testid="chat-message-row"]`, which is the symptom. With the fix they pass.

Worth noting for review: my first unit test passed while the bug was live, because it called `mergeRoomMessages`, which does not filter. The render path goes through `mergeMessagesWithStreamOverlay`, which does. That is why the test now runs at the component seam.

I checked the other two `content.trim()` sites in that file: they only drive overlay de-duplication and require a visible body, so a quote-only row is unaffected. Apple has no equivalent filter.

Web chat suite: 1150 passing.

mrosberghaus:
## Grok review

Second pass on `99a199c30` vs `main` (`5fd2ec9f7`). Did not re-apply `c778bd4de` — it is already on HEAD.

### Fixed

### Remaining
- Minor: Apple `MessageQuoteView` still only makes the author name tappable. Spec says clicking the card. Same as existing same-room quotes. Left as-is on purpose (shared composer/timeline view).
- Minor: `isSavedQuoteMessage` is `quote != null`, so any empty-body quote is kept, not only send-to-self. Regular POST still requires a non-empty body, so this only matters for these saved rows.
- Minor: when You already exists, send-to-self does not call `invalidateChatRoomMessageReaders`, so other tabs may not bump the You row until the next rooms fetch. Create-on-demand still publishes `chat_rooms_changed`. Toast/Open cover the in-session case.
- Skipped / disagreed: Apple Open after first create — already fixed in `c778bd4de` (`refreshRooms` before Open). Auth, cross-user isolation, and blob ownership still check out.

### Assessment

Core still writes only into the caller’s Self Direct after a membership read on the source room; other users cannot copy into your notes; quote attachments keep the original public blob URLs (no ownership transfer). The empty-body transcript drop is covered by `mergeMessagesWithStreamOverlay` plus the RoomsClient test.

mrosberghaus:
Fixed the sidebar note in `467c7cceb`; left the other two as-is.

**Fixed — stale You row in other tabs.** The route now publishes `chat_rooms_changed` on every send, not only when the Self Direct was just created. A Self Direct has exactly one member, so the caller's own tabs are the entire audience: a new room gets listed, an existing one gets reordered. That is simpler than reaching for `invalidateChatRoomMessageReaders`, which would re-read a roster that is known to be `[callerId]`. The route test now asserts the publish for both the created and the existing case.

**Not changed — `isSavedQuoteMessage` is `quote != null`.** That breadth is deliberate. The rule the transcript needs is "a quote is content in its own right", not "this row came from send-to-self". Narrowing it to the source `roomId` would leave a same-room quote with an empty body invisible for the same wrong reason. As you note, `POST /messages` still requires a non-empty body, so today these saved rows are the only ones affected.

**Not changed — Apple quote card tap target.** Still the shared composer/timeline view, so widening it would change every quote card. Worth a separate change.

Core chat suite: 609 passing, no database needed.

PR #3923, opened 2026-08-24
feat(core): retry failed channel coworker mentions

mrosberghaus:
Closes https://linear.app/masumi/issue/SOK-861/retry-a-failed-channel-coworker-mention

## Mock — failure + Retry

Failed coworker mention with Retry

- **You (mentioner):** Thought sparkle + `Failed to reply` on one line. **Retry** in the outbound-style footer under it.
- **Everyone else:** sparkle + `Failed to reply`, no Retry.
- [Open the PNG](https://github.com/masumi-network/sokosumi/blob/664a8c835/docs/plans/sok-861-failed-mention-retry.png)

## Spec
- Mentioner-only Retry on the failed coworker shell. Other members see the shell without Retry.
- Reuses the existing ChatRoomMention and Thought shell (`responseMessageId`); does not create a second coworker message.
- `POST /v1/chats/rooms/{id}/messages/{messageId}/mentions/{mentionId}/retry` resets `failed` → `pending`, publishes `mention_status`, and `waitUntil` dispatches again.
- Room sees live Thought, then a reply or another failure. Success still has no mention chrome on the parent.
- Out of scope: anyone-in-the-room Retry, timeout changes, replacing channel dispatch with the 1:1 stream route.

## Verify
- `pnpm --filter core check` + `pnpm core:test`
- `pnpm web:check` + `pnpm web:test`

### Review notes - medium (human review)

| Area | Location | Finding |
|------|----------|---------|
| R7 | `rooms-client.tsx` `isCurrentUserMentionerOfFailedShell` | Retry is hidden if the source human message is not in the loaded timeline/thread set. |
| R7 | `post.ts` then dispatch | Other members keep the failed shell until dispatch republishes the placeholder; the mentioner is optimistic. |

PR #4738, opened 2026-09-17
fix(chat): keep failed coworker mention shells in the transcript

mrosberghaus:
## What

A coworker mention that fails in a channel or thread now keeps its bubble in the transcript, so the row can render **Failed to reply** and the mentioner-only **Retry** button.

## Why

Core rewrites a failed mention's coworker placeholder to `content: ""` with `metadata.mention_failed: true` precisely so the bubble survives for fail + Retry (`failMentionThoughtPlaceholder`, #3923 / SOK-861). The row (`room-message-row.tsx` → `FailedMentionActions`) and the retry action (`handleRetryMention` → `POST …/mentions/{mentionId}/retry`) are already wired. But `shouldKeepPersistedMessage` in `merge-room-messages.ts` only kept shells matching `isPersistedMentionThoughtShell`, which #3923 narrowed to `mention_failed !== true` — so `mergeMessagesWithStreamOverlay` dropped the failed shell from `displayMessages` and `displayThreadMessages`. On production the bubble shows "Thinking" and then vanishes; nobody ever sees Retry.

Reproduced by running the function: `[human, failedShell, thinkingShell]` → `["m1", "m3"]`.

Found while auditing current web for Apple parity row 37 (coworker mention execution status); the Apple slice waits on this fix.

## Change

- `merge-room-messages.ts`: the persisted-shell predicate also keeps `isFailedMentionThoughtShell` rows (renamed `isStreamingCoworkerPlaceholder` → `isMentionCoworkerShell`). No other filtering changed.
- `merge-room-messages.test.ts`: new case proving failed shells stay with the overlay idle and active, next to a streaming shell.

## Verification

- `pnpm --filter web test "src/app/(app)/chat/utils"` — 33 files, 304 tests pass
- `pnpm --filter web test "src/app/(app)/chat/components"` — 70 files, 684 tests pass
- `biome check` on both files — clean

Not verified: a live failed mention in the running UI (needs a coworker configured to fail); the row rendering and retry action are covered by their existing tests.


mrosberghaus:
## Grok review

Reviewed `d381c6821..c8b41555b` (`merge-room-messages.ts` + test). Predicates, row Retry UI, and `handleRetryMention` checked against Core `failMentionThoughtPlaceholder`.

### Fixed

### Remaining
- Minor: `merge-room-messages.test.ts` asserts kept ids only, not that `mention_failed` is still on the row. Filter-only merge, so it cannot strip metadata.
- Minor: no merge-level case for `mention_failed: true` without `mention_id`. `isFailedMentionThoughtShell` already rejects that in `coworker-thought.test.ts`; empty streaming-without-`mention_id` still drops.
- Skipped / disagreed: none.

### Assessment
Ready to merge: Yes

Failed shells stay via `isFailedMentionThoughtShell` (needs `mention_id`); other empty coworker rows still drop; Retry stays mentioner-only and `pendingMentionRetriesRef` still blocks a double POST.

mrosberghaus:
Response to the Grok review.

**Test asserted ids only** — added in the follow-up commit: the kept row still carries `mention_failed: true` and `mention_id`, which is what the row reads to render Failed to reply / Retry.

**No merge-level case for `mention_failed` without `mention_id`** — added: an empty failed coworker row without `mention_id` is still dropped, mirroring the existing streaming-without-`mention_id` case.

22 tests pass; Biome clean.

PR #4739, opened 2026-09-17
feat(web): mark the open room on the collapsed rail's right edge

mrosberghaus:
## What the reader sees

Collapsed to icons, you could not tell what you had open. Rest, hover and selected all rendered the same colour: `--muted`, `--accent` and `--sidebar-accent` are the identical 15% in dark, so the Channel tile's own plate, the hover wash and the active wash all landed on it and only the box grew.

Selection moves off the fill onto a 20px primary bar on the rail's **right** edge, mirroring Chat's attention pill on the left. Attention arrives on one edge; what you have open is marked on the other, pointing at the panel it opened. A room can be unread *and* open at once without the two marks arguing.

Because the fill is now free, the Channel tile keeps the muted plate its lock/globe corner mark sits on — no restyle of the mark itself. Hover takes a 1px ring instead, which reads over a filled tile without repainting it.

| State | Before | After |
| --- | --- | --- |
| Rest | tile plate, 15% | unchanged |
| Hover | fill, 15% — same colour, bigger box | 1px neutral ring |
| Selected | fill, 15% — identical to hover | 20px primary bar, right edge |

**This covers the whole rail, not just channels.** Nav rows (Projects, Tasks, Soko Bot, Drive, …) had the same collision — active and hover both painted `bg-sidebar-accent`. Nav had it more narrowly than Chat: its glyph carries no plate of its own, so rest was already distinct and only two states shared the fill, not three.

So the rules live on `sidebarMenuButtonVariants` and a shared `SidebarRailSelectionBar`, rather than a second copy beside the nav list. Chat keeps only what is Chat's: the attention pill.

## Notes

- **Geometry had to be reconciled first.** Nav put its horizontal padding on the `SidebarMenuItem` and Chat puts it on the `SidebarGroup`, so the two item boxes were 56px and 40px wide for the same 8px inset — a shared `-right-2` would have overshot the rail on nav. Nav's padding moves up to the group so both are 40px, and its separator, which spans the rail rather than the row, takes that padding back with `-mx-2`.
- The collapsed overrides outrank the base `hover:` and `data-[active=true]:` rules on **specificity**, not source order — `.…:is(:where(.group)[data-collapsible="icon"] *):hover` is (0,3,0) against `.hover\:bg-sidebar-accent:hover` at (0,2,0). Tailwind decides source order, so that is not something to lean on.
- The bar is decorative; `aria-current="page"` already states it. No new message keys.
- How the bar looks is pinned on the primitive that owns it; each list only tests *when* it shows.
- New **Rail selection bar** entry in `CONTEXT.md`, beside the existing Rail attention pill.

## Verification

- `pnpm --filter web test` — 778 files, 6326 pass, 1 skipped. New coverage: the primitive's classes and the bar's appearance (`ui/__tests__/sidebar-rail-selection.test.tsx`), plus when it shows for Chat rows and for nav rows.
- `pnpm check` and `pnpm typecheck` clean (precommit ran both).
- Generated the Tailwind CSS for the touched files and read the emitted selectors to confirm each utility exists and the specificity margin holds, rather than assuming it.
- **Not yet viewed in the running app** — reviewed in generated CSS and jsdom only. Worth a look at the rail before merge.
- One unrelated flake seen once in a full run, `job-share-modal.test.tsx`; passes in isolation and in the runs either side, and touches nothing here.


mrosberghaus:
## Grok review

### Fixed
- Important: Soko Bot sat on the collapsed rail with `isActive` but never rendered `SidebarRailSelectionBar`, so the hover/active fill override left an open Soko Bot with no mark (`018e2fbc1` — `apps/web/src/app/(app)/components/sidebar/components/personal-assistant-nav.client.tsx`)

### Remaining
- Minor: `sidebarMenuButtonVariants` still paints `active:bg-sidebar-accent` on the icon rail, so a press flashes the same 15% wash this change removed from hover/selected (`apps/web/src/components/ui/sidebar.tsx`)
- Minor: `CONTEXT.md` says the bar is owned by `SidebarMenuButton`; look lives there, but *when* it shows is still an opt-in on each list
- Minor: nav tests only assert bar count, not which destination (`menu-items.test.tsx`)
- Skipped: moving the bar into `SidebarMenuButton` — the PR split look vs when on purpose; filling the missing Soko Bot call site matches Chat and nav
- Skipped: expanded-row restyle — out of scope

### Assessment
The right-edge bar, hover ring, and nav padding geometry match the spec for Chat and dest rows; Soko Bot now uses the same mark. Worth a look at the collapsed rail in the running app before merge — jsdom cannot prove the 15% collision is gone.

mrosberghaus:
All three remaining notes fixed in `8f351b822`.

**`active:bg-sidebar-accent` on the icon rail** — confirmed, and it was a bare `:active` (press), separate from the `data-[active=true]` I had overridden. A press flashed the exact 15% the rail had just stopped reading. A press now deepens the hover ring (`active:ring-2`) instead of reaching back for the fill, so that colour never appears on the rail at all. Expanded keeps all three fill rules.

**`CONTEXT.md` ownership wording** — correct, my sentence conflated the two. Both the entry and the component doc now say plainly that the primitive owns how the mark looks while each list decides *when* it shows, because only the list knows what "open" means for its rows.

**Nav tests asserting count, not destination** — correct, and it was the one way this could fail without looking broken. They now assert the href of the link in the same `li` as the mark: `["/tasks"]` and, from `/projects/project-1`, `["/projects"]`.

---

On the Soko Bot fix: verified the geometry it depends on, which the review didn't mention. `personal-assistant-nav` uses `SidebarGroup` with the default `p-2` and no item padding, so its `SidebarMenuItem` is the same 40px box as a Chat row's and `-right-2` lands flush on the rail. If it had followed the nav list's old pattern (group `p-0` + item `px-2`) the bar would have sat 8px outside the rail.

Also swept every other `SidebarMenuButton isActive` call site for the same gap — the remaining three are all `ChatRoomSidebarRow`, which renders the mark itself. Soko Bot was the only one missing.

Verification: full web suite green (778 files, 6329 tests), Biome and repo typecheck clean, and I re-read the generated Tailwind to confirm the three new `active:` utilities emit with the specificity to beat the base rule.

Agreed on the last point — jsdom can't prove the collision is gone. Still worth a look at the running rail before merge.

PR #4740, opened 2026-09-17
fix(web): paint the sidebar in its persisted collapsed state

mrosberghaus:
## User-facing impact

If you keep the sidebar collapsed, reloading the app used to show the full 14rem sidebar for the entire skeleton phase, then snap it to the rail once hydration finished. Now the rail is there from the first painted frame.

## Why it happened

The app shell is a static Instant Nav prerender, so `(app)/layout.tsx` cannot call `cookies()` and every sidebar streams in expanded. `SidebarProvider` restored the `sidebar_state` cookie in `useLayoutEffect`, which only runs **after** hydration — seconds after the shell paints.

## Approach

A blocking boot script applies the preference **before paint**: it reads the cookie and rewrites `data-state` / `data-collapsible` on each sidebar as it streams in. Hydration lands on the same values, so React never has to correct anything, and the layout stays sync — the static shell and Instant Nav are untouched. Same pre-paint pattern next-themes already uses in the root layout.

The alternative — making the layout `async` and reading `cookies()` — was rejected because it opts the whole app shell out of the prerendered Instant Nav shell (#3601).

## Changes

- `lib/ui-preferences/sidebar-state.ts` (new) — cookie name, parse/serialize, and the boot script source. Sits next to the existing `tasks-view-mode` / `files-view-mode` preferences.
- `app/(app)/components/sidebar-boot-script.tsx` (new) — renders it, with the same `data-cfasync="false"` as the next-themes script.
- `app/(app)/layout.tsx` — mounts it above `SidebarProvider`.
- `components/ui/sidebar.tsx` — emits `data-collapsible-mode` (the mode is not recoverable from `data-collapsible` while expanded), adds `suppressHydrationWarning` on that node, retires the boot observer once React takes over, and reads the cookie through the shared helpers instead of its own inline parsing.

## Verification

- `pnpm --filter web test` — 778 files, 6325 passed, 1 skipped. 9 new tests in `sidebar-state.test.ts` cover sidebars already in the markup and streamed in later, both cookie values, no cookie, non-collapsible sidebars, and the handoff to React.
- `pnpm typecheck`, `pnpm check` — clean.
- Recorded a reload on the local stack with the boot script off and on. Off: the skeleton frame paints a 140px sidebar, then snaps to the 35px rail. On: the skeleton frame is already the rail. Screenshots in the thread.


mrosberghaus:
## Grok review

Range: `5fd2ec9f`..`635f432bf` (`origin/main` merge-base).

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/components/sidebar-boot-script.tsx` — inline `<script>` has no `id`. React 19 de-dupes inline scripts by `id`; `ConsentModeInit` already uses one. A second insert can leak the first `MutationObserver` because `window.__sokosumiSidebarBootStop` is overwritten.
- Minor: `apps/web/src/lib/ui-preferences/sidebar-state.ts` — boot IIFE is not idempotent. Disconnect any existing stop-handle before attaching a new observer if this ever runs twice.
- Minor: `apps/web/src/lib/ui-preferences/sidebar-state.test.ts` — tests build a fake node. They would still pass if `data-collapsible-mode` were dropped from `sidebar.tsx`.
- Minor: `apps/web/src/components/ui/sidebar.tsx` — controlled `open` returns from `useLayoutEffect` before calling the boot stop. Unused in this app (`defaultOpen` only).
- Skipped / disagreed: switch this to `next/script` `beforeInteractive` — that strategy only works in the root layout, and a blocking tag next to the sidebar is the right streaming point (same idea as next-themes). `cookies()` in `(app)/layout.tsx` stays rejected for Instant Nav.

### Assessment
Ready to merge: Yes

Blocking pre-paint cookie rewrite is the right fix for a static Instant Nav shell; hydration handoff, mobile Sheet skip, and CI (`Test Web`, typecheck, biome) all line up.

mrosberghaus:
Pushed `f97eb2042` (merged as `371b8201d` after picking up the `main` merge on the branch).

**Boot IIFE not idempotent** — fixed in `sidebar-state.ts`. It now calls any existing `__sokosumiSidebarBootStop` before attaching its own observer, so a second run cannot leave the first one live. New test `retires the previous observer when it runs twice`.

**Missing `id`** — added (`id="sokosumi-sidebar-state"`). One correction on the mechanism though: React 19 de-dupes hoisted `<script src>` by source, not inline `dangerouslySetInnerHTML` scripts by `id` — that de-dupe is `next/script`'s behavior, and `ConsentModeInit` gets it because it uses `next/script` with `strategy="beforeInteractive"`. So the `id` here is for consistency and CSP/debugging, not the leak guard. The idempotency fix above is what actually closes the double-run hole, and it holds no matter how the tag ends up on the page.

**Controlled `open` returns before the boot stop** — fixed, the stop call moved above the early return. You're right it's unused today (`grep` finds only `defaultOpen` call sites), but React owns the attributes in the controlled case too, so leaving the observer running was a latent leak either way. Note the related gap I did *not* close: the script collapses any `[data-slot="sidebar"][data-state="expanded"][data-collapsible-mode]`, including a controlled one that is meant to stay open. There's no DOM marker for "controlled", and inventing one for zero call sites is YAGNI — worth revisiting if a controlled sidebar ever ships.

**Tests build a fake node** — this was the one with teeth, fixed properly. New `components/ui/__tests__/sidebar-boot-script-contract.test.tsx` renders the real `Sidebar` through `SidebarProvider` and asserts both halves of the contract: the expanded markup carries `data-collapsible-mode="icon"` with a blank `data-collapsible`, and the boot script collapses that real markup. Verified it has teeth by deleting `data-collapsible-mode` from `sidebar.tsx` — both tests go red, where the old stand-in tests stayed green.

Full web suite: 779 files, 6330 passed, 1 skipped. `pnpm check` and `pnpm typecheck` clean.

PR #4743, opened 2026-09-17
fix(web): tell the account chip apart from a chat room

mrosberghaus:
## What

The sidebar footer had no boundary of its own, so at the scroll position where a DM row landed flush against the account chip, the two read as one list: same circular avatar, same presence dot, same bold name over a muted second line. The collapsed rail was worse — there the chip is only an avatar in a column of avatars, and it also used the wrong interaction language.

## Changes

- **Footer hairline.** `SidebarFooter` carries `border-t`, the twin of `SidebarHeader`'s `border-b`, so the rule is continuous across the sidebar and the footer reads as chrome. Applies in both states, so the collapsed rail gets the boundary too.
- **Rounded-square avatar.** Rooms stay circles; the account is a square. One shape says "this is you, not a conversation" at a glance, expanded and collapsed.
- **Rail hover fixed.** On the collapsed rail the chip now adopts the rail's own language from `sidebarMenuButtonVariants`: `ring-1` on hover, `ring-2` when the popover is open, both on transparent. It previously kept its expanded `bg-sidebar-accent` fill, which read as a *selected room* and — being a `rounded-lg` rect flush under a 32px circle — showed at all four corners. The avatar drops to `size-6` inside the same 32px target so the ring clears it by 4px, exactly as a room tile's does.
- **No shape morph.** The Suspense skeleton and the popover's own avatar follow the new shape, so the avatar does not change when credits stream in or when the popover opens.

## Deliberately not included

A scroll shadow on the footer. It needs a scroll listener on `SidebarContent` to appear only when the list is actually scrolled; a permanent shadow looks wrong on a short list. The hairline alone fixes the flush-row case.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — clean (ran in the pre-commit hook across all 11 packages)
- `pnpm --filter web test 'src/app/(app)/components/sidebar'` — 10 files, 69 tests passing

Not exercised in a running browser: sign-in fails against this worktree's database. The change is CSS-class-only; reviewer eyeball on the preview deploy is worth a minute, specifically hovering the collapsed rail's account avatar.


mrosberghaus:
## Grok review

### Fixed
- Important: rail size and hover sat behind JS `isCollapsed`, so SSR and the credits Suspense swap ignored the boot-collapsed group and flashed a 32px square (`2d56fb1ae` — `apps/web/src/app/(app)/components/sidebar/components/sidebar-account-chip.client.tsx`)

### Remaining
- Minor: collapsed trigger stays `rounded-lg`; room tiles use `rounded-md`, so the rail ring is slightly rounder than `sidebarMenuButtonVariants`
- Minor: collapsed presence mark stays `size-2.5`; DM faces on the same 24px rail use `size-2`
- Minor: `transition-colors` is still on the chip while rail hover is a ring

### Assessment
Hairline, square avatar, and rail rings match the PR; group-data classes stop the 24↔32 flash when credits stream in on a collapsed cookie.

mrosberghaus:
Two of the three minors held up and are fixed in `e0b1f1a`-range (latest commit on the branch); the third I'm pushing back on.

**Fixed**

- **Rail radius.** Confirmed: `sidebarMenuButtonVariants` is `rounded-md`, the chip's base is `rounded-lg`, and nothing overrode it collapsed — so the account's hover ring traced 8px where every room's traces 6px. Added `group-data-[collapsible=icon]:rounded-md`.
- **Presence mark size.** Confirmed and slightly worse than described: `direct-room-avatar-stack` draws its mark at `size-2` in *both* states, while its face goes `size-5` → `size-6` on the rail. So on the rail two 24px faces sit in the same column with a 10px mark on the account and an 8px mark on every DM. Added `group-data-[collapsible=icon]:size-2`. Expanded stays `size-2.5`, which is what `PresenceDot`'s own sizing note specifies for a 32px face.

**Disagreed: `transition-colors` on the chip**

`transition-colors` covers `color`, `background-color`, `border-color`, `outline-color`, `text-decoration-color`, `fill`, `stroke`. A Tailwind ring is `box-shadow`, which is not in that set — so the property is already inert on the rail, and the ring snaps there. Room buttons behave identically: `sidebarMenuButtonVariants` declares `transition-[width,height,padding]`, also no `box-shadow`, so their rings snap too. The states are already consistent.

Removing it would mean *adding* `group-data-[collapsible=icon]:transition-none` to suppress a property that has no effect, while the base `transition-colors` still has to stay for the expanded fill. That's a class added to cancel a no-op, so I've left it.

**Verification:** `pnpm check` clean, `pnpm typecheck` clean across all 11 packages, `pnpm --filter web test 'src/app/(app)/components/sidebar'` 10 files / 70 tests passing.

On the important fix — agreed on the mechanism, and the `group-data-[collapsible=icon]` classes are the right call: they match what `SidebarAccountChipFallback` already did, which is exactly why the swap flashed.

Still outstanding on my side: none of this has been seen in a browser. The preprod preview requires a sign-in I can't complete from here (1Password relay unreachable), so the rail hover is verified by tests and class inspection only.

PR #4744, opened 2026-09-17
feat(web): give the sidebar skeleton a shape in both states

mrosberghaus:
## What changed

The cold-boot sidebar painted an empty panel — a header hairline and nothing else — and the streamed room list fell back to full-width bars that the 56px rail had no shape for.

**The boot shell renders the real nav instead of placeholders for it.** `SidebarLogo`, `CustomTrigger` and `MenuItems` are static client components with no session behind them, and `MenuItems` already soft-reads its dialog providers for exactly this frame, so nothing in the nav redraws when the streamed sidebar replaces it. Soko Bot's nav and the Calendar row stay out: both are gated on membership this frame has not loaded, and a row that appears once is cheaper than one that appears and then goes away.

**Only the two session-backed surfaces are placeholders now**, and the room list's is shared between the boot shell and the streamed sidebar. `SidebarChatListSkeleton` is built from the same sidebar primitives the real list uses, so the rail's `group-data-[collapsible=icon]` rules apply to it for free: the section header drops outright and each row keeps only its leading mark, the same 24px tile `ChannelRoomMark` collapses to.

`min-w-10` is what centres that mark on the rail. Every real rail item gets it from `sidebarMenuButtonVariants`, and without it the group's own `p-2` left the tiles 4px left of the nav icons above — caught by measuring, not by eye.

The account chip's placeholder was already collapse-aware; it moves to the `Skeleton` primitive so the sidebar pulses one grey rather than `bg-muted` beside `bg-accent`.

## User-facing impact

On a cold load, and on every Instant Nav frame where the session is still streaming, the sidebar now shows its finished shape on the first frame in **both** the expanded panel and the collapsed rail, instead of a blank column (boot) or bars the rail has no shape for (streaming).

## Files

| File | Change |
| --- | --- |
| `app-sidebar-fallback.tsx` | Empty panel → real logo, trigger, nav, footer chip |
| `sidebar/components/sidebar-chat-list-skeleton.tsx` | New, collapse-aware, shared by both call sites |
| `private-cached-app-sidebar.tsx` | Dropped the local `SidebarChatListFallback` |
| `sidebar-deferred-account.tsx` | Hand-rolled pulses → `Skeleton` primitive |
| `app-sidebar-fallback.test.tsx` | New: asserts the real nav links render, and that only the session-backed surfaces are skeletons |

## Verification

- `pnpm check` — clean
- `pnpm --filter web test` — 6389 passed, 1 skipped, 784 files
- Both states driven in a browser against a throwaway preview route (since deleted). Every rail mark measured at x=28 in a 56px rail, matching the nav icons and the account avatar.

## Screenshots

Not attached — `gh` cannot upload images to a PR body. Both states were captured and reviewed during the change; drag them in here if you want them on the PR.


mrosberghaus:
## Grok review

### Fixed
- Important: cookie-collapsed first paint still used React `state` for the logo, so the 123px wordmark sat in the 56px rail until `useLayoutEffect`. Both states now stay in the DOM and follow `data-collapsible` from the boot script (`19f5968e3` — `apps/web/src/app/(app)/components/sidebar/components/sidebar-logo.client.tsx`)
- Important: no test pinned the collapsed room-skeleton geometry (`min-w-10`, header hide, 24px mark) this PR exists to protect (`19f5968e3` — `apps/web/src/app/(app)/components/app-sidebar-fallback.test.tsx`)

### Remaining
- Minor: `app-sidebar-fallback.test.tsx` does not assert `/drive`, which `MenuItems` renders on desktop
- Minor: skeleton count `15` is a magic number (5 rows × 2 + header 2 + account 3)
- Skipped / disagreed: omitting Soko Bot and Calendar from the boot shell — intentional; a row that appears then vanishes is worse than one that appears once
- Skipped / disagreed: shadcn `SidebarMenuSkeleton` — wrong rail geometry and random widths (hydration mismatch)

### Assessment
Boot shell now matches the finished sidebar in both states; leftover notes are polish, not shape.

mrosberghaus:
Both remaining notes are fixed in `7809f5284`.

- **`/drive` not asserted** — added, with a note that it is desktop-only and `use-mobile` is mocked to desktop above it.
- **Magic `15`** — the room list's marks are now counted against its own row count (`rows.length * 2 + 2`) and the footer's separately, so adding a placeholder row no longer means recomputing a total by hand.

On the logo fix: that is a real bug and it belongs in this PR. I had seen the `state` branch while building the boot shell and wrongly filed it as pre-existing behaviour — but this PR is what puts `SidebarLogo` on the cold-load path, which is exactly when the boot script's `data-collapsible` is ahead of React.

One thing the new tests cannot prove, since happy-dom does not run Tailwind: they assert the class *string* is present, which would still pass if a stacked `group-data-[collapsible=icon]:md:flex` variant emitted no rule at all. Verified in a browser against both states instead:

| | `data-collapsible` | rail width | expand button | wordmark |
| --- | --- | --- | --- | --- |
| Collapsed | `icon` | 56px | `display: flex`, 32px | `display: none`, 0px |
| Expanded | `""` | 224px | `display: none` | `display: flex`, 123px |

The variant does emit, and no wordmark ever occupies the rail.

On the two you marked skipped — agreed on both, and for the same reasons you gave.

`pnpm check`, `pnpm typecheck`, `pnpm --filter web test` (6395 passed, 784 files) all green.

PR #4746, opened 2026-09-17
fix(web): ring every collapsed rail item the same way

mrosberghaus:
## What changed

Every item on the collapsed sidebar rail now hovers the same way: one 32px square, a 1px `sidebar-ring` ring on hover, a 2px ring on press, no fill.

Before this, four things differed:

- **Logo / expand button** still painted a `sidebar-accent` fill on hover while every other rail item had moved to a ring (#4739). It now rings and deepens on press like the rest.
- **Nav links and chat rooms** carried a `min-w-10!` from the 2024 hover-to-expand work, so their rings drew 40×32 while the logo's and the account chip's drew 32×32. The minimum width is gone; the primitive's `size-8!` now means what it says, and `mx-auto` centres the square in the group's 40px, which the minimum width used to do by filling it. The room skeleton from #4744 follows the same rule, so its marks stay on the nav icons' centre line.
- **Soko Bots** rang `rounded-lg` beside `rounded-md` neighbours. Collapsed it is `rounded-md` now; expanded it keeps its `rounded-lg` card.
- **Archived channel rows** were the one non-row element without `group-data-[collapsible=icon]:hidden`, so they leaked into the rail as a clipped glyph plus name when the section had been left open.

`CONTEXT.md` (Rail selection bar) records the 32px square.

## Verification

- `pnpm --filter web test` — 6398 passed, 1 skipped (rebased on #4744)
- `pnpm check`, `pnpm --filter web typecheck` — clean
- New tests: `sidebar-logo.client.test.tsx` (ring, no fill), rail selection test asserts the centred square, Soko Bots test asserts the collapsed radius, the sidebar fallback test follows the skeleton's `mx-auto`

No screenshots from the running app; the change is class-level and the tests pin it.


mrosberghaus:
## Grok review

### Fixed
- Important: collapsed `SidebarMenuButton` stayed 32×40 because `size-8!` sets height, not min-height, so `min-h-10` still won. Added `group-data-[collapsible=icon]:min-h-8!` (`469bf0e9e` — `apps/web/src/components/ui/sidebar.tsx`). Skeleton rows got the same floor.

### Remaining
- Minor: `organization-chat-list.client.tsx`: archived rows now use `group-data-[collapsible=icon]:hidden`, but no test pins that (the other three rail bugs in this PR have tests).
- Skipped: `p-3!` icon offset and Soko Bots `hover:border-primary` on the rail — already fixed in `0c368400e` while this review was in flight; kept that `p-0!` + `justify-center` pairing instead of swapping it for `p-2!`.

### Assessment
Every collapsed rail item is now the same 32px square with the same hover/press ring; leftover is a missing archived-row hide test.

mrosberghaus:
Addressed the remaining note from the Grok review.

- **Archived-row hide test**: added in `3d226f4ad`. `organization-chat-list-section-visibility.test.tsx` renders an archived channel through the shared harness (which now takes `archivedRooms`) and asserts the row carries `group-data-[collapsible=icon]:hidden`.

One correction to the review's "Skipped" line, since it predates a later commit: Soko Bots' `hover:border-primary` is deliberately back on the rail as of `fee3f621e`. It keeps its expanded hover there (primary border and accent fill, with the shared neutral ring suppressed) so the featured entry hovers the same way in both states. `CONTEXT.md` records it as the one exception to the ring rule.

PR #4758, opened 2026-09-18
fix(web): estimate chat row height from text length and list width

mrosberghaus:
## What changes for users

Scrolling up through a chat on iPhone makes messages jump: they shift down in steps while you scroll, then the view jumps once when the scroll stops. Desktop does not show it. This PR makes those jumps smaller. It does not remove them; see "Follow-up".

## Why mobile and desktop differ

- A message above the screen mounts at an estimated height, then gets measured. If it is taller, everything below moves down by the difference and the scroll position has to be corrected.
- On desktop `@tanstack/virtual-core` corrects in the same frame, so nothing is visible.
- On iOS WebKit, writing the scroll position during a touch scroll kills the momentum. The library detects iOS and holds every correction until about 150 ms after the scroll settles (`applyScrollAdjustment` / `_flushIosDeferredIfReady`). Every estimate error is therefore visible.
- The estimate was a flat 80px per message. On a phone the same text wraps to several more lines than on desktop, so the error per row is much larger there.

## The change

`estimateTranscriptRowHeight` now counts the lines the body wraps to, from the message length and the scroller width, capped at the 16-line clamp. `TranscriptViewport` passes `scroller.clientWidth`. Room and thread lists both use it.

## Verification

- `pnpm --filter web test "src/app/(app)/chat/utils/transcript-viewport-model.test.ts" "src/app/(app)/chat/components/transcript-viewport.test.tsx"` — 37 passed (4 new estimator tests)
- `pnpm check` and `pnpm typecheck` — clean (pre-commit hook)
- Not measured on a device yet. The estimate is a heuristic (about half an em per character), so small errors remain and still show on iOS.

## Follow-up (separate PR)

A throwaway prototype on the same `virtual-core` build compared the current top-anchored list with a bottom-anchored one (`flex-col-reverse` scroller, no scroll writes). Stepped scroll up, 150 steps of 60px, in Safari on the iPhone 17 simulator: current layout 10 jumps, largest 124px, 467px jump on settle; bottom-anchored 0 / 0 / 0. Desktop Chromium reads 0 for both. Porting `TranscriptViewport` to that layout would remove the mobile/desktop difference entirely.

No screenshots: no visual change, only scroll behavior.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/utils/transcript-viewport-model.ts`: `md:text-sm` is a **viewport** breakpoint; the estimator uses `listWidth < 768`. A `lg:w-[420px]` thread on iPad/desktop therefore assumes 16px type while CSS is 14px (slight overestimate). Phone portrait — the jump bug — still matches.
- Minor: same file `ROW_INSET_PX = 86` models room `px-5`; thread content is `px-4` (`thread-panel.tsx`). A few extra characters per line, not a new jump class.
- Minor: `transcript-viewport-model.test.ts` never asserts a **long** body at `listWidth` 0 (or omitted) stays 80px. The `<= 0` branch does that; a regression would only show up on real rows.
- Skipped / disagreed: quotes, markdown images, coworker thought chrome, hover `pr-64` gutter — still uncounted. Same as the old flat 80px. This PR only sizes body wrap; that is the iOS error that mattered.

### Assessment
Width-aware line count is the right fix for iOS deferred scroll correction; leftover jumps are heuristic error, not a logic bug. CI Test Web is green. Device measurement is still open, as the PR says.

mrosberghaus:
Reply to the [Grok review](https://github.com/masumi-network/sokosumi/pull/4758#issuecomment-5726215811). All three notes checked against the code; two fixed in 83307b691.

1. **`md:text-sm` is a window breakpoint — fixed.** Confirmed: the thread panel is `lg:w-[420px]` (`thread-panel.tsx:308`), so on a wide window the estimator assumed 16px type where CSS sets 14px. On an iPad that overestimate is visible, because iOS defers the correction there too. The estimator now takes `{ listWidth, viewportWidth }` and picks the type size from the window width against `MOBILE_BREAKPOINT`. New test: a 420px list on a 1440px window wraps at 47 characters, not 41.
2. **`ROW_INSET_PX = 86` models the room's `px-5`, the thread is `px-4` — not changed.** Confirmed, but the error is 8px of about 334px: one character per line. That is below the noise of the half-an-em-per-character assumption the whole estimate rests on, so a second inset constant would add a knob without a measurable effect.
3. **No test for a long body with unknown width — fixed.** Added: a 10,000-character body stays 80px both with no space passed and with `listWidth: 0`.

Skipped items (quotes, markdown images, coworker chrome, hover gutter): agreed, out of scope here.

Tests: 39 passed across the model and viewport files; `pnpm check` and `pnpm typecheck` clean in the pre-commit hook.

PR #4757, opened 2026-09-18
feat(web): show pending invitations on the collapsed rail

mrosberghaus:
## What

A pending External invitation was hidden outright once the sidebar collapsed to icons, so a collapsed reader had no sign of it. It was the one attention state with no mark on the rail.

## User-facing impact

- Collapsed, each pending invitation shows as the Channel tile the room will have once joined (initials plus the globe corner), under the 8px primary Rail attention pill. Mention weight, because an invitation is addressed to the reader.
- The tooltip and the accessible name reuse the existing "Invitation to {name} from {organization}" string, so there are no new translation keys.
- Pressing it expands the sidebar, where Accept and Decline already live.
- Expanded, nothing changes: the button is CSS-hidden there.

`CONTEXT.md` records the rule under "Rail attention pill".

Known limit, not addressed here: a closed External section still hides its rows on the rail, invitations included.

## Verification

- `pnpm --filter web test src/components/chat "src/app/(app)/chat"` (158 files, 1505 tests passed), including 3 new tests beside `pending-invitation-rail-button.tsx`
- `pnpm --filter web typecheck`
- `pnpm check` and `pnpm typecheck` via the pre-commit hook

**Not verified in a running app.** The worktree database has no pending invitation to show. Reviewer: please check the rail on the preview deploy with a real pending External invitation, in light and dark.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4759, opened 2026-09-18
fix(web): show closed sections' rooms on the collapsed rail

mrosberghaus:
## What

The collapsed rail hides every section header. A reader who closed Channels, External or Direct Messages while expanded, and then collapsed the sidebar, lost those rooms from the rail with no way to bring them back, unread ones and pending invitations included.

## User-facing impact

- Collapsed to icons, Channels, External and Direct Messages always show their rooms, whatever was closed while expanded.
- Expanding again restores the reader's own open and closed sections unchanged.
- The phone sheet is excluded: it always shows headers, so closed sections stay closed there.
- Archived is untouched; its rows are hidden on the rail anyway.

Section open state is not persisted across page loads, so this only affected a reader who closed a section and then collapsed the sidebar in the same session.

This also closes the known limit noted in #4757: a pending invitation in a closed External section now shows on the rail.

## Tests

- The list test harness stubbed the collapsible as always open. It now honours `open` and `onOpenChange`, which the new test needs.
- One existing Archived test relied on the old stub, so it opens the section first.
- New test: close External and the room leaves; collapse and it returns; on a phone it stays closed.

## Verification

- `pnpm --filter web test src/components/chat "src/app/(app)/chat" "src/app/(app)/components"` (208 files, 1812 tests passed), run after merging `origin/main`
- `pnpm check` and `pnpm typecheck` via the pre-commit hook

**Not verified in a running app.** To check on the preview deploy: close a section, collapse the sidebar, and confirm its rooms are on the rail; expand and confirm the section is still closed.


mrosberghaus:
Closing in favour of rail section headers: each section keeps a 32px toggle on the collapsed rail, so a closed section can be reopened there and the rail should honour the reader's closed state rather than force every section open. The new header also carries the attention pill when a closed section holds something unread, which covers the case this PR fixed. The test harness change here (collapsible stub honours open and closed) moves over to the new branch.

PR #4761, opened 2026-09-18
feat(web): keep chat section headers on the collapsed rail

mrosberghaus:
## What

The collapsed rail hid every chat section heading. That left three problems: a section closed while expanded could not be reopened from the rail, Channel tiles ran straight into Direct avatars with nothing between them, and every room jumped up by the heading's height when the sidebar toggled.

## User-facing impact

- Channels, External and Direct Messages each keep a 32px icon square on the rail (`#`, a building, a speech bubble). The tooltip names the section and a click opens or closes it.
- A closed section's icon is dimmed.
- A closed section that holds something for the reader says so on its heading, by the same rules its rooms follow: the Rail attention pill on the rail, a bold title when expanded. A pending External invitation counts as a mention. The expanded sidebar had this gap too, so this fixes both states.
- Archived gets no rail header; its rows never show on the rail.
- The loading skeleton has the same square, so the rail keeps its shape while loading.

`CONTEXT.md` gains a "Rail section header" entry.

## The toggle jump

The heading's share of the jump is gone: a 32px square now stands where the 32px title row stood. Some movement remains, because room rows are 40px expanded and 32px on the rail. That is a separate decision and is not changed here.

## Replaces #4759

#4759 forced every section open on the rail. With a toggle on the rail that rule is wrong, not just redundant, so it was closed. Its closing note said the test harness change would move over; it turned out not to be needed here, so it did not.

## Tests

- New `chat-sidebar-section-header.test.tsx`: the rail square is CSS-shown only on the rail, toggles the section, shows the pill only when closed and holding something, and is absent without an icon.
- `room-attention.test.ts`: `resolveSectionAttention` (read, unread, marked unread, mention wins, muted stays quiet, pending invitation).
- `app-sidebar-fallback.test.tsx` pinned the old "header hidden on the rail" rule and now pins the square.

## Verification

- `pnpm --filter web test src/components/chat "src/app/(app)/chat" "src/app/(app)/components" src/components/ui` (221 files, 1964 tests passed)
- `pnpm check` and `pnpm typecheck` via the pre-commit hook

**Not verified in a running app.** To check on the preview deploy, in light and dark: collapse the sidebar and confirm the three squares and their tooltips; close a section from the rail and from the expanded heading; with an unread room in a closed section, confirm the pill (rail) and the bold title (expanded); toggle the sidebar and watch how far the rooms move.


mrosberghaus:
## Grok review

### Fixed
- Important: closed-section attention was visual-only. `RailAttentionPill` is `aria-hidden`; room rows already put `RoomMentions.railMention` / `RoomUnread.railUnread` in the accessible name. Closed headings now reuse those strings so assistive tech still hears the signal after the rooms hide (`6d6167a0b` — `apps/web/src/components/chat/chat-sidebar-section-header.tsx`).

### Remaining
- Minor: `apps/web/src/components/chat/__tests__/organization-chat-list-harness.tsx`: `SidebarMenuButton` is mocked to `null`, so list tests cannot catch a missing `railIcon` on Channels / External / Direct Messages.
- Minor: `apps/web/src/components/chat/chat-sidebar-section-header.tsx`: expanded `CollapsibleTrigger` still has no `focus-visible` ring (pre-existing).
- Minor: `tooltip={{ children }}` accepts a `ReactNode`; every caller passes a string. A string `tooltip` would match room rows.
- Skipped / disagreed: two headings in the DOM, CSS shows one — same pattern as a Channel glyph vs tile, not a duplicate-control bug.
- Skipped / disagreed: `opacity-60` on a closed rail icon — matches muted room rows; it is not a color-token opacity modifier.
- Skipped / disagreed: remaining 40px→32px room-row jump when the sidebar toggles — called out in the PR as out of scope.

### Assessment
Channels, External, and Direct Messages keep a 32px named square on the rail that toggles the section; closed attention now matches the room-row a11y contract. Remaining items are test isolation and pre-existing chrome.

mrosberghaus:
Re: the Grok review above. All three Minor items are fixed in 612b971dc.

- **Reviewer's fix (`6d6167a0b`):** read and kept. The pill is `aria-hidden`, so a closed heading had no spoken signal; reusing the room row's `railMention` / `railUnread` strings is the right contract.
- **Harness mocked `SidebarMenuButton` to `null`:** it now renders a bare marker carrying the tooltip. A new list test asserts Channels, External and Direct Messages each get a rail header, in that order, and Archived gets none. It stays a marker rather than the real button because the square's children repeat the title and would double every `getByText`.
- **No `focus-visible` ring on the expanded trigger:** pre-existing, but one class in a file this PR already rewrites, so it is fixed here (`focus-visible:ring-2` on `ring-sidebar-ring`, the rail's own ring colour).
- **`tooltip={{ children }}`:** the header's `children` is now typed `string`, which is what every caller passes, and the tooltip takes it as a string like room rows do.
- **Skipped items** (two headings in the DOM, `opacity-60`, the 40px→32px row jump): agreed, no change.

Verification: `pnpm --filter web test src/components/chat "src/app/(app)/chat" "src/app/(app)/components"` (209 files, 1828 tests passed), `pnpm --filter web typecheck`, pre-commit `pnpm check` and `pnpm typecheck`. Still not checked in a running app.

PR #4764, opened 2026-09-18
feat(chat): add a Pinned sidebar section with reorderable rooms

mrosberghaus:
## What changes for users

- The chat sidebar gets a **Pinned** section above Channels. It holds pinned channels, DMs and external rooms, in the reader's own order. A pinned room leaves its old section, and activity never moves it. The section is hidden when nothing is pinned.
- Pinned rooms can be reordered in a **reorder mode**, entered from a toggle on the Pinned heading (shown with two or more pins). Each pinned row then shows a grip handle in place of its menu and stops being a link. The handle drags with mouse or touch and takes the up / down arrow keys. Outside the mode nothing drags.
- The per-row pin icon is gone; the section header says it. Unpin stays in the row menu.

## How

- **Core:** new `PUT /v1/chats/rooms/starred` with `{ roomIds }`. It rewrites `starredAt`, the key every client already sorts pins by (oldest first), one millisecond apart and all in the past. No migration. The Apple app and older web builds show the new order unchanged.
- Lenient on a stale list: ids the caller has not starred are ignored, starred rooms left out keep their order after the listed ones, and the call never stars or unstars a room. It only writes to rows that are still starred, and writes nothing when the order already matches.
- **Web:** `partitionRoomsForSidebar` gains a `pinned` bucket; the pinned branch leaves the activity comparator. Drag uses the already-installed `@dnd-kit/core` (no new dependency). The update is optimistic; the latest reorder wins, and a failure reloads the list from Core with a toast.
- `CONTEXT.md`: `starredAt` is now documented as a sort key, not the time of pinning.

## Not in this PR

- Apple app: pins stay inline there (already in the new order). The Pinned section in Swift is a follow-up.

## Verification

- `pnpm check`, `pnpm typecheck`, `pnpm test` pass.
- New tests: Core route (mocked Prisma, 7), partition and sort, row reorder mode, list-level Pinned section (mode toggle, arrow-key reorder, failure reload), move helper.
- Checked in a headless browser against the real list component with fixture rooms: outside the mode a drag does nothing; in the mode a click on a row does not navigate, and a handle drag reorders with rows stepping aside. Arrow keys are covered by unit tests. Touch drag is not yet checked on a device.
- Not yet checked against a signed-in session with real data. Reviewer or preview deploy: pin three rooms, drag one, reload, confirm the order holds.


mrosberghaus:
## Grok review

### Fixed
- Important: `PUT /v1/chats/rooms/starred` loaded every starred membership for the user, so a reorder in one sidebar rewrote other workspaces' exclusive pins and returned those room ids (`ab2222d0e` — `apps/core/src/routes/v1/chats/rooms/starred/put.ts`). The read now uses the same membership-visible active set as `GET /chats/rooms`. Shared rooms (personal DMs, guests, matched) still share one `starredAt` and can move in every sidebar they appear in.

### Remaining
- Minor: `apps/web/src/components/chat/chat-room-sidebar-row.tsx`: reorder mode still renders a `Link` (`pointer-events-none` + `tabIndex={-1}`). Pointer users cannot navigate; AT still sees a link. Collapsed rail correctly keeps the link via CSS, which is why this is not a JS `aria-hidden`.
- Minor: `apps/web/src/components/chat/organization-chat-list.client.tsx`: `pinnedReorderMode` stays true if the pin count drops below 2 (toggle hides). Pinning another room later re-enters reorder mode.
- Minor: `apps/core/src/schemas/chat-room.schema.ts` `ChatRoom.starredAt` still says "when the current user starred this room"; `CONTEXT.md` now calls it a sort key.
- Minor: starred reorder still does one `updateMany` per room inside an interactive transaction (capped at 500). Fine for real pin lists; a single `CASE` update if that cap is ever hit.
- Skipped: collapsing the membership-visible `OR` with `GET /chats/rooms` into one helper — two copies, GET still owns the list filter.

### Assessment
Auth is session-user only, order persists on `starredAt`, collapsed rail still opens rooms, and exclusive pins no longer leak across workspaces.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/4764#issuecomment-5727757380). Addressed in `a22f8823a`.

**The workspace-scoping fix (`ab2222d0e`)**: checked and kept. Web sends only the session cookie and Core takes the active organization from the session, the same way `GET /chats/rooms` does, so the scope reaches the route from web. Route tests pass (9).

**Remaining notes**

- **Two copies of the membership-visible filter (marked Skipped):** changed. It is an access filter, and the reorder is now scoped by it, so a drift between the copies would make the route write to rooms the list does not show. Extracted `membershipVisibleActiveRoomWhere` into `rooms/helpers.ts`; `GET /chats/rooms` and `PUT /chats/rooms/starred` both use it. The `GET` filter is unchanged clause for clause; the Core chat route tests pass (618).
- **Organization header:** the reorder route now declares `X-Organization-Slug` like the list it is scoped with. Client regenerated.
- **`pinnedReorderMode` survives the pin count dropping below 2:** fixed. The mode ends when its toggle leaves the screen (section closed, or fewer than two pins), so it cannot resume by itself. New test fails without the fix.
- **`ChatRoom.starredAt` description:** fixed; it now says sort key, matching `CONTEXT.md`. `CONTEXT.md` also records that a reorder is scoped to one sidebar and that shared rooms move in each.
- **Link still exposed to assistive tech in reorder mode:** left as is. The same element has to stay a working link on the collapsed rail, which only CSS can tell apart, so `aria-disabled` or `aria-hidden` on it would break the rail. The cost is small: the link is out of the tab order, and a screen-reader user who activates it opens the room, which is harmless. The handle carries the reorder instructions in its name.
- **One `updateMany` per room:** left as is, as suggested. The no-op check already skips all writes when the order is unchanged.

Checks: `pnpm check`, `pnpm typecheck`, Core chat routes, web chat and client tests pass.

PR #4768, opened 2026-09-18
fix(web): anchor the chat transcript to the bottom so it never jumps on iOS

mrosberghaus:
## What the reader gets

Scrolling up through a chat on an iPhone no longer jumps. Desktop and mobile share one code path.

## Why it jumped

On iOS, `@tanstack/virtual-core` defers every scroll correction until a touch scroll settles, so each row-height estimate error above the view showed on screen, then one flush jump. Background: #4758, prototype on `prototype/chat-bottom-anchored-scroller`.

## What changed

- The message scroller (`CHAT_MESSAGE_LIST_SCROLLER_CLASS`, room shell and thread panel) is `flex-col-reverse`: `scrollTop` is 0 at the newest message and negative above it. Growth above the view no longer moves it, so no correction is needed. Its one child is `shrink-0`.
- `TranscriptViewport` maps offsets with a custom `observeElementOffset` / `scrollToFn` and sets `virtualizer.shouldAdjustScrollPositionOnItemSizeChange` on the instance (as an option it is ignored).
- Removed: the clamped-scroll finish in `onChange`, the `ResizeObserver` re-pin, `pinToEndAfterRowsChange`, and the open-time `scrollToEnd`. The browser does all four now.
- **Not in the original plan:** growth at or below the top edge (a new or streaming message while reading history, unmeasured rows below after a landing) does move a bottom-anchored view. It is put back with one write. While a scroll or touch is under way the list is pulled down by a negative bottom margin instead, and that becomes one write when the scroll ends, so iOS momentum survives.
- `rowHoldAfterRowsChange` now returns the first surviving row from the top edge down, and keeps it exactly where it was.
- New `__tests__/transcript-viewport-virtualizer-contract.test.ts` pins the library behavior this relies on, so a `virtual-core` upgrade fails there.

## Measurements

Throwaway `/share` route with the real `TranscriptViewport`, 600 rows of 40–300px against the 80px estimate, 150 steps of 60px with a finger held down (route removed before this PR).

| Setup | Before | After |
| --- | --- | --- |
| Chromium, iPhone user agent, scroll up | 62 jumps, largest 635px | 0 |
| Chromium, iPhone user agent, scroll down after a landing | n/a | 0 |
| Chromium, desktop user agent, up and down | 0 | 0 |
| Safari, iPhone 17 simulator, 10 real flicks up | 4–10 (prototype numbers) | 0 in 925 scrolled frames |
| Safari, iPhone 17 simulator, 3 real flicks down after a landing | n/a | 0 in 307 frames, momentum kept |
| Prototype `?mode=reverse&auto=1`, desktop Safari 27 | | 0 |

Checked one at a time in the same route, both user agents: row hold on prepend (0px moved), prepend at the live edge, append at the end / 100px above it / far above it, `landOnMessage` (centered, marked), `scrollToMessage` (smooth, centered), jump-to-latest pill, short transcript pinned to the bottom, composer growing and shrinking, PageUp / ArrowDown / wheel direction.

## Not done

- **Firefox is not measured.** It is not installed on the machine this was built on.
- While growth is held back by the margin, `scrollTop` 0 stops short of the newest rows by that amount until the scroll ends (about 150 ms).

## Verification

- `pnpm test`: web 6419 passed. One Core test (`src/routes/sync/index.test.ts`, cron auth) timed out at 5 s in the full run and passes alone (42/42); no Core code is touched.
- Still open before marking ready: a real iPhone on the Vercel preview.


mrosberghaus:
## Grok review

Reviewed `a9262a435..72abbb7f7` (8 files). Requirements: iOS Safari jump gone; keyboard/composer inset; new messages still pin; thread vs room.

### Fixed

### Remaining
- Minor: Firefox unmeasured (PR already says so). Modern Gecko uses negative `scrollTop` on `column-reverse` ([bug 1851749](https://bugzilla.mozilla.org/show_bug.cgi?id=1851749)), so this is a verification gap, not a known break.
- Minor: `transcript-viewport.tsx` jump-to-latest is `sticky bottom-3` inside a `flex-col-reverse` scroller — known CSS footgun. Author listed the pill as checked; still worth the real iPhone pass already in Not done.
- Minor: `thread-panel.tsx` content `shrink-0` is not in `room-shell-scroller-overflow.test.ts` (room class is). Same clamp if that class loses `shrink-0`.
- Minor: `keepRowsInPlace` can set `margin-bottom: -${delta}px` with a fractional `delta` (whole-pixel layout rule). Tests use integer row heights.
- Skipped / disagreed: live-edge `scrollToFn` dropping `adjustments` looks like it would write `scrollTop` away from 0 — it does not. `resizeItem` calls `scrollToFn` before `directDomUpdates` grows the sizer, so `offset - maxScroll` is a no-op; `onChange` then reads `topOffsetOf` after the height write.

### Assessment

The reverse-flex scroller, `shrink-0` child (room + thread), mapped `observeElementOffset` / `scrollToFn`, instance-level size-change veto, and deferred negative margin under touch match the requirement. Keep the PR draft until a real iPhone on the Vercel preview.

mrosberghaus:
Reply to the [Grok review](https://github.com/masumi-network/sokosumi/pull/4768#issuecomment-5727183781):

- **Thread `shrink-0` not covered** — fixed in 4d0ce124e. Room and thread now build their list content class from one base, `CHAT_MESSAGE_LIST_CONTENT_CLASS`, next to the scroller class. `room-shell-scroller-overflow.test.ts` asserts `shrink-0` on that base and that `thread-panel.tsx` uses it.
- **Fractional `margin-bottom`** — not reachable, no change. The value is a sum of size deltas. Measured sizes are `Math.round(borderBoxSize)` or `offsetHeight` in virtual-core's `measureElement`. Estimates from `estimateTranscriptRowHeight` are sums of integer constants. The rows-change path rounds explicitly (`Math.round(scroller.scrollTop - target)`). Rounding inside `keepRowsInPlace` as well would make the margin and the later scroll write disagree by the rounding error.
- **`sticky bottom-3` pill in a reversed flex scroller** — no change. The pill sits inside the scroller's one child, which is a normal `flex-col`, so it is not a flex item of the reversed container. It rendered pinned above the composer in Safari on the iPhone 17 simulator and in Chromium. It stays on the real-iPhone checklist.
- **Firefox** — still unmeasured. The PR stays draft until the real-iPhone pass on the Vercel preview.
- **Live-edge `scrollToFn` dropping `adjustments`** — agreed with the analysis. The contract test pins that call (`skips the veto at the live edge and hands the growth over as adjustments`).

PR #4773, opened 2026-09-18
fix(web): debounce the chat hover action pill

mrosberghaus:
## What

Scrolling a transcript drags a stationary pointer across row after row, and the hover action pill (quick reactions + More) flashed in and out at every message it passed.

Two parts, because the first alone was not enough:

1. **A 200ms reveal delay.** The pill waits before fading in; leaving the row clears it immediately, so it still disappears at once.
2. **Nothing at all while the transcript is scrolling.** The delay only covers rows that pass under the pointer faster than 200ms — a trackpad scroll leaves each row sitting there for longer, so the pill still flashed its way down the list. The scroller now carries `data-chat-scrolling` for as long as a scroll runs, and the pill is held back while that mark is on.

After the scroll settles the pill fades in on whichever row the pointer was left on.

## How

- `room-message-row.tsx`: `transition-opacity` → `transition-[opacity,pointer-events] transition-discrete` plus `[@media(hover:hover)]:group-hover:delay-200`. `pointer-events` rides the same delayed transition — otherwise the pill would be invisible but clickable for 200ms over the row above, exactly the hazard the surrounding code already guards against.
- `use-quiet-hover-while-scrolling.ts`: sets the mark on scroll, clears it 150ms after the last event so momentum gaps do not let the pill through. Plain DOM attribute writes, no re-render per scroll event. Wired into the room transcript (`rooms-client`) and the thread panel.
- `globals.css`: one unlayered rule holding the pill back under the mark. Unlayered beats Tailwind's `@layer utilities` regardless of specificity — the same mechanism the neighbouring `data-search-landed` rules already rely on. Its `transition-delay: 0ms` drops the pill the instant a scroll starts, while the way back keeps the 200ms reveal delay.

Keyboard access is untouched: `focus-within:opacity-100` carries no delay. Touch (`hover:none`) is unaffected — the pill is `hidden` there and the long-press sheet owns message actions.

## Verification

- `pnpm check` (biome), `pnpm typecheck` — clean
- `pnpm --filter web test src/app/(app)/chat` — 1168 passed, including 4 new tests for the scroll mark (set on scroll, held across momentum gaps, cleared and unlistened on unmount, inert until the scroller mounts)
- Confirmed Tailwind 4.3.3 emits all three new utilities, that `tailwind-merge` 3.7.0 drops none of them, and that utilities land in `@layer utilities` so the globals.css rule wins


mrosberghaus:
## Grok review

### Fixed
- Important: unlayered `[data-chat-scrolling]` hid a keyboard-focused hover pill (`502f672af` — `apps/web/src/app/globals.css`). The rule now skips `:focus-within`, so `focus-within:opacity-100` still shows, and a source contract test locks the selector.

### Remaining
- Minor: `apps/web/src/app/(app)/chat/hooks/use-quiet-hover-while-scrolling.test.ts`: settle test advances 1000ms instead of asserting the 150ms window.
- Minor: overflow/More trigger can still hide if the transcript scrolls while the menu is open (focus is in the portal, so `:focus-within` does not keep the pill).
- Skipped: 200ms hover delay and 150ms scroll settle — match the PR; no change.

### Assessment
Delay plus the scroll mark stops the flash; keyboard-focused pills stay visible during transcript scroll.

mrosberghaus:
Both remaining notes addressed in `40e2a416b`, and the `:focus-within` fix in `502f672af` is right — I had judged that case acceptable and it isn't.

**Open menu (fixed).** Reachable, but not the way the note describes: `DropdownMenu` is modal, so a wheel scroll can't reach the transcript while More is open. What does reach it is a live message pinning the transcript to the bottom — the scroll mark then hid the pill the open menu is anchored to. The rule now also skips a pill with an expanded trigger:

```css
[data-chat-scrolling]
[data-message-actions="hover"]:not(:focus-within):not(:has([aria-expanded="true"])) {
```

`:has([aria-expanded="true"])` covers the emoji picker too (also a Radix trigger, and its Popover is non-modal, so that one *is* wheel-reachable). Verified both primitives render `aria-expanded` from their open state rather than assuming it.

**Settle window (changed, not as asked).** Advancing 1000ms did prove almost nothing, so the test now brackets the window — still marked at 100ms, gone by 300ms. Not pinned at exactly 150ms on purpose: that number is a tuning knob, and a test asserting it would have to be edited on every tweak while proving nothing about behaviour. The bracket locks what actually matters — the mark outlives the gap between two wheel ticks, and it clears while the pointer is still where the reader left it.

One note on the contract test that came with `502f672af`: it matched the selector as a literal substring, which broke as soon as biome wrapped the longer selector across lines. It now normalises whitespace first, and asserts both skips.

`pnpm --filter web test src/app/(app)/chat` — 1169 passed. `pnpm check` and `pnpm typecheck` clean.

PR #4774, opened 2026-09-18
fix(web): dismiss the keyboard when the reader drags the chat transcript

mrosberghaus:
## User-facing impact

On iOS Safari, dragging a chat transcript now closes the keyboard. Before, it stayed up over the messages the reader was scrolling back to read, and the composer kept focus with no way to dismiss it short of tapping elsewhere.

Dragging means reading, not typing — this is the platform's own `keyboardDismissMode = .onDrag` behaviour, which the web app did not have. It applies to both the room transcript and the thread panel.

## Implementation

A `touchmove` listener on the scroller blurs the focused `[contenteditable="true"]`.

- **`touchmove`, not `scroll`** — closing the keyboard resizes the visual viewport, which scrolls a bottom-anchored scroller. A `scroll` binding would re-blur on every keyboard open. `touchmove` only fires for a real finger drag, never for momentum or a resize.
- **Any drag dismisses**, not only upward — matching iOS. Down-drags at the live edge move nothing anyway.
- Lives in `TranscriptViewport`, which already owns the scroller's touch listeners, so the room and the thread panel share one implementation.

Matched by attribute rather than `isContentEditable`, so the rule is the markup React writes.

## Out of scope

Two other points were raised alongside this and deliberately not changed:

- **Dismissing the composer on send** — every major mobile chat client keeps the keyboard up after sending, because people send bursts. Dismissing costs a re-tap per message.
- **The `^ v ✓` bar above the keyboard** — that is iOS Safari's native form-navigation accessory view, not our composer. Nothing in this repo renders or can remove it.

## Verification

- `pnpm --filter web test "src/app/(app)/chat"` — 126 files, 1166 tests passed
- `pnpm check` — clean
- `pnpm typecheck` — clean

New tests in `transcript-viewport.test.tsx`: a drag blurs the composer, a tap without a drag does not. Verified red before green — with the listener removed the drag test fails and the other 23 still pass.

Not yet verified on a physical iPhone; happy-dom cannot prove the keyboard actually retracts, only that focus leaves the editor.


mrosberghaus:
## Grok review

Range `1d8f517ca`..`5b8848eb5` against the PR title/body (dismiss keyboard on transcript drag). Branch name still says close-on-send; that is out of scope in the body and is not in the diff.

### Fixed
- Important: in-row message edit lives inside the transcript scroller. A `touchmove` on that editor was blurring it, which also cancels an unchanged edit via `MessageEditComposer` `onBlur`. Skip blur when the drag target is inside the focused editor (`5b8848eb5` — `apps/web/src/app/(app)/chat/components/transcript-viewport.tsx`).

### Remaining
- Minor: `transcript-viewport.test.tsx` still does `createRef()` during render in `HarnessWithComposer` (test-only).
- Minor: no movement threshold — a 1px wobble on the transcript still blurs. Left as-is; the PR asked for any drag, matching iOS `keyboardDismissMode = .onDrag`.
- Skipped / disagreed: dismiss-on-send — not a gap. Body explicitly keeps the keyboard up after send.
- Skipped: physical iPhone not run here. happy-dom only proves focus left the editor.

### Assessment
Dismiss-on-drag is in `TranscriptViewport` (room + thread), does not steal the bottom composer while typing, and in-row edit stays focused when the finger is on the editor. One iPhone pass still wanted.

mrosberghaus:
Verified the in-row edit finding — it's real, and the fix in `5b8848eb5` only covered half of it. Pushed `fa654f181`.

**In-row edit blur.** Confirmed: `ComposerWysiwygEditor` in `room-message-row.tsx:1809` cancels an unchanged edit `onBlur`, and it renders inside the transcript scroller, so my `touchmove` handler was closing open edit boxes. Real bug I introduced.

But guarding on `event.target` only covers a drag *on* the editor. A drag anywhere else in the transcript — scrolling back to re-read the conversation before typing your edit, which is the normal thing to do mid-edit — still blurred it and still closed the edit box. Test for that case fails against `5b8848eb5`.

Replaced the target check with a scroller check:

```ts
if (
  active instanceof HTMLElement &&
  active.matches(EDITOR_SELECTOR) &&
  !scroller.contains(active)
) {
  active.blur();
}
```

The room composer (`room-shell-layout.tsx:101`) and the thread composer (`thread-panel.tsx:405`) both render *after* the scroller, so "outside the scroller" is exactly the composer and nothing else. Every editor inside the transcript is left alone whatever the finger is on. Also drops the `event` parameter.

**`createRef()` during render** — fixed. The ref was never read; `ref={null}`.

**Movement threshold** — agreed, leaving it. `onDrag` is the platform rule and a wobble costs a re-tap, not data.

**Branch name** — correct, it's stale. The branch was cut before the scope narrowed to dismiss-on-drag; renaming it now would orphan this PR.

`pnpm --filter web test "src/app/(app)/chat"` — 126 files, 1168 tests passed. Biome and typecheck clean. Still not run on a physical iPhone.

PR #4775, opened 2026-09-18
fix(web): give the mobile chat list 44px touch targets

mrosberghaus:
## Impact

On a phone the chat list rows and their `…`, `+` and browse controls were 32px, well under the 44pt touch target. Below `md` they are now 44px. Desktop density is unchanged.

- Room rows, section headers and archived rows are 44px tall.
- The room menu, browse-channels and archived restore/menu buttons keep their 32px box (so `…` stays in the `+` column) and reach 44px through a pseudo-element. The header's two controls get a gap so their targets do not overlap.
- The whole section header height toggles the section.
- Archived row controls were `opacity-0` until hover, so invisible on touch. They now always show on touch.

## Verification

- `pnpm --filter web test src/components/chat` (369 passed, one new test)
- `pnpm check`, `pnpm typecheck` (pre-commit hook)
- Not yet checked on a device: verify the chat list on the Vercel preview from a phone.


mrosberghaus:
## Grok review

### Fixed
- Important: compose `+` hid its 44px hit from `sm` (640px); the mobile list is `md` (768px) (`8d76f6ff8` — `apps/web/src/app/(app)/chat/components/chat-compose-dialog.ts`)
- Important: leading-mark tests substring-matched `md:size-5` / `md:h-5`, so a missing 28px mobile slot still passed (`8d76f6ff8` — `channel-discoverability-icon.test.tsx`, `direct-room-avatar-stack.test.tsx`, `chat-room-sidebar-row.test.tsx`)

### Remaining
- Minor: `SidebarChatListSkeleton` is still `h-10` / `min-h-10` / `size-5`, so the loading list is 4px shorter than live 44px rows
- Minor: the room-menu test asserts class tokens, not computed 44×44 geometry
- Skipped / disagreed: `after:` without `content-['']` — Tailwind 4 sets `content: var(--tw-content)` on `after`/`before` variants; the `+` still spells `before:content-['']` because that class predates this PR
- Skipped: `+` stays `size-7` (28px box + 8px `before`) rather than the 32px `…` box — 4px visual mismatch is pre-existing; 44px hit is what the PR asked for

### Assessment
44px targets land below `md` on rows, menus, browse, archived, and now `+`. Desktop `md:` density is unchanged. Not checked on a phone.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/4775#issuecomment-5727750657)

- **Skeleton 4px short** — fixed in `7e8604659`. `SidebarChatListSkeleton` now uses the live mobile geometry below `md`: 44px header and rows, 28px leading mark, 12px gap, chevron centred in the leading column. `md` and the collapsed rail are unchanged.
- **Room-menu test asserts class tokens, not 44×44 geometry** — not changing. The web suite runs in happy-dom, which has no layout engine, so `getBoundingClientRect()` is all zeros and pseudo-elements have no box. The file's existing tests pin class tokens for the same reason. Measured geometry would need a browser test, which this PR does not add.
- `after:` without `content-['']` and the `+` staying `size-7`: agree with skipping both.

`8d76f6ff8` pulled and verified: 673 tests pass across `src/components/chat` and `src/app/(app)/components`. Still not checked on a phone.

PR #4771, opened 2026-09-18
feat(apple): show chat notifications while the app runs

mrosberghaus:
## Feature

Apple chat parity **row 40** (the last Todo row): chat notifications while the native app runs.

- **Settings → Notifications**: the Chat group of web's account-wide notification matrix. Preset menu (Most / Essential / In app / Off, Custom when the cells match none) and one reach per kind (every room message, mentions, direct messages): Off / In app / In app and banner. Optimistic with rollback and inline "Could not save that change…", disabled while saving. A write that leaves a banner on asks macOS for notification permission (web's moment: the press on a push cell) and records the account-wide `pushOptIn` consent Core requires before it marks any event `osBanner`; a write that leaves no banner on in any group releases it. A denied OS permission is explained with a link to System Settings.
- **Events**: chat-kind `notification_created` rows arrive on the user notifications channel, whose name is read from the Ably token capability (production `notifications:all:user_{id}`, branch-scoped on previews) and only subscribed when granted.
- **Banners**: an unread event with `osBanner` raises a local notification only while OS permission is granted and the app is unfocused (web `document.hasFocus()` → the app is not frontmost, or no chat window is active). One banner per room (`sokosumi-room:{roomId}`); each arrival replaces it and carries the waiting count, with web's titles and bodies. An already-read event (another device, web, or reading the room) takes the banner down before any display gate.
- **Opening**: attempts `PATCH /notifications/{id}/read`, switches to the notification's workspace when it names another one, then navigates to the room, message or reply thread — even if marking read fails.
- Room and thread mute are applied by Core at fan-out (web has no client gate), so a muted room sends no event.

## Web sources audited (at `8890ecc28`)

[...]

PR #4778, opened 2026-09-18
feat(apple): show reactions before Core confirms

mrosberghaus:
## Feature

Apple chat parity row **20c — pending reactions** ([ADR 0032](https://github.com/masumi-network/sokosumi/blob/main/docs/adr/0032-reactions-are-idempotent-and-shown-before-confirm.md)). A reaction tap on macOS now shows at once (count, own-reaction highlight, your name at the end of the reactor list) instead of waiting for Core. The last tap per message and emoji wins, realtime patches and refreshed pages never flicker the tap back, and a failure rolls back only that emoji with the existing error alert. Rooms, thread parents and replies.

## Web sources audited

- `apps/web/src/app/(app)/chat/utils/pending-reactions.ts` (+ `pending-reactions.test.ts`)
- `apps/web/src/app/(app)/chat/components/rooms-client.tsx` (`handleToggleReaction`, `overlayReactions`, `mergeConfirmedReactionIntoState`) and `__tests__/rooms-client-pending-reactions.test.tsx`
- `apps/web/src/app/(app)/chat/components/room-message-row.tsx` (emoji-use recording)
- `packages/utils/src/chat-room-reactions.ts` (`MAX_LISTED_CHAT_REACTION_REACTORS`)

Web and the ADR agree; nothing was ambiguous and no API is missing.

## Included

- `SokosumiChat/PendingReactions.swift`: pure, UI-free model. Tap-ordered intents, `tap` / `intent(…after:)` / `settle`, and the overlay with web's rules (idempotent, count ± 1, entry dropped at zero, viewer listed last while under the cap of 20 and a name is known).
- `WorkspaceState.toggleReaction`: one request loop per message and emoji over the idempotent `PUT`/`DELETE …/reactions/{emoji}` routes. Only the newest intent is sent (on, off, on is one request; on, off is two), each response merges only that emoji, a failure drops only that emoji's intent, 401 still signs out, and the request outlives leaving the room. `reset()` clears intents.
- `displayedTranscript`, `displayedThreadReplies` and the new `displayedThreadParent` apply the overlay; confirmed rows are written only by Core responses, pages and realtime, so re-application after a realtime patch holds by construction.
- The row stays a consumer: `pendingReactionEmoji` / `pendingEmoji` and their disabled states are gone because every tap must register. Appearance and the selected accessibility traits are unchanged.
- Deleted the path this replaces: the duplicate-tap guard, generation guards, update-after-success and the one-`GET` same-emoji reconciliation from slice 20.
- PARITY: row 20c in review with evidence; row 40 flipped to Done (#4771) with its checkpoint and the two review follow-ups (`ac4eb5445`, `b6a17ff1f`).

## Native adaptations / excluded

- Quick-reaction history (20b) is still recorded only after Core confirms an add, once per request chain; taps absorbed by a running request record nothing (web records at tap time, once per emoji). Removing still teaches nothing.
- The error stays the row's alert, not a global toast; if the row is gone when the failure arrives only the rollback is visible.
- No Core contract, dependency, OpenAPI snapshot, generated-file, project-setting or `apps/web` change. No new fixtures: the row draws the same views from the same DTO.

## How to test

1. In a room and in a reply thread (parent and a reply), click a quick reaction or a chip: count and highlight change immediately.
2. Click the same emoji several times quickly: the last click is what stays and what Core holds after a reload.
3. React from web on the same message while the native request is slow: the native tap does not flicker.
4. Go offline and tap: only that emoji snaps back, with the error alert.
5. Add an emoji and see it move up in the quick reactions; remove one and see no change.

- `swift test --package-path Packages/SokosumiChat` — 530 passed (new `PendingReactionsTests`)
- `swift test --package-path Packages/SokosumiWorkspace` — 115 passed (seven new tests with paused `PUT`/`DELETE`: instant display in room/reply/thread parent, last tap wins with one request at a time, on/off/on sends one request, realtime patch mid-flight plus per-emoji rollback with two pending emoji, leaving and re-entering the room, failure after leaving)
- `SokosumiAuth` 35, `SokosumiRealtime` 31, `CoreAPI` 1 passed
- `mint run swiftformat --lint --cache ignore .` — 0/367 need formatting; `mint run swiftlint lint --strict --no-cache` — 0 violations
- `xcodebuild … build` (macOS arm64, ad-hoc) — BUILD SUCCEEDED, no warnings in hand-written sources; `xcodebuild … test -only-testing:SokosumiTests -enableCodeCoverage NO` — TEST SUCCEEDED, 95 tests
- `swift build --package-path Packages/SokosumiWorkspace --triple arm64-apple-ios17.0 --sdk "$(xcrun --sdk iphoneos --show-sdk-path)"` — Build complete

Unverified: no live reaction was sent and the app was not launched, so perceived tap latency in the running app, rapid physical clicks on a chip that disappears at zero, the alert after a real Core failure, Core's realtime echo racing the `PUT` response, and VoiceOver were not exercised. The web audit is source-based.


mrosberghaus:
## Grok review

### Fixed
- Important: reaction chips painted the last `PreparedTranscript` snapshot, so the overlay waited on a detached markdown prepare and a second tap could invert last-tap-wins (`fbe1a957d` — `apps/apple/Sokosumi/Chat/Timeline/RoomTimelineView.swift`, `apps/apple/Sokosumi/Chat/Threads/ReplyThreadView.swift`)

### Remaining
- Minor: `PendingReactions.swift`: `maxListedReactionReactors = 20` is a copy of `MAX_LISTED_CHAT_REACTION_REACTORS`; Apple packages cannot import `@sokosumi/utils`. Tests lock the cap.
- Minor: `PendingReaction.reacted` is a mutable `var` on a `Hashable` struct. Safe today because intents live in an array, not a `Set`.
- Minor: `.task(id: preparationInput)` still re-prepares the transcript on every tap because `Input.messages` includes the overlay. Extra work only; chips no longer wait on it.
- Skipped: quick-reaction history still records after Core confirms, and the error stays the row alert — documented native adaptations vs web, not bugs.
- Skipped Minors: no live app launch, so VoiceOver, a chip disappearing at zero, and a real Core 403 were not re-checked.

### Assessment
ADR 0032 holds: last tap wins, one in-flight request per message/emoji, overlay on top of realtime, per-emoji rollback, request outlives leaving the room. Chips now follow the live overlay instead of the markdown snapshot.

mrosberghaus:
Response to the Grok review.

**Chips painted the last `PreparedTranscript` snapshot** — confirmed fixed in fbe1a95: rows take reactions from the live overlay, so a tap shows before the markdown prepare and a second tap can't invert last-tap-wins. Verified locally on that head: strict SwiftFormat/SwiftLint clean, macOS app tests pass.

**`maxListedReactionReactors = 20` copies `MAX_LISTED_CHAT_REACTION_REACTORS`** — leaving as is. Swift packages cannot import `@sokosumi/utils`; the value is part of the wire behavior (Core truncates the reactor list at the same cap), the constant names its origin, and the tests lock it.

**`PendingReaction.reacted` is a `var` on a `Hashable` struct** — leaving as is. It is a value type mutated in place inside an array (`intents[index].reacted.toggle()`); a `Set` element cannot be mutated in place, so the hash can never go stale. Making it `let` would turn the toggle into remove-and-reinsert and lose tap order, which the model keeps on purpose.

**`.task(id: preparationInput)` re-prepares on every tap** — leaving as is. `PreparedTranscript.prepare(_:reusing:)` reuses parsed documents, so the pass is an identity check per row; it also converges the snapshot's reactions with the overlay so the live map in fbe1a95 stays a bridge rather than a second source. Excluding reactions from `Input` equality would touch room and thread preparation for a cost nobody has measured; perceived tap latency is listed as unverified.

Quick-reaction history after Core confirms and the row alert for errors stay the documented native adaptations.

PR #4781, opened 2026-09-18
feat(web): crossfade the chat mention badge with the room menu

mrosberghaus:
## What a reader sees

A chat room with unread @mentions used to spend its row three times over: the
badge, the gap before it, and the hole that kept the badge from jumping when
the room menu faded in. The name ended **52px** in from the row's edge, where
every other row ends at **36px**.

The badge now stands in the room menu's column and fades out as the `…` fades
in — the same swap the muted bell already does. So:

- **Badge present** → the name is shortened once, by the 16px hole the badge and
  the menu share, and it holds there in every state. Nothing reflows on hover.
- **No badge** → unchanged. Full name width at rest, hole opens with the button.

That is **16px back for the name** on every mentioned room, and one fewer
special case in the trailing-spacer rule.

## Details

- `MentionBadge` is now `aria-hidden` decoration in the trailing cluster,
  absolutely positioned at `right-1` so a 20px badge lands on the `…` glyph's
  centre. At its widest (`99+`, 27px) it reaches 31px in, and the name ends at
  36px, so it clears in every state.
- The translated mention text moved to `MentionAnnouncement`, an `sr-only` span
  still inside the `<Link>`, so a screen reader hears the mention with the room
  name. Losing that was the one real risk in moving the badge out of the link.
- Badge, bell and `…` all gained `motion-safe:transition-opacity
  motion-safe:duration-150`, so the swap reads as one move and stills for a
  reader who asked for that.
- Touch has no hover, so the badge stays in flow beside the menu there, as
  pin/mute does (`[@media(hover:none)]:w-14`). Reorder mode does the same,
  because its handle never fades.

## Trade-off taken

Hovering a room hides its mention count — the number goes quiet exactly while
you are deciding whether to open it. The bold name and the reader's opt-in
unread count both stay, on the one row the cursor is on. Discord's DM list makes
the same call.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — clean
- `pnpm --filter web test src/components/chat` — 36 files, 388 tests pass

Four new tests cover the crossfade tokens, the badge living outside the link,
the announcement staying inside it, and the reorder case. The spacer test for
badged rows now asserts the hole is held open in every state — it fails if the
badge ever rests at `size-0`, which would run the name under it.

## Screenshots

Rest and hover, badged and plain rows, are easiest to check by hovering the
sidebar on a room with unread mentions.


mrosberghaus:
## Grok review

Range `0e6cec0a93..d020a09917` (2 files). Requirements: badge/menu crossfade, collapsed vs expanded, keyboard access to the menu, no jump.

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/chat-room-sidebar-row.tsx` (`MentionAnnouncement`): outer span is already `sr-only`; the inner `sr-only` span is leftover from the old visible pill.
- Minor: same file, `99+` geometry comments disagree (`27px` vs `24px`; “reaches 31px”). Badge `right-1` is relative to the cluster that is already `right-1`, so a wide badge lands ~35px in. Name still ends at 36px (`px-3` + `md:gap-2` + `size-4`).
- Minor: `chat-room-sidebar-row.test.tsx` menu fade-in test asserts hover opacity, not `group-focus-within` (the badge test does). Keyboard path is in the source: `group-focus-within/room-row` on both layers.

Skipped / disagreed: none.

### Assessment

Crossfade matches the muted bell (absolute, `pointer-events-none`, hover/focus-within/open). Announcement stays inside the link; visual badge is `aria-hidden` in the trailing cluster. Collapsed rail still hides both numbers. Touch and reorder keep the badge in flow with a `w-14` hole. Keyboard: focusing the row reveals the `…` via `focus-within`, so the menu stays a tab stop. No name jump on a badged row — spacer holds `size-4` in every hover-capable state.

mrosberghaus:
All three fixed in 49447a848.

**Badge geometry** — this was worse than the note said, so I fixed the mechanism
rather than the comment. `right-1` resolves against the trailing cluster, which
is itself `right-1`, so a `99+` pill reached 35px in against a name ending at
36px. And no fixed edge offset can centre every pill, because its width *is* the
reader's unread count: `right-1` centres `2` and leaves `99+` 3.5px off.

The pill now rides the same `size-8 md:size-7` box the muted bell uses, pinned at
`right-0`. The box spans 4–32px in whatever is inside it, so the pill sits on the
`…` glyph's centre at every width and the widest clears the name by 4px. That also
makes the badge reserve the bell's width on touch, so the two share one hole rule
(`w-16`) instead of the badge carrying its own `w-14`.

**Nested `sr-only`** — correct, leftover from when the outer span was the visible
pill. One span now. The collapsed-rail test asserted the rule on `parentElement`,
which only held because of that nesting, so it now asserts on the announcement
itself; the unread count still asserts on its parent, where its own wrapper
carries it.

**Menu fade test** — pinned. It now asserts `group-focus-within` and
`data-[state=open]` alongside the hover path, matching what the badge test
already did. The source path was right; the test just didn't cover it.

Two assertions moved as a consequence of the new box: `aria-hidden` sits on the
badge's box rather than the pill, so `getByText("2")`/`getByText("99+")` now
assert via `closest("[aria-hidden]")`. The text is still hidden from assistive
technology by inheritance.

`pnpm check`, `pnpm --filter web typecheck`, and 391 chat tests all green.

PR #4783, opened 2026-09-18
feat(web): turn a pasted Message link into the pending quote

mrosberghaus:
## What changes for users

Pasting a Message link into the room or thread composer turns it into a quote. The pasted URL is replaced by the usual quote chip right away, and the message can be sent like that, with no text of its own. The sent message carries the normal quote card, which opens the original message. Removing the chip puts the link back as plain text.

The link becomes a quote only when the sender can read the message and every user member of the room can also read the source room, so a snippet never reaches someone who cannot follow the link. Anything else stays a plain link, with no prompt and no error. A link pasted while another quote is pending also stays a link.

Linear: [SOK-1106](https://linear.app/masumi/issue/SOK-1106/paste-a-message-link-in-the-composer-to-quote-that-message) (spec is the first comment). Apple follow-up: SOK-1109.

## How

- **Core, cross-room quotes:** `POST /v1/chats/rooms/{id}/messages` takes an optional `quote.roomId`. A user sender may quote from another room they read when the target room's user members are all members of the source room (the sender's Self Direct passes this rule). Every refusal is the same 400 "Quoted message not found". Coworker and Soko Bot senders stay same-room. No migration; the quote still lives in message metadata.
- **Core, quote-only messages:** `content` may be empty when `quote` is set. Without a quote an empty body is still refused. Coworker and Soko Bot prompts now carry the quoted message above the sender's words, which is all they get when the body is empty.
- **Shared:** `canQuoteIntoRoom` in `@sokosumi/utils` is the one audience rule Core and web both use. Core is the authority; web only decides whether to convert the paste.
- **Web:** `parseChatRoomMessageLink` is the inverse of `chatRoomMessageHref`. The session composer resolves a pasted link and swaps it for the pending quote. The link is removed in the editor itself (new `removeLastText` on the editor handle), because a focused editor ignores outside value changes. The source room id travels through the optimistic message, the outbound queue, the server action and the service. `onRestorePendingQuote` is renamed `onSetPendingQuote`. A pinned quote-only message shows what was quoted instead of a blank body.
- **Generated client:** regenerated from the schema changes, not hand-edited.
- **Docs:** `CONTEXT.md` gains a Quote entry.

## Limits

1. In a coworker 1:1 (the stream send path) a quote still needs text with it, and only same-room links convert. That Core route requires words and takes same-room quotes only.
2. Paste is detected around the composer rather than inside the editor, so a clipboard holding only HTML never converts. Copy link puts plain text on the clipboard, so that path works.
3. A notification for a quote-only message has no preview text; it shows the author and room line, the same as for a message with no previewable body.
4. Quote-only messages cannot be found by message search, which matches on the body.
5. The Apple composer cannot send a quote with an empty body yet (tracked with SOK-1109).

## Verification

- Core suite: 5848 passed. Web suite: 6493 passed, locale parity ok.
- New tests: Core send-message route (cross-room and empty-body cases, mocked Prisma), audience rule, link parser, link-to-quote resolver, AI prompt text, Soko Bot dispatch with a quote-only message, session composer paste flow.
- Offer-row version was tried on the preview; the convert-on-paste version has not been checked in a browser yet. No screenshots.


mrosberghaus:
## Grok review

Reviewer ran against `origin/main` merge-base `efcceedc5` … `ed2309cba`. Requirements taken from the PR title and body (convert-on-paste; Linear still describes an accept/decline offer row). Skills `requesting-code-review` and `receiving-code-review` were present in this worktree.

### Fixed
- Important: quote-only room messages were invisible in coworker/Soko Bot CONTEXT (`ed2309cba` — `apps/core/src/services/chat-room-mention-context.ts`). The mention prompt already used `roomMessagePromptText`; surrounding history still passed the empty body. CONTEXT now uses the same helper, so a later @mention still sees the snippet.

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/room-session-composer.tsx` — `quotedLink` is dropped on submit, so a restored quote after `onSend` `{ ok: false }` no longer puts the URL back on dismiss. Classic send is optimistic (`ok: true`) and does not hit this path.
- Minor: `packages/utils/src/chat-room-quote-audience.ts` — `canQuoteIntoRoom([], source)` is `true` because `[].every` is true. Core write access always includes the sender, so this is belt-and-suspenders only.
- Minor: comments and the PR title still say “offer”; the shipped composer converts on paste (matches the PR body and `CONTEXT.md`).
- Minor: no dedicated Soko Bot cross-room 400 test; coworker and Soko Bot share the same refusal branch.
- Skipped / disagreed: Linear stories 2–3 / 27–28 (offer row, keyboard/SR on the offer) — superseded by convert-on-paste in the PR body. Send-to-yourself still uses its own endpoint; it already shares `buildRoomQuoteSnapshot`, which is the snapshot that matters.

### Assessment

Cross-room quotes are membership-gated on Core (uniform 400, no existence leak), web only converts when the source room is already in the sender’s catalog, and quote markdown still goes through `sanitizeMarkdown`. Remaining notes are polish, not merge blockers.

mrosberghaus:
Reply to the [Grok review](https://github.com/masumi-network/sokosumi/pull/4783#issuecomment-5729489486). Pulled `ed2309cba` (quote-only messages in CONTEXT); the remaining notes are handled in `20bae933a`.

- **`quotedLink` dropped on submit:** fixed. The pasted link now travels in the send snapshot and is restored with the quote, so removing a quote that came back after a failed send gives the URL back. The failed-send composer test now removes the restored chip and checks for the link; it failed before the change.
- **Soko Bot cross-room 400:** added a route test next to the coworker one. Same branch, but now each actor has its own case.
- **"offer" wording:** the PR title is now the convert-on-paste commit subject, and the one stale comment (`message-link-quote.ts`) is reworded. No other comment in the diff still says offer.
- **`canQuoteIntoRoom([], source)` is `true`:** left as is. The rule answers "can every reader of the target follow the link", and with no readers nobody receives the snippet, so `true` is the right answer rather than an accident of `every`. Core never reaches it with an empty target, because the write-access check requires the sender to be a member. On web a wrong `true` only means the paste converts and Core then refuses.
- **Skipped items:** agreed. The offer-row stories were replaced by convert-on-paste on the author's decision, and Send to yourself already shares `buildRoomQuoteSnapshot`.

Checks after the change: Core send-message route 48 passed, web session composer 16 passed, web typecheck clean.

PR #4785, opened 2026-09-18
feat(apple): add a reorderable Pinned sidebar section

mrosberghaus:
## What

Apple chat parity row **26a**: a reorderable **Pinned** sidebar section, matching web #4764.

- Every pinned room of any kind (channel, private channel, Direct, external) lists in a **Pinned** section above Channels, in every workspace, hidden when nothing is pinned. A pinned room leaves the section it would otherwise sit in and keeps that section's leading mark (lock/number, globe, Direct avatars).
- Order is the reader's own: oldest `starredAt` first, id tie-break. Activity, mute rank and public/private no longer move a pin; the activity comparator lost its pinned branch.
- The per-row pin glyph is gone. Unpin stays in the context menu; pin/mute exclusivity is unchanged.
- With two or more pins and the section open, a toggle on the heading enters a reorder mode: Pinned rows stop navigating, the context menu is not offered, rows move by native `List` drag (`onMove`) or a handle menu with Move up / Move down. Closing the section, dropping below two pins or switching workspace ends the mode, and it does not come back by itself.
- Reorder is optimistic, the latest reorder wins, a list read during the request keeps the order, and a failure of the latest request shows the existing sidebar alert and reloads the room list from Core. 401 signs out.
- Closed section headings now show their rooms' attention (bold, plus an accessibility value). Apple had none for any section; the headings share one view, so all sections get it.
- Selects the deployed `PUT /chats/rooms/starred` into the Apple Core snapshot via `scripts/update-core-api.py` (additive: the path, `ReorderStarredChatRoomsRequest`, `StarredChatRoomOrder`, and Core's new description on `ChatRoom.starredAt`).

## Web sources

`partition-rooms-for-sidebar.ts`, `chat-room-activity-sort.ts`, `organization-chat-list.client.tsx`, `chat-room-sidebar-row.tsx`, `pinned-rooms-dnd.tsx`, `organization-chat-list.actions.ts`, `use-organization-chat-rooms.ts`, `room-attention.ts`, and Core `routes/v1/chats/rooms/starred/put.ts`.

## Native differences / not included

- No bare Up/Down key reorder: arrow keys keep moving the `List` selection; keyboard and VoiceOver users use the handle menu.
- On success Apple applies Core's returned sort keys (web ignores the body; same order).
- No changes to `apps/web`, `apps/core`, Core contracts, dependencies, pinned tools or workflows.

## How to test

Pin a channel, a private channel, a Direct and an external room; each moves to Pinned in pin order without a glyph. Post in a lower pin: it does not move. Collapse Pinned with an unread room inside: the heading turns bold. Click the heading's arrows, drag rows, use a handle's Move up / Move down, confirm a click on a Pinned row does not open it and right-click offers nothing, then click the checkmark. Unpin down to one room while in the mode: the mode ends. Reload web: same order. Go offline and reorder: alert, then Core's order after the reload.

- `swift test` — SokosumiChat 539, SokosumiWorkspace 116, SokosumiAuth 35, SokosumiRealtime 31, CoreAPI 1, all passed (new partition/sort, section attention, paused-PUT optimistic / latest-wins / failure-reload / stale-context / unpinned-meanwhile / mode tests).
- `mint run swiftformat --lint --cache ignore .` — 0/369; `mint run swiftlint lint --strict --no-cache` — 0 violations.
- macOS `xcodebuild … build` — succeeded, no warnings in hand-written sources; `… test -only-testing:SokosumiTests -enableCodeCoverage NO` — 96 tests, 0 failures.
- `swift build --package-path Packages/SokosumiWorkspace --triple arm64-apple-ios17.0 --sdk "$(xcrun --sdk iphoneos --show-sdk-path)"` — complete.
- Light/dark fixtures of the real sidebar inspected: `docs/screenshots/sidebar-pinned-{open,reorder,closed}-{light,dark}.png`.

Unverified: the app was not launched and no live reorder was sent — real pointer drag through `onMove`, the handle menu by click/Full Keyboard Access/VoiceOver, the alert after a real failure, and Core's lenient semantics against production. Details in `apps/apple/PARITY.md`, slice 26a.


mrosberghaus:
## Grok review

Range: `354a978ec` (`origin/main`) .. `15fafe5aa` (`claude/apple-pinned-section`). Requirements from the PR title/body (parity row 26a / web #4764).

### Fixed

### Remaining
- Minor: `ConversationSidebarView.swift` — reorder mode still attaches `.contextMenu { if pinned == nil { … } }`. An empty builder can still install the gesture; omit the modifier when `pinned != nil` if a right-click should be a true no-op.
- Minor: `apps/apple/PARITY.md` checkpoint still says draft #4785; the PR is open. The 2026-09-15 delta row still leads with “Apple still lists pins inline” and only appends the implemented note.
- Minor: `WorkspaceStateTests.pinnedReorderShowsAtOnceAndAFailureReloadsFromCore` 401 case does not assert sign-out (only “no alert, no reload”).
- Skipped / disagreed: guest Direct in Pinned uses `sidebarRoomKind` (globe / External) vs web `room.kind === "direct"` (avatars). Row 26a asked for the section the room left; the partition test pins that. Not changing it.
- Unverified (already in the PR): live `onMove` drag (optimistic apply is hopped off the List update, so a snap-back is possible), handle menu via click/FKA/VoiceOver, real failure alert, Core’s lenient PUT against production.

### Assessment

Pinned section, `starredAt`-only order, org-slug PUT, latest-wins reorder, 401 sign-out, and workspace-switch invalidation match the spec; leftover notes are docs/gesture polish and the live-drag check the PR already flagged.

mrosberghaus:
Response to the Grok review.

**Delta-table entry led with the old state** — fixed in 524a76a: the #4764 entry now leads with what slice 26a closed and keeps the previous behavior as history. The checkpoint's "draft #4785" is accurate; the PR is still a draft.

**`.contextMenu` with an empty builder in reorder mode** — leaving as is. SwiftUI documents empty menu content as "no context menu", which is the behavior relied on here. Making the modifier itself conditional would put the row behind an `if`/`else`, changing the row's structural identity exactly when the mode toggles and rows are being moved — the wrong trade for a `List` that is reordering. A real right-click in the mode is part of the live pass already listed as unverified.

**401 case does not assert sign-out** — leaving as is. `ephemeralState`'s `AuthState` is built with `restoreSession: false` and an empty token store, so it is never signed in and there is no state change to assert at the workspace level. The test covers what is observable there (no alert, no reload); the `401 → .unauthorized` mapping and `signOutIfUnauthorized` are covered where the other sidebar actions cover them.

Agreed with the skip on guest Directs in Pinned: the row shows the section the room left, and the partition test pins that.

Live `onMove` drag, the handle menu via click/FKA/VoiceOver, a real failure alert and Core's lenient PUT against production stay listed as unverified.

PR #4787, opened 2026-09-18
feat(web): let the OS banner through while Sokosumi is focused

schaier-io:
Closes SOK-957.

## What a reader gets

A reader with Sokosumi open in another window heard nothing when a notification arrived. The push service worker skips its own banner while a Sokosumi page is focused, and on the web the banner is the only thing that makes a sound.

This adds one switch in Account → Notifications: **let the banner through while Sokosumi is open**. It is off by default, so a reader who never touches it sees nothing change. When it is on, the worker stops skipping, the banner appears, and the device plays its own notification sound.

The reader does not choose that sound, and cannot. The Notification API has no `sound` option, a service worker has no audio API, and nothing on the web exposes the operating system's own notification sound. That is why the original ticket's custom-sound feature was dropped.

The sound still follows the delivery grid. A category with its push channel off sends no push at all, so a silenced group stays silent.

## Second fix, same area

Repeat messages in one chat room share a banner tag and replaced each other silently. Those replacements now carry `renotify`.

The one replacement that must stay silent still does. A push catching up with a banner the page already drew for the same notification is told apart by notification id rather than by timing, because the two transports have no fixed order. `renotify` is Chromium-only; other engines ignore the key.

## Two races left open, on purpose

Both are documented in the code that owns them, rather than claimed away:

1. Turning the switch **on** is not heard in a second Sokosumi window until that window's session refreshes, because `authClient.updateUser` refreshes only the tab that wrote. Turning it **off** is immediate in every window. Closing the other direction needs a channel between tabs that nothing else here needs.
2. A push arriving late for a notification the page already replaced re-alerts with older content. The tag collapse did that replacement before this change; what is new is the sound on it. Ordering by `createdAt` would miss every banner drawn before that field existed.

## Schema

`packages/database/prisma/migrations/20260918090000_user_banner_while_focused` adds `user.bannerWhileFocused BOOLEAN NOT NULL DEFAULT false`. Additive with a safe default, so it is backward compatible with the currently deployed Core. No data script.

## Verification

- `pnpm test` — web `794 passed (794)` files, `6504 passed | 1 skipped`; core `530 passed | 6 skipped (536)`, `5836 passed | 37 skipped`; all packages green.
- `pnpm typecheck` — 20/20 tasks.
- `pnpm check` — 4205 files, no fixes applied.

One Core file, `src/routes/sync/index.test.ts`, timed out once under machine load during a full run and passes alone (`42 passed`) and on a clean re-run of the whole Core suite. It is not in this diff.

New behaviour is pinned by tests that were mutation-checked, not just observed green: removing `aria-describedby`, swapping the two toast keys, and flipping the session fallback each fail a named test.

## Review

Six rounds of a standards and spec review loop. The last full pass reported nothing on either axis. Findings fixed along the way: duplicated switch markup extracted into one shared row, a missing `aria-describedby`, an untested session-loading branch, an untested toast wording, a Core test that asserted nothing its name promised, a Prisma fixture in an unrelated Core route test that made `GET /me/preferences` return 500, and four comments that claimed more than the code did.

## Out of scope, reported not fixed

`apps/core/src/routes/v1/users/[id]/preferences/preferences.test.ts:239` — the pre-existing `showRoomUnreadCount` test has the same weakness as the one fixed here: it cannot see a notification cell being written. Untouched, since it is not part of this change.

`apps/web/public/ably-push-sw.js` is 723 of a 750-line ceiling. The next change to it will need a split.


schaier-io:
Superseded by #4789. The `bannerWhileFocused` preference in this PR was cut on review: the banner stays suppressed while a Sokosumi tab is focused, with no user choice, and SOK-957 is cancelled with the reasoning recorded on the ticket. The re-alert fix was always independent of it and continues in #4789 under SOK-1111.

PR #4790, opened 2026-09-18
fix(web): hold sidebar rows in place when the sidebar toggles

mrosberghaus:
Toggling the app sidebar between the expanded panel and the collapsed rail redrew the list instead of narrowing it. Rows were 40px expanded and 32px on the rail, Soko Bots was a 48px card and a 32px square, and the separator after New Task existed only when expanded — by the first Chat room the list had already moved about 97px. Marks slid sideways too (a room's mark 2px off the rail's centre, a section's chevron 2px, the account face 4px), and labels started on two columns, 44px for nav and 48px for room names.

Toggle the sidebar and an ordinary row stays where it was, under your cursor. Only the panel's width changes.

One rule now — a **Sidebar row**, owned by the row primitive:

- **One height in both states**: 32px with a pointer, 44px on the phone sheet, which is never the rail.
- **One 24px leading slot** (`SidebarRowSlot`), centred 28px from the sidebar's left edge in both states.
- **One label column** at 48px — nav labels, section titles, room names and a section's empty line alike.

Soko Bots becomes an ordinary row: one 24px face (the first workspace bot, the bot icon when there is none), no border, no rail hover exception. The three-face stack is gone — it read as coloured dots collapsed and broke the label column expanded.

Archived gains a rail square, so Direct Messages no longer jumps 40px on a toggle. Its rows stay off the rail (they are static, and Restore and Delete need a menu no 32px square has room for), so pressing it expands the sidebar and opens the section — the way a pending invitation's tile expands to its Accept and Decline.

Three items still differ between states, by design: a guest room's host organisation line, a pending invitation's card, and a section's empty text. Their marks still use the slot and the column.

## Measured in a browser

A throwaway page (not committed) rendered the sidebar with fixture data; a headless browser recorded every row's top edge and mark centre in both states.

| row | expanded top | collapsed top | delta |
| --- | --- | --- | --- |
| Soko Bots … History | 72–361 | same | 0 |
| announcements / general | 442 / 514 | same | 0 |
| invitation card | 586 | 586 | 0 |
| partners (guest row) | 663 | 618 | −45 |
| "No direct messages yet" | 786 | 730 | −56 |

Marks: **28px centre, 24px wide** in both states, account face with them. Label column: **48px** for nav labels, section titles, room names and the empty line.

Only the two exceptions present in the fixture move anything below them (invitation card −45, guest row −11).

## Decisions worth a reviewer's eye

- **44px on touch, not the spec's 40px.** #4775 landed hours before SOK-1107 and gave mobile chat rows 44px targets; unifying on 40px would have reverted it. The slot is still 24px everywhere, so phone marks go 28→24 while the row stays 44.
- **The axis is 28px, pinned with a 4px offset rather than `mx-auto`.** The rail's 1px right border leaves its own centre on 27.5px; the expanded panel's has to be a whole pixel, so the rail square gives up half a pixel of its own symmetry.

## Verification

- `pnpm check`, `pnpm --filter web typecheck`, `pnpm --filter web test` — 792 test files pass.
- New: a row-geometry regression test at the primitive, a contract test asserting every row type uses the shared slot, and tests for Archived's rail square and the separator on the rail.
- `CONTEXT.md` gains a **Sidebar row** entry and names what is not one.

Closes SOK-1107


mrosberghaus:
## Grok review

Range reviewed: `d1a041620` (merge-base with `origin/main`) … `d55371d11` (`sok-1107-sidebar-rows-keep-their-place`). 29 files, web sidebar only.

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/chat-sidebar-section-header.test.tsx` — test still named “stays off the rail without an icon, as Archived does”. Archived now has a rail square; the no-icon path is real, the Archived analogy is not.
- Minor: geometry tests assert Tailwind class tokens (`h-11`, `md:h-8`, `min-w-6`, `ml-1`), not layout. A padding or `SidebarGroup` change would still pass. The PR’s browser table is the real measurement; it is not in the repo.
- Minor: `sidebar-row-slot-contract.test.tsx` counts `> 20` slots and does not pin that Archived’s rail square is one of them (the section-visibility test covers that separately).
- Skipped: an *open* Archived section still hides its rows on the rail, so Direct Messages moves by those rows’ height. That matches CONTEXT.md (rows stay off the rail; Restore/Delete need the expanded menu).
- Skipped: the 64px header logo is not on the 28px axis (expanded centre ~24px, rail ~27.5px). CONTEXT.md excludes the header from Sidebar row.

### Assessment

One height (`h-11 md:h-8`), one 24px `SidebarRowSlot` on a 28px axis (`px-2` expanded, `ml-1` + `w-8` on the rail), one 48px label column (`gap-2` / `pl-10` empty line). Soko Bots is an ordinary row; the New Task separator stays on the rail; Archived’s square expands rather than toggling over nothing; a group Direct grows the slot rightward off a fixed left edge so the first face does not slide. Mobile keeps 44px (`h-11`; the Sheet is never the rail). `useSidebar()` in `OrganizationChatList` is safe: `(app)/layout.tsx` already wraps `SidebarProvider`, and the rail square is `display: none` on the mobile `/chat` list.

No color-token, generated-file, or whole-pixel layout literals in the new geometry (the 4px `ml-1` offset is the whole-pixel answer to the rail’s 27.5px centre).

mrosberghaus:
Addressed in `73e0fc4ab`.

**Minor 1 — stale test name.** It was worse than stale: with Archived carrying a square, all five call sites pass a `railIcon`, so the optional path was dead code, not just a bad analogy. `railIcon` is required now, the branch and its test are gone, and the type says what `CONTEXT.md` says — a section without a square lets everything under it jump on a toggle.

**Minor 3 — contract test didn't pin Archived's square.** Fixed, and the check found my own error on the way: I first asserted four squares and it renders five (Pinned, Channels, External, Archived, Direct Messages). It now names `archivedChannels` and asserts the exact count rather than a `> 20` floor that a missing section wouldn't dent.

**Minor 2 — class tokens, not layout.** Correct, and I'm leaving it. There's no e2e harness in the repo, and the web suite runs on happy-dom, which has no layout engine — `getBoundingClientRect()` returns zeros, so a layout assertion can't be written with what's here. Introducing Playwright for one geometry spec is more than this PR should carry. The measurement is real but out-of-repo: the numbers in the description came from a headless browser against a throwaway page, and each commit body records what it measured. If you want it guarded in CI, that's its own ticket and I'd rather it cover the whole sidebar than this change alone.

**Both Skipped items** — agreed on both, no action. An open Archived section moving Direct Messages by its rows' height follows from rows staying off the rail, and the header is excluded from Sidebar row by `CONTEXT.md`.

PR #4793, opened 2026-09-18
fix(web): keep the mobile keyboard open after sending a chat message

mrosberghaus:
## User-facing impact

On iOS, tapping Send in the chat composer blurred the editor a beat later and the software keyboard slid away, so every message cost a second tap to keep typing. The keyboard now stays open and the caret stays in the composer, ready for the next message.

## Root cause

The send button already guarded the tap with `preventDefault` on `pointerdown`, added earlier for exactly this bug. **That guard does nothing on iOS.**

Measured on iOS 27 WebKit (iOS Simulator, standalone login-free page instrumented with `visualViewport` and focus/blur logging):

| Setup | Result |
| --- | --- |
| Button whose pointerdown handler *only* calls `preventDefault()` | editor fires `blur`, viewport 484 → 714 (keyboard closes) |
| `innerHTML = ""` on the focused editor from a timer, after real typing | no blur, viewport stays 384 |
| Button whose **touchstart** handler calls `preventDefault()` | no blur, viewport stays 384 |
| touchstart guard **+** the composer's real `innerHTML` clear | no blur, keyboard stays up, caret kept in the emptied editor |

WebKit moves focus on the **touch**, not on the compatibility mouse event, so `pointerdown` is too late. The composer's own clear of the editor is innocent.

## The change

- `preventDefault()` on `touchstart` for the send button. React registers `onTouchStart` passively at the root, where `preventDefault` is a silent no-op, so this is a native `{ passive: false }` listener.
- The button moves into a `RoomComposerSendButton` that mounts only when the composer has no `submitControl` of its own. That satisfies `apps/web/.cursor/rules/effects.mdc` (conditional mounting over guarding inside an effect, `useMountEffect` over ad-hoc `useEffect(fn, [])`) and means the listener's lifetime matches the button's rather than the composer's — with the listener on the parent, a `submitControl` swap would have silently shipped with no guard.
- The `pointerdown` guard stays; it is what holds focus on desktop.

Three tests in `room-message-composer-send.test.tsx`: the guard itself, the guard surviving a `submitControl` swap, and one tap still submitting exactly once now that the synthesized `click` is suppressed on touch.

## Verification

- `pnpm check` — clean
- `turbo typecheck` — clean
- `pnpm --filter web test` — 6,465 passed, 1 skipped
- Mechanism verified on iOS 27 WebKit as tabulated above; each new test confirmed red without the fix

> [!IMPORTANT]
> **Device verification inside Sokosumi is still pending.** The mechanism was proven on a standalone page using the identical elements and event path, but the worktree DB rejects the vault account, so the real Send button was never tapped in the app. Please confirm on the Vercel preview from a phone or simulator before this is marked ready for review.

## Known gap, out of scope

Tapping Send while it is *disabled* still blurs the editor: `Button` carries `disabled:pointer-events-none`, so the touch falls through to unguarded composer chrome. Pre-existing and a different interaction from the one reported here.

The other composer toolbar buttons (format, emoji, attach, mention) rely on the same ineffective `onMouseDown` guard or none at all, so tapping them mid-message closes the keyboard too. Tracked separately.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/room-message-composer.tsx` — `submitControl` has no production caller. `RoomComposerSendButton` plus `sendBlockedRef` / `onPrepareSubmitRef` exist for a Stop/Send swap that never happens. The mount listener only `preventDefault`s, so those refs are unused by it.
- Minor: in-app iOS Send is still unverified (already in the PR body). The new tests prove `preventDefault` on `touchstart` and single submit in jsdom; they cannot prove the OSK stays up.
- Skipped: submit from `touchstart` so Send does not depend on `pointerdown` after cancel — would double-post unless `pointerdown` shared the 400ms guard. Safari fires `pointerdown` before `touchstart`; the current split matches that. Don't change without a device repro.
- Skipped: toolbar / disabled-Send blur — already called out in the PR; different interaction.

### Assessment
Room and thread both send through `RoomMessageComposer`; desktop pointer/click is unchanged; native `{passive:false}` `touchstart` is the right React 19 workaround. Confirm one Send tap on the Vercel preview from a phone or simulator before marking ready.

mrosberghaus:
Addressed in f377d76.

**Dead refs — correct, and they were mine.** `sendBlockedRef` / `onPrepareSubmitRef` are gone. The comment justifying them ("the listener below is bound once on mount, so it must not close over a stale render's values") was simply wrong: the mount listener only calls `preventDefault()` and never reaches `requestComposerSubmit`. Its two callers are React handlers that close over fresh props each render. The tell was in the same file — `handleSendPointerDown` read `sendBlocked` directly while `requestComposerSubmit` read the ref. `requestComposerSubmit` now reads the props.

**`submitControl` has no production caller — confirmed, but not removing it here.** `grep` finds only the prop declaration, the render site, and my test. I'm leaving it because it is pre-existing API surface, not something this PR introduced, and deleting it properly means the prop, its doc comment, and the branch at the render site — a refactor of the composer's interface that doesn't belong in a mobile keyboard fix. Happy to do it as a follow-up; say the word.

I'm keeping `RoomComposerSendButton` regardless. While `submitControl ?? (…)` exists, the button is conditionally rendered, and `apps/web/.cursor/rules/effects.mdc` calls for conditional mounting over guarding inside an effect, with `useMountEffect` in the child. Putting the listener on the parent is what produced the latent bug two reviewers flagged: a `submitControl` swap would have shipped with no guard.

**Skipped items — agreed.** Submitting from `touchstart` would need `pointerdown` to share the 400ms guard or it double-posts, and Safari fires `pointerdown` first, so the current split matches the platform. Not changing it without a device repro. Toolbar and disabled-Send blur are separate interactions and stay out of this PR.

**In-app iOS verification is still open.** Still the one gap. `main` is now merged in, `pnpm check`, turbo `typecheck` and 6,527 web tests are green on the merged base. I'll confirm one Send tap on the Vercel preview from a simulator before this moves out of draft.

PR #4794, opened 2026-09-18
fix(web): keep composer code block text and toggle the block off

mrosberghaus:
## User-facing impact

Three bugs in the chat composer's code block button, reported from a screen recording. All three came from one cause: `codeBlock` was the only toolbar format with no toggle helper — it blind-inserted `<pre><code></code></pre>` via `execCommand("insertHTML")`, and Chrome leaves the caret on the `pre` rather than inside the `code`.

| Bug | Cause |
| --- | --- |
| The composer drew the block as a grey sliver | The empty `code` had no line, so the `pre` collapsed to just its padding (16px) |
| The sent message showed an empty block in the room | Typing landed as a `pre` child *beside* the empty `code`. `htmlToMarkdown` read the empty `code`, emitted an empty fence, and **silently discarded the author's text** |
| Clicking the button again did not remove the block | No unwrap branch existed, unlike `code` and `quote` |

The second one is content loss: text people typed never reached the room.

## Changes

- **New `apps/web/src/lib/utils/composer-wysiwyg-code-block.ts`** — the missing sibling of the existing `composer-wysiwyg-blockquote.ts` and `composer-wysiwyg-code-format.ts` modules. Builds the block by hand and owns the caret, so text always stays inside `code`; unwraps when the caret is already in a block, preserving newlines.
- **`composer-wysiwyg-editor.tsx`** — the `codeBlock` case now routes through the toggle and calls `handleInput()` / `publishActiveFormats()`, matching `code` and `quote`.
- **`composer-markdown-dom.ts`** — `getCodeContent` now reads the whole `pre` rather than its `code` child, so text contentEditable parks elsewhere can no longer be dropped. It also reads `br` and `div`-split lines back as the newlines they draw (previously it leaned on `innerText`, which is undefined under happy-dom), and trims the trailing contentEditable break.

## Verification

- `pnpm --filter web test` — **795 files, 6496 passed**, 1 skipped
- `pnpm --filter web typecheck` and `pnpm check` clean; full `pnpm precommit` gate green
- 12 new tests: 7 in `composer-wysiwyg-code-block.test.ts`, 4 serializer regressions in `composer-markdown-dom.test.ts` (including the exact `<pre>what about now.<code></code></pre>` DOM Chrome produces), 1 div-split line case

Behavior confirmed in **real Chrome** by bundling the shipped module against a contentEditable harness, since happy-dom does not model contentEditable caret placement:

```
open:  <pre><code></code></pre>  16.0px  →  <pre><code><br></code></pre>  32.5px
type:  <pre>what about now.<code></code></pre>  →  <pre><code>what about now.</code></pre>
click: second block inserted  →  unwrapped to plain text, newlines kept
```

Multi-line entry and wrapping a selection were verified the same way.

## Left out of scope

- **No keyboard exit from a code block.** Enter inserts a newline (correct) and `composer-wysiwyg-arrow-exit.ts:56` explicitly refuses to arrow out of `code` inside `pre`, so the toolbar toggle is now the only way out. Blockquote already has `tryExitComposerBlockquoteOnEmptyLine`; a matching Enter-on-empty-last-line escape would be a reasonable follow-up.
- The `br` → newline walk now exists in three places (the new module, the serializer, and `rangeToPrefixedText` in the blockquote module). Worth extracting, but no bug drives it and it would touch quote behavior.


mrosberghaus:
## Grok review

### Fixed
- Important: empty unwrap placed the caret at the start of the editor, because the last node is a `br` and its parent is the whole editor (`923111819` — `apps/web/src/lib/utils/composer-wysiwyg-code-block.ts`)

### Remaining
- Minor: `toggleComposerInlineCode` still `insertHTML`s a `<pre><code>` for a multiline selection, so the caret can sit on the `pre`. The serializer now reads the whole `pre`, so typed text is no longer dropped.
- Minor: unwrap's `readCodeBlockText` only treats `<br>` as a newline; `getCodeContent` also handles `div`/`p`. Composer Enter goes through `insertLineBreak`, so typed blocks unwrap correctly. Pasted div-split lines could concatenate on toggle-off.
- Minor: `getCodeContent` always strips one trailing newline. That drops CE's caret `<br>` (correct) and also an intentional last blank line from markdown-sourced fences on round-trip.
- Skipped: extracting the `br` → newline walk (three copies) — no bug drives it, and it would touch quote behavior.
- Skipped: Enter-to-exit a code block — called out of scope in the PR; toolbar toggle is the way out.

### Assessment
The three reported bugs (collapsed empty block, dropped typed text, no unwrap) are fixed and covered; leftover notes are edge-path polish, not send-path content loss.

mrosberghaus:
Two of the three remaining notes are fixed in `fc60de77e`. The third I'd rather leave for a follow-up — reasoning below.

### Fixed: trailing newline strip dropped an intentional blank last line

Confirmed and worse than cosmetic — it was silent content mutation on edit. `markdownToHtml("```\nfoo\n\n```")` produces `<pre><code>foo\n</code></pre>`, with the newline in the **text node**, and the unconditional strip round-tripped it down to `` ```\nfoo\n``` ``.

The trailing newline is now stripped only when it came from **structure** (a `br` or a block close), which is what contentEditable adds as scaffolding. A newline that came from text is the author's own blank line and survives. Covered by `keeps an author's blank last line in a code block` and a `markdownToHtml` → `htmlToMarkdown` round-trip test.

### Fixed: `readCodeBlockText` vs `getCodeContent` drift

Fixed by extracting the single reader both need, `readComposerCodeText` in `apps/web/src/lib/utils/composer-code-text.ts`, rather than teaching the `div`/`p` rule twice.

You're right that the practical exposure was low — I checked, and `toggleComposerCodeBlock` is only reachable from the chat composer, where nothing currently produces div-split lines inside a `pre` (`markdownToHtml` emits text newlines, Enter goes through `insertLineBreak` → `br`, and paste is flattened by `composerPastedHtmlToPlainText`). But the trailing-newline fix had to touch that walker anyway, and the drift is the kind that rewrites text on the way out, so consolidating was cheaper than keeping two. New test: `keeps the newlines of a div-split block when unwrapping`.

This also closes the "three copies" note for these two. `rangeToPrefixedText` in the blockquote module stays separate — it extracts a *range prefix* and does no block handling, so it isn't the same function.

### Not fixing here: `toggleComposerInlineCode`'s multiline `insertHTML`

Confirmed real — the inline `code` button still does `insertHtmlAtSelection('<pre><code>…</code></pre>')` for a multiline selection, so the caret can land on the `pre`.

Leaving it out because it's a **pre-existing path in a different module** that this PR strictly improved rather than regressed: before, that caret position meant the next keystroke was dropped by the serializer; now the text survives, so what's left is caret placement only. Folding it in means changing the inline-code button's semantics (delegating to the block toggle would also make it *unwrap* when the selection is already in a block), which is a behavior call that deserves its own diff and its own tests rather than riding along in a bugfix.

Happy to do it next if you'd rather it land here.

### Verification

`pnpm --filter web test` — **797 files, 6548 passed**, 1 skipped. Typecheck and `biome check` clean. Chrome re-verified against the rebuilt module: open/type, toggle-off, multiline, multiline toggle-off and wrap-selection all unchanged.

Also confirmed your `923111819` caret fix is right — `setStartAfter(last)` is correct where my `placeCaretAtStartOf(last.parentNode)` walked up to the editor for a `br`.

PR #4797, opened 2026-09-18
feat(apple): turn a pasted Message link into the pending quote

mrosberghaus:
Closes [SOK-1109](https://linear.app/masumi/issue/SOK-1109/paste-a-message-link-in-the-apple-composer-to-quote-that-message).

## What changes for the user

Pasting exactly one Message link on the configured web origin into the Mac composer now replaces the pasted URL with the usual pending quote chip, immediately and with no offer row. Removing the chip puts the link back at the caret. The sent message carries the normal quote card, including the cross-room source link.

This is parity with web after #4783 (SOK-1106); the issue's earlier accept/decline design was dropped to match what web shipped.

- The conversion happens only when the sender can read the message and the quote is allowed: same room, the sender's Self Direct, or every user member of the target room is also a member of the source room. Anything else stays plain text, with no conversion and no error.
- Core re-checks the audience at send time; the client check only decides whether the paste converts.
- One quote per message: a paste never replaces a quote chosen from the message menu.
- A quote may now be the whole message, so send is enabled with an empty body — except in the coworker 1:1 stream, which needs words to answer and takes same-room quotes only.
- The thread composer behaves like the room composer. A hand-typed link, or one inside longer pasted text, stays plain.

## How it works

`MacComposerTextInput.pasteText(from:)` stays the one plain-text paste entry point and now reports the inserted text with the editor's own removal, so the swap happens in the text view and the words and caret around the link stay as typed. `ChatComposerView` resolves it through `WorkspaceState.messageLinkQuote`, which parses with the existing `ChatLink`, reads the message with the existing `ChatService.getMessage`, and applies the audience rule against the rosters the app already holds. It commits only when the resolve wins its own paste generation, no quote was chosen meanwhile, the room is still open and the link is still in the draft.

The `CoreAPI` snapshot was refreshed with `scripts/update-core-api.py` from the current Core specification. The only contract delta is SOK-1106's: `quote.roomId`, and `content` no longer `minLength: 1`. No Core or web files change.

## Not covered

After a failed coworker-stream send, the restored draft brings its quote chip back but not the link text it replaced, so removing that chip inserts nothing. Apple restores that quote through `DirectStreamSession` rather than a composer snapshot, and web's own fix for this is still unmerged on the SOK-1106 branch. Recorded in `PARITY.md` and moved to Out of scope on the issue.

## Verification

- `xcodebuild test -only-testing:SokosumiTests` (macOS, Debug) — passed. Note: coverage must be off locally (`-enableCodeCoverage NO`); with it on, the tree-sitter C targets fail to link on `___llvm_profile_runtime`, unrelated to this change.
- `swift test` for all five packages — 548 / 118 / 35 / 31 / 1 passed.
- `mint run swiftformat --lint .` and `mint run swiftlint lint --strict` — clean.
- `pnpm check && pnpm typecheck` via the pre-commit hook — passed.

Not yet verified live: the paste, the chip and the restored link in a running app against real rooms.


mrosberghaus:
## Grok review

### Fixed
- Important (test gap): `removeLastText` / `insertAtCaret` had no coverage (`174b3a450` — `apps/apple/SokosumiTests/Chat/Composer/MacComposerTextInputTests.swift`)

### Remaining
- Minor: `WorkspaceStateTests.pastedMessageLinkQuotesAcrossRoomsAndSendsWithoutABody` uses `roomsBody` with empty `userMembers`, so `canQuoteIntoRoom` is vacuously true. Audience is covered in `MessageLinkQuoteTests`.
- Minor: `pasteAsPlainText` / `pasteAsRichText` still skip `pasteText(from:)`, so Paste and Match Style will not convert a Message link.
- Skipped / disagreed: leftover composer newline adding a leading space on chip dismiss — after removing a sole pasted URL, `captureDraft()` is empty, so `draft.isEmpty` matches web.
- Already disclosed: live paste/chip/restore not run against real rooms; coworker-stream failed-send restore omits the replaced link (parity with unmerged web).

### Assessment

Audience, membership, and fail-closed paste match web and Core; client never quotes a room the sender cannot see, and Core re-checks at send. Still a draft until live Mac paste is accepted.

mrosberghaus:
Both remaining minors are fixed in `942b4f519`.

**Paste and Match Style.** Correct, and it was worse than the quote swap: only `paste(_:)` was overridden, so `pasteAsPlainText:` and `pasteAsRichText:` reached `NSTextView` directly and skipped the composer's whole paste path — plain-text normalization, file and image attaching, and now the link swap. Both now delegate to `pasteText(from: .general)`. No test: the overrides are three-line delegations with no branching, and asserting them would mean writing to `NSPasteboard.general` and clobbering the developer's clipboard mid-run. `pasteText(from:)` itself stays covered by `pastesTextWithoutImportingClipboardFormatting` and `pasteSwapRemovesTheInsertedTextAndLeavesABlankDraft`.

**Vacuous audience check.** `roomsBody` now takes a roster. `pastedMessageLinkQuotesAcrossRoomsAndSendsWithoutABody` gives the target `user_1` and the source `user_1` + `peer`, so the rule is actually exercised, and a new `pastedMessageLinkStaysPlainForAReaderOutsideTheSourceRoom` puts an `outsider` in the target and asserts the paste stays plain with no message read at all. That also pins the wiring: the previous version would have passed even if the wrong rooms array were handed to the resolver.

**Leftover-newline disagreement.** Your reading is right, and your own `pasteSwapRemovesTheInsertedTextAndLeavesABlankDraft` asserts it — `captureDraft()` is empty after the sole pasted URL is removed, so the dismiss path takes the no-leading-space branch, matching web.

Re-ran after the changes: app tests, SokosumiChat (548) and SokosumiWorkspace (119), strict SwiftLint, SwiftFormat lint, and `pnpm check && pnpm typecheck`. Still draft pending live Mac paste acceptance.

PR #4802, opened 2026-09-18
fix(web): collapse the sidebar on one clock

mrosberghaus:
## Summary

Toggling the app sidebar reads as one panel getting narrower: panel, rows, and names share 200ms linear, and names leave with the narrowing edge instead of clipping to 1px on the first frame.

https://linear.app/masumi/issue/SOK-1114/make-the-sidebar-collapse-read-as-one-movement

## Changes

- Shared `duration-200 ease-linear` on the gap, the fixed panel, and the row button
- Row transition includes `margin` so the rail square's 4px offset eases
- Row names use `SIDEBAR_ROW_LABEL_CLASS` instead of `sr-only`: out of flex flow, clipped by the panel, opacity waits 200ms

## Verification

- `pnpm check`, `pnpm typecheck`, `pnpm --filter web test` — 796 files, 6543 passed
- Browser: 28px mark axis held in both states; mid-collapse names still visible while width travels


mrosberghaus:
## Grok review

### Fixed
- Important: Projects row still used `sr-only`, so that label clipped on frame one while every other row eased max-width (`ba549b9ce` — `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.tsx`)

### Remaining
- Minor: `sidebar-chat-list-skeleton.tsx` still hides name bars with `group-data-[collapsible=icon]:hidden`, so placeholders pop on frame one if you collapse while the chat list is loading.
- Minor: the `SIDEBAR_ROW_LABEL_CLASS` unit test does not assert `motion-reduce:transition-none` (the class already includes it via `SIDEBAR_COLLAPSE_TRANSITION`).
- Skipped / disagreed: account chip, section headers, and pending-invitation rail still use `hidden` or a dual-render `sr-only` copy — those are SOK-1107 two-state rail controls, not collapsing row names.

### Assessment
Panel, rows, and names now share 200ms linear with reduced-motion intact; the 28px mark axis holds in both states. Only leftover nits are skeleton name pop and a test assertion gap.

mrosberghaus:
Addressed the two remaining nits from https://github.com/masumi-network/sokosumi/pull/4802#issuecomment-5732377019 in 9ee250d12:

- Skeleton name bars use `SIDEBAR_ROW_LABEL_CLASS` (`max-w-0` on the same 200ms linear clock) instead of `hidden`, with `flex-none` so the ragged widths stay. Collapsing while the chat list is still loading no longer pops the placeholders.
- Label-class test now asserts `motion-reduce:transition-none`.

Left account chip, section headers, and pending-invitation rail as-is — those are two-state rail controls, not collapsing row names.

PR #4806, opened 2026-09-18
fix(web): handle server-action rejections at fire-and-forget call sites

schaier-io:
Closes the client half of Sentry issue **SOKOSUMI-E3** — `Error: An unexpected response was received from the server.`, 1130 events, ongoing in `vercel-production`.

Linear: [SOK-1116](https://linear.app/masumi/issue/SOK-1116/server-action-transport-failures-surface-as-unhandled-rejections)

## User-facing impact

None visible on the happy path. When a server-action POST fails at the transport layer, two admin surfaces currently get stuck: the matched-channels panel spins forever, and the scenario lab leaves its version picker disabled. Both now recover. The other three sites already behaved correctly and only leaked a rejection to Sentry.

## What the error is

Next.js throws error E394 in `fetchServerAction` when a server-action POST comes back as neither an RSC content type nor a redirect (`next/dist/client/components/router-reducer/reducers/server-action-reducer.js:143`). The reducer delivers that rejection to the caller's promise (same file, line 332). A call site that dispatches an action as `void …then(…)` with no rejection handler therefore turns it into an unhandled rejection.

Five call sites had none:

- `apps/web/src/app/(app)/chat/hooks/use-unread-thread-count.ts`
- `apps/web/src/app/(app)/chat/components/thread-mute-button.tsx`
- `apps/web/src/app/(app)/personal-assistant/components/skills-section.client.tsx`
- `apps/web/src/components/admin/matched-channels/matched-channels-hub.tsx`
- `apps/web/src/components/admin/soko-bots/scenario-lab.client.tsx`

`pinned-messages-panel.tsx` and `organization-chat-list.client.tsx` already passed a rejection handler and are untouched.

## Out of scope

Why the response was not RSC is not determined. SOK-1116 records three candidates (proxy sign-in redirect, a Next.js PPR invariant seen in the same trace, deploy skew) and the measurement that would separate them. That work belongs in its own PR.

## Verification

- `pnpm --filter web test "src/app/(app)/chat" src/components/admin/soko-bots` → `Test Files 135 passed (135)`, `Tests 1214 passed (1214)`
- `pnpm --filter web check` → `Checked 2301 files. No fixes applied.`
- `pnpm --filter web typecheck` → clean
- Mutation check: reverting only the `use-unread-thread-count.ts` hunk while keeping the new test makes the run exit 1 with `Unhandled Rejection`. Note the six assertions still pass without the fix — the guard is Vitest's unhandled-error detection failing the run, not an assertion.

## Blind spot

The five sites were found by grepping for `void <name>Action(`. A site that stores the promise in a variable or dispatches through a wrapper would not match. This is a floor, not a total.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when chat data requests fail, preserving unread counts and keeping mute controls from displaying uncertain states.
  * Browse results in the personal assistant now remain visible when a request is rejected.
  * Admin channel lists now exit loading state and show an error when requests fail.
  * Version switching in the scenario lab now resets correctly after a failed request.

* **Tests**
  * Added regression coverage for preserving unread counts after failed requests.


schaier-io:
@coderabbitai

PR #4809, opened 2026-09-18
fix(web): serialize DOCX export so concurrent requests keep their own globals

schaier-io:
Closes SOK-1017 (security review finding OTHER-07, Low).

## Impact

The DOCX export route replaced process globals: `withDocxExportFetchGuard` swapped `globalThis.fetch` and `setupDomContext` swapped `window`, `document`, `HTMLElement` and `SVGElement`. Fluid Compute reuses one function instance across concurrent requests, so two simultaneous exports traded SSRF guards and deleted each other's DOM mid-conversion. A user could get a failed export, or one running with the wrong guard.

`mdast2docx` and the `@m2d/*` plugins read those globals ambiently and accept no injected `fetch` or `Window`, so the conversion now runs one at a time per instance. That trades export throughput for correctness, which is the stopgap the ticket describes.

Each remote image fetch now carries a 10s timeout. `ssrfSafeFetch` sets none and `@m2d/image` calls `fetch` with no init, so a host that accepts the connection and never answers would otherwise hang the conversion and hold the lock for every later export on that instance.

`setupDomContext` no longer returns a no-op cleanup when a document already exists. Handing a caller another caller's document, with a cleanup that restores nothing, is what made the interference invisible.

## Scope

This bounds exports against each other only. While a conversion runs, the swapped `globalThis.fetch` still reaches unrelated requests on the same instance. Removing that swap needs an injected image resolver, tracked in SOK-1121. The lock is also unbounded and ignores `request.signal`; a concurrency cap belongs with SOK-1012.

## Verification

- `pnpm --filter web test` — 801 files, 6630 passed, 1 skipped
- Mutation checks, each reverted afterwards: bypassing the lock fails 2 tests, dropping the image-fetch timeout fails 2, restoring the no-op cleanup branch fails 3
- Abort behaviour measured against a local server: a host that accepts and never answers rejects in 303ms, a dripping body in 301ms

No UI change, so no screenshots. No schema change and no migration.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved DOCX export reliability when multiple exports are requested at the same time.
  - Added a 10-second timeout for remote images included in DOCX exports, preventing stalled requests.
  - Ensured individual exports remain isolated, reducing interference between concurrent or nested export operations.

- **Tests**
  - Expanded coverage for concurrent exports, timeout handling, request cancellation, and DOM cleanup.


schaier-io:
@coderabbitai review

PR #4821, opened 2026-09-18
fix(apple): let test builds link with code coverage on

mrosberghaus:
## Problem

Xcode Cloud's "Pull Requests" **Test** action fails on every Apple PR (first seen on #4820) with 2 build errors and 0 test failures:

```
Undefined symbol: ___llvm_profile_runtime
Linker command failed … TreeSitterYAMLdynamic-product / TreeSitterTypeScriptdynamic-product
```

GitHub Actions never saw it because `apple.yml` passes `-enableCodeCoverage NO`. Xcode Cloud tests with coverage on.

## Reproduction (the loop used for diagnosis)

From `apps/apple`, fresh derived data, red on `main` in ~50 s:

```
xcodebuild -project Sokosumi.xcodeproj -scheme Sokosumi -configuration Debug \
  -destination 'platform=macOS,arch=arm64' -skipPackagePluginValidation \
  DEVELOPMENT_TEAM= CODE_SIGN_IDENTITY=- -derivedDataPath <fresh> \
  -enableCodeCoverage YES build-for-testing
```

## Cause

`SokosumiTests` is hosted in the app (`TEST_HOST` / `BUNDLE_LOADER`) but also declared and linked the same package products the app links (`CoreAPI`, `SokosumiAuth`, `SokosumiChat`, `SokosumiRealtime`, `SokosumiWorkspace`, `OpenAPIRuntime`, plus `HTTPTypes`). When two targets in one build graph link the same static package products, Xcode rebuilds **every** package product as a dynamic framework — 74 `*dynamic-product` targets here. With coverage on, the C-family package sources (tree-sitter grammars, msgpack, …) are compiled with profile instrumentation and reference `___llvm_profile_runtime`, but those dynamic-framework link lines carry no profile runtime. YAML and TypeScript were just the first two links to run; all of them would fail.

## Hypotheses tested

1. **Limit coverage to the app target in the scheme** (`onlyGenerateCoverageForSpecifiedTargets`) — falsified. In a clean build the package C targets are still instrumented; the setting only scopes the report.
2. **Remove the package products from the test target's Frameworks phase only** — still red, 74 dynamic products remain while `packageProductDependencies` lists them.
3. **Test target declares no package products at all** — green: 0 dynamic products; packages link statically into the app, whose link step has the profile runtime; the hosted test bundle binds to the app's copy and still `import`s the modules.

## Change

- `project.pbxproj`: `SokosumiTests` loses its seven package product build files and its `packageProductDependencies` entries (21 lines removed). Nothing else.
- `apps/apple/AGENTS.md`: a gotcha explaining why the test target must not declare package products, and the command that checks it.

## Verification (local, Xcode 27)

- The red command above: **TEST BUILD SUCCEEDED**, 0 `dynamic-product` targets.
- `xcodebuild … -enableCodeCoverage YES test -only-testing:SokosumiTests`: **TEST SUCCEEDED**, 181 passed test cases, 0 failed, profile data produced.
- The GitHub CI form (`-enableCodeCoverage NO`): **TEST SUCCEEDED**.
- Release build: succeeds. Shipped app size is unchanged (276 MB on `main` vs 278 MB here, 0 embedded frameworks in both) — the app was always linked statically; only test build graphs went dynamic.

No regression test is added: the seam is the build graph, and Xcode Cloud's Test action on every PR is the guard. This PR's own Xcode Cloud run is the real confirmation — I could not run Xcode Cloud locally.


mrosberghaus:
Superseded by #4824, which merges this branch in — `project.pbxproj` and the AGENTS.md gotcha ship there.

#4824 converts `apps/apple` to an Xcode workspace so the `Sokosumi` scheme runs the app tests and all five package suites in one action. That makes this fix load-bearing rather than optional: with the package test targets in the scheme and coverage on, a fresh build produced **826** `dynamic-product` targets and failed to link; with this branch merged in, **0** and `TEST BUILD SUCCEEDED`.

**Close this PR once #4824 merges** — its diff will be empty by then. GitHub will not do it automatically: closing keywords only close issues, and the repository squash-merges, so this branch's tip never becomes an ancestor of `main`.

PR #4823, opened 2026-09-18
feat(web): rank sidebar projects by last visit and cap the list

mrosberghaus:
## What changed

The sidebar's Projects disclosure now shows **five whole rows, ordered by what you last opened**, and no longer scrolls inside itself.

Before, the list was capped at `max-h-64` with its own scrollbar, so a project row was sliced through the middle right above Tasks, and expanding it pushed Tasks / Calendar / Files / History down by as much as 16rem. It also walked *every* page of the workspace to fill that box.

Core sorts projects by latest activity, so a teammate touching a project reorders your sidebar. This ranks by your own visits instead, so the list holds still.

## User-facing impact

- Five whole rows, never a half row bleeding into Tasks.
- The order stops reshuffling when other people work.
- The project you're in is reliably near the top.
- **Behavior loss worth knowing:** the old list eventually loaded every project as you scrolled. Now it's five, and the rest are reached through the `Projects` row above the disclosure, which still links to `/projects`.

## Notes for review

- `order-sidebar-projects.ts` holds the ordering rule as a pure, tested function: pins first in the reader's order, last-visited filling to five, **pins claim slots rather than add rows**, no project listed twice (the rule `partition-rooms-for-sidebar.ts` already uses for chat).
- `pinnedIds` is wired but passed empty — **pins are a follow-up PR** with the Core pin routes (server-side, matching chat pins). The rule is already covered by tests.
- `use-recent-projects.ts` deliberately mirrors the existing `use-frequently-used-emojis.ts` store shape (`useSyncExternalStore` over localStorage, cross-tab via the `storage` event).
- Visits are per-browser on purpose: they're derived, so losing them on another device costs nothing. A *pin* is explicit work, which is why pins go to the server instead.
- The disclosure drops `useInfiniteQuery` → `useQuery`. The scroll handler, focus-driven paging, `useLoadWhenVisible` usage and the inner scroller are deleted (net −62 lines in that component).
- No new translation keys. I dropped the "Show all N projects" row from the original mock: the `Projects` row directly above already links to the same place, and Core's response carries no workspace total, so the count would have been invented.

## Verification

- `pnpm --filter web test` — 805 files, 6673 tests pass
- Not yet exercised in a running browser; testing on the Vercel preview.

## Follow-up

Server-side project pins: new per-user pin table + Core routes + client regen. `Project` has no per-user membership row today, unlike `ChatRoomUserMember.starredAt`. The pin endpoint should return pinned projects **directly** rather than relying on them appearing in the single page this PR fetches.


mrosberghaus:
## Grok review

### Fixed
- Important: last-visit log leaked across users and workspaces on a shared browser (`70fa3cf9d` — `apps/web/src/hooks/use-recent-projects.ts`). Storage is now `userId` + `organizationId`, matching the query key. Record waits for session. Tests cover same-org other user, other org, and personal vs org.

### Remaining
- Minor: `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.tsx`: skeleton is 3 rows (`SKELETON_NAME_WIDTHS`); cap is 5. A reader who left the disclosure open can still see height jump when names land.
- Minor: ranking can only promote visits that appear on the first Core page (`PROJECTS_PAGE_LIMIT` = 20). A quiet project just opened in a large workspace stays off the disclosure; `/projects` is the overflow path (same as the pin follow-up).
- Minor: `PROJECTS_EXPANDED_STORAGE_KEY` is still a shared browser preference, not scoped. Not a visit leak.
- Skipped / disagreed: pins rendering past the cap — intentional for the pin follow-up; this PR passes `pinnedIds: []`.
- Skipped / disagreed: paging the full workspace again — that is the behavior this PR removed. Remaining projects stay behind the `Projects` row → `/projects`.

### Assessment
Visit ranking no longer crosses users or workspaces; the five-row cap still reaches the rest through the Projects link, including on the collapsed rail.

mrosberghaus:
Went through the three remaining notes.

**Skeleton height (fixed — `fe64f841c`)**

Correct, and the doc comment above it claimed the opposite of what the code did. The stand-in now sizes off the reader's scoped visit log, floored at 3 and bounded by `SIDEBAR_PROJECT_ROW_CAP`, instead of a fixed 3. Only someone who has used the disclosure ever sees this, so their own history is the closest thing to a row count we hold before the page lands.

It can still fall short — someone with two visits and ten projects gets 3 rows and 5 arrive. Making it exact means remembering the last rendered count in scoped storage, which is a real option, but it buys accuracy on a sub-second placeholder for another persisted key. The comment now states the estimate instead of promising no resize. Three tests cover it: history-sized, capped, and the no-history floor.

**First-page ranking limit (no change — known trade-off)**

Agreed and already called out in the PR description. A visit can only be promoted if the project is on the first page of `PROJECTS_PAGE_LIMIT` = 20. Fixing it properly means Core returning the reader's specific projects rather than hoping they land on page one — which is exactly what the pin follow-up needs, since a pin has the same problem and matters more (a pin is explicit; a visit is derived). Folding both into that PR rather than half-solving it here.

**`PROJECTS_EXPANDED_STORAGE_KEY` scoping (disagree — consistent as-is)**

Checked the established pattern before changing it. The sidebar's own collapsed state is `sidebar_state`, a plain unscoped cookie (`apps/web/src/lib/ui-preferences/sidebar-state.ts:29`) shared across every user and org on the browser. Chrome preferences here are per-browser by design; it's the *data* that got scoped in `70fa3cf9d`. Scoping this one key would make the disclosure behave differently from the sidebar that contains it, for no privacy gain — as the review notes, it isn't a visit leak. Leaving it.

Verification on the merged state: 805 files, 6691 tests pass, typecheck and Biome clean.

PR #4825, opened 2026-09-18
chore(deps): bump turbo to 2.11.1

mrosberghaus:
## What

Two commits:

1. **`chore(deps): bump turbo to 2.11.1`** — bumps the pinned `turbo` devDependency from `2.10.13` to `2.11.1` (current `latest`).
2. **`chore(deps): declare pnpm pin via devEngines.packageManager`** — the 2.11 deprecation migration.

Release notes: https://turborepo.dev/blog/2-11

## User-facing impact

None at runtime — build tooling only. What the team gets:

- **Up to 4x faster time-to-first-task** than 2.9 (Vercel's largest monorepo saw 45%; smallest repo tested saw 74% faster startup). Should shave time off `pnpm typecheck` / `pnpm test` / CI.
- `turbo prune --production` — a new flag to exclude workspace packages only reachable through `devDependencies`. Not used here yet, but available if we ever want leaner Docker images.
- Experimental Cargo / uv / `go.work` workspace discovery (not relevant to this repo today).

No `turbo.json` changes are required.

## The `devEngines.packageManager` migration

2.11 deprecates the top-level `packageManager` field in favour of Node's `devEngines.packageManager`. The pnpm 12.4.2 pin is now declared there and is the source of truth.

**The legacy top-level `packageManager` field is deliberately kept.** Vercel's [package-managers docs](https://vercel.com/docs/package-managers) state that under Corepack it reads `packageManager` — `devEngines` is not mentioned. This repo needs Corepack at all because Vercel's native pnpm support tops out at 10 and we're on 12. Dropping the field risks breaking deploys, so it stays until Vercel supports `devEngines`.

`devEngines.packageManager` uses **`onFail: "warn"`, deliberately not `error`**. npm validates `devEngines` against itself, so with `error` (or with the key absent, which defaults to error) npm rejects the whole tree with `EBADDEVENGINES` — that breaks `npx` at the repo root and from `apps/apple`, where the documented skill install in `AGENTS.md` walks up to the root `package.json`. Measured on npm 11.19.1:

| `onFail` | `npx` at repo root |
| --- | --- |
| `"error"` / absent | fails `EBADDEVENGINES` |
| `"warn"` / `"ignore"` | works |

Under `warn`, a mismatch between the two fields is a Corepack **warning** and the legacy `packageManager` value wins — so the two must be bumped together. No CI drift guard was added: there's no Renovate/Dependabot here, pnpm bumps are manual, and both values sit on adjacent lines of one file. `AGENTS.md` records the constraint.

`scripts/cloud-agent-db/ensure-pnpm.sh` now composes its Corepack spec from `devEngines.packageManager`, falling back to the legacy field, so it keeps working once that field is eventually dropped. `README.md` and `AGENTS.md` point at the new field.

## Verification

Run locally on this branch (post-rebase onto the `main` merge):

| Check | Result |
| --- | --- |
| `pnpm exec turbo --version` | `2.11.1` |
| `pnpm exec turbo ls` | all 11 packages discovered |
| `corepack pnpm --version` | `12.4.2` |
| `pnpm install --frozen-lockfile` | lockfile up to date |
| `pnpm check` (Biome) | 4232 files, no fixes applied |
| `pnpm typecheck` | 20/20 tasks pass |
| `pnpm --filter web test` | 802 files, 6654 tests pass |
| `pnpm --filter @sokosumi/core test` | 531 files, 5865 tests pass |
| `bash -n scripts/cloud-agent-db/ensure-pnpm.sh` | syntax OK; spec resolves to `pnpm@12.4.2` |

`ensure-pnpm.sh` was not executed end-to-end locally — it runs `rm -rf ~/.local/share/pnpm/.tools/pnpm`, which is intended for the cloud-agent VM, not a dev machine. The one changed line (the spec extraction) was verified in isolation.

### Known pre-existing failure (not caused by this PR)

`apps/cli/test/tui/status-app.test.ts` hangs with `Promise resolution is still pending but the event loop has already resolved`.

This is unrelated to the upgrade: the test runs under `tsx --test`, which turbo only shells out to. While verifying, the same test was observed hanging **simultaneously in four other local worktrees still on turbo 2.10.13**. This PR touches nothing under `apps/cli`.


mrosberghaus:
## Grok review

Reviewed `061379c3725d6d6c81299679d0e355b01569d297`..`11bfeb33fbc653219600c71c265304d1dab050a6` (`package.json` + `pnpm-lock.yaml` only) against the PR title/body, pinned-deps rule, ADR 0008, and Turborepo 2.11 notes.

### Fixed

### Remaining
- Skipped / disagreed: migrate root `packageManager` → `devEngines.packageManager` in this PR — 2.11 still honors the top-level field; removal is a future major. `scripts/cloud-agent-db/ensure-pnpm.sh` and Vercel filtered installs still read `packageManager`. Leave it.

### Assessment
Ready to merge: Yes

Exact pin `turbo@2.11.1` matches npm `latest`; lockfile specifier and all six `@turbo/*` optional binaries moved in lockstep; no leftover `2.10.13`; `turbo.json` correctly unchanged (Cargo/Python discovery is `futureFlags`-gated). CI already green on Typecheck, Build, Test Core/CLI/Packages.

mrosberghaus:
Replying to the Grok review above.

## The review covered a stale range

It reviewed `061379c37..11bfeb33f` — "`package.json` + `pnpm-lock.yaml` only". The `devEngines` migration landed afterwards in `cc09912ba`, so the change it argues against was not in the diff it saw.

## On the recommendation itself

> Skipped / disagreed: migrate root `packageManager` → `devEngines.packageManager` in this PR — 2.11 still honors the top-level field; removal is a future major. `scripts/cloud-agent-db/ensure-pnpm.sh` and Vercel filtered installs still read `packageManager`. Leave it.

Both facts are correct, and both were the reason the top-level field was **kept**, not removed. `package.json` carries both fields; `ensure-pnpm.sh` reads `devEngines` with a fallback to the legacy field. The migration was requested explicitly after that deploy risk was raised.

So the conclusion ("leave it") doesn't follow from those two premises — nothing was taken away. **But digging into it surfaced a real defect the review didn't name, and the instinct to be cautious was warranted.**

## The actual bug: `onFail: "error"` broke `npx` repo-wide

npm validates `devEngines` against *itself*. Since the declared manager is pnpm and npm is not pnpm, npm rejects the entire tree with `EBADDEVENGINES`. Measured on npm 11.19.1:

| `onFail` | `npx` at repo root |
| --- | --- |
| `"error"` | ❌ `EBADDEVENGINES` |
| absent (defaults to error) | ❌ `EBADDEVENGINES` |
| `"warn"` | ✅ works |
| `"ignore"` | ✅ works |

This had a concrete victim. [AGENTS.md:277](https://github.com/masumi-network/sokosumi/blob/main/AGENTS.md#L277) documents the Apple skill install as `npx skills add … -s swiftui-expert-skill -y` run with `apps/apple` as cwd. `apps/apple` has no `package.json`, so npm walks up to the root and hit the error. Verified broken, then verified fixed.

Fixed in `f5eba30ae` by setting `onFail: "warn"`. Corepack still resolves pnpm 12.4.2; `npx` works from both the root and `apps/apple`.

## Correction to my own PR body

I had claimed Corepack "hard-errors when the two disagree, so they cannot drift". That was true under `onFail: "error"` and is **not** true under `warn` — a mismatch is now a warning and the legacy `packageManager` value wins:

```
! Corepack validation warning: "packageManager" field is set to "pnpm@10.0.0" which does not
  match the value defined in "devEngines.packageManager" for "pnpm" of "12.4.2"
10.0.0
```

So: bump both fields together. I did not add a CI drift guard — there's no Renovate or Dependabot in this repo, pnpm bumps are manual, and the two values sit on adjacent lines of one file. `AGENTS.md` now records the constraint and why `warn` is deliberate.

Side benefit: `warn` also removes the hard-failure mode on the one path I can't verify locally — Vercel. Preview deploys only trigger on `apps/web/**`, `apps/core/**`, `packages/**`, so none ran here. Vercel reads the untouched top-level `packageManager`, and under `warn` nothing in that path can hard-fail regardless.

## Verification after the fix

`corepack pnpm --version` → 12.4.2 · `npx` OK at root and from `apps/apple` · `pnpm install --frozen-lockfile` clean · `turbo ls` 11 packages · Biome 4360 files clean · `pnpm typecheck` 20/20.

Worth a `/deploy` before merge to confirm Vercel on a real preview build.

PR #4826, opened 2026-09-18
feat(web): match chat sidebar section motion to the Projects disclosure

mrosberghaus:
## What

The chat sidebar's sections — Pinned, Channels, External, Archived, Direct Messages — snapped open and closed while the Projects disclosure directly above them slid. They now run the **same** motion as the one shipped in #4823.

## User-facing impact

- Opening or closing a chat section slides: its height travels with the rows, so every section below it moves with them instead of jumping.
- The section chevron turns on the same 200ms ease-out as the Projects chevron (was Tailwind's default 150ms `ease`).
- `motion-safe:` gates the height animation and `motion-reduce:` still stops the chevron, so reduced-motion readers get the old instant toggle.

## How

New `ChatSidebarSectionContent` lives beside `ChatSidebarSectionHeader` — the section's other half — and replaces all five bare `<CollapsibleContent>` in `organization-chat-list.client.tsx`. It carries the classes verbatim from `projects-menu-item.tsx`, plus two things the chat sidebar needs and Projects does not:

- **`-mx-2 px-2`** — the height animation needs `overflow-hidden`, which would otherwise clip a room's `RailAttentionPill`. The pill sits at `-left-2`, in the sidebar group's padding, outside the rows' box. The negative margin moves the clip edge out by exactly that gutter; the padding puts the rows back where they were.
- **Pinned drops the clip while reordering** — a dragged row translates with the pointer past the section's box (`useDraggable` transform in place, no `DragOverlay`), so `overflow-hidden` would cut it off mid-drag. The section cannot be mid-animation while its rows are being reordered: the reorder toggle only exists on an open heading.

## Verification

- `pnpm --filter web test src/components/chat` — 400 passed (36 files)
- `pnpm check` — clean

Two new tests in `chat-sidebar-section-header.test.tsx` lock the animation classes (mirroring the assertion in `projects-menu-item.test.tsx`) and the gutter that keeps the rail pill visible.

Not verified in a browser: the change is class-only, and this worktree's database rejects sign-in.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `organization-chat-list.client.tsx` — the Pinned comment says the section cannot be mid-animation while reordering because the toggle only exists on an open heading. The reorder control lives on the heading itself, so it is still available when Pinned is closed, and the heading remains a `CollapsibleTrigger`. Closing (or opening) while `isReorderingPinned` keeps `overflow-visible`, so the height animation will not clip the rows for 200ms.
- Minor: `chat-sidebar-section-header.test.tsx` — no assertion that `cn(..., "overflow-visible")` actually drops `overflow-hidden` (twMerge) while Pinned is reordering.
- Skipped / disagreed: first-paint `animate-collapsible-down` on default-open sections (Pinned, Channels, External, DMs) — Radix `CollapsibleContent` already sets inline `animation-name: none` on mount via `isMountAnimationPreventedRef`, then restores it on the next open/close. Same classes as Projects are safe here.
- Not verified in a browser (author noted the same). Class tokens, Radix mount skip, twMerge overflow override, and the `-mx-2`/`px-2` gutter vs `RailAttentionPill` (`-left-2`) and `SidebarRailSelectionBar` (`-right-2`) checked in code only.

### Assessment
Ready to merge: Yes

Chat sections reuse the Projects disclosure motion (`motion-safe` height keyframes, 200ms ease-out chevron, `motion-reduce` on the chevron), widen the clip box so rail pills and the selection bar stay visible, and drop the clip while Pinned rows drag. CI is green.

mrosberghaus:
Both minor findings addressed, one by fixing the comment rather than the code.

**Finding 1 — closing while reordering keeps `overflow-visible`:** not reproducible. `canReorderPinned` is `pinnedOpen && pinned.length > 1` ([organization-chat-list.client.tsx:266](https://github.com/masumi-network/sokosumi/blob/main/apps/web/src/components/chat/organization-chat-list.client.tsx#L266)), and `isReorderingPinned` is `pinnedReorderMode && canReorderPinned`. The render that flips `pinnedOpen` to `false` is already the render where `isReorderingPinned` reads `false`, so the collapse clips exactly as every other section does — no dependence on the render-phase reset at line 270, the conjunction alone settles it. Two consequences of the same `pinnedOpen` term: the reorder control is `createAction={canReorderPinned ? … : undefined}`, so it is also not rendered on a closed heading.

The comment was the problem: it justified the escape hatch by *where the toggle lives* instead of *what gates the flag*, which is the weaker of the two facts and the one that reads as an invariant someone could break. Rewritten to name `canReorderPinned` carrying `pinnedOpen`.

**Finding 2 — no assertion that twMerge drops `overflow-hidden`:** fair, and it was the load-bearing assumption of the whole escape hatch. Added `lets a caller drop the clip, which Pinned needs while its rows drag` in `chat-sidebar-section-header.test.tsx`: renders `ChatSidebarSectionContent` with `className="overflow-visible"` and asserts the class list contains `overflow-visible` and **not** `overflow-hidden`. Swapping `cn` for a plain join now fails it.

`pnpm --filter web test src/components/chat` — 401 passed (36 files). Typecheck and `pnpm check` clean.

Agreed on the first-paint point: Radix's `isMountAnimationPreventedRef` is why Projects can carry these classes on a default-open section, and the same holds here.

PR #4827, opened 2026-09-18
refactor(web): shape the sidebar chat skeleton like the real room list

mrosberghaus:
## What

The sidebar's room-list placeholder drew one anonymous section of five identical rows, each with a 24px square mark. The list it stands in for has two sections every reader always has, and neither kind of room draws that square expanded — so the shape that arrived was not the shape that had been held.

It now stands in for **Channels** and **Direct Messages**, three rows each:

- **Channel rows** carry `ChannelRoomMark`'s two shapes: a 16px kind glyph beside the name expanded, a 24px tile on the rail.
- **Direct rows** carry `DirectRoomAvatarStack`'s: 24px round faces in both states, with one group Direct whose second face grows its slot to the right and drops off the rail.
- Pinned, External and Archived are left out. Each depends on data the boot frame has not loaded, and a section that appeared and then went away would move every row under it.

Row geometry is unchanged — still built from `SIDEBAR_ROW_CLASS` plus the rail pad classes, so the 28px leading axis and the 48px label column hold in both states.

## User-facing impact

The placeholder the reader sees while the room list streams in now resolves into the list that actually arrives: same section break, same marks, same columns. Less of a visible re-shuffle on the swap.

## Tests

Two assertions in `app-sidebar-fallback.test.tsx` pinned the old shape by a total — `rows * 2 + 2` skeletons, and a `w-20` heading width. No single number covers a Channel's two marks and a group Direct's two faces, so both now assert per row and per heading instead.

## Verification

- `pnpm --filter web test src/app/(app)/components/__tests__/sidebar-row-slot-contract.test.tsx` — 4/4
- `pnpm --filter web test src/app/(app)/components/app-sidebar-fallback.test.tsx` — 5/5
- `pnpm check && pnpm typecheck` (precommit) — clean


mrosberghaus:
## Grok review

### Fixed
- Important: tests now pin Channel glyph+tile, Direct face counts/overlap, and both heading widths (`2d303a3fe` — `apps/web/src/app/(app)/components/app-sidebar-fallback.test.tsx`). The previous loop only required `> 0` skeletons in a slot, so a 24px square per row would still have passed.

### Remaining
- Minor: `ChannelMarkSkeleton` shows the collapsed tile with `group-data-[collapsible=icon]:block`; `ChannelRoomMark` uses `inline-flex`. Same 24px box in the flex slot — no jump.
- Minor: rail/expanded geometry tests still sample the first Channel row only, not the wider group Direct slot. The placeholder test now covers that row's faces.
- Skipped / disagreed: Channels is org-only (`hasOrganization` in `organization-chat-list.client.tsx`), so a personal workspace still loses a section on swap. Boot fallback has no org signal. Not adding a `hasOrganization` prop — same class of guess as omitting Pinned, and the inner Suspense is short.

### Assessment
Ready to merge: Yes

Skeleton now holds Channels + Direct Messages with the real marks in both sidebar states; row geometry is still the shared primitives, so the swap should not reshuffle columns.

mrosberghaus:
Picked up `2d303a3fe` — the tighter placeholder assertions run green here (`app-sidebar-fallback.test.tsx` 5/5). The earlier loop was indeed a `> 0` check that a 24px square would have passed.

### Personal workspace — implemented rather than skipped

`17bffa653`. The reasoning for skipping doesn't hold at the one call site that matters: `PrivateCachedAppSidebar` already has `activeOrganizationId` in scope, immediately beside the `Suspense` boundary whose fallback this is. That is a known value, not a guess — unlike Pinned/External/Archived, which depend on the room payload the boundary is waiting for.

And the Suspense is not short there: it wraps `getPrivateCachedChatListChrome`, a Core API round trip. A personal-workspace reader was getting a Channels heading plus three rows that then vanish, sliding Direct Messages up — precisely the jump this PR set out to remove.

So `SidebarChatListSkeleton` now takes `hasOrganization`, defaulting to `true`. The boot shell (`AppSidebarFallback`) has no session and genuinely cannot know, so it keeps the default and behaves exactly as before. Covered by a colocated `sidebar-chat-list-skeleton.test.tsx` (3 tests: org, personal, default).

### Not changing

- **`block` vs `inline-flex` on the collapsed tile** — `SidebarRowSlot` is `inline-flex`, so its children are flex items and the outer `display` is discarded either way. `ChannelRoomMark` needs `inline-flex` because it centres initials and a corner mark; the skeleton holds nothing to centre. Matching the token would only mean editing the assertion you just pinned it with, for no rendered difference.
- **Geometry tests sampling one Channel row** — every row comes from the one `RoomRowSkeleton`, so looping would assert the same class string three times. The group Direct's wider slot is the row-level difference, and the placeholder test covers its faces.

Verified: `pnpm --filter web typecheck` clean; `sidebar-chat-list-skeleton` + `app-sidebar-fallback` + `sidebar-row-slot-contract` 12/12; `pnpm check && pnpm typecheck` clean.

PR #4834, opened 2026-09-19
feat(web): open recent projects in a flyout from both sidebar states

mrosberghaus:
## What this changes

The Projects disclosure pushed Tasks, Calendar, Files and History down by five rows every time it opened, and the collapsed rail could not show recent projects at all — a 56px rail has nowhere to put them.

Both sidebar states now answer hover with a flyout panel anchored to the Projects row:

- **Vertical movement is zero.** The sidebar is a fixed list of rows that never reflows, in either state and across the collapse toggle.
- **The collapsed rail reaches recents for the first time.** Previously the only way to a project from the rail was to expand the sidebar first.
- **The panel floats over the page**, so its height costs the sidebar nothing. This matters because `MenuItems` and the chat list share one scroll container — an inline list competes with the reader's rooms for the same scroll.

## Design notes

Reuses `HoverCard`, already the app's flyout primitive (`chat-participant-hover-card`), on the same 200ms open / 100ms close beat. The open delay is what keeps a sweep up the sidebar towards the logo from flashing the panel.

The rail tooltip is **conditional, not removed**. Where there is a flyout it is the hover hint, headed with the same label, and a tooltip would race it to the same spot. A workspace with no projects has no panel, so there the tooltip still names the icon.

The chevron stays as a static `aria-hidden` indicator pointing at the panel — it has no open/closed state of its own, so it never animates the row.

Deletes the disclosure machinery this makes dead: the `Collapsible`, `PROJECTS_EXPANDED_STORAGE_KEY` and its read/write pair, `disclosable`, the `animate-collapsible-*` classes, and `skeletonRows()` — which existed only to guess how tall the inline list would be before it pushed everything down.

`SIDEBAR_PROJECT_ROW_CAP` stays at 5. Height is free in a flyout so it could go higher, but that is a separate call.

## Verification

- `pnpm --filter web test` — 806 files, 6690 passed
- `pnpm --filter web messages:parity` — de/es ok; removed `expandProjects` / `collapseProjects` from all three locales

Tests open the panel by **focus** rather than hover: Radix opens on either, focus has no delay to wait out, and it pins the keyboard path as real behaviour rather than leaving the pointer as the only way in.

**Not verified in a browser.** How the 200ms delay and the panel placement actually feel is what the preview is for.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/components/sidebar/components/order-sidebar-projects.ts` still describes the old inline disclosure (file comment + `SIDEBAR_PROJECT_ROW_CAP` comment).
- Minor: the flyout `allProjects` footer (`projects-menu-item.tsx`) is a raw `Link` with no `focus-visible` ring; project rows get one from `SidebarMenuSubButton`.
- Skipped / disagreed: recents are pointer-only. HoverCard content is portaled to `body`, so Tab from the Projects row moves to the next sidebar item and the 100ms close delay shuts the panel; the chevron is `aria-hidden` and touch has no hover. `/projects` still works for keyboard and the mobile sheet. Restoring a chevron control or swapping to Popover would undo the PR's hover-only, zero-reflow choice — leaving that as a product call, not a code fix here.

### Assessment
Desktop hover flyout matches the spec (no sidebar reflow, collapsed rail recents, dead disclosure/storage gone, CI green). Recents are not a keyboard or touch control; confirm that is acceptable before merge.

Ad-hoc review only — general-purpose reviewer subagent dispatch was unavailable.

mrosberghaus:
Both Minor items fixed in `32c7c7f99`.

**Stale disclosure wording** — `order-sidebar-projects.ts` described a control that no longer exists, and pointed at the `Projects` row for the projects the cap leaves out. The panel's own `All projects` footer does that now; the file comment and the `SIDEBAR_PROJECT_ROW_CAP` comment both say so.

**Missing focus ring** — correct, that footer was the one row in the panel without one. Now carries the sidebar's convention, `outline-hidden ring-sidebar-ring focus-visible:ring-2`, matching what `SidebarMenuSubButton` gives the project rows.

---

**On the keyboard finding: the conclusion is right, and I'd go further than "product call".**

Confirmed both halves in this branch:

1. `HoverCardContent` wraps `HoverCardPrimitive.Portal` (`components/ui/hover-card.tsx`), so the panel is appended at the end of `<body>` — after the whole page, not after the sidebar row.
2. Focus leaving the trigger closes it. That is already asserted in this PR's own suite: `opens without a loading state and does not refetch when the reader comes back` does `fireEvent.blur(trigger)` and then waits for the rows to leave the document.

Together those mean the panel cannot be entered by keyboard — moving focus is precisely what dismisses it. And because Radix HoverCard *opens on focus*, a keyboard user currently gets a panel that appears and then cannot be reached, which is worse than one that never opened.

Where I disagree is the framing that fixing it "would undo the PR's hover-only, zero-reflow choice". A `DropdownMenu` or `Popover` opened on `pointerenter` keeps both properties — it is still hover-triggered and still an overlay, so the sidebar still never reflows. The hover-only behaviour is a consequence of picking HoverCard, which is Radix's preview primitive, not of the design.

Worth noting this is a **regression**, not a pre-existing gap: the inline disclosure this PR removes rendered its rows inline, so in the expanded sidebar they were tab-reachable. No destination is unreachable today — `/projects` still gets you everywhere — but the shortcut is now mouse-only.

Raising it with the author before changing the primitive, since it is the one decision on this PR that was explicitly chosen.

mrosberghaus:
Keyboard access fixed in `78cd7a15a`. Full web suite green (806 files, 6696 tests).

**One deviation from what I proposed:** anchored `Popover`, not `DropdownMenu`.

`DropdownMenuTrigger` claims Enter and Space to toggle the menu, and it is composed so that a Slot child cannot decline them without `preventDefault()` — which would also cancel the link's navigation. That would have turned the row from a link into a menu button, changing what click and Enter do. `PopoverAnchor` contributes positioning and no behaviour, so the row stays a link that navigates to `/projects` exactly as before, and the open/close logic is explicit rather than fought with.

What that buys, and what it costs: the panel no longer gets roving focus or type-ahead for free, so the keys are wired by hand. For a five-row list of links that is a fair trade against changing the row's semantics.

Behaviour now:

| | before | after |
| --- | --- | --- |
| Hover row | opens | opens, unchanged (200ms/100ms) |
| Focus row | opened a panel that could not be entered | nothing; `aria-expanded="false"` |
| ArrowRight / ArrowDown | — | opens, focus moves into the rows |
| Enter | opened the panel | navigates to `/projects` |
| Escape | — | closes, focus returns to the row |
| Pointer open | — | never takes focus off what the reader was in |

`ArrowRight` because it is the way the chevron points and the side the panel comes out on; `ArrowDown` accepted too since a reader may reach for it.

Four tests cover the paths that were previously unreachable, including one asserting Enter is *not* `defaultPrevented` so the row still navigates, and one asserting a pointer open leaves `document.activeElement` alone.

Still hover-triggered, still an overlay, so the zero-reflow property this PR exists for is unchanged.

mrosberghaus:
## Grok review

### Fixed
- Important: keyboard-opened flyout restored focus to the Projects row on click/focus outside, so leaving the panel yanked the caret back (`8fa5434e2` — `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.tsx`)

### Remaining
- Minor: `menu-items.test.tsx` still says the flyout is "headed by the same label" as the row; the heading is `recentProjects` now.
- Minor: Radix non-modal `FocusScope` still loops Tab inside the panel (`loop: true`). Escape is the way out; that is the dialog default, not a trap.
- Skipped / disagreed: `aria-haspopup` on the row — Enter must keep navigating to `/projects`, and `haspopup` would tell a screen reader the opposite.

### Assessment
Hover still opens without reflow, keyboard still enters via ArrowRight/ArrowDown, and focus now stays where the reader went after they left.

Ad-hoc review only — general-purpose reviewer subagent dispatch was unavailable.

mrosberghaus:
Reviewed `8fa5434e2` before building on it — the finding and the fix both hold.

The bug was real and mine: `openedByPointer` was the only thing gating the restore, so a *keyboard* open followed by a click anywhere else closed the panel and pulled focus back to the Projects row, out of whatever the reader had just clicked. `interactedOutside` separates the two dismissals correctly, and it is reset on keyboard open so a stale `true` cannot suppress a later Escape. Verified locally: 115 sidebar tests, full web suite, lint and typecheck all green on the merged tree.

**Stale comment** — fixed in `52241cdf4`. `menu-items.test.tsx` was the last place still saying the flyout is "headed by the same label"; the reason for passing no tooltip there is unchanged, only the heading it referred to.

**Tab looping** — verified rather than assumed, and the disposition is right. Radix hands `FocusScope` `loop: true` unconditionally from `PopoverContent`, and the guard is `if (!loop && !trapped) return`, so the wrap runs even though `modal={false}` gives us `trapFocus: false`. Tab does cycle inside the panel.

Leaving it, with the reasoning stated so the next reader does not re-litigate it: `loop` is not exposed through `PopoverContent`, so changing it means editing the shared `components/ui/popover.tsx` and altering every popover in the app to fix one flyout. Escape is wired, returns focus to the row, and is covered by a test. Worth revisiting only if Tab-out is wanted app-wide.

**`aria-haspopup`** — agreed, and for the reason given. Activation opens nothing; `ArrowRight`/`ArrowDown` do. `aria-expanded` alone reports state without promising that Enter opens the panel, which is what actually happens.

PR #4835, opened 2026-09-19
fix(web): keep sidebar row classes off the client boundary

mrosberghaus:
## What the reader sees

The sidebar's chat skeleton — the placeholder rooms on the boot shell and on every workspace switch — drew each row as a bare block: the leading mark and the name stacked on two lines, both flush at the group's padding instead of the 16px mark axis and the 48px label column the nav rows directly above them use. The collapsed account chip was off its axis for the same reason.

| | mark x | label x | row height |
| --- | --- | --- | --- |
| Nav row (correct) | 16px | 48px | 32px |
| Skeleton row (before) | ~8px | ~8px, next line | ~44px, unstyled |
| Skeleton row (after) | 16px | 48px | 32px |

## Cause

`components/ui/sidebar.tsx` is a `"use client"` module. `sidebar-chat-list-skeleton.tsx` and `sidebar-deferred-account.tsx` are Server Components that imported its row-shape string constants.

Components cross the RSC boundary; plain values do not — the import resolves to a client reference, not the string. So `cn(SIDEBAR_ROW_CLASS, SIDEBAR_RAIL_SQUARE_CLASS, SIDEBAR_ROW_RAIL_PAD_CLASS)` returned `""` and the row lost its `flex`, its `h-8` and its `px-2`.

It failed silently: no error, no type error, and the real `OrganizationChatList` beside it is a client component, so the identical constants worked there. Only the two Server Components were affected.

## Change

- New `src/components/ui/sidebar-classes.ts` — no `"use client"` — owns the six shape constants (`SIDEBAR_ROW_CLASS`, `SIDEBAR_RAIL_SQUARE_CLASS`, `SIDEBAR_ROW_RAIL_PAD_CLASS`, `SIDEBAR_ROW_LABEL_CLASS`, `SIDEBAR_ROW_LABEL_INSET_CLASS`, `SIDEBAR_COLLAPSE_TRANSITION`), doc comments carried over intact.
- All 11 importers point at it directly. `sidebar.tsx` imports them like everyone else and no longer exports them, so a Server Component cannot pick the broken path back up.
- `sidebar-classes.test.ts` locks both halves: the module stays off the client side, and `sidebar.tsx` stays out of the re-export business.
- `apps/web/AGENTS.md` records the trap under Server Components First.

No behaviour change for the client components — same strings, same classes.

## Verification

- Browser, both sidebar states (CONTEXT.md, **Sidebar row**): expanded, all eight skeleton rows measure mark 16px / label 48px / 32px tall, matching the nav rows; collapsed, every mark and the account chip sit on the 28px rail axis.
- `pnpm --filter web test` — 807 files, 6699 passed, 1 skipped
- `pnpm check` and `pnpm typecheck` — clean


mrosberghaus:
## Grok review

Ad-hoc review only — requesting-code-review subagent dispatch was unavailable. Review followed `code-reviewer.md` in this process.

### Fixed
- Important: the client-boundary test was a no-op. First non-empty line of `sidebar-classes.ts` is `/**`, so `"use client"` after the doc comment (valid Next.js) would still pass; the re-export check only looked at the last `export {` block. Now strips comments (same helper as other web contract tests), scans every export list, and rejects `export const` / `export * from` (`8c971a4f7` — `apps/web/src/components/ui/sidebar-classes.test.ts`)

### Remaining
- Minor: `apps/web/src/components/ui/sidebar.tsx`: extra blank line left where the constants were deleted (after `SidebarRowSlot`)
- Skipped / disagreed: no render-test for `flex` / `h-11` / `px-2` on skeleton rows — Vitest does not emulate the RSC client-reference, so that would not have caught the original bug

### Assessment
The move to `sidebar-classes.ts` is the right fix; every importer points at it, and `sidebar.tsx` no longer re-exports the strings.

mrosberghaus:
Both review items handled.

**Client-boundary test (`8c971a4f7`)** — the catch is correct and I verified the new guard actually bites. My version checked the first non-empty *line*, which is `/**`, so a `"use client"` placed after the doc comment would have sailed through; and it only scanned the last `export {` block. Mutation-checked the replacement against exactly those two holes:

| Mutation | Result |
| --- | --- |
| `"use client";` inserted after the doc comment in `sidebar-classes.ts` | ✗ fails (my version passed) |
| `export { SIDEBAR_ROW_CLASS };` in a non-final export block in `sidebar.tsx` | ✗ fails (my version passed) |

Both restored after; the suite is green.

**Blank line (`9fd785deb`)** — fixed. The comment directly under it also said "the row shape **above**", which the move made false, so that now points at `sidebar-classes.ts`. Grepped the two files for other positional references; the remaining `above`/`below` are about sidebar rows, not code position.

**Render test for `flex` / `h-11` / `px-2`** — agreed, no action. Vitest doesn't emulate the RSC client-reference, so such a test would have rendered the constants as real strings and passed against the broken code. That's why the fix is verified in a browser instead: expanded, all eight skeleton rows measure mark 16px / label 48px / 32px tall against the nav rows; collapsed, every mark and the account chip sit on the 28px rail axis.

`pnpm --filter web test` 807 files / 6699 passed, `pnpm check` and `pnpm typecheck` clean.

PR #4727, opened 2026-09-17
fix(web): grow collapsed sidebar room marks to 24px

mrosberghaus:
## What

In the collapsed sidebar rail, DM avatar fallbacks like "MA", "NM" or "WM" touched the rim of their 20px circle while "AT" or "DE" sat fine. Inter's M and W run about 40% wider than A or E, and a circle's usable width at cap height is well short of its diameter, so no single font size fit every pair at 20px.

- Collapsed to icons, the DM face and the channel tile now measure **24px**. Type is unchanged: 9px on the DM face, 10px on the channel tile, which sits on a square and had room already. Every initials pair fits with even margin. The expanded list is unchanged.
- The row centres that mark in the collapsed button: padding drops to zero, the name goes `sr-only` (the link keeps its accessible name, the tooltip adds none), and the trailing spacer hides so it cannot push the mark off centre on touch.
- A group direct shows only its **first face** when collapsed. The 40px button could not hold a stack of 20px faces either (they were clipped); at 24px it cannot at all. The button tooltip already names everyone.

## Why 24px at 9px

Compared in an Inter mock at 4x: letter-spacing barely helps, 8px type fits every pair but makes "AT" or "IL" tiny, 24px at 10px has "WM" touching again. 24px at 9px holds every pair with even margin.

## Verification

- `pnpm --filter web test` on the three touched test files: 46 passed
- `pnpm check`, `pnpm typecheck`: clean (pre-commit)
- Tailwind order confirmed by compiling both utilities: `group-data-[collapsible=icon]:px-0!` emits after the sidebar button's `p-3!`, so it wins.
- Not verified in the running app in this worktree (local sign-in blocked by Turnstile); please check on the Vercel preview with the sidebar collapsed, including a 3-person group DM.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4837, opened 2026-09-19
fix(web): drop sidebar room marks to 20px

mrosberghaus:
## What

Every mark in the app sidebar goes from 24px to 20px: Direct faces, the Soko Bot nav face, and the Channel tile on the collapsed rail.

A Direct's face filled the 24px mark slot edge to edge while the Channel row beside it carried a 16px hairline glyph — 75% of a 32px row against 50% — so the Directs outweighed every other row in a mixed Pinned section.

| | Before | After |
| --- | --- | --- |
| DM face / Soko Bot face | 24px | 20px |
| Channel tile (rail) | 24px | 20px |
| Tile corner mark | 12px badge, 10px glyph | 10px badge, 8px glyph |
| Face overlap in a stack | 8px | 6px |
| Initials fallback type | 9px | 8px |
| Group stack width | 56px | 50px |

Unchanged: row height (32px / 44px on the sheet), the 24px mark slot, the 48px label column, the presence dot.

## Why these two knock-ons

- **Initials fallback to 8px.** 24px was bought in #4790 to hold a wide pair like `WM` at 9px without touching the rim. A pair is the rare case — most faces are photographs — so it takes the smaller type rather than every face taking a larger circle.
- **A stacked row pays 2px of leading padding.** The slot centres a mark narrower than itself, so a lone 20px face sits 2px in while a stack, wider than the slot, starts flush at its edge. Without the padding a group row's first face would sit 2px left of every 1:1 face under it. Pinned by a new test.

## Invariants kept

- #4790 (SOK-1107): one size in both states, so no mark resizes on the sidebar toggle.
- #4727: the rail's tile and face are still the same size as each other — they move together instead of the tile being left behind at 24px.

The loading skeleton mirrors the real marks, so it moved too.

## Verification

- `pnpm --filter web test` — 806 files, 6698 passed, 1 skipped
- `pnpm check` — clean
- `pnpm typecheck` — clean

## Screenshots

Reviewed against a before/after mock at true size and 3× (the sidebar's real numbers: 32px rows, 24px slot, 48px label column), including the `WM` initials-fallback case and the collapsed rail. Happy to attach stills if useful.


mrosberghaus:
## Grok review

### Fixed
- Important: group Direct `pl-0.5` stayed on the collapsed rail. Extra faces are `display: none` there, so the slot centred a 22px mark and the first face sat 1px off every 1:1 face and Channel tile (`9145505c5` — `apps/web/src/components/chat/direct-room-avatar-stack.tsx`)

### Remaining
- Minor: `apps/apple/Sokosumi/Chat/Sidebar/ConversationSidebarView.swift`: Apple overlap is still 8pt and the comment still says web `-ml-2`; web is `-ml-1.5` (6px)
- Minor: `apps/web/src/components/ui/sidebar.tsx`: `SidebarRowSlot` comment still calls the avatar/tile "24px"
- Skipped: CONTEXT.md Channel-tile sentence is a run-on — still true for the expanded list

Ad-hoc review only — requesting-code-review dispatcher was unavailable.

### Assessment
20px marks, 8px initials, and expanded stack padding match the PR; the rail pad is now dropped with the extra faces, and Apple overlap is the leftover platform split.

mrosberghaus:
Review items addressed in 6a4f7668f.

**Rail padding (already fixed in 9145505c5)** — verified the reasoning: extra faces are `display: none` on the rail, so the stack is a 20px face plus a 2px pad, and the slot centred that 22px mark at 19px against every 1:1 face's 18px. `group-data-[collapsible=icon]:pl-0` is right. The skeleton carries the same fix, which matters because it mirrors the real mark.

**Apple overlap** — fixed, 8pt → 6pt, and the doc comment now cites `-ml-1.5`. Apple's `faceSize` was already 20 (it never followed web up to 24px in #4790), so the overlap was the only number left out of step. `apps/apple/PARITY.md` says current web takes precedence, so following rather than recording the split. SwiftFormat lint and strict SwiftLint pass on the file.

**`SidebarRowSlot` comment** — fixed, and one more than reported: it listed "a 16px nav glyph, a 20px kind glyph, a 24px avatar or Channel tile". The kind glyph is `ChannelKindGlyph`, which has been `size-4` (16px) all along, so that one was wrong before this PR too. Now: 16px nav or kind glyph, 20px avatar or tile. It also says why the box stays 24px when nothing fills it — it is the column the label is measured from, so marks can change size without moving names.

**CONTEXT.md run-on** — taking this one rather than skipping it, since the run-on is mine. Split into three sentences, and it now records the rail dropping the stack's padding.

Verification: `pnpm check` clean, `pnpm typecheck` clean, `pnpm --filter web test src/components/chat src/app/(app)/components src/components/ui` → 103 files, 944 passed.

PR #4839, opened 2026-09-19
feat(web): tell the reader when this browser stopped receiving push

schaier-io:
Closes [SOK-929](https://linear.app/masumi/issue/SOK-929/tell-the-reader-when-this-browser-stops-receiving-push).

## User-facing impact

A browser can lose its Web Push subscription on its own: the push service expires the endpoint, or the browser drops it with its site data. `healPushSubscription` already repairs that on app open where it can. When it cannot, the browser just goes quiet, and the only place that says so is the account notifications page, which a reader who does not know push stopped has no reason to visit.

Now the notification bell panel and `/notifications` both say it, on the card that is already there for browser permission states, with one press that re-subscribes in place. No prompt: the notice only appears when the permission is already granted.

The notice is shown only for a browser Ably still holds a registration for, which is the same signal the repair reads. A reader who runs push on their phone alone is never told their laptop is broken.

## What changed

- `apps/web/src/lib/ably/push-repair-outcome.client.ts` (new): a small store holding `pending | quiet | healthy`, read entirely from this browser, so it costs no request.
- `apps/web/src/lib/ably/push-self-heal.client.ts`: records that outcome when the repair settles, either way, and reads the Ably registration **before** the repair runs.
- `apps/web/src/app/(app)/components/notification-browser-permission-primer.tsx`: new granted-but-quiet branch, subscribed with `useSyncExternalStore`, plus the restore press.
- `apps/web/messages/{en,de,es}.json`: four keys in `Components.NotificationCenter`.

## The subtle part

`activatePush` calls `client.push.deactivate()` halfway through its round, which clears Ably's stored device token. A repair that fails after that point therefore destroys the evidence that this browser ever had push. Reading the registration afterwards would classify the one reader this notice exists for as a reader who never turned push on, and a failed press on the card would make the card vanish as though it had worked. So the registration is read before the attempt and carried past it, in the repair and in the press. Both paths have a regression test.

## Verification

- `pnpm --filter web test` — 808 files, 6714 passed, 1 skipped.
- `pnpm typecheck` — 20/20 successful.
- `pnpm check` — clean.
- Mutation-checked: reverting the new guard fails the new store test; stubbing out the primer branch fails the new primer test.

## Known limits (not fixed here)

- A browser whose site data was cleared loses the registration together with the subscription, so it cannot be told apart from one that never turned push on.
- A subscription that dies mid-session is only noticed at the next app open.
- The bell itself carries no mark for this state, so the reader still has to open the panel. Worth a follow-up if we want the notice to reach further.

## Summary by CodeRabbit

* **New Features**
  * Added browser-specific notifications when push notifications stop delivering.
  * Added an option for signed-in users to restore push notifications, with progress feedback.
  * Added localized support in English, German, and Spanish.
  * Browsers with healthy, unsupported, or ineligible push setups no longer show unnecessary recovery prompts.
* **Bug Fixes**
  * Push recovery status now remains accurate after failed or successful restoration attempts.


schaier-io:
@coderabbitai

PR #4849, opened 2026-09-19
feat(apple): publish a notarized rolling DMG from main

mrosberghaus:
## What this gives you

A permanent public download link for the newest macOS build:

**`https://github.com/masumi-network/sokosumi/releases/download/apple-latest/Sokosumi.dmg`**

Every Apple change that reaches `main` replaces the single `apple-latest` prerelease behind that URL. The repo is public, so no GitHub account is needed to download.

## Why it has to be signed and notarized

The app is `ENABLE_APP_SANDBOX = YES` with `ENABLE_HARDENED_RUNTIME = YES`. An ad-hoc-signed build that someone *downloads* gets quarantined, and Gatekeeper refuses it — the right-click→Open bypass is gone on current macOS. So the job archives with a **Developer ID Application** certificate, exports via `apps/apple/ExportOptions.plist`, builds the image with `hdiutil` (no third-party packaging dependency), notarizes with `notarytool --wait` so a rejection fails the job rather than shipping, and staples the ticket.

## Design notes

- **`publish` needs `test` and `lint` green first.** A rolling download that ships a broken build is worse than no download.
- **One tag, rewritten.** `gh release upload --clobber` keeps the URL stable; recreating the release would mint a new URL and break every link already shared.
- **`Xcode release build` loses its `push` leg.** `publish` archives Release on the merge commit anyway, so keeping both compiled Release twice per Apple merge. It keeps the pull-request leg, which is where it actually blocks a break.
- **Not a required check.** `publish` never runs on a PR, so it could never report there — the same trap `apps/apple/docs/xcode-cloud-required-check.md` documents for Xcode Cloud.

## ⚠️ Blocked: this has never run

There is **no Developer ID Application certificate** yet — the machine has only two `Apple Development` identities. Until one exists, the archive, notarize and staple steps are unexercised. Merging this is safe (the job simply fails on `main` until the secrets exist), but it won't produce a DMG.

**Run `apps/apple/scripts/setup-release-signing.sh` first.** It walks through creating the certificate and an App Store Connect API key, then sets all five secrets and live-checks them against Apple's notary service. Written with the `wizard` skill, because the credential steps need an Apple Developer account role no agent can hold.

## Verification

| Check | Result |
| --- | --- |
| `actionlint` | clean apart from the pre-existing `xcode-27` runner-label warning |
| `plutil -lint ExportOptions.plist` | OK |
| `bash -n setup-release-signing.sh` | OK |
| `shellcheck -S warning` | only `SC2034` on the wizard template's own unused `RED` |
| Secret-name parity | all five `set_secret` names match the five `secrets.*` references exactly |
| Archive / notarize / staple | **not run** — no Developer ID certificate |

## Channel ownership

Settled: **Xcode Cloud stays, for App Store and TestFlight.** GitHub Actions owns the direct-download DMG. Both now build releases, so `apps/apple/AGENTS.md` records which owns which as a "do not blur them" rule — they are not interchangeable (different certificates, different review paths, and Actions cannot upload to App Store Connect the way Xcode Cloud does).


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/AGENTS.md`: still says the DMG is built with `hdiutil`; the job uses `diskutil image create from`.
- Minor: `apps/apple/AGENTS.md`: “every `xcodebuild` invocation” in CI is ad-hoc — true for test/lint/`Xcode release build`, false for `publish`.
- Minor: `apps/apple/AGENTS.md`: the three required gates are described as also running on path-filtered `main` pushes; `Xcode release build` is pull-request-only now.
- Minor: `apps/apple/README.md`: “every change to `main`” publishes; only `apps/apple/**` and `.github/workflows/apple.yml` do.
- Skipped: first `main` runs will fail until `setup-release-signing.sh` sets the five secrets — documented in the PR, and `Publish macOS DMG` is not a required check.
- Skipped: not signing the DMG itself — the app is self-contained; users drag it to Applications, so Gatekeeper path randomization on an unsigned image does not apply.

Ad-hoc review only — general-purpose reviewer subagent was unavailable.

### Assessment
Security split is right (no Developer ID on PR jobs), URL-stable `--clobber` release is right, `diskutil image create from` does accept a folder. The DMG still will not exist until the signing wizard has been run.

mrosberghaus:
Three of the four fixed in `aa384de0`. The fourth was already stale.

**Not actioned — `hdiutil` in `apps/apple/AGENTS.md`.** Already said `dmgbuild` at review time. The premise is stale too: `diskutil image create from` was replaced by dmgbuild in `a69ae7af`, because it **strips `.DS_Store`** when packaging a folder, and that file carries the window size, icon positions and view mode. Verified by repackaging a known-good styled volume with it — `.VolumeIcon.icns` and `.background.tiff` survived, `.DS_Store` did not. Two further consequences: Finder then sorts alphabetically, putting `Applications` to the *left* of the app so the drag reads backwards, and a staged `.VolumeIcon.icns` never takes, because a volume icon needs the Finder custom-icon flag set on the mounted volume rather than the file merely copied in.

**Fixed — ad-hoc signing claim.** The strongest of the four. It was not only false for `publish`, its "keep those flags when adding CI steps" would have stripped the signature off a downloadable build if anyone followed it there. Now states the split, with `Publish macOS DMG` named as an exception that has to stay one:

```
Xcode test          DEVELOPMENT_TEAM=  CODE_SIGN_IDENTITY=-
Xcode release build DEVELOPMENT_TEAM=  CODE_SIGN_IDENTITY=-
Publish macOS DMG   DEVELOPMENT_TEAM=$TEAM_ID  CODE_SIGN_STYLE=Manual
                    CODE_SIGN_IDENTITY="Developer ID Application"
```

**Fixed — gate triggers.** The three no longer share one. `Xcode test` and `Swift lint and format` run on Apple PRs, dispatch and path-filtered `main` pushes; `Xcode release build` is pull-requests-only since the signing boundary landed. Now spelled out per job.

**Fixed — README.** Only `apps/apple/**` and the workflow itself publish, per `on.push.paths`.

**Skipped items:** agreed on not signing the disk image — it is notarized and stapled, and the app inside is Developer ID signed. The "first `main` runs will fail" note is now moot: all five secrets are set, and `Publish macOS DMG` has completed end to end on dispatch (archive 11.9 min, notarize 32 min, 24 MB artifact, no release touched).

PR #4868, opened 2026-09-19
feat(web): fit jobs panes to content width, not the window

mrosberghaus:
## Why

The jobs list + detail two-pane layout switched on a `lg:` viewport media query, but the panes live in the window **minus the sidebar** (14rem expanded, 3.5rem as the rail). At a 1024px window with the sidebar open:

```
1024 − 224 (sidebar) − 288 (list) − 64 (gutters) − 16 (gap) = 432px detail pane
```

432px is phone width — narrower than the modal the breakpoint replaced. The `lg:` threshold promised "two panes fit" and delivered that at its own edge.

## What changed

**Two-pane fit is now a content-width decision**

- `jobs/layout.tsx` — the panes row is its own named query container (`@container/jobs-panes`); every `lg:` variant became `@4xl/jobs-panes:` (56rem of row, not 64rem of window). Containment starts at `md:` because `container-type` implies `contain: layout`, which would otherwise make the `md:hidden` fixed jobs header position against the row instead of the viewport.
- `use-jobs-two-pane-fit.ts` (new) — a `ResizeObserver` on that same row, for the two decisions CSS cannot make: auto-selecting the first job into the right pane, and rendering the modal only when the right pane cannot show. Both `matchMedia("(min-width: 1024px)")` / `"(max-width: 1023px)"` calls are gone.

**Sidebar shows as the rail on a narrow window**

- Below a 1024px window the sidebar is the rail whatever `sidebar_state` says. Opening it there is session-local and never writes the cookie, so a detour through a narrow window does not leave a wide one collapsed; widening restores the stored preference.
- `SIDEBAR_BOOT_SCRIPT` applies the same rule pre-paint, so there is no expanded-then-collapse flash on a tablet.

## User-facing impact

- Jobs detail no longer opens in a cramped pane on tablets, small laptop windows, or half-snapped desktop windows. It stays a full-width modal until there is real room, and a collapsed sidebar now earns two panes at a window width an expanded one does not.
- The sidebar stops eating 14rem of a narrow window, without touching the preference that applies on a large one.

## Verification

- `pnpm --filter web test` — 810 files, 6732 passed, 1 skipped

New coverage: a contract test locks `JOBS_TWO_PANE_MIN_WIDTH` to the `@4xl/jobs-panes` variant (separate mechanisms that can silently disagree), the redirect/modal tests now assert the sidebar-squeezed case that used to pass, and four sidebar cases cover the compact window — including that the cookie stays `false` after opening the sidebar on a narrow one.

## Reviewer notes

- **No browser proof yet.** The jobs page is auth-gated and sign-in from this worktree fails against the worktree DB, so this is verified at the unit level only. Worth a look at the 1024–1280 band with the sidebar both expanded and collapsed.
- **Left alone deliberately:** `@modal/[jobId]/loading.tsx` still uses `lg:hidden`. It renders outside the query container so it cannot read the row width, and it only affects a loading overlay for a few hundred ms.
- No schema changes, no migrations, no translation keys touched.


mrosberghaus:
## Grok review

### Fixed
- Critical: `@4xl/jobs-panes:flex-row` sat on the `@container/jobs-panes` node. An element cannot query its own containment context, so the panes would stay `flex-col` after the list/detail slots had already switched. Row styles now live on a descendant; the outer box stays the named container the hook measures (`ef3bfdcea` — `apps/web/src/app/(app)/agents/[agentId]/jobs/layout.tsx`).
- Important: `jobs-list` still used viewport `lg:w-72` / `lg:border-r`, so a 1024px window with a stacked row pinned the list to 18rem. Those utilities now follow `@4xl/jobs-panes` (`ef3bfdcea` — `apps/web/src/app/(app)/agents/[agentId]/jobs/components/jobs-list.tsx`).

### Remaining
- Minor: `apps/web/src/app/(app)/agents/[agentId]/jobs/@modal/[jobId]/loading.tsx` still uses `lg:hidden`. Author left this: the overlay sits outside the query container and only covers a few hundred ms.
- Minor: `JobLayoutSkeleton` still uses `lg:h-[calc(100dvh-6rem)]` (Suspense fallback, not the panes row).
- Minor: boot script keys compact off `window.innerWidth`; React uses `matchMedia('(max-width: 1023px)')`. Those can disagree by a scrollbar at the 1024 edge.
- Skipped: no browser proof in the 1024–1280 band — author already flagged sign-in against this worktree DB.

### Assessment
Two-pane fit is now a real content-width decision (CSS on a descendant, JS on the same box); leftover `lg:` is the modal loading overlay and the layout skeleton.

mrosberghaus:
Worked the three remaining notes. One was a real bug; two I'm leaving, with reasoning.

### Fixed: boot script / React disagreement (`6c49965f4`)

Correct that the two could disagree at the edge — though I'd put the mechanism differently. Media Queries Level 4 defines the `width` feature as including the scrollbar, same as `window.innerWidth`, so a classic scrollbar shouldn't split them. What can is rounding: `innerWidth` is an integer, and a fractional viewport width (browser zoom, fractional device pixel ratio) rounds to one side while `(max-width: 1023px)` evaluates the real value on the other.

Either way the mechanism doesn't need settling, because the fix removes the question. Both sides now evaluate one exported string:

```ts
export const SIDEBAR_COMPACT_MEDIA_QUERY = `(max-width: ${SIDEBAR_COMPACT_BREAKPOINT - 1}px)`;
```

The boot script interpolates it; `useIsSidebarCompact` subscribes to it. They can't land on opposite sides of a breakpoint they no longer compute separately. The contract test asserts the script carries that exact string and no longer mentions `innerWidth` — an *equivalent* query would reintroduce exactly the drift this removes.

### Not fixing: the two leftover `lg:`

Both are outside the query container, not oversights inside it. A container query only styles **descendants** of the container, so neither can be converted:

- `@modal/[jobId]/loading.tsx` renders into the `@modal` slot, a sibling of the panes row.
- `JobLayoutSkeleton` is the Suspense fallback that renders *instead of* the container — at that point the container doesn't exist yet.

Converting either needs JS measurement (`useJobsTwoPaneFit`) in a component whose whole job is to paint instantly while something loads. Not worth it for a few hundred ms of loading state.

`header.tsx`'s `lg:gap-6 xl:gap-8` stays too — a spacing ramp, not a pane-fit decision.

### Still open: no browser proof

Standing. The preprod preview uses a real Turnstile key, so the sign-in form shows a human check I won't answer. [#4872](https://github.com/masumi-network/sokosumi/pull/4872) makes `doctor` report `turnstile=live` so the next agent hitting this gets told why instead of guessing — but it doesn't unblock this PR. The 1024–1280 band is unverified in a browser, and given that the container-query bug in this PR was invisible to source-level unit tests, that gap is worth closing before merge rather than after.

### Verification

- `pnpm --filter web test` — 810 files, 6735 passed, 1 skipped
- `pnpm check` clean; `pnpm typecheck` 18/18

PR #4872, opened 2026-09-19
feat(dx): use turnstile test keys locally and report the key in doctor

mrosberghaus:
Follow-up to [#4871](https://github.com/masumi-network/sokosumi/pull/4871), which documented Cloudflare's Turnstile test keys in the `.env.example` files. Documenting them still left every new checkout to paste them in by hand, and left an agent that hit a challenge with no way to tell why.

## 1. `env:bootstrap` sets them

`scripts/local-env/bootstrap.mjs` now turns each commented-out Turnstile key into Cloudflare's always-passes value. Two guards keep it honest:

- **Only the key's own commented assignment is rewritten**, so `apps/web/.env` gets the site key and `apps/core/.env` gets the secret — the web example *names* `TURNSTILE_SECRET_KEY` in prose, and that must not become an assignment web never reads.
- **Only when nothing has set it.** A real key someone put there on purpose survives untouched.

The value goes on its own line under a comment rather than as a trailing `KEY=value # note`, since not every dotenv parser strips those.

## 2. `verify-sokosumi doctor` reports it

New line: `turnstile=off|test|live`.

This closes a real blind spot rather than adding a nicety. `fixture_auth` calls Core's `POST /auth/sign-in/email` **directly**, so it reports `ok` while the browser sign-in form is still blocked behind a captcha — the harness would tell an agent auth works while the UI is unreachable. `turnstile=` is what tells those two states apart, and `live` warns explicitly that the challenge is not something an agent should answer.

This is not hypothetical: it is exactly how a verification attempt failed on a Vercel preview earlier today. Credentials filled correctly, submit no-op, and the only clue was an unchecked Cloudflare checkbox in the snapshot.

## 3. Docs

`features/sign-in.md` gains a precondition and a gotcha (with the symptom to recognise), `features/sign-up.md` the same gotcha, and `SKILL.md` lists `turnstile=` alongside `fixture_auth=`.

## Verification

- `node --test scripts/local-env/bootstrap.test.mjs` — 22 passed, 0 failed (5 new cases: each file gets only its own key, an already-set key survives, the note stays off the value line)
- End-to-end against the **real committed** `.env.example` files in a throwaway root:

```
web  NEXT_PUBLIC_TURNSTILE_SITE_KEY = 1x00000000000000000000AA
core TURNSTILE_SECRET_KEY           = 1x0000000000000000000000000000000AA
core leaked a site key?             = false
web leaked the secret?              = false
```

- Every `turnstile_status` branch exercised against throwaway `.env` files — no `.env`, commented, empty, `1x`, `2x`, `3x`, real key, quoted real key — each returning the right classification.
- `bash -n` and `shellcheck -S warning` clean on the changed function (pre-existing `SC2209` warnings elsewhere in the file are untouched).
- `pnpm check` clean; `pnpm typecheck` 18/18.

## Note

The always-passes secret makes Cloudflare's `siteverify` succeed for **any** token. It is written only to a local `.env`, never defaulted in `env.ts` — see #4871 for why that distinction matters.


mrosberghaus:
## Grok review

### Fixed
- Critical: writing Cloudflare's 1x secret turns Core captcha on; dummy `siteverify` omits `action`, so `expectedAction: "auth"` 403s every local UI token (`fef9b1887` — `apps/core/src/lib/auth-captcha.ts`)
- Critical: harness `POST /auth/sign-in/email` sent no `x-captcha-response`, so `fixture_auth` and cookie sign-in 400 once the secret is set (`fef9b1887` — `.cursor/skills/verify-sokosumi/bin/verify-sokosumi`)
- Important: doctor told agents to `pnpm env:bootstrap` to clear a live key; bootstrap does not overwrite a set key (`fef9b1887` — doctor warn + sign-in/sign-up recipes)

### Remaining
- Minor: 2x/3x still report `turnstile=test` (now with a warn). Recipes require `turnstile=test (always passes)` or `off`.
- Minor: doctor reads only `apps/web/.env`. A live Core secret with a 1x site key is a different failure (API 403), not the widget.
- Skipped / disagreed: none.

### Assessment
Dummy 1x keys now actually pass local UI and harness auth; live keys still need a manual `.env` swap.

mrosberghaus:
Both remaining notes worked, plus the gap the critical fix exposed. Pushed as `d59722430`.

First, confirming the regression was mine and the diagnosis is right. `createAuthCaptchaPlugin` returns `{ id: "captcha-disabled" }` when the secret is falsy, so writing the 1x secret flipped Core's captcha **on** for every local checkout. I reasoned about the widget half of the pair and never about the server half — enabling verification was the actual effect of that line.

### Minor 1: 2x/3x reporting as `turnstile=test`

Real problem for the intended consumer. An agent matching on the value read `test` and proceeded; the prefix in the parenthetical is prose. Split into a machine-readable value:

```
turnstile_site=off | test-pass | test-block | test-interactive | live
```

Recipes now check `turnstile_site=test-pass` or `off` — string equality, nothing to parse out of a sentence.

### Minor 2: doctor reads only `apps/web/.env`

I'd rate this above minor now, because the critical fix is what makes it reachable. A live Core secret under a 1x site key produces a widget that solves itself and then an API 403 — which reads as a broken app, not a misconfiguration, and is the single hardest state here to diagnose from the outside. Doctor now reads both files:

```
turnstile_site=test-pass
turnstile_secret=live
warn: site key passes but TURNSTILE_SECRET_KEY in apps/core/.env is live — Core will reject the dummy token (API 403), not the widget.
```

### Also: guard the dummy secret in deployed environments

Not in the review, but it follows from the fix. `validateEnv()` warns when `TURNSTILE_SECRET_KEY` is **unset** in a deployed environment. It said nothing when the value is Cloudflare's always-passes secret — which is the quieter and worse case, since siteverify then accepts every token including forged ones, and the endpoints look protected. Before this PR that was unlikely; now every local checkout carries that exact string, which is how it ends up pasted into a deployment.

`TURNSTILE_ALWAYS_PASS_SECRET` moved to `@sokosumi/utils` beside `AUTH_CAPTCHA_ACTION`, so `env.ts` can compare against it without importing `better-auth` into environment validation.

It warns rather than fails, matching the existing unset case. Say the word if you'd rather it `process.exit(1)` — I'd support that, since unlike unset it is actively misleading.

### Verification

- `pnpm --filter web test` — 810 files, 6735 passed, 1 skipped
- `pnpm --filter @sokosumi/core test` — 532 files, 5868 passed, 39 skipped (3 new env cases)
- `pnpm --filter @sokosumi/utils test` — 70 files, 678 passed
- `node --test scripts/local-env/bootstrap.test.mjs` — 23 passed
- `pnpm check` clean; `pnpm typecheck` 18/18
- Every site-key × secret-key combination exercised against throwaway `.env` pairs: both test-pass, no files, site-pass/secret-off, site-pass/secret-live, both live, 2x, 3x, and a quoted value.

Branch also carries the `main` merge that brought in #4868.

mrosberghaus:
Production now refuses to boot on the always-passes secret; previews keep the warning (`48a90b35f`).

```
NODE_ENV=production / VERCEL_ENV=production  → console.error + process.exit(1)
VERCEL_ENV=preview                           → warn, boots
local                                        → silent (expected setup)
```

Split out as `isProductionEnvironment()` beside the existing `isDeployedEnvironment()`, rather than making the whole deployed set fatal. Previews are the one deployment where a test key is defensible — it is what would let an agent drive the sign-in form without answering a human check — so failing there would close that door along with the hole.

An unset secret keeps its warning in both. The two cases are not the same severity: unset is honestly off and an operator can reason about it, while this one serves a captcha that passes everything. That asymmetry is why only this case is fatal.

`apps/core/.env.example` now spells out the three outcomes, so nobody has to read `env.ts` to find out production will not start.

### Verification

- `pnpm --filter @sokosumi/core test` — 532 files, 5869 passed, 39 skipped
- `pnpm --filter web test` — 810 files, 6735 passed, 1 skipped
- `pnpm --filter @sokosumi/utils test` — 70 files, 678 passed
- `node --test scripts/local-env/bootstrap.test.mjs` — 23 passed
- `pnpm check` clean; `pnpm typecheck` 18/18

New cases cover all four environments: production and Vercel production both exit 1 (via the existing `expectInvalidEnvironment` helper), preview warns and still returns the secret, local stays silent.

PR #4873, opened 2026-09-19
refactor(database): consume the package from source instead of built declarations

mrosberghaus:
Implements steps 1–3 of [ADR 0035](https://github.com/masumi-network/sokosumi/blob/main/docs/adr/0035-database-consumed-from-source.md). Step 4 (folding the package into `apps/core`) is deliberately not part of this.

Linear: [SOK-1126](https://linear.app/masumi/issue/SOK-1126)

## User-facing impact

**Two live bugs were hiding behind the type hole and are fixed here.** Prisma rejects a bare `null` for a nullable `Json?` column — clearing one requires `Prisma.DbNull` — but both of these wrote `null` and failed at query time:

1. **Deleting a chat message.** The soft-delete tombstone wrote `metadata: null`.
2. **Clearing a coworker's metadata.** `PATCH /v1/coworkers/{id}` with `metadata: null` did the same.

Neither was caught because the declaration emit had already widened the relevant Prisma types to `any`.

Everything else in this PR is a type-level change with no runtime effect.

## Why

`@sokosumi/database` compiled itself with `tsc` and handed Core the emitted `.d.ts`. Declaration emit widens **18** of Prisma's runtime value re-exports to `any` — `Decimal`, `DbNull`, `JsonNull`, `AnyNull`, `Sql`, `empty`, `join`, and the five `PrismaClient*Error` classes:

```
source   prismaNamespaceBrowser.ts:23   export const Decimal = runtime.Decimal
emitted  prismaNamespaceBrowser.d.ts:4  export declare const Decimal: any;
```

Because Prisma defines `NullableJsonNullValueInput` and its siblings in terms of `typeof JsonNull`, those unions collapsed to `any`, and **every JSON-column write in Core was typechecking against `any`**.

## What changed

All five entry points move to source in one commit — mapping only the root while the other four still resolve to `dist` was measured at 553 errors, because it puts two instantiations of the Prisma types in one program. There is no valid intermediate state.

The package's `build` and `clean` scripts, its `dist`, and `pnpm database:build` are gone. `prisma:generate` is untouched, and the package keeps its own `typecheck`. Core's `vercel-build` drops the `tsc` step.

Core's `tsup` needs both halves together:

```ts
noExternal: ["@sokosumi/database"],
external: ["pg", "@prisma/client", "@prisma/adapter-pg"],
```

`noExternal` alone inlines `pg`, which is CommonJS, and the bundle dies at import with `Dynamic require of "events" is not supported`. Keeping the drivers external also stops the 4.37 MB wasm query compiler being inlined as base64.

### The 31 exposed errors, fixed at the type level

No `as any`, no double casts — a cast here would reinstate exactly the blindness this removes.

- JSON metadata producers (`mergeChatRoomMessageMetadata`, `thoughtMetadataFields`, the assistant metadata builder, `membershipMetadata`, the mention-stream payload) now declare `Prisma.InputJsonObject` / `Record<string, InputJsonValue | null>`. Tightening the first one alone cleared 16 cascading errors.
- `calculateAverageExecutionTime` typed its raw-query column as `typeof Prisma.Decimal` — the *constructor*, not an instance — so `.toNumber()` resolved against `any`. Now `PrismaRaw.Decimal`, the server namespace that exists for raw SQL.
- `sanitizePersistedValue` returns JSON on every branch but declared `unknown`. It and `RuntimeEvent.data` now say so, so the Soko Bot event log typechecks against Prisma with no conversion inserted.
- Shapes persisted verbatim into JSON columns (`PersistedChatUi*Part`, `ScenarioCheck`) become object **type aliases**. Only an alias carries TypeScript's implicit index signature, which is what makes it checkable against `Prisma.InputJsonValue`; an interface silently fails that check. Root `AGENTS.md` records this carve-out to the "prefer interfaces" rule so it is not "fixed" back.
- Core's two copies of a private `jsonInput` are extracted to `@/helpers/prisma-json`, alongside a new `nullableJsonInput`.

### Docs

`packages/database/AGENTS.md`, `apps/core/AGENTS.md`, root `AGENTS.md`, and [ADR 0008](https://github.com/masumi-network/sokosumi/blob/main/docs/adr/0008-turbo-task-runner-on-pnpm.md) all described the old build; updated in this change.

## Verification

- `pnpm check` and `pnpm typecheck` — green
- Core **5862** tests, web **6720**, packages **1734**, CI scripts **67** — all pass
- `tsc --listFiles` resolves **zero** files from `packages/database/dist`
- A probe bundled with the production `tsup` settings returns `P1001` on `prisma.user.count()` and `P2010` on `` $queryRaw`SELECT 1` `` against a dead port. Both prove the wasm query compiler loaded and compiled SQL before the socket failed — a wasm that failed to load gives an engine-initialisation error instead, which is the failure this check exists to catch.

## Review notes

**One test changed.** `delete.test.ts` asserted the mock received `metadata: null` — the bug it now asserts against. No other test needed touching, which is the evidence that no behaviour moved.

**Ongoing cost, accepted deliberately.** Core's program grows by ~129 files from the package's source, and `skipLibCheck` no longer shields them. They compile clean under Core's tsconfig today, but a future `prisma generate` emitting source Core's config rejects would block Core's typecheck rather than being absorbed into a `.d.ts`. That is the price of the checking.

Unrelated: `apps/cli/test/tui/status-app.test.ts` hangs the root `pnpm test` and does not terminate on its own. It predates this branch and `apps/cli` has no dependency on `@sokosumi/database`. Filed as [SOK-1127](https://linear.app/masumi/issue/SOK-1127).


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/README.md` still says Core `vercel-build` runs `@sokosumi/database` `build` (`tsc`) before tsup.
- Minor: `packages/database/README.md` still documents `pnpm run build` / `clean` and “generate then rebuild”.
- Skipped / disagreed: turbo `^prisma:generate` after removing database `build` — Test Core logs still run `@sokosumi/database:prisma:generate` from the stub `build` DAG (`dependsOn: ["^build", "prisma:generate"]`).
- Ad-hoc review only — general-purpose reviewer subagent dispatch was unavailable.

### Assessment
Source consumption, `Prisma.DbNull` on message delete / coworker metadata clear, and Core `tsup` `noExternal` + driver `external` match ADR 0035; leftover READMEs still describe the old `tsc` step.

mrosberghaus:
Both Minor findings were correct — I under-scoped the docs. SOK-1126 named four files to update and I updated exactly those, which missed the two READMEs.

**Fixed** (`ea9a79aeb`, plus the merge):
- `apps/core/README.md` — the `vercel-build` steps no longer list the database `tsc`; step 1 is now `prisma:generate` alone and step 2 notes that `tsup` inlines the package from source.
- `packages/database/README.md` — dropped `pnpm run build` / `clean`, the `dist/` entry in the tree diagram, the "No Output Files Generated" section, and "regenerate then rebuild". The Building and Troubleshooting sections now say there is no build and point at ADR 0035.

A repo-wide sweep for `database:build`, `@sokosumi/database run build`, `clean`, and `database/dist` now returns only ADR 0034/0035's own historical prose, which is meant to read in past tense.

**On the turbo `^prisma:generate` point — you were right to skip it, and I verified it rather than taking it on trust.** Deleted `packages/database/src/generated` entirely and ran `pnpm typecheck`:

```
@sokosumi/database:prisma:generate: cache bypass, force executing
@sokosumi/database:prisma:generate: ✔ Generated Prisma Client (7.10.0) to ./src/generated/prisma
 Tasks:    17 successful, 17 total
```

The edge survives because the now-script-less `build` task still exists as a node in the DAG, and `build` itself declares `dependsOn: ["^build", "prisma:generate"]` — so `^build` from Core still reaches database's `prisma:generate` through the stub. Generation works from a cold checkout. No change needed.

One thing that came out of the merge rather than the review: #4869 landed `.cursor/rules/prisma-json-columns.mdc` while this branch was open, which is the canonical home for this guidance. The interface-vs-type-alias exception I'd put in `AGENTS.md` moved into that rule file, with `AGENTS.md` keeping a one-line pointer like the other rules.

Re-verified after merging `main` in twice: `pnpm check` and `pnpm typecheck` green, Core **5869** tests, web **6735**, CI scripts **67** — all pass.

PR #4882, opened 2026-09-20
fix(web): opt blocking routes out of instant navigation validation

mrosberghaus:
## What

`/signin` logged this on every request in `next dev`:

```
Error: Route "/signin": Next.js encountered URL data during prerendering or a navigation.
Learn more: https://nextjs.org/docs/messages/instant-shell-url-data
```

…even though `(auth)/layout.tsx` already declares `export const instant = false`.

## Why the layout's opt-out didn't cover it

Next reads `instant` per segment, in two independent places:

| | Walks up to parents? | Effect here |
| --- | --- | --- |
| `isPageAllowedToBlock` | yes | the layout's opt-out **did** apply — these routes were already blocking, build was fine |
| `anySegmentNeedsInstantValidation` | no — keeps walking past an `instant === false` segment | the page has no explicit config, so it gets *implicit* validation and logs |

That split is deliberate, not a bug. It's what lets a blocking layout host instant pages — the [docs](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config/instant#disabling-instant) show `instant = false` on the layout with `instant = true` on the page beneath. So the opt-out has to be repeated on each page that should block.

## Change

`export const instant = false;` on the 27 pages that read `params`/`searchParams` in the page body **and are meant to block**:

- **`(auth)` (5), `(app)/admin` (12), personal-assistant OAuth return hop** — their layouts already declare it. Auth entry must stay fully dynamic (POST/RSC to `/signin`|`/signup` is served 405 off a prerendered shell, per the comment in `(auth)/layout.tsx`); admin keeps its must-block gates.
- **`join/[token]`, `accept-invitation/[id]`, `share/[token]`, `share/jobs/[token]`** — reached by an external link, never by in-app navigation, and the whole page body is the token-resolved result. A `<Suspense>` boundary would leave an empty shell, which defeats the point. `share/jobs/[token]` is a bare `redirect()`.

## User-facing impact

None. Every touched route still builds as `◐ (Partial Prerender)` — `instant = false` silences the validator without disabling prerendering. This removes dev-log noise on routes that were already, deliberately, blocking.

## Deliberately not fixed here

- **29 routes still report the insight** (tasks, projects, agents, chat, calendar, connections, organizations, developer, billing, history, welcome, personal-assistant). Those are real in-app destinations where an App Shell pays off, so they want the `<Suspense>` treatment — see `apps/web/.claude/skills/next-partial-prefetching-adoption` step 5. Left visible on purpose as the worklist.
- **`share/[token]` will keep logging a second insight** (E1429): it reads `params` in `generateMetadata()` for the OG title. The only fixes are static metadata or a `connection()` marker, and a share link's title genuinely has to come from the token.
- **Pre-existing `blocking-prerender-crypto`** on `/join/[token]`, `/accept-invitation/[id]` and `/share/[token]`: `crypto.randomUUID()` in `src/lib/clients/utils/core-request-id.ts:16`, hit by every `coreClient` call. A/B'd — it fires with and without this change, so it's untouched here and worth its own issue.

## Inventory correction

The first pass counted routes with `grep "await searchParams|await params"`, which misses
`await Promise.all([params, searchParams])` and a `searchParams` promise handed to a helper
(`await getRedirectQueryString(searchParams)`). Five blocking pages were missed that way and
are now included; scanning for the prop declaration instead finds **61** route files that read
URL data — 27 opted out here, 5 already behind `<Suspense>`, 29 remaining.

## Verification

- **A/B on `/signin`** in `next dev`: commenting the export out brought E1439 straight back; restoring it → 0 error lines. Same clean result on `/signup`, `/forgot-password`, `/reset-password` and all four token routes (all 200).
- `pnpm --filter web check` — clean (2316 files)
- `pnpm web:build` — exit 0, every touched route `◐`


mrosberghaus:
## Grok review

Ad-hoc review only — general-purpose reviewer subagent was unavailable (no Task dispatcher in this child). Skills were loaded from the repo; review ran in this worktree against `origin/main` merge-base `a33af282a`.

### Fixed
- Important: three admin Soko Bot pages still awaited `params`/`searchParams` with no page-level `instant = false` (the `[id]/advanced` sibling already had it). Layout opt-out does not stop per-page URL-data validation. (`8ec40cff3` — `apps/web/src/app/(app)/admin/soko-bots/[id]/page.tsx`, `versions/[slug]/page.tsx`, `versions/new/page.tsx`)
- Important: `/login` and `/register` await `searchParams` then `redirect()` (same shape as `share/jobs/[token]`). (`8ec40cff3` — `apps/web/src/app/(auth)/login/page.tsx`, `register/page.tsx`)

### Remaining
- Minor: the same two-line comment is copy-pasted on every opted-out page; the export is the signal.
- Skipped / disagreed: 24 in-app destinations (tasks, projects, agents, chat, …) still log E1439 — PR's documented `<Suspense>` worklist, not this change.
- Skipped / disagreed: `share/[token]` `generateMetadata` E1429 and token-route `blocking-prerender-crypto` — pre-existing, called out in the PR body.
- Skipped / disagreed: `admin/enterprise-contracts/page.tsx` — already reads `searchParams` behind `<Suspense>`, correctly omitted.

### Assessment
The original 22 opt-outs match Next's per-segment `instant` rule (layout `false` does not cover the page for URL-data validation); the five missed blocking pages now match that same rule, and leftover in-app insights are the documented worklist.

mrosberghaus:
Verified both fixes and they're correct — the miss was a flaw in how I built the inventory, not just five stragglers.

I scanned with `grep "await searchParams\|await params"`, which cannot see either shape you found:

- `login/page.tsx` — `await getRedirectQueryString(searchParams)`, the promise handed to a helper
- `admin/soko-bots/[id]/page.tsx` — `await Promise.all([params, searchParams])`

Re-scanning on the **prop declaration** instead of the await spelling changes the numbers: **61** route files read URL data, not 50. 27 are opted out (22 + your 5), 5 sit behind `<Suspense>`, **29 remain** — not the 24 the PR body claimed. Five were invisible to the old grep and are still unfixed: `projects/[projectId]/page.tsx`, `personal-assistant/page.tsx`, `history/page.tsx`, `(welcome)/page.tsx`, `billing/page.tsx`. PR body updated with the corrected counts and the reason.

**Minor (comment noise)** — agreed, removed in `4c8e8907e`. 27 pages now carry a bare `export const instant = false;`, matching the pre-existing `admin/soko-bots/lab/page.tsx`. #4881 had just done the same to packages; the per-segment rationale stays in the commit that introduced the opt-outs.

**`admin/enterprise-contracts/page.tsx`** — you're right, correctly omitted. It passes `props.searchParams` into a Suspense-wrapped `EnterpriseContractsContent`. My re-scan initially flagged it because its page export is `async`; the structure is fine.

The other three skips I agree with as scoped.

One correction to the PR body that came out of this, unrelated to your review: I'd written that the remaining routes need "an `instant()` regression test per step 5." That's wrong — step 5 is a dev-only sweep gated on typecheck; `instant()` tests belong to the preservation path and the optimizer skill, and this repo has no Playwright at all. Removed that claim.

PR #4884, opened 2026-09-20
fix(core): load the sync router in a hook, not the first test

mrosberghaus:
`pnpm test` at the repo root fails on `main`. One Core test times out:

```
× returns 401 for missing cron auth   5013ms
✓ returns 401 for invalid cron auth    545ms
✓ returns 409 when lock is already held  1ms
```

## Cause

`createApp()` did `await import("./index")`, which pulls in Core's whole module graph. The file never calls `vi.resetModules()`, so that import is **cached** — only the *first* test pays for it, inside its own 5s budget. Every later test reuses the module and runs in ~1ms.

Measured on an idle machine: first test **1532ms**, rest 0–1ms. Under full-suite worker contention it crosses 5s and fails. The same run reports `1606 modules evaluated 12247 times · 499.38s, 86% of tracked time`, which is the contention that pushes it over.

This surfaced only recently because it was hidden: until #4875 the CLI TUI suite hung the root `pnpm test` outright, so nobody got far enough to see a Core failure. CI never saw it either — `Test Core` runs Core alone, without the cross-workspace contention.

## Fix

Load the router in `beforeAll`, so the cost lands in setup rather than racing a per-test timeout. `createApp()` no longer needs to be `async`.

| | before | after |
|---|---|---|
| first test | 1532ms | **3ms** |
| `index.test.ts` | 41 pass | 41 pass |
| root `pnpm test` | exit 1, 135s | **exit 0, 91s** |

## Scope

**Not applied to the other Core files that re-import inside tests** — `lib/auth.test.ts`, `clients/stripe.client.test.ts`, `routes/v1/index.test.ts` and friends all call `vi.resetModules()`, so their re-imports are the point and hoisting would break them. `sync/index.test.ts` was the one file where the import was incidental rather than intentional, which is exactly why it was the one that flaked.

**Not a timeout bump.** Raising `testTimeout` hides the cost instead of moving it, and leaves the next person to rediscover this. `isolate: false` (which vitest itself suggests, ~23s faster) is a suite-wide performance change that shouldn't ride along on a flake fix — worth considering separately.

## Verification

- `pnpm check` + `pnpm typecheck` green; hook ran on the commit, no bypass
- `apps/core` `src/routes/sync/index.test.ts` — 41 pass
- root `pnpm test` — 17/17 tasks, exit 0


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/src/routes/sync/index.test.ts` — the six-line comment above `beforeAll` is flake history. One line is enough (`import once in beforeAll so the graph load is hookTimeout, not testTimeout`).
- Skipped: static `import syncRouter from "./index"` (vi.mock is already hoisted) would drop `beforeAll`/`let` entirely. Not a bug; the hook is the stated fix and it works because Vitest's default `hookTimeout` is 10s vs `testTimeout` 5s.

### Assessment
Ready to merge: Yes
The timeout moved from the first `it` (5s) into `beforeAll` (10s). Same cached module as before; no `resetModules()` here, so sharing the router is correct. Other Core files that re-import after `vi.resetModules()` were rightly left alone.

Ad-hoc review only — nested general-purpose dispatch was unavailable. Install: npx skills add obra/superpowers -s requesting-code-review -y.

mrosberghaus:
Both items addressed, but the second one changed the outcome — the skipped suggestion is not viable, and finding that out is what the comment now records.

**On the static import.** Marked "Skipped: not a bug", but it would have been strictly simpler than the hook, so I tried it rather than take the skip. It fails:

```
ReferenceError: Cannot access 'LOCK_TIMEOUT_MS' before initialization
 ❯ getEnv src/routes/sync/index.test.ts:63
 ❯ getSokoBotRuntime src/lib/soko-bot/factory.ts:10
 ❯ new SokoBotControlPlane src/services/soko-bot-control-plane.service.ts:554
 ❯ src/services/soko-bot-events-sync.service.ts:4
```

`vi.mock` is hoisted, but the module-level consts its factory closes over are not. The router's graph calls the mocked `getEnv` at module scope, so a static import evaluates it inside the temporal dead zone. The deferred import is load-bearing — presumably why it was written that way originally, and the reason this is a hook rather than a plain import.

So the framing in the review is slightly off: it is not that `beforeAll` works *because* `hookTimeout` is 10s vs `testTimeout` 5s. It is that the import must be deferred at all, and a hook is the only place to defer it to that is not a timed test. The 10s-vs-5s difference is a side effect, not the mechanism. If it were only about the budget, raising `testTimeout` would have worked — and it would have left a 1.5s graph load sitting in a test.

**On the comment.** Rewritten, and it is now two lines longer than the one-liner you suggested rather than shorter. The six lines of flake history were right to cut — the commit message and this PR carry that. What replaced them is the constraint a reader cannot recover by looking: that static is not an option, and why. Without it the next person makes exactly the edit above and gets a `ReferenceError` with no hint that it was already tried.

Verified after merging `main` in: file 41/41, `pnpm check` and `pnpm typecheck` green, hook run on both commits with no bypass.

Separately, on something not in this PR: vitest's own hint suggests `isolate: false` for ~23s off the Core suite. I measured it rather than filing it — **212s and 164 failing files**, versus 61s green on the default. Import time drops 85% to 1% but worker time goes to 73%, so the suite is parallelism-bound, not import-bound. Not worth a ticket; recording it here so the hint does not get picked up later.

PR #4888, opened 2026-09-20
test(web): guard that every URL-data route picks an instant strategy

mrosberghaus:
Retrospective follow-ups to #4882. Two independent changes, both aimed at the same failure: the inventory for that PR was built with a grep that could not see two common shapes, so five blocking pages shipped without an opt-out and the published route counts were wrong twice.

## 1. Point at the web-scoped Next.js skills

`apps/web/.claude/skills/` holds twelve Next/Vercel skills — including `next-partial-prefetching-adoption`, which is the spec for instant-navigation work, and `next-dev-loop`, which reads the dev overlay instead of hand-rolled curl. No `AGENTS.md` named that directory.

They register only once a file under `apps/web/` is open, so an agent starting from the repo root does not know they exist. During #4882 that cost a hand-written plan — spec, tickets, a Playwright rig — that the skill already specified better and cheaper.

Mirrors the existing `### Evlog (Core only)` subsection. Four lines.

## 2. Guard that every URL-data route picks a strategy

`apps/web/src/lib/utils/__tests__/instant-navigation-routes.test.ts`, alongside the thirteen existing filesystem guards (`color-tokens`, `whole-pixels`, …).

Every `page.tsx`/`layout.tsx` that declares a `params`/`searchParams` prop must either export `instant = false` or appear in one of two lists — the routes that already hand the promise to a Suspense-wrapped child (4), or the worklist still to migrate (29). A new route has to pick a side on day one.

The guard keys on the **prop declaration**, not on `await params` / `await searchParams`. That grep is what built the original inventory, and it cannot see:

```ts
await Promise.all([params, searchParams])        // admin/soko-bots/[id]/page.tsx
await getRedirectQueryString(searchParams)       // (auth)/login/page.tsx
```

A second case keeps the lists from rotting: a listed path that no longer exists, no longer reads URL data, or has since gained the export fails the test. That is the usual way an allowlist dies, so it is checked rather than trusted.

The worklist doubles as the live inventory for the remaining instant-navigation work — entries move to the Suspense list as each feature area is migrated.

## Verification

- Added a probe route using the `Promise.all` shape with no opt-out → guard fails, naming it. Removed the probe → passes. A guard that only passes on a clean tree proves nothing, so this is the case that matters.
- `pnpm --filter web test src/lib/utils/__tests__/instant-navigation-routes.test.ts` — 2 passed, against `main` with #4882 already merged.
- `pnpm --filter web check`, `pnpm --filter web typecheck` — clean.


mrosberghaus:
## Grok review

Ad-hoc review only — requesting-code-review subagent dispatch was unavailable.

### Fixed
- Important: `(app)/agents/[agentId]/jobs/layout.tsx` already forwards `params` into a `<Suspense>` child (`JobLayoutInner`) and was still on `INSTANT_WORKLIST`, so the remaining-work count was off by one (`1383ae768` — `apps/web/src/lib/utils/__tests__/instant-navigation-routes.test.ts`)

### Remaining
- Minor: `apps/web/src/lib/utils/__tests__/instant-navigation-routes.test.ts`: no in-repo fixture for the `Promise.all` / helper-pass shapes the regex exists to catch. The live probe was deleted. Four worklist files that never write `await params` still fail the stale-list case if the regex is weakened to await-only.
- Minor: `AGENTS.md`: new Next.js subsection points at `apps/web/.claude/skills/`; Evlog/Apple and the paragraph above use `.agents/skills/` (`.claude/skills` is the Claude Code symlink to that tree).
- Minor: `(app)/agents/[agentId]/layout.tsx` is on `INSTANT_WORKLIST` only because `generateMetadata` types `params`; the layout default export does not read URL data.
- Skipped / disagreed: add a `<Suspense>` assertion on `SUSPENSE_WRAPPED` — current five files all have the boundary; extra rot check is optional.

### Assessment
Guard keys on the prop declaration, inventory is complete (28 opt-out + 5 suspense-wrapped + 28 worklist), and the one misclassified App Shell layout is now on the migrated list.

mrosberghaus:
All four addressed in `b48273fee`. Verified each first.

**`jobs/layout.tsx`** — confirmed. `JobLayout` returns `<Suspense fallback={<JobLayoutSkeleton />}><JobLayoutInner params={params}>`, so it forwards rather than awaits. Your move to `SUSPENSE_WRAPPED` is right.

**`agents/[agentId]/layout.tsx`** — this was the one that mattered. Confirmed: the default export is **sync** and reads nothing from the URL; only `generateMetadata` types `params`. So it isn't `instant-shell-url-data` at all, it's `blocking-prerender-metadata-runtime`, and neither of my two categories can fix it — metadata cannot go behind the page's `<Suspense>`, and `instant = false` does not quiet it. My two-way split had no correct slot for it, which made the worklist count wrong in the other direction.

Added a `METADATA_ONLY` set so the worklist stays an honest count of what the Suspense treatment actually fixes. Scanned for others in the same shape — it is the only one.

**Regex fixture** — fair; the probe that proved the regex was deleted with it, so nothing guarded the guard. Added a case pinning `URL_DATA_PROP` against both shapes that defeated the original grep, plus non-matches (`useSearchParams()`, a bare `type Params = Promise<…>`). Weakening the regex now fails here rather than silently in a future inventory.

**`AGENTS.md` path** — correct, and worse than a style mismatch: only `apps/web/.agents/skills/` is tracked in git. `.claude/skills` is the local Claude Code mirror and is absent from a fresh clone, so the pointer I wrote would have dangled for anyone who had not run the tooling. Now `.agents/`, matching Evlog and Apple.

**Suspense assertion — implemented rather than skipped.** One disagreement: it is not redundant with the other rot checks. A `SUSPENSE_WRAPPED` entry that loses its boundary still reads URL data and still has no `instant` export, so every existing case reads it as settled while the route is back to blocking. That is the only rot the list cannot otherwise catch, and it is three lines. If you would rather not have it, say so and I will drop it.

Verification: 3 tests pass, `pnpm --filter web check` and `typecheck` clean.

PR #4905, opened 2026-09-20
fix(core): release idle database connections before suspension

mrosberghaus:
Core now attaches its PostgreSQL pool to Vercel's Fluid lifecycle so idle connections can expire before the instance suspends. The shared Prisma factory accepts an externally owned pool while retaining its connection-string path for scripts and tests. Both paths preserve TCP keepalive with the existing 10-second initial delay; adapter pool and transaction-connection errors report to Sentry.

This is preventative hardening explicitly requested after measurement found substantial connection headroom. The baseline and decision are recorded on [SOK-1129](https://linear.app/masumi/issue/SOK-1129). No pool sizing, schema, or transaction policy changes.

Validation:
- Factory behavior tests: 4 passed, covering both input paths, external pool lifetime, idle pool errors, and transaction-connection errors.
- Full workspace test run: Web and packages passed; Core exposed an incomplete Vercel mock. After preserving real exports in that mock, the full Core rerun passed: 529 files, 5,864 tests, 39 skipped tests.
- `pnpm check`, `pnpm typecheck`, and `pnpm exec turbo run build --filter=@sokosumi/core` passed. All commits also passed the normal precommit hook.
- Independent standards and spec reviews: no findings.

After deployment, compare Neon idle connections and pooler waits with the ticket's baseline under comparable traffic. This operational verification is pending; local tests do not establish production suspension behavior.


mrosberghaus:
## Grok review

Ad-hoc review only — general-purpose reviewer subagent was unavailable.

### Fixed

### Remaining
- Minor: `packages/database/AGENTS.md` still documents `createPrismaClient(databaseUrl: string)` and the pre-change Core singleton. README was updated; this agent-facing copy was not.
- Minor: most Core tests still stub `@vercel/functions` as `{ waitUntil }` only. `evlog.test.ts` now spreads the real module so `attachDatabasePool` exists; the others stay safe only because they mock `@/lib/db/prisma`. A future test that loads the real singleton and stubs `waitUntil` will fail the same way.
- Minor: `createPrismaClient` tests cover `onPoolError` and pool lifetime, not `onConnectionError`.
- Skipped / disagreed: a unit test that `attachDatabasePool` was called — SOK-1129 testing decisions reject construction-shape tests; the factory seam tests plus the Core suite are the intended net.

### Assessment
Core owns the `pg.Pool`, attaches Fluid idle cleanup, keeps keepalive on both factory paths, and leaves `@sokosumi/database` off Vercel; CI is green. Remaining items are docs and test-hygiene, not production defects.

mrosberghaus:
Follow-up to [this review](https://github.com/masumi-network/sokosumi/pull/4905#issuecomment-5749748073), addressed in b3b0315f882a30e94f8ae8a857227eab908db748:

- Updated `packages/database/AGENTS.md` with the pool-or-URL factory signature, error callbacks, and Core-owned pool lifecycle.
- Added a transaction-connection error test that exercises the real Prisma adapter and verifies `onConnectionError` receives the emitted error.
- Left the remaining `waitUntil`-only mocks unchanged: those tests replace `@/lib/db/prisma`, so they do not load the real singleton or require `attachDatabasePool`. There is no current failure to fix in those tests.
- Did not add a construction-shape assertion for `attachDatabasePool`, consistent with the ticket's testing decisions.

Validation: all four factory tests pass; repo-wide `pnpm check` and `pnpm typecheck` pass through the normal precommit hook. Post-deployment verification of idle connection cleanup remains pending.

PR #4909, opened 2026-09-20
feat(web): surface project activity and drop empty count pills

mrosberghaus:
Three commits: the row redesign, the name filter, then the header placement.

## What changes for the user

The projects list gave each row an avatar, a name, a briefing line and two count pills — and nothing else. Most rows carried a `0` jobs pill that said nothing, the suitcase icon named nothing a sighted user could read, and there was no way to narrow a workspace's projects at all.

### Rows

- **Count pills only appear when they count something.** A pill at zero is dropped, and when both are zero the wrapper goes too, so no stray gap is left before the timestamp.
- **Each pill carries a tooltip** naming what it counts, using the existing Radix `Tooltip`. The `aria-label` ("Tasks: 26") is unchanged.
- **The jobs icon is now `Bot`, not `Briefcase`.** History already pairs `ListTodo` (tasks) with the agent icon and its `Bot` fallback (jobs); `Briefcase` had no other use in the app and read as "job listing" rather than "agent run".
- **Every row stamps its last activity on the right**, which makes the list's existing activity ordering legible. The visible text is a bare relative time ("2 days ago"); its hover title names the measure ("Last activity: 18 Sept 2026 at 09:05").
- **A project with no activity yet reads "Created about 1 month ago".** Core floors `lastActivityAt` at `createdAt`, so a bare stamp there would imply work that never happened.
- **Rows without a briefing no longer render an em-dash placeholder line** and collapse to a single line.

### List header

- **A filter field narrows the list by project name**, sitting in a header row inside the list card, divided from the rows it controls. The card wraps the header plus either the rows or the no-matches message, so the filter stays reachable when a search returns nothing. A workspace with no projects at all skips the header — there is nothing there to filter.
- **Filtering runs in Core, not the browser.** The list pages with "Load more", so a client-side filter would only ever search the pages already fetched — a project on a later page would look like no match at all. `GET /projects` gains a `q` parameter, applied inside the ranked activity query and mirrored onto the count so the pagination total describes the same set as the rows. ILIKE wildcards in the term are escaped, so a search for `50%` stays a literal substring.
- **"Load more" carries the active filter** into later pages.
- **The sort indicator is plain text**, not a control: the route has one fixed ordering and its cursor encodes `lastActivityAt`, so a chip would promise a menu that cannot exist yet.

## Implementation notes

- Core: `lastActivityAt` added to `ProjectListItem`, returned from the ranked activity row the list query already computes; `q` added to the list route's query schema. Web client regenerated with `generate:core:snapshot`.
- `q` is debounced through nuqs with `shallow: false`, so the server refetches rather than filtering what is already on screen. Follows the existing `jobs-search.tsx` pattern.
- `TimeAgo` gains an optional `titlePrefix` so a caller can name what its timestamp measures.
- `previewProjectBriefing` is removed: the em-dash fallback was its only reason to exist, and its single call site now uses `stripMarkdownToText` directly.
- The loading skeleton mirrors the new right-hand cluster (one pill plus a stamp) to keep the Instant-nav swap stable.
- Translations added to `en`, `de` and `es` with real translations, parity check passing.

## Known limitations

Two accessibility gaps I chose not to fix here, both worth a reviewer's opinion:

- The pill tooltip is **mouse-only** — `TooltipTrigger asChild` wraps a non-focusable `<span>`, so keyboard and touch users get only the `aria-label`. This matches the existing pattern in `job-list-item.tsx`; making pills focusable would add two tab stops per row.
- Radix's `aria-describedby` repeats the tooltip word already in the `aria-label`, so a screen reader hears "Tasks: 2 … Tasks".

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 17/17
- `pnpm test` — 17/17 (web 815 files, core 537)

New tests: Core covers `lastActivityAt` pass-through, `q` reaching both the ranked query and the count, and wildcard escaping. Web covers pills hidden at zero (each side), the activity stamp and its title prefix, the `Created` fallback, the absent em-dash, the filter writing/clearing `q`, and the page forwarding a trimmed `q` to Core.

The row changes and the header layout were reviewed against the preview deployment with real data. **The filter field has not been exercised in a browser yet** — the debounce-plus-server-refetch path is the part unit tests cover least well, so it is worth typing into on the preview before this leaves draft.


mrosberghaus:
## Grok review

### Fixed
- Important: `GET /projects` pagination total used Prisma `contains` (unescaped ILIKE) while the ranked query escaped `%`/`_`. A search for `50%` could count a different set than the rows. Count now uses the same `ESCAPE` clause (`31d7c0530` — `apps/core/src/routes/v1/projects/get.ts`, `apps/core/src/helpers/project-activity.ts`).
- Important: `shallow: false` refetch remounted the filter. The page `key` reset `ProjectsView` after every `q` change, and the update was not a transition, so the route Suspense fallback stole focus. `startTransition` keeps the current tree; appended pages reset in render instead of remounting the view (`31d7c0530` — `apps/web/src/app/(app)/projects/components/projects-filter.tsx`, `apps/web/src/app/(app)/projects/components/projects-view.tsx`, `apps/web/src/app/(app)/projects/(root)/page.tsx`).
- Minor: filter input now has `aria-label` (placeholder is not a name). One-line while editing that file.

### Remaining
- Minor: `apps/web/src/app/(app)/projects/components/projects-loading-view.tsx`: Instant skeleton still has no filter header row. First paint of `/projects` will shift rows down ~40px when the live header appears. Card `min-h-[320px]` contains the outer swap.
- Skipped / disagreed: mouse-only pill tooltips — matches `job-list-item.tsx`; making pills focusable adds two tab stops per row, as the PR already notes.
- Skipped / disagreed: Radix `aria-describedby` repeating the `aria-label` — same existing tooltip pattern; not introduced as a new control role.

### Assessment
Name filter now searches the same set it counts, and typing into it should keep focus. Worth a pass on the preview (the PR already flagged that path) and a glance at Instant-nav header CLS if that swap still looks jumpy.

mrosberghaus:
Both important findings were real and are in `31d7c0530`.

The escape mismatch was the worse of the two, and my Core test made it invisible: it asserted `projectCountMock` was called with `{ name: { contains: "autumn", mode: "insensitive" } }`, which pinned the unescaped path as if it were the intended behaviour. The wildcard test I wrote only checked the ranked query's values, so `%` diverging between the rows and the total passed a green suite. Pinning the *shape* of the count call rather than its agreement with the ranked query was the mistake.

The remount/focus one is the path I flagged in the PR body as unverified, and it was in fact broken.

**Skeleton header** — fixed in `3567d5277`. The skeleton mirrored the card but not its header, so the first paint dropped every row by the header's height (~53px: `py-2.5` + an `h-8` control + the 1px border) once the live filter appeared; `min-h-[320px]` held the card's outer box but not the rows inside it. The skeleton now carries a header row with the live geometry, in skeletons only so there is no copy to flash, and `projects-loading-view.test.tsx` pins `border-b` / `px-4` / `py-2.5` plus the `h-8` block against the live header so the two cannot drift apart again.

**The two skipped items** — agreed, for the reasons given. The mouse-only tooltip and the duplicated `aria-describedby` are both inherited from the existing `job-list-item.tsx` pattern; changing them here would set a new precedent for every pill in the app, which belongs in its own change rather than riding along with this one.

**Still unverified:** nobody has typed into the filter in a browser. The focus fix in `31d7c0530` makes that more likely to hold, but "more likely" is not proof, and it is the one path in this PR with no browser evidence behind it.

PR #4910, opened 2026-09-20
feat(database): record the email sent for a notification

schaier-io:
Two nullable columns on Notification for SOK-1090. emailId is the Resend id
of the email one row was covered by, so a later unread row about the same
thing can see the reader was already told. emailScheduledAt is when a held
email leaves Resend, null for one sent at once, so a read before that
instant can cancel it by id.

Nothing writes them yet. The Core fixture that builds a full Notification
row gains the two fields so the layer typechecks alone.

---

## Summary by CodeRabbit

- **New Features**
  - Notifications now retain email delivery identifiers and scheduled delivery times.
  - Scheduled notification emails can be distinguished from immediate messages, improving delivery status visibility and handling of unread notifications.
- **Bug Fixes**
  - Notification records now consistently include email delivery metadata, reducing inconsistencies when working with scheduled or previously sent notification emails.


schaier-io:
@coderabbitai review

PR #4913, opened 2026-09-20
feat(notifications): mail what is addressed to the reader

schaier-io:
Switches the notification emails on (SOK-1090). A mention, a direct
message, a task that stopped for the reader, a task that finished and a
request for a workspace they manage now join the reminder in
`NOTIFICATION_EMAIL_CATEGORIES`, which is what Core's delivery reads and
what the dispatch in the commit below acts on. Email defaults to on for
each, as the reminder's already did.

The account page draws the email cell on those rows instead of the "not
sent" marker. Every message in a room and the other task updates keep
the marker, and it no longer says "yet": those two rows have no email
waiting to be built, and the word read as a promise of one.

A preset speaks for Sokosumi and the device only. A press of one writes
In app and Push on every kind of the group and carries each kind's email
cell over as it was, and the rail reads the group by those two channels.
Before, a preset wrote the whole channel set, which would have switched
a reader's emails off on every press, whichever preset it was: no
preset's reach names the inbox. A row with its email on would have read
as Custom for the same reason, because the reader's cells were compared
whole against a set that names the two device channels only.

The email column hint now says what happens to an email that is still
waiting: it is dropped when the reader reads its notification in
Sokosumi. Reading does not unsend one that already went, which is what a
reader who was away gets, and what every reminder gets: reminders are
mailed by their own sync and never wait.

---

## Summary by CodeRabbit

- **New Features**
  - Email notifications are now available by default for task attention, task completion, chat mentions, direct messages, and system notifications.
  - Notification presets now preserve existing email preferences while updating in-app and OS-banner settings.
  - Email remains unavailable for task updates and room messages.

- **Documentation**
  - Updated English, German, and Spanish notification settings text to clarify email availability and pending-email behavior.


schaier-io:
@coderabbitai review

PR #4916, opened 2026-09-20
feat(chat): show who is typing in a room

mrosberghaus:
Closes [SOK-1131](https://linear.app/masumi/issue/SOK-1131/show-who-is-typing-in-a-room)

## What changes for people

When you send a message you can now see whether anyone is answering. A room says who is composing, on a line by the composer:

- `Patrick is typing…`
- `Patrick and Andreas are typing…`
- `Several people are typing…` (past two)

Every room, Channels and Directs alike — the 1:1 Direct is where it matters most, because that is the case where you are waiting on one specific person. Humans only: a coworker's turn is already announced by Thought, and saying it twice in two vocabularies would be worse than not saying it.

**Placement:** below the composer from `md` up, absolutely positioned in the bottom padding the composer form already has, so the transcript keeps its full height. On narrow layouts it is a row above the composer instead, holding its height so the composer never moves under your thumb mid-sentence — there the pad is 12px, or the home-indicator inset, and is dropped entirely with the keyboard open. One element repositioned by breakpoint, so there is one live region rather than two. No animation, so it never reads as a coworker thinking.

## How it works

Architecture is fixed by [ADR-0033](docs/adr/0033-typing-rides-its-own-ephemeral-channel.md); the vocabulary (**Typing**, **Typing line**) is in `CONTEXT.md`.

Typing publishes as an **ephemeral** Ably message on a second per-room channel, `chat_typing:room_{roomId}`, granted `publish` + `subscribe`. The message channel stays **subscribe-only**.

**Why a second channel.** Ably capabilities are per channel and operation, never per event name. Granting `publish` on the message channel would let any room member publish a forged `chat_room_message`, which every client renders as a genuine message from whoever the payload names — live, convincing, and unpersisted, so it leaves no trace. A separate channel makes that impossible to send rather than possible to filter. There is a Core test asserting `publish` appears on typing channels and nowhere else.

**Timings** (Ably Chat's own defaults): heartbeat on a 10s throttle, receiver-side expiry at 12s. The explicit stop is an optimisation — correctness rests on the expiry, so a closed tab or dropped connection clears itself.

**Four stops:** send, clear-to-empty, blur, leaving the room.

**Stays silent for:** a restored Draft, an emoji the toolbar inserts, attachments, and the Thread composer. Typing is about text the person typed, so it never lies.

Readers aggregate by `userId` (not Ably `clientId`, so one person on two devices is one typist), drop themselves, and keep start order so the line does not resort while you read it.

**The entire Core diff is the capability map** — no route, table, migration, DTO or OpenAPI change.

## Also in this PR

- **Vitest `maxWorkers: "50%"` locally** on web and Core (CI keeps full parallelism). A full happy-dom suite taking all but one core starves WindowServer badly enough that its watchdog kills it and the desktop freezes. Requested by the author after hitting this.
- **`ably-capability-map.ts`** extracts capability-map parsing that the room and organization parsers had each copied, rather than adding a third copy. Behaviour-preserving; both existing test files pass unchanged.

## Deliberate deviations

- The line goes through a new `typingLine` slot on the composer chrome, **not** the `aboveEditor` slot the spec named. That slot sits inside the bordered card above your own input; a line about other people does not belong there.
- `parentMessageId` ships on the wire always `null`. It is groundwork, which normally earns deletion — it stays so Thread-scoped Typing is additive rather than a wire break, and so readers on an older build drop Thread events from day one. Reasoning recorded in ADR-0033.
- The collapse-past-two counts **resolved** names, so three typists where one cannot be named renders two names rather than "Several people". Naming what you know beats a vaguer line.

## Verification

| Check | Result |
| --- | --- |
| `pnpm --filter web test` | 797 files, 6511 passed |
| `pnpm --filter core test` | 530 files, 5834 passed |
| `pnpm --filter @sokosumi/utils test` | 674 passed |
| `pnpm check` (Biome, 4211 files) | clean |
| `pnpm typecheck` | 20/20 tasks clean |
| `pnpm --filter web messages:parity` | de, es ok |

Three defects were found and fixed during the work, each **mutation-verified** so the test genuinely fails without the fix:

1. **Infinite render loop** in `RoomTypingProvider` — the island reported upward from an effect keyed on array identity while the provider replaced state unconditionally. Burned 126s and aborted the test worker; would have locked a browser tab.
2. **`useAbly()` called from the composer** — the composer renders outside `LazyAblyProvider`, where that hook throws. Caught by the existing suite; fixed by moving Ably into a sibling island, mirroring `OrgPresenceProvider`.
3. **Emoji picker announced Typing** — `insertText` fires a native input event indistinguishable from a keystroke. Fixed with an `onToolbarInsert` signal.

## Not verified here

**No part of this was exercised against a live Ably connection.** The capability grant, the ephemeral flag, and the publish/subscribe round trip are proven by tests and types only. Worth a manual two-browser check on the preview deploy before merge.

## Platform

Web only. Recorded as Apple parity row 36a; the native client needs the typing channel in its own capability handling first.

## Follow-ups (not yet filed)

- Thread-scoped Typing — the `parentMessageId` groundwork is already on the wire
- Apple client participation


mrosberghaus:
## Grok review

### Fixed
- Important: typing hook used raw `channel.subscribe` / `channel.detach`. Room switches remount the island (`key={selectedRoom.id}`), and those calls reject on Ably attach/detach races (90000/409) and offline give-up (80003) — the same unhandled-rejection class as SOKOSUMI-QK/QQ/RN. Message and presence channels already go through `safeSubscribeChannel` / `safeDetachChannel`. (`51e25100a` — `apps/web/src/lib/ably/use-room-typing.tsx`)

### Remaining
- Minor: `apps/web/src/lib/ably/use-room-typing.tsx`: receive side trusts payload `userId` and never checks `message.clientId`. A room member can impersonate or `stopped`-clear anyone else's line until the next heartbeat. Presence already derives identity from `${userId}:${instanceId}` (`parseUserIdFromAblyPresenceClientId`). Left as-is: ADR-0033's security boundary is the second channel (so a room message cannot be forged), not typing-identity authenticity among members who can already send real messages.
- Minor: `apps/web/src/app/(app)/chat/components/room-session-composer.tsx`: `toolbarInsertRef` is a one-shot flag. If `insertText` produces no `onValueChange` (failed or no-op insert), the next real keystroke is swallowed and does not announce Typing.
- Minor: `apps/web/src/app/(app)/chat/components/room-composer.tsx`: emoji and quote-restore go through `onToolbarInsert`; `ComposerAddLinkDialog` / `insertLink` do not, so a toolbar link into an empty composer can announce Typing.
- Skipped / disagreed: treating payload-`userId` spoofing as Important — same ADR tradeoff as not putting a client-side identity guard on a client-published signal. Channel isolation is the invariant this PR is for.

### Assessment
Architecture, tests, and the publish-vs-subscribe split match the PR body and ADR-0033; the remaining notes are polish, not blockers. Live Ably round-trip is still unverified here, as the PR already says.

mrosberghaus:
All three remaining notes were real and are fixed in `42365c554`. Each was verified against the codebase first.

### `userId` spoofing — fixed, and I disagree with the reason for skipping it

The review left this as-is on the grounds that ADR-0033's boundary is the second channel. That boundary stops a member forging a room **message**; it does nothing about forging typing **attribution**. A member can already post as themselves — what they cannot do is make a post look like it came from someone else. Typing let them do exactly that, including a `stopped` to clear another person's line.

"They can already send real messages" isn't the same permission. Sending as yourself and sending as me are different things, and only one of them was possible before this commit.

The repo already had the precedent: `parseUserIdFromAblyPresenceClientId`, whose own doc says it exists "so peers cannot spoof via free-form clientId (token still binds the full clientId)". Ably stamps `clientId` from the token, so it is the one field a sender cannot choose. The receiver now derives the user from it and requires the payload to agree, so a client that gets its own identity wrong fails closed rather than announcing under the wrong name.

Three tests, **mutation-checked** — all three fail with the guard removed:
- a member publishing a teammate's `userId` is ignored
- a forged `stopped` does not clear the real typist's line
- an event with a missing or malformed `clientId` is dropped

Side effect: that helper was only reachable by deep import, so it now leaves the `@sokosumi/utils` barrel.

### One-shot flag swallowing the next keystroke — fixed

Confirmed as described. An insert that failed or no-opped left `toolbarInsertRef` set, and the next genuine keystroke was eaten. It now clears in a `queueMicrotask`, so only the insert's own synchronous change can claim it. The failure direction is now announcing Typing once too often rather than going silent while somebody is genuinely typing, which is the right way round.

### `ComposerAddLinkDialog` / `insertLink` — fixed

Confirmed: `handleLinkSave` was a third programmatic-insert path after the emoji picker and quote-restore, so a toolbar link into an empty composer announced Typing. It now goes through `onToolbarInsert` like the others.

### On the attach/detach race

Correct catch, and the right precedent — `safeSubscribeChannel` / `safeDetachChannel` are what the message and presence channels already use. Kept as pushed in `51e25100a`.

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter web test` | 820 files, 6809 passed |
| `pnpm typecheck` | 17/17 |
| `pnpm check` (Biome) | clean, 4228 files |

Still unverified, as the PR body says: no part of this has run against a live Ably connection. The capability grant, the ephemeral flag and the publish/subscribe round trip are proven by tests and types only, and that is the check to do on the preview deploy before this goes out of draft.

mrosberghaus:
## Grok review

### Fixed

Already on HEAD from the last pass:
- Important: typing attach/detach goes through `safeSubscribeChannel` / `safeDetachChannel` (`51e25100a` — `apps/web/src/lib/ably/use-room-typing.tsx`)
- Minor (author promoted): receive side binds `userId` to Ably `clientId` via `parseUserIdFromAblyPresenceClientId` (`42365c554` — `apps/web/src/lib/ably/use-room-typing.tsx`)
- Minor: `toolbarInsertRef` clears in a microtask so a no-op insert cannot swallow the next keystroke (`42365c554` — `apps/web/src/app/(app)/chat/components/room-session-composer.tsx`)
- Minor: `handleLinkSave` goes through `onToolbarInsert` (`42365c554` — `apps/web/src/app/(app)/chat/components/room-composer.tsx`)

### Remaining
- Minor: `apps/web/src/lib/ably/room-typing-model.ts`: `liveTypistIds` filters expired typists but `applyTypingEvent` never prunes them. A typist who vanished at 12s and then heartbeats again keeps the old `startedAtMs` and can jump back in front of people who started later.
- Minor: `apps/web/src/lib/ably/use-room-typing.tsx`: a `start` that lands before `canPublish` (authorize still in flight) is dropped and not retried. The next genuine keystroke after attach is what first announces.
- Minor: `apps/web/src/app/(app)/chat/components/room-session-composer.typing.test.tsx`: emoji no-op is covered; toolbar link insert is not.
- Skipped / disagreed: none of the prior findings. Identity spoofing was left as Minor last time; the author fixed it, and the guard is correct for this codebase (token `clientId` is `{userId}:{instanceId}`).

### Assessment
Ready to merge: Yes
Capability split, ephemeral publish, island/provider shape, and tests match the PR body and ADR-0033. Live Ably round-trip is still unverified here, as the PR already says.

mrosberghaus:
## Grok review

One-shot fresh review of PR 4916 (session request).

Range: `fd980a25fefc88464dc557659c4a37317f24eccb`..`4647ffa5f607669fdf79899ee8a1f2c888555d04` (`main` merge-base). Re-read the current diff; prior Grok notes were not treated as done.

### Fixed

Already on HEAD from earlier passes:
- Important: typing attach/detach goes through `safeSubscribeChannel` / `safeDetachChannel` (`51e25100a` — `apps/web/src/lib/ably/use-room-typing.tsx`)
- Minor (author promoted): receive side binds `userId` to Ably `clientId` via `parseUserIdFromAblyPresenceClientId` (`42365c554` — `apps/web/src/lib/ably/use-room-typing.tsx`)
- Minor: `toolbarInsertRef` clears in a microtask so a no-op insert cannot swallow the next keystroke (`42365c554` — `apps/web/src/app/(app)/chat/components/room-session-composer.tsx`)
- Minor: `handleLinkSave` goes through `onToolbarInsert` (`42365c554` — `apps/web/src/app/(app)/chat/components/room-composer.tsx`)

### Remaining
- Minor: `apps/web/src/lib/ably/room-typing-model.ts`: `liveTypistIds` filters expired typists but `applyTypingEvent` never prunes them. A typist who vanished at 12s and then heartbeats again keeps the old `startedAtMs` and can jump back in front of people who started during the gap. Explicit `stopped` is correct; only the expiry-then-restart path is wrong, and it needs two overlapping typists to show.
- Minor: `apps/web/src/lib/ably/use-room-typing.tsx`: a `start` that lands before `canPublish` (authorize still in flight, or the island not mounted yet) is dropped and not retried. The next genuine keystroke after attach is what first announces.
- Minor: `CONTEXT.md`, `apps/apple/PARITY.md` row 36a, and the PR body still say the line sits above the composer. After `4be22fd80` it is above on narrow layouts and absolutely positioned in the form's bottom padding from `md` up.
- Minor: `apps/web/src/app/(app)/chat/components/room-session-composer.typing.test.tsx` covers emoji no-op; toolbar link insert is still untested.
- Skipped / disagreed: promoting the expiry-order gap to Important — name order is the documented invariant, but the path is pause-without-blur for 12s with a second typist, and blur/send/clear already remove the row. Not a merge blocker.

### Assessment
Ready to merge: Yes

Second channel, subscribe-only messages, token `clientId` binding, island/provider shape, and the four stops match the PR body and ADR-0033. Live Ably round-trip is still unverified here, as the PR already says.

mrosberghaus:
Three of the four fixed in `0b499f00c`; pushing back on the fourth.

### Expiry-then-restart reordering — real, fixed

Confirmed as described. `applyTypingEvent` matched an existing typist whatever the state of their entry, so `startedAtMs` and their array position survived an expiry. Pat starts, goes quiet past 12s, Kim starts during the gap, Pat types again — Pat was still named first despite Kim having typed continuously.

An entry past its expiry is one every reader has already stopped seeing, so it is now dropped and appended with a fresh `startedAtMs`. A pause *inside* the window still refreshes the heartbeat and keeps its place.

Two tests. The reordering one fails if the expiry check is removed — checked by removing it.

### Stale docs — fixed

Correct, and mine: I moved the line in `4be22fd80` and left `CONTEXT.md` and parity row 36a describing the old placement. Both now say below-the-composer on `md` and up, above it on narrow layouts, and the glossary records the returning-typist ordering. The PR body above is updated too.

### Toolbar link insert untested — fixed

Correct. Emoji had a test, `handleLinkSave` didn't. It has one now, mutation-checked by removing the `onToolbarInsert` call.

### Not taken: `start` dropped before `canPublish`

This is the intended behaviour, not a gap, and it is why `publish` returns a boolean.

A keystroke landing while `authorize` is still in flight must **not** record "we are announced". If it did, it would open a 10s throttle window against a publish nobody received — so the typist would look idle for the full ten seconds rather than for one keystroke. Returning false keeps the window closed so the next keystroke announces immediately.

The only case that loses anything is typing exactly one character during the sub-second attach and then stopping, which the 12s expiry clears regardless. Retrying on attach would mean replaying composer state into the publish path to cover a window that closes by itself.

Happy to revisit if you have a case where the first keystroke is the only one.

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter web test` | 824 files, 6889 passed |
| `pnpm typecheck` | 17/17 |
| `pnpm check` (Biome) | clean, 4257 files |

Live Ably round-trip still unverified, as the PR body says — that and how the line reads against a real virtualized transcript are the two checks left for the preview deploy.

PR #4918, opened 2026-09-20
fix(web): align task, job and history row geometry with their cards

mrosberghaus:
Applies the row geometry fixed for the projects list in #4909 to the task, job, history and drive lists. CSS only — no behaviour, no data, no copy.

## What changes for the user

Two defects, both visible on hover:

- **Separators stopped short of the card edge.** The jobs list padded its divider wrapper (`divide-y px-2`), so every line was inset 8px while the rows cancelled the padding again with `-mx-2`. Dividers now meet the card edge.
- **The hover highlight was a rounded rectangle between straight lines.** Task, job and history rows carried `rounded-lg` under a `hover:bg-*` fill, so the highlight rounded its corners inside square dividers, and its 8px radius did not match the card's 12px at the first and last row. The rows are square now; the card's `overflow-hidden` rounds the ends.

## Files

| Surface | Change |
| --- | --- |
| Tasks → jobs list | two divider wrappers lose `px-2`; the row loses `-mx-2` and `rounded-lg` |
| Tasks → task list | row `rounded-lg` → `rounded-none` |
| History list | row `rounded-lg` → `rounded-none` |
| Drive list view | divider wrapper loses `px-2`; row loses `-mx-2` and `rounded-lg` |
| Drive recents panel | day wrapper loses `px-2`, day headings move to `px-4` |

### Hover fill

Squaring the rows exposed a second problem: several of them hovered to a fill that was not a step away from their card.

- Task, job and history rows used `hover:bg-card-background` — the *same* colour as the card they sit on, so the highlight was invisible. Now `hover:bg-card-background-hover` (fixed in `afa461ccd`).
- Drive list rows used `hover:bg-background` — the page colour, three steps below the card. Now the same pinned step.

### Scope

Only the **list** branch of the drive view-mode helpers changes; the grid branch is untouched. Content stays at the same 16px inset — the row's `px-2` plus the body's `px-2` replace what the container's padding used to provide.

## Verification

- `pnpm check` — clean
- `pnpm typecheck` — 17/17
- `pnpm test` — 17/17; drive + tasks + history suites specifically: 88 files, 738 tests

Tests were added with the hover-fill fix, pinning the hover class on the history, task and job rows so the no-contrast case cannot come back. The geometry changes themselves are Tailwind class strings with nothing pinning them.


mrosberghaus:
## Grok review

### Fixed
- Important: task, job, and history rows sit on `--card-background` but used `hover:bg-card-background`, so the square hover this PR shaped was the same fill as the card. Switched to `hover:bg-card-background-hover` and pinned the class the way the projects list already does (`afa461ccd` — `apps/web/src/app/(app)/tasks/components/job-list-item.tsx`, `task-list-item.tsx`, `apps/web/src/app/(app)/history/components/history-list-item.tsx`).

### Remaining
- Minor: `apps/web/src/app/(app)/drive/components/drive-view-layout.ts` still has the inset divide (`px-2`) plus `-mx-2 rounded-lg` cancel pair. Author scoped this out; it needs the list branch split first.
- Minor: `apps/web/src/app/(app)/projects/components/project-needs-attention-section.tsx` still uses `-mx-2 rounded-lg` on its rows.
- Skipped / disagreed: matching the projects list from #4909 — that list still insets dividers (`PROJECTS_BROWSE_DIVIDE_CLASS` keeps `px-2`) and rounds hover (`md:rounded-lg`). Full-bleed square rows plus the card's `overflow-hidden` is the right geometry for these three lists; it is a different pattern, not a copy of the projects rows.

### Assessment
Jobs dividers now meet the card edge, rows are square, and hover is actually visible against the card.

mrosberghaus:
The hover-fill catch was right and is in — those three rows hovered to the same colour as their card, so squaring them left a highlight with no contrast. That also prompted me to check drive, which hovered to `--background`, the page colour, three steps below its card; it now uses the same pinned step.

On the three remaining notes:

**Drive** — fixed in `b1abc1df6`, pushed after this review was written. `driveItemsListClass` is now `divide-border divide-y`, and the list row is `relative flex items-center gap-1 rounded-none px-2 hover:bg-card-background-hover` with no `-mx-2`. The helpers already branched on view mode, so no split was needed and the grid branch is untouched. Content stays at the same 16px inset: the row's `px-2` plus the body's `px-2` replace what the container's padding provided.

**`project-needs-attention-section.tsx`** — correct for this branch, which is cut from `main`. That row is already `rounded-none` with no `-mx-2` on #4909. Changing it here would collide with that PR, so it stays as-is and lands with #4909.

**The "skipped / disagreed" item** — the premise is out of date. It reads `PROJECTS_BROWSE_DIVIDE_CLASS` and `project-list-item.tsx` from `main`, where #4909 has not merged. On #4909's head:

- `PROJECTS_BROWSE_DIVIDE_CLASS` is `"divide-border divide-y"` — no `px-2` (`d47724ca0`)
- the row is `"flex min-w-0 flex-row items-center gap-4 rounded-none px-4 py-3 transition-colors"` — square at every breakpoint, `md:rounded-lg` removed (`78bb085e6`)

So full-bleed square rows on `overflow-hidden` is exactly what #4909 established, and this PR applies that same pattern rather than a different one. Worth re-reading against #4909's head rather than `main` if the geometry is being compared.

PR #4921, opened 2026-09-20
feat(web): put New project on the line it acts on

mrosberghaus:
The projects index carried a row above the list whose only job was holding the create button, so the page opened with a button floating over empty space and the list card started a row lower than it needed to.

**"New project" now sits in the list card's header row, beside the filter it acts on.**

### What changed

- `projects-view.tsx` — the button moved into the header row, `hidden md:inline-flex`. Below md the mobile FAB stays the create control, exactly as before.
- The sort label yields across md–lg (`inline md:hidden lg:inline`). Nothing in the row wraps, so with the button sharing the line a longer locale (`es`: "Ordenado por actividad reciente" + "Nuevo proyecto") would otherwise squeeze the filter to roughly 100px, of which the icon padding eats 64px.
- `projects-loading-view.tsx` — the Instant skeleton moved with it, so the swap keeps the same 52px header and does not drop every row by its height.
- `constants.ts` — the header row geometry became `PROJECTS_BROWSE_HEADER_ROW_CLASS`, paired in the Instant contract test, so live view and skeleton cannot drift apart and quietly reintroduce swap jank.

### Fixed along the way

An empty workspace used to render **two** "New project" buttons: one in the orphan row, one in the empty state. Now one.

### Verification

- `tsc --noEmit` clean, biome clean.
- Full web suite: 6823 passed, 1 skipped, 818 files.
- New `projects-view.test.tsx` (7 tests) covers the button's place in the header row, that it is the only create control on the browse screen, the desktop-only band, the sort label's yield band, and both empty states.

Not verified in a browser: this worktree has no sign-in, so the layout was checked against the tests and a static mock rather than a running app.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/projects/components/projects-loading-view.tsx`: create skeleton is `md:block`; live button is `md:inline-flex`. Harmless as flex items (both blockified, both `h-8`), but the Instant pair is not the same display type.
- Minor: `projects-instant-loading-contract.test.ts`: the new pairing test only greps `PROJECTS_BROWSE_HEADER_ROW_CLASS`. An unused import would still pass. Loading-view tests check the header tokens on the DOM; the live view test does not.
- Minor: create skeleton stays `w-[7.25rem]` (English sm-button guess) while live `AddProjectButton` grows with locale (`Neues Projekt` / `Nuevo proyecto`). Height swap stays 52px; the filter can still shift horizontally on Instant replace.
- Minor: header is nowrap inside `overflow-hidden` card. Filter shrinks first (`Input` has `min-w-0`); 200% zoom / a long locale at the lg edge was not checked in a browser.
- Skipped / disagreed: empty-state mobile still shows in-card create plus the FAB. That predates this PR; the duplicate this change removes is the desktop orphan row.

### Assessment
Ready to merge: Yes
Create moved onto the browse header it acts on, Instant geometry is shared via `PROJECTS_BROWSE_HEADER_ROW_CLASS`, the empty-workspace double button is gone, and tests lock placement, the desktop-only band, the sort yield, and both empty states.

mrosberghaus:
Pushed `72215f0a2` for two of the four.

**Fixed**

- Create skeleton is now `md:inline-flex`, matching the live button. Both are flex items of the header row so nothing rendered differently, but that file exists to keep the pair identical.
- `projects-view.test.tsx` now asserts every token of `PROJECTS_BROWSE_HEADER_ROW_CLASS` on the rendered header, so an unused import no longer satisfies the pairing. The contract test can only grep source; the skeleton side already had its DOM check, the live side did not.

**Not fixing**

- The `w-[7.25rem]` create skeleton against a locale-sized button: the fix would be rendering the real label in the skeleton, and the file's whole constraint is no copy, so no locale flash. The sort label skeleton (`w-32`) has carried the same approximation since before this PR. Vertical geometry is what the Instant contract guards, and both sides resolve to the same 52px header; the residual is a horizontal nudge of the sort label on replace.
- 200% zoom and the long-locale lg edge: correct that I did not check these in a browser, and I still cannot — this worktree has no sign-in for the preview deploy. Stated as unverified rather than claimed.

Agreed on the skipped item: the mobile empty state showing both an in-card create and the FAB predates this PR. Filed separately rather than widened into this diff.

PR #4922, opened 2026-09-20
feat(core): give projects a personal Pin

mrosberghaus:
Core half of project Pins: a personal, private mark on a project that the sidebar's Projects flyout will render first. Web is a follow-up PR, because the flyout and the list-row button need the regenerated client types this PR produces.

Naming follows ADR-0017 — **star in the API and the database, Pin in the UI**.

## What's here

- **`project_star`** — one row per user and project. `@@unique([userId, projectId])`, `@@index([userId, starredAt])`, cascade on both FKs, and deliberately no workspace column: a project belongs to exactly one workspace, so org filtering falls out of the read. Unlimited per reader; the flyout caps what it draws, not what is stored.
- **`POST` / `DELETE /v1/projects/{id}/star`** — both idempotent. The upsert passes `update: {}`, so re-starring keeps the original `starredAt` instead of moving the project to the end of the reader's Pins on a double click. Unstarring something never starred is a no-op; the 404 is about the project, not the Pin.
- **`GET /v1/projects/starred`** — the reader's Pins in the active workspace, oldest first. This is the one that matters: a Pinned project is usually a quiet one, and the projects list is ordered by activity, so a Pin regularly falls outside the first page. Without this endpoint the flyout would silently drop it.
- **`starredAt` on `ProjectListItem`** — resolved in the list route by one indexed lookup inside the existing `Promise.all`. No raw-SQL join was needed; the route already re-fetches its page. It sits on the list item rather than on the shared `projectSchema`, which would have forced a decision at ten call sites that have no interest in the reader's Pin.

A Pin belongs to a person, so coworker and vendor contexts are refused on the writes and read every row as unpinned rather than seeing the bound user's Pins as their own.

Closing a project keeps its Pin row and filters it out on read, so reopening restores it — rather than deleting Pins as an un-undoable side effect of closing.

## Verification

- 86 project tests pass, 14 of them new (`star.routes.test.ts`, `starred/get.test.ts`, plus two list-route cases for the resolved and unresolved `starredAt` paths).
- `tsc --noEmit` clean across the monorepo, biome clean.
- No DB-backed tests: the new route tests mock Prisma beside the route, matching the house pattern.

## Note for reviewers

The design record (ADR-0036 plus the CONTEXT.md glossary entry defining **Pinned project** and **Recently visited project**) is on `claude/flyout-project-list-alignment-c992bd` and lands separately, so it is not in this diff.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/src/routes/v1/projects/get.test.ts` — list-route Pin leak is untested. Writes and `GET /starred` 403 a coworker; the list *allows* a bound coworker and relies on `actor === "user"` so `starredAt` stays null. A later cleanup that stars by `requireAuthorizedUserContext().userId` would leak the bound user's Pins. Add a coworker-with-binding case that expects `starredAt: null` and no `projectStar.findMany`.
- Minor: `apps/core/src/routes/v1/projects/starred/get.ts` — response is `projectSchema`, so no `starredAt`. Order is only the JSON array. The flyout must not re-sort after merge; POST already returns the stamp if the client needs it.
- Minor: `packages/database/prisma/schema.prisma` `ProjectStar` — unique key is `(userId, projectId)`, so `ON DELETE CASCADE` from `project` has no `projectId`-leading index. `NoticeAcknowledgment` indexes the non-leading FK. Harmless until pin volume is large.
- Minor: `apps/core/src/routes/v1/projects/get.ts` — the ILIKE/`q` comment now sits on the Pin `readerUserId` block, not the count query it documents.
- Minor: `GET /projects/starred` `take: 50` with no `nextCursor`. Oldest-five flyout still works; a later “all Pins” view would drop newest pins past 50 with no signal.

Skipped / disagreed: none. No Critical or Important defects on this range.

### Assessment
Ready to merge: Yes

Core Pin write/read matches the PR: idempotent star/unstar, personal (coworker refused on writes and `GET /starred`), list `starredAt` only for a user actor, closed projects kept on the row and omitted from the starred list.

mrosberghaus:
Four of the five are fixed in d3cbe57f8. Pushing back on the fifth.

**Pin leak untested (list route)** — fixed. Added a bound-coworker case that mocks the binding and grant lookups so the request reaches the handler, stubs `project_star` as *returning a Pin*, and asserts the response carries `starredAt: null` and that `projectStar.findMany` was never called. A test that passes because the table is empty would prove nothing, so I mutation-checked it: replacing the gate with `(await requireAuthorizedUserContext(c.var.authContext)).userId` — the exact cleanup described — turns it red, and it goes green again when the gate is restored.

**`GET /starred` order lives only in the array** — fixed, and worth fixing rather than leaving to client discipline. The response is now `StarredProject` (`projectSchema` extended with `starredAt`), so a client that merges the list into a map can recover Pin order instead of depending on JSON array position. Relying on "the flyout must not re-sort after merge" is exactly the kind of rule that holds until the merge is written by someone who never read the endpoint.

**Cascade has no `projectId`-leading index** — fixed. `@@index([projectId])` on `ProjectStar` plus the matching `CREATE INDEX` in the migration, same shape as `NoticeAcknowledgment`'s non-leading FK index. The migration is unmerged and has never been applied, so it was edited in place rather than followed by a second one.

**Misplaced ILIKE comment** — fixed. Moved back onto the count branch it documents; the Pin comment stays with the `readerUserId` block.

**`take: 50` with no cursor** — not changing this, but I have recorded the reasoning at the constant. There is no "all Pins" view, and ADR-0036 decided the flyout draws five, so a cursor here would be pagination built for a caller that does not exist. The sharp edge is worth naming, though, and it is the opposite of what a reader might assume: the order is `starredAt` ascending, so truncation hides a reader's *newest* Pins past 50, never the oldest — and the flyout takes its five from the near end, so it cannot be affected. The comment now says that, and says to add a cursor if an "all Pins" view is ever built, since that view is where the loss would be silent. Happy to add the cursor now if you would rather not rely on a future reader seeing that note.

Verification: 98 project tests pass, `tsc --noEmit` clean across the monorepo, biome clean.

Note on the branch: `main` was merged in twice while this sat open, so the diff now also carries #4921, #4923 and the pnpm 12.5.1 bump. ADR-0036 landed with #4923 and is on this branch as a result.

PR #4927, opened 2026-09-21
feat(web): Pin projects from the list - the header and the flyout

mrosberghaus:
Web half of project Pins. Stacked on #4922 — review that one first; this PR's diff is only the web layer.

Design: [ADR-0036](docs/adr/0036-project-pins.md), already on `main`. Product UI says Pin throughout; only the generated calls say star, per [ADR-0017](docs/adr/0017-sidebar-pin-api-is-star.md).

## What's here

- **Regenerated Core client** — `postProjectsByIdStar`, `deleteProjectsByIdStar`, `getProjectsStarred`, the `StarredProject` type and `starredAt` on `ProjectListItem`. Its own commit, so it does not drown the hand-written diff.
- **Client, service and actions** — `pinProject` / `unpinProject` / `getPinnedProjects`, matching service methods, and `pinProjectAction`, `unpinProjectAction`, `loadPinnedProjects`. The last carries the same workspace-scope guard `loadMoreProjects` uses, so a Pin list cannot outlive an org switch.
- **The flyout draws Pins first, under headings.** The panel fetches Pins alongside the activity page and merges them *before* ordering. `orderSidebarProjects` caps Pins as well, so the popover stays five rows however much a reader has Pinned.
- **A Pin button on every list row and on the project header**, always rendered rather than hover-revealed, because a hover-only control does not exist on a touch device. Hidden on a closed project unless a Pin already exists there.
- **`pinnedProjects`, `projectsPanel`, `list.pin` and `list.unpin`** in `en`, `de` and `es`.

## Five things worth a reviewer's eye

**The merge is the feature.** The flyout merges the Pin list into the activity page *before* ordering, and that is the whole reason `GET /v1/projects/starred` exists in #4922: a Pinned project is usually a quiet one, so looking only at page one drops it silently. There is a test for exactly that case.

**The list row is a flex pair, not one big link.** A `<button>` inside an `<a>` is invalid HTML and the click would navigate instead of Pinning. The button sits beside the link, and the hover fill moved from the link up to the row — otherwise the row's right edge stays unlit while the pointer is over it. A test asserts the button has no anchor ancestor.

**The header derives its state rather than reading it.** `GET /projects/{id}` returns the plain project, which deliberately carries no `starredAt` — putting it there would force the question on ten routes that return a project and do not care. The header reads the Pin list through a `usePinnedProjects` hook the flyout also uses, so one fetch serves both and a toggle on either updates the other.

**`ProjectPinButton` reads `QueryClientContext`, not `useQueryClient`.** The latter throws without a provider, which broke nine existing tests — every test rendering a project row would have needed a `QueryClientProvider`. Refreshing the sidebar is best-effort; the button's job is to Pin.

**The button holds committed state, not `useOptimistic`.** That was the first attempt and was wrong here: an optimistic value is dropped when the transition ends and falls back to the row's server-rendered `starredAt`, which nothing refreshes, so a fresh Pin flipped blue and snapped straight back to grey while the write had in fact succeeded. State is now re-seeded whenever the server sends a different truth, so a navigation or a refetched list still wins.

## Two decisions reversed while building, with ADR-0036 amended to match

**The flyout groups now carry headings.** ADR-0036 originally argued a hairline divider was enough at five rows. On the preview that was plainly wrong: the panel's one heading read "Recent projects" directly above a Pinned row, so the label was lying about the first group. The panel's accessible name is now its own string, since naming a panel after one of its groups mislabels the other.

**A Pinned row moves hue, fill and ground together.** One step of grey plus a filled icon did not read at all — at 16px a filled lucide `Pin` is barely distinguishable from an outlined one. It pairs `text-primary` with `bg-primary-quinary`, the same pairing the task-created badge uses. Solid ramp steps, not opacity modifiers: `color-tokens.test.ts` enforces that rule absolutely, and caught the first attempt in CI.

The control stays on the row's trailing edge. The list is ordered by activity, so Pins are scattered through it and nobody scans the column for them; the leading edge already belongs to the avatar, which is the row's identity anchor.

## Verification

Full web suite: 827 files, 6931 tests passing. Monorepo `tsc --noEmit` clean, biome clean, `messages:parity` green.

Two notes on the tests. Adding `loadPinnedProjects` to the flyout's action mock mattered more than it looks: without it the new query errored silently, Pins fell back to `[]`, and all 117 existing sidebar tests passed while covering none of the new path. And the snap-back bug got through because the existing tests asserted the action was *called*, never what the control showed afterwards — the regression test for it is the feedback loop that reproduced it.


mrosberghaus:
## Grok review

### Fixed
- Important: a closed project's header could not Unpin. `GET /v1/projects/starred` dropped closed Pins, the flyout already hid them, and a quiet closed row is often off page one of the list, so the Pin was stranded. The starred list now returns closed Pins; the flyout still drops them (`a55e3e9e4` — `apps/core/src/routes/v1/projects/starred/get.ts`, `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.tsx`, `apps/web/src/app/(app)/projects/components/project-detail-pin-button.tsx`).
- Important: a Pin-list fetch error painted an empty Pin group with no retry. `isError` only watched the activity page. Retry now refetches both (`a55e3e9e4` — `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.tsx`).

### Remaining
- Minor: `apps/web/src/app/(app)/projects/actions.test.ts`: `pinProjectAction`, `unpinProjectAction`, and `loadPinnedProjects` (including the workspace-scope guard) have no action tests.
- Minor: `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.test.tsx`: the locale-key check still omits `pinnedProjects` and `projectsPanel`.
- Minor: `apps/web/src/app/(app)/projects/components/project-pin-button.tsx`: hit target is `size-8` (32px). Meets WCAG 2.5.8 24px; below the 44px touch aim.
- Minor: Pin/unpin failure only reverts the icon. No toast.
- Skipped / disagreed: not treating `startTransition(async () => …)` `isPending` as a blocker. Same pattern as `ProjectsView` load-more; both Core calls are idempotent.

### Assessment
Web Pin layer matches ADR-0036: merge-before-order, flex-pair row, committed button state, solid color tokens. Remaining items are tests and polish.

mrosberghaus:
All four Minor items are done. No pushback on any of them.

First, on the two Important ones already fixed in `a55e3e9e4`: that was a real hole in my own reasoning, not just an edge case. I hid the Pin button on a closed project *unless a Pin already existed*, specifically so closing could not strand one — but the header derives `isPinned` from the starred list, and that list excluded closed projects. So the header could never learn a closed project was Pinned, the button stayed hidden, and the Pin was stranded by exactly the guard meant to prevent it. Moving the filter out of Core and into the flyout is the right split: the endpoint reports the truth, each surface decides what to show.

**Action tests** — `pinProjectAction` and `unpinProjectAction` now cover the happy path and the unauthenticated path, and `loadPinnedProjects` covers both the list and the workspace-scope guard. The guard test asserts the rejection by message, so it fails if the guard is removed rather than passing on a coincidence.

**Locale-key check** — `pinnedProjects` and `projectsPanel` added to the `it.each([en, de, es])` list.

**Hit target** — the circle stays 32px and the hit area grows to 44px via `before:-inset-1.5`. Worth stating the consequence that made this more than a guideline: the row behind the button is a link, so a near miss on a phone did not do nothing, it navigated to the project.

**Toast on failure** — added, with `list.pinError` in all three locales. Agreed on the reasoning: the icon returning to where it was, with nothing said, reads as a bug rather than as a write that did not land. It uses the same `sonner` path `ProjectDetailActions` already uses for close and delete failures.

Noted on `startTransition` / `isPending` — no change made, matching your call.

Verification: full web suite 826 files, 6940 tests; monorepo `tsc --noEmit` clean, biome clean, `messages:parity` green.

PR #4948, opened 2026-09-21
feat(chat): show who has read a room with Seen by

mrosberghaus:
Implements [SOK-1110](https://linear.app/masumi/issue/SOK-1110/seen-by-show-who-has-read-a-chat-room).

A sender could not tell whether anyone had looked at what they posted. Sokosumi already knew — every member's **Room last-read** is stored and advanced automatically — but that knowledge was spent entirely on the reader's own sidebar bold and unread count.

**Seen by** shows it back to the room, on two surfaces:

- **Room header** — a stack of the three most recent readers plus a `+N`. Opening it (click or keyboard) lists every reader with when they last read, then the members who have **not** read at all, so "who is missing" is a fact to read rather than one to derive.
- **Transcript** — the same faces, smaller, trailing the newest message's last line. Only the newest message ever carries them; a row under every message was rejected as scrollback noise.

Nothing is asked of the reader: no receipt toggle, no "mark as read" button, no change to how reading works. A reader's private **Marked unread** stays private — it does not rewind Room last-read and does not make them show as unread to anyone else.

## How it works

- **No schema change, no migration, no new endpoint.** `ChatRoomReadState` already stores one Room last-read per member per room.
- **Core** adds the room's read-state relation to the shared `chatRoomInclude`, so every room-returning route carries receipts without its own change, in the same query as the roster. The synchronous `mapChatRoom` attaches each member's mark to their roster entry as a nullable `lastReadAt`.
- **Realtime** rides the existing room channel under its own event name, published best-effort after commit and marked `ephemeral` — a receipt is derived, so it has no business in Ably history, rewind or resume. No capability change.
- **Web** merges payload and events in one `useRoomReadReceipts` hook; both surfaces consume the hook, neither reads the DTO or the channel. An out-of-order event never rewinds a mark, so a dropped connection keeps the values it last delivered and the next payload heals it without a reload.

## The guest boundary, and a bug the review caught

Read times must not cross the organization boundary on an External channel. The mapper emits `null` for everyone when the **viewer** is a guest — but that only covered the payload. Ably capabilities are per channel and never per subscriber, so a guest holding `subscribe` on the room channel received every read event: they opened a room to an empty stack and then watched host members' faces and exact read times appear live.

Core now stays silent on any room with a guest on its roster. Host members there fall back to the mark their next room payload carries — the same degradation as a dropped connection. The hook also drops read events when the viewer is a guest, so a client that somehow meets one still shows nothing.

This is the same class of problem that put Typing on its own channel ([ADR-0033](docs/adr/0033-typing-rides-its-own-ephemeral-channel.md)). Staying on the room channel is still right for the publish direction — Core publishes and clients only subscribe, so nothing needed widening — but subscribe breadth was itself the leak. Worth knowing if receipts on External channels ever need to be live.

## Scope

Humans only; Coworkers and Soko Bots never get a face. Room-level only — a face means "read the room", never "read your Thread reply". Web only; the DTO field is additive so the native clients can adopt it later. No backfill: a member with no read-state row simply reads as never-read.

## Verification

- `pnpm typecheck` — clean
- `pnpm check` (Biome) — clean
- `pnpm --filter web test` — 829 files, 6972 passed
- `pnpm --filter core test` — 543 files, 6071 passed
- Locale parity across `en` / `de` / `es`, with real translations

New tests, all without a database or a live Ably connection: the room mapper's guest rule and roster-driven mapping; the mark-read route publishing exactly one event, staying silent with a guest on the room, and surviving a publish failure; the hook's seeding, merge, out-of-order protection and guest guard; the avatar stack's cap, ordering, accessible name, keyboard path and 44px touch target; and the transcript faces appearing on the newest message only.

**Not verified in a browser.** The trailing-faces layout and the reserved gutter are proven by tests and a static prototype, not by the running app — worth a look before this leaves draft. Note the gutter narrows every line of the newest message by the faces' width, which is inherent to reserving space this way.


mrosberghaus:
## Grok review

Range: `bf8afac93` (`origin/main`) .. `f88814d22`. In-process against requesting-code-review (no spawn tool in this child).

### Fixed
Nothing pushed — no verified Critical/Important product bugs.

### Remaining
- Important: `roster-row-ideas.html` — 485-line scratch prototype at the repo root (hex palette, unreferenced). Drop it before this leaves draft; it will otherwise land on `main`.
- Minor: `apps/web/src/app/(app)/chat/components/room-message-row.tsx` — comment still claims a 16rem hover gutter for the action pill; that gutter was removed. Faces are `end-2`, which matches the CSS, not the comment.
- Minor: header stack is 4 faces of the full roster (viewer + machines, recency-sorted); transcript `READ_RECEIPT_FACE_CAP` is 3. `CONTEXT.md` still says “three most recent readers” in the header.
- Minor: mobile `after:-inset-3.5` on the transcript control can overlap the last letters of the newest line when `+N` is showing. `md:after:hidden` hides that on pointer.
- Skipped / declined to judge: live receipts off for every member of a room that has a guest — documented, three-layer (mapper + no publish + hook). Browser layout of trailing faces / gutter — author already flagged; not verified here.

### Assessment

Core guest boundary, roster-driven mapping, out-of-order merge, and newest-message-only faces hold up; drop the root HTML and glance at the corner layout in a browser before this leaves draft.

mrosberghaus:
All four addressed in `b88270d4b`.

**`roster-row-ideas.html`** — correct, and worse than the note says: it was already *tracked*, committed by accident in `00e9843c2`, so it was on course for `main` rather than merely sitting in the working tree. Untracked with `git rm --cached` so it stays on disk beside the other scratch prototypes, which are untracked and stay that way.

**Stale 16rem comment** — confirmed. The comment justified anchoring the faces to the row rather than the column by pointing at the hover gutter, and a later commit in the same branch deleted that gutter. The reasoning it preserved is now void; `end-2` is simply the action pill's edge, so the two share a vertical line. Comment rewritten to say that and nothing more.

**`CONTEXT.md` on the header stack** — correct, and wrong on two counts rather than one. It said "three most recent readers"; the header is four faces, and they are the *whole roster* — viewer and machines included — merely ordered by read recency, with a `+N` that counts the room rather than the readers. That is deliberate (one stack, two facts) but the glossary described the transcript's rule instead. Entry now distinguishes the two surfaces.

**`after:-inset-3.5` on touch** — reproduced by arithmetic. Three faces plus the `+N` is 52px; the hit area reaches 14px further left, so the corner occupies 66px, and the reserve was `pe-14` = 56px. 10px of overlap onto a full-width last line, below `md` only. Reserve raised to `pe-20` = 80px, which clears the target rather than the faces. The comment now carries the measurement so the next person changing the face size can see what the number is made of.

On the two you skipped: the guest boundary is three-layer on purpose (mapper nulls it, the route declines to publish when a guest is on the roster, the hook guards both the event path and the derivation) — belt and braces because any one of them failing alone leaks read times across an org boundary. The corner layout has since been checked in a browser and moved twice off the back of it: the faces were landing in the middle of the hover gutter, and the gutter itself is now gone so the transcript uses the full width.

Not in this PR, flagged for the record: the roster changes on this branch — viewer pinned first, people/Coworkers split, recency sort, row redesign — are beyond SOK-1110's spec, as is the full-width transcript change. Happy to split either out if you'd rather review them separately.

PR #4962, opened 2026-09-21
perf(apple): skip retry-source copies for ordinary messages

mrosberghaus:
Room and thread rendering checks retry eligibility for ordinary messages while scrolling. The coordinator previously constructed the entire retry-source array before the existing eligibility predicate rejected those rows. It now checks for a failed mention shell with a source ID first; valid failed shells retain the existing loaded-source and ownership checks.

At 2,000 messages, the actual check drops from 0.117340 to 0.000271 ms. A counter in the retained disposable harness fails before the change and passes afterward: 2,100 ordinary checks construct 2,100 source collections before and zero after. The first 120 native scroll events likewise drop from 4,022 source collections to zero, with nearly identical row-body counts.

One paired Time Profiler capture shows about 5% less sampled main-thread work, but effectively unchanged phase duration. This is a focused removal of measured work, not a claim that all scrolling sluggishness is fixed. The prepared-message projection remains a measured contributor, and M6 stays open. ScrollView, LazyVStack, rendering, geometry behavior, dependencies and Xcode/test package linkage are unchanged.

Validation: 858 workspace tests pass (99 app plus all five package suites), including new ordinary/thinking/orphan/missing-source cases and sources loaded only in the thread parent or replies. The Release harness and iOS 17 Workspace build pass, as do pinned SwiftFormat and strict uncached SwiftLint. Six unprofiled plain/mixed cases at 50/500/2,000 messages reach the top with zero source collections, unchanged preparation frequency and inactive distant sentinels. Harness extraction/replay and diff checks pass.

Extends the account-free measurement harness and retains the before/after evidence from #4957. Binary traces remain local; capture/analysis commands and parsed metrics are in the docs. Commits use HUSKY=0 as requested.


mrosberghaus:
## Grok review

Reviewed `8d7ed372c16101d7cb984d373073a59992bc25e1..54f6ddcf5ad4bbd7f3b9497831ba289587fbfbdf` against the PR title/body. Product change is the early failed-shell-with-source-ID guard in `WorkspaceState.canRetryMention` so ordinary rows do not evaluate `mentionRetrySources`.

### Fixed

### Remaining
- Minor: `apps/apple/docs/scrolling-performance-harness.md` (`setup.py`): the `M6MentionRetryProbe` patches sit after `print(out)`. They still run before the process exits; they just look appended after the “done” line.
- Minor: `apps/apple/Packages/SokosumiWorkspace/Sources/SokosumiWorkspace/WorkspaceState+Mentions.swift`: the `.failed(_, _?)` match duplicates `CoworkerMentionShell.canRetry`. An `@autoclosure` on `sources` would avoid the copy without a second parse; not needed for this slice.
- Minor: `apps/apple/PARITY.md`: “draft #4962” will read stale after merge.
- Skipped / declined: overall M6 scroll feel, `List` conversion, and the prepared-message projection — the PR says those stay open.
- Skipped / declined: whether the ~5% sampled-weight drop is user-visible — phase duration is documented as unchanged.

### Assessment

Guard preserves existing eligibility (room, failed shell with source ID, loaded source, mentioner-only), is shared by room and thread, and is covered by ordinary/thinking/orphan/missing-source plus thread-parent/reply tests. Harness numbers in `scrolling-retry-guard-results.json` match the write-up (2,000-message check 0.117340 → 0.000271 ms; first 120 wheel events 4,022 → 0 source copies). Wait for Apple CI; do not treat this as M6 closed.

PR #4959, opened 2026-09-21
fix(apple): drop deleted messages from the transcript like web

mrosberghaus:
## What changed

Apple parity row **19a** (user decision 2026-09-21: Apple follows current web). A deleted message now leaves the room transcript and thread replies; "This message was deleted" remains only on a deleted thread parent rendered as the thread root.

- `SokosumiChat/OutboundRoomMessage.swift`: new `shouldKeepPersistedMessage`, a port of web's whole predicate (membership, non-blank body, quote, coworker mention shell). The existing `displayedTranscript(messages:shells:)` filters with it. That function was already the single display merge behind `WorkspaceState.displayedTranscript` and `ThreadSession.displayedReplies`, so history pages, gap/jump windows, full realtime events, id-envelope tombstones and the HTTP delete response are all covered in one place. No view filters anything and there is no second path.
- State keeps the tombstoned row (as web's `messagesState` does), so realtime patches, reply-count reconciliation and the open thread's parent keep working unchanged.
- `isHiddenSokoBotMentionShell` (slice 38) was the same predicate narrowed to one sender; the full predicate subsumes it, so it is removed.
- `WorkspaceState+MessageNavigation.swift`: `openMessage` returns `.unavailable` for a target the transcript drops, so a quote card or Message link to a deleted message cannot load a window and then wait for a row that never appears.
- `MessageRowView`'s tombstone branch stays; `ReplyThreadView` renders the thread root through it. Nothing became unreachable, so no rendering code was removed.
- `PARITY.md`: row 19a "In review" with evidence, row 26b Done (#4946 merged 2026-09-21, VoiceOver note kept), checkpoint and work order advanced to 13a, new "Slice 19a" notes.

No `apps/web` or `apps/core` change, no Core snapshot, dependency, project or entitlement change.

## Web sources audited (at `e0cf0a511`)

- `apps/web/src/app/(app)/chat/utils/merge-room-messages.ts`: `shouldKeepPersistedMessage` :173, `isMentionCoworkerShell` :163, `mergeMessagesWithStreamOverlay` :195 (filters at display time), `applyFullChatRoomMessageEvent` :21 (tombstones still merge into state).
- `apps/web/src/app/(app)/chat/components/rooms-client.tsx`: room display :1142 and thread replies :1181 both go through the filter; `threadParentMessage` does not; `handleDeleteMessage` :2121 is not optimistic.
- `apps/web/src/app/(app)/chat/components/room-message-row.tsx` :2605: the remaining tombstone copy.
[...]

PR #4965, opened 2026-09-21
fix(apple): open the mention picker without typing into the draft

mrosberghaus:
Apple parity row **13a**. Built on `origin/main` `26f8e9e0f` (includes #4962).

## What changed

The composer's mention button typed a separator plus "@" into the draft, so dismissing the list left stray characters in the message. It now opens the list over the unchanged draft and selection, and only accepting a row edits the text.

- `Packages/SokosumiChat/.../ComposerMention.swift`: new `ComposerMentionPicker` beside `ComposerReferenceTrigger`. It holds the button-opened flag and answers two questions for both paths: `query(typed:)` (empty while the button opened the list, the typed mention query otherwise) and `insertion(in:selection:typed:)` (the typed trigger's range, or for the button the selection's end plus the separator rule).
- `MacComposerCommands`: `beginMention` became `openMentionPicker` (focus, open, refresh; no text edit). `refreshSuggestions` takes the mention query from the picker and leaves the typed path's trigger and dismissal tracking as it was. `dismissSuggestions` (Escape, `textDidEndEditing`) and `acceptMention` close the picker; `closeMentionPicker` closes only a button-opened list.
- `MacComposerTextInput`: `acceptMention(_:picker:)` asks the picker where to insert and calls the existing `insertReference`, which gained a leading separator and now places the caret after what it inserted. No second insertion path. `updateNSView` closes a button-opened list when the draft is replaced from outside (send, draft swap).
- Room, thread and inline edit composers all render `ComposerTextInput`, so all three change together. There is no iOS composer: every input-owning file under `Chat/Composer/` is `#if os(macOS)` and the project has no iOS target.
- `PARITY.md`: row 13a in review with evidence, row 19a Done (#4959 merged; its "live deletion unverified" note kept), work order advanced so 21a is next.

## Web sources audited (at `26f8e9e0f`)

- `apps/web/src/components/chat/composer-wysiwyg-editor.tsx`
  - `openMentions` :949 — focuses, sets `manualMentionOpenRef`, opens with `query: ""`, inserts nothing, does not toggle.
  - `handleInput` :556 — re-opens a button-opened list with `query: ""` before looking for any typed trigger; `getLiveSuggestionAtCaret` :980 and `syncSuggestionsWithCaret` :997 agree, so typing does not filter and caret moves do not close.
  - `insertMention` :605 — typed `@query` first (replaced, no separator); else for the button path the caret, after a space when the previous character is not whitespace (:629); chip plus `shouldAppendTrailingSpace`.
  - `closeSuggestions` :418 clears the flag: Escape :1163, blur after 150 ms :1296, select, emoticon/shortcode conversion.
- `apps/web/src/components/ui/mention-textarea-utils.ts` — `getCaretOffset` :341 (selection end), `isWhitespaceChar` :90, `shouldAppendTrailingSpace` :94, `getActiveTrigger` :431.
- `apps/web/src/app/(app)/chat/components/room-composer.tsx` :835-848 — the button; it prevents its own mousedown so the editor keeps its selection.
- `composer-wysiwyg-editor.test.tsx` :340 — web's own pin: `hello` + button + Alice gives `hello @Alice`.

## Behaviour decisions

Mirrored web:
- Press: nothing typed, selection unchanged, empty query, first row highlighted. A second press does not toggle.
- Typing under a button-opened list goes into the draft and does **not** filter the list; the text stays on dismiss and on select. This is what web's code does in three places, so it is mirrored rather than "fixed".
- Select: a typed `@query` at the caret is replaced (also under a button-opened list); otherwise the chip lands at the selection's end with selected text kept, after a space when the previous character is not whitespace, with the usual trailing space. A parameterized test asserts text, serialized draft and caret equal the typed "@" path.
- Channel and emoji rows are hidden under a button-opened list; caret moves do not close it.

Kept Apple:
- The button does nothing while an input method has marked text (the typed path's existing guard).
- An emoji conversion does not close the list (web closes it as a side effect of an early return).
- A button-opened list closes when the draft is replaced from outside; web gets that from the editor's blur, and the Send button here takes no focus.

Row wording vs web: the row says the picker opens "at the caret". On web that is the trigger offset; in the room composer the list is placed on the composer card, which is where Apple's panel already sits for both paths. Nothing was ambiguous.

## Tests

- Removed `MacComposerTextInputTests.mentionShortcutStartsQueryAtSelection`: it asserted the defect (`"Hello @\n"`, caret 7). Replaced by `buttonOpensTheListAndLeavesDraftAndSelectionAlone` with the opposite expectation over an empty draft, end of text, mid-word and a non-empty selection. No other existing test changed.
- New package suite `ComposerMentionPickerTests` (state boundary) and app suite `MacComposerMentionPickerTests` (glue), including a regression test of the typed path's filter, Escape and re-open.
- New fixture `MentionButtonPickerFixtureTests`: the real `ComposerTextInput` with "Hello", the real button pressed by a mouse down/up, light and dark.

## Verification (from `apps/apple`, on the merged tree)

- `swift test --package-path Packages/<each>`: CoreAPI 1, SokosumiAuth 35, SokosumiChat 567, SokosumiRealtime 48, SokosumiWorkspace 125 — all passed.
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi -configuration Debug -destination 'platform=macOS,arch=arm64' -skipPackagePluginValidation DEVELOPMENT_TEAM= CODE_SIGN_IDENTITY=- -enableCodeCoverage NO` — TEST SUCCEEDED on the first run, 890 tests, 0 failed.
- Same invocation with `build` — BUILD SUCCEEDED.
- `mint run swiftformat --lint .` — 0/382 files; `mint run swiftlint lint --strict` — 0 violations in 382 files.
- Fixtures inspected in light and dark (`apps/apple/docs/screenshots/composer-mention-button-picker-{light,dark}.png`): list open above the composer, editor reads "Hello" with no "@". The images were recorded as test attachments and exported from the result bundle, so no unsandboxed build was needed.

## Unverified

The app was not launched. Not exercised live: the button pressed with the mouse while the editor holds a selection, clicking a row with the mouse, Send with the list open, switching rooms with the list open, the button in the thread and edit composers, and marked text. Whether a click on the transcript takes first responder from the editor, and so closes the list the way web's blur does, is not known; the typed "@" list has the same dependency today.

No web, Core, API snapshot, dependency, project or entitlement change.


mrosberghaus:
## Grok review

Review range `26f8e9e0f..46be20c64` against `origin/main`. Requirements: PR title/body and PARITY row 13a (mention button must not type into the draft).

### Fixed

### Remaining
- Minor: `MacComposerMentionPickerTests.swift` fixture starts with an unfocused editor and never asserts the suggestion overlay (only `"Hello"`, no `"@"`, and first responder). Web pins the real click path: editor focused → blur → `openMentions` still open after the deferred close (`composer-wysiwyg-editor.test.tsx` “keeps the mention listbox open when openMentions runs after editor blur”). Apple `textDidEndEditing` hops `dismissSuggestions` onto the next turn, so if a toolbar click resigns the `NSTextView` before `openMentionPicker`, that Task would close the list the button just opened. Not reproduced here; the PR already lists focused-editor mouse press as unverified.
- Minor: `MacComposerMentionPickerTests.swift` click point is hardcoded (`12+12+28+12+28+12+14`). A padding or `@ScaledMetric` change can hit the emoji button instead; that button also focuses the editor, so the current assertions would still pass.
- Minor: PARITY row 13a still cites `beginMention` at `MacComposerCommands.swift:142`. That line is now `struct LinkEditor`.
- Skipped / disagreed: treating the untested focused-click path as a must-fix. AppKit does not always resign an `NSTextView` when a non-focusable SwiftUI `Button` is clicked (unlike contenteditable). No code change without a failing pin.
- Declined to judge (outside this slice / already recorded): live Send with the list open, transcript click-outside, thread/edit composers, marked text, emoji conversion not closing the list (explicit Apple keep).

### Assessment
Ready to merge: Yes

Root cause is gone: the button no longer inserts a separator plus `"@"`. `ComposerMentionPicker` in `SokosumiChat` owns the button-opened flag and answers query plus insertion for both paths; `acceptMention` still goes through `insertReference`. Package tests cover query/separator/selection; app tests invert the old defect and pin typed-path filter/Escape. CI `Xcode test` is green. Remaining notes are test/docs nits, not merge blockers.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/4965#issuecomment-5760521620) — follow-ups are in `f0195c1e2`.

- **Blur closing the button-opened list** — pinned, and the pin was red, so this got a source fix. `MentionButtonPickerBlurTests/buttonPressedRightAfterABlurKeepsItsList` drives the real `Coordinator.textDidEndEditing` (`window.makeFirstResponder(nil)`), then `openMentionPicker`, then lets the deferred work run: `mentionOptions` came back `[]` instead of `["anna", "bob"]`. Fix mirrors web, where `openMentions` clears `blurTimeoutRef` (`composer-wysiwyg-editor.tsx`:954-956): the deferred dismissal is now a cancellable `Task` (`MacComposerCommands.dismissSuggestionsAfterBlur()`), and `openMentionPicker` cancels it. `plainBlurStillClosesATypedAndAButtonOpenedList` passes before and after, so an ordinary blur still closes both kinds of list.
  On whether a real click can hit this: in the test host it cannot. The fixture now focuses the editor and sends a real mouse event to the button; zero `NSText.didEndEditingNotification`s were posted and the editor stayed first responder, which matches your note that a plain SwiftUI `Button` does not take focus. The race needs a blur and a press in the same turn (programmatic or possibly keyboard paths, not tested). The fix stays because it is the web behavior and costs one stored task.
- **Hardcoded click point** — partly fixed. The point is now derived from the editor's real frame plus `ComposerLayout`'s insets rather than window constants; it still encodes 28-point buttons and 12-point gaps, and the test comment says so. An accessibility lookup was not possible: the hosted composer's `NSHostingView` reports an `AXGroup` with no children in the test host. The weakness you named is closed differently: the test asserts `suggestionKeyHandler?(125)` is false before the click and true after it, which only holds while a suggestion list is showing, so a click landing on the emoji button now fails.
- **PARITY cited `beginMention` at `MacComposerCommands.swift:142`** — fixed; row 13a describes the old defect in the past tense and points at `openMentionPicker`. A "Review follow-up (13a)" note records the above.

Verification on `f0195c1e2`: SokosumiChat 567 passed; full workspace `xcodebuild test` 892 tests, 0 failed, run twice (the second after moving a test helper to file scope for `large_tuple`/`nesting`); strict SwiftLint/SwiftFormat clean. Still not exercised in a running app: a mouse press with a selection, clicking a row, Send with the list open.

PR #4970, opened 2026-09-21
perf(apple): reuse transcript projection within view updates

mrosberghaus:
Scrolling rebuilt the same live prepared-message projection four times per room-view evaluation and three times per thread-view evaluation. Both views now compute it once and share it with rows, jump readiness, and the last-message observer. Every update still reads live data; the existing SwiftUI scroll containers, state ownership, and preparation lifecycle remain in place.

The retained account-free harness compares 12 cases per version: room/thread × 50/500/2,000 messages × plain/mixed content. Its deterministic check fails before and passes after. In the 2,000-message plain room, repeat-scroll projection time falls from **383.784 to 132.235 ms** for 101 view evaluations. A paired Time Profiler capture reduces projection sample weight from **396 to 125 ms**, but total sampled main-thread work and phase duration are effectively unchanged. **An overall scrolling speedup is not established; M6 remains Partial.** One linear projection per update and live interaction acceptance remain open.

Validation:

- Full workspace suite: **867 tests passed** before and after; final app-only rerun: **100 passed**. Includes 16 content-growth cases, four rich/media scroll cases, and four light/dark room/thread jumps requested before preparation completes. All four jump snapshots inspected. Jump tests also verify the room/thread navigation state and recognize the requested message in the captured bitmap; a wrong-target control fails all four rendered-target checks.
- Both Release harness builds, iOS 17 Workspace build, pinned SwiftFormat, and strict uncached SwiftLint pass. No project linkage or dependency changes.
- All 12 updated benchmark cases pass projection, scrolling, preparation, distant-row, and retry-copy checks. Inlined harness recreates measured probes byte for byte; original 21-case data validator still passes.

Numbers, method, startup-preparation differences, raw phase summaries and trace bounds are retained in `scrolling-performance-audit.md`, `scrolling-performance-harness.md`, and `scrolling-projection-results.json`.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/Sokosumi/Chat/Timeline/RoomTimelineView.swift`: hoist comments narrate the change (`Share the live projection…`, `Existing view composition, now parameterized…`) instead of a non-obvious constraint.
- Minor: `apps/apple/SokosumiTests/Chat/Timeline/TranscriptScrollingTests.swift`: thread/dark jump cases still assert `state.messageJump == nil` (room jump) and `distanceFromBottom > 400`, not `thread.jumpTarget` or highlight. Landing is a manual PNG check.
- Minor: `apps/apple/PARITY.md`: “projection reuse awaiting review” goes stale on merge.
- Skipped / disagreed: extract `messageList` into a new `View` type — this PR keeps scroll-container and state ownership; `overlaying` already returns a new array every update, so a child struct would still invalidate.

Declined to treat as in-scope: persistent projection cache, `List` conversion, and live-room FPS. The PR states those remain open and M6 stays Partial.

### Assessment
Ready to merge: Yes

Room and thread compute `preparedMessages` once per view evaluation and pass that live snapshot to rows, jump readiness, and last-message observers. That matches the claimed 4→1 / 3→1 split without changing overlay semantics, jump-before-prepare, or the ScrollView/LazyVStack lifecycle. The deterministic `--projection` check is the right gate; overall scrolling speedup is correctly not claimed.

PR #4979, opened 2026-09-21
feat(core): let developers create a Vendor they administer

Sarthib7:
## Summary
- Core: `POST /v1/vendors` for signed-in developers; caller becomes `admin` (platform `POST /v1/admin/vendors` unchanged).
- CLI: `sokosumi vendors create --name NAME --slug SLUG`.
- CLI: `coworkers register --create-vendor --confirm-create-vendor --vendor-name NAME --vendor-slug SLUG` creates then registers (not both with `--vendor-id`).
- Gate copy points at `vendors create` plus existing-admin invite.

Web Developer create form / OpenAPI client regen still follow-up (`generate:core:snapshot` currently fails on main for an unrelated utils export).

## Test plan
- [x] Core `vitest run src/routes/v1/vendors/post.test.ts`
- [x] CLI registration-authority + coworkers + vendors tests
- [ ] Manual: `POST /v1/vendors` as non-admin user → 201 + admin on `/v1/vendors/me`
- [ ] Manual: `sokosumi vendors create` then `vendors me`
- [ ] Manual: `coworkers register --create-vendor --confirm-create-vendor --vendor-name … --vendor-slug … --name …`


Sarthib7 (commented):
## Local E2E verification + advisor review

Verified developer Vendor create locally end-to-end against Core on `127.0.0.1:8787` (Postgres 17, all 351 migrations applied). CI on the head is green.

**Test plan (all three items pass, DB-grounded):**

1. `POST /v1/vendors` as a non-admin user → 201, `role: admin` on `/v1/vendors/me`. Confirmed `vendor` + `vendor_member(role=admin)` rows.
2. `sokosumi vendors create` then `vendors me` → same vendor, `role: admin`. (Headless key via `SOKOSUMI_API_KEY`; the `--api-key` flag does not exist; use the env var or `auth login --api-key-stdin`.)
3. `coworkers register --create-vendor --confirm-create-vendor --vendor-name … --vendor-slug … --name …` → coworker created and linked to the new vendor with the caller as vendor admin. **Note:** this only completes as a platform admin, because `POST /v1/coworkers` is still `requireAdminAuthContext` (`apps/core/src/routes/v1/coworkers/post.ts`). A plain developer 403s here, after the vendor already committed.

### The endpoint itself is correct

`apps/core/src/routes/v1/vendors/post.ts` creates vendor + admin membership in one `prisma.$transaction`, derives `userId` from the auth context (no mass-assignment/IDOR), hardcodes `role: "admin"`, reuses `createVendorRequestSchema`, maps slug conflicts to 409, and leaves `/v1/admin/vendors` untouched.

### Blocking concern: the new capability boundary, not the code

Making vendor-admin **self-service for every authenticated user** transitively exposes the pre-existing `POST /v1/vendors/{id}/members` route:

- A vendor admin adds members **by email**, case-insensitive (`helpers/vendor-membership.ts:164-166`), returning **404 vs 201** → registered-email enumeration oracle.
- The 201 body returns the target's **id, email, name** (`helpers/vendor.ts:28-31`) → PII disclosure of any registered user.
- The add happens with **no invitation/consent** (`[id]/members/post.ts:91-106`).

Before this PR, vendor-admin was a curated set (platform-admin-created). This PR opens that oracle to every account. Suggested fixes: gate member-add behind an invitation/consent flow, make the lookup response uniform, or restrict member-add on self-created vendors.

### Secondary

- **Unbounded vendor creation** into a globally visible list (`GET /v1/vendors` returns all vendors to any user; slugs globally unique; no per-user cap or route rate limit) → slug/name squatting. `registration-authority.ts:49` now funnels every blocked developer here.
- **CLI `register --create-vendor` orphans a vendor** for non-admins: the vendor commits, then `POST /v1/coworkers` 403s, and there is no developer delete path. Reproduced. `register` output also omits the created vendor, so a retry 409s on the slug with no hint to switch to `--vendor-id`.
- **Logo strings** unvalidated/unbounded: `vendorLogoSchema = z.string().nullable()`, no `.max()`, no URL check, no `bodyLimit`; persisted and served via `GET /v1/vendors`. Hardening.

### Polish (non-blocking)

- `vendors.ts:17-23` reimplements `optionString` instead of the shared `command-helpers` one (drops last-wins repeat-flag behavior).
- Missing-flag errors don't name `--name`/`--slug` (`vendors.ts:51-58`).
- Dead `vendor.role ?? "admin"` fallback (`vendors.ts:60-65`).
- Flag-conflict check runs after a network round-trip (`coworkers.ts:202-206`).
- `post.test.ts` has no invalid-slug (422) or unauthenticated (401) case.

mrosberghaus (changes requested):
## Review

Do not merge this as it stands. The vendor-plus-admin write is the right shape, and `POST /v1/admin/vendors` is unchanged. The CLI still rejects a foreign `--vendor-id` before create. Two blockers below.

Checked at `2ee31a4ffaba4bbe3ed7ccaa14b59427a7394d61`.

### What holds

- `POST /v1/vendors` creates the vendor and the caller's `admin` membership in one transaction. `userId` comes from auth context. `role` is hardcoded `"admin"`, not taken from the body.
- Slug clashes return 409 via `isSlugUniqueConstraintError`.
- The handler reuses `createVendorRequestSchema`, `mapVendor`, and `vendorLogoCreateData`.
- `discover` only lists `vendors create`. `--create-vendor` still requires `--confirm-create-vendor` and still cannot be combined with `--vendor-id`.
- Tests run from a detached worktree at that commit: Core `vitest run src/routes/v1/vendors/post.test.ts` (3 passed); CLI `tsx --test` on `registration-authority.test.ts`, `coworkers.test.ts`, `vendors-workspaces.test.ts`, and `bin.test.ts` (35 passed). Those tests mock Prisma and HTTP. They do not cover the 403 on coworker create.

### Must fix

1. **`coworkers register --create-vendor` writes a vendor, then fails for the developer it addresses.**

   Order in `apps/cli/src/cli/commands/coworkers.ts`: confirm, workspace check, `POST /v1/vendors`, then `POST /v1/coworkers`. Coworker create still calls `requireAdminAuthContext` (platform `User.role`, not vendor admin) at `apps/core/src/routes/v1/coworkers/post.ts:79`. A normal signed-in user gets 403 after the vendor row is committed. There is no developer-facing vendor delete (only the platform-admin delete). A retry with the same slug is 409 at `apps/core/src/routes/v1/vendors/post.ts` and does not continue registration. The same command also creates the vendor before the local coworker `name` check, so a platform admin who omits `--name` orphans a vendor too.

   Gate copy in `apps/cli/src/cli/registration-authority.ts` tells that blocked developer to run this command. Before this PR, confirm refused and wrote nothing.

   Keep `register` from calling `POST /v1/vendors` until a vendor admin can create a coworker. Point the gate at `sokosumi vendors create` and say coworker create is still platform-admin only. If the chain stays, require coworker `--name` first, and only create the vendor when the follow-up can succeed.

2. **Any user actor becomes vendor admin, and that role is a user-directory oracle.**

   `apps/core/src/routes/v1/vendors/post.ts` only checks that the caller is a user (session, user API key, or OAuth). The new test expects `role: "user"` to get 201. There is no separate developer role, so this matches the PR test plan, and it is also the problem: vendor admin used to be someone a platform admin appointed. One call now grants `admin`.

   `POST /v1/vendors/{id}/members` then looks up any email, case-insensitive, with no invite and no rate limit (`apps/core/src/helpers/vendor-membership.ts`, `resolveUserIdFromIdentity`). Unknown email is 404 `"User not found"`. A hit inserts a member with no consent and returns `id`, `email`, and `name`. A second probe of the same person is 409, which also confirms the account. The caller can pass `role: "admin"` and appoint more people who can do the same lookup.

   The lookup itself is existing behavior for appointed vendor admins. This route is what makes it available to every account. Hold self-service admin until add-by-email stops distinguishing registered users: same response whether or not the email exists, no name or email in the body until that person accepts, and a rate limit. A pending invite is enough. An immediate membership row is not.

### Should fix

**"One-time" create is unlimited, and every row is on the global vendor list.**

The OpenAPI text says one-time cold-start. Nothing caps vendors per user, and nothing requires the organization workspace that registration already requires. `slug` is globally unique, so any account can take a name or fill the namespace. `GET /v1/vendors` is the grant picker and returns every vendor, with no pagination. New rows show up for every signed-in user. Grant approval itself stays with the org owner or admin, so this is squatting and list abuse, not an automatic workspace grant. Logo strings are unbounded `z.string()`, unlike name and slug (max 120), and this route now accepts them from any user. A same-slug retry by the owner is 409 instead of the vendor they already administer.

Require an organization workspace, cap how many vendors one user may administer (start at 1), keep self-serve vendors off the global picker until a platform admin lists them, cap logo length, and if the caller is already admin of that slug, return that vendor instead of 409.

### Later

- `vendors create` help omits `--name` and `--slug` because command usage is `""` (`apps/cli/src/cli/index.ts`). Missing flags throw `"Vendor name is required"` instead of the usage line. Put `--name NAME --slug SLUG` in `COMMAND_USAGE`.
- `vendor.role ?? "admin"` in `apps/cli/src/cli/commands/vendors.ts` never runs. `createVendor` already throws when `role` is not `admin`.
- `apps/cli/docs/developer-cli-implementation-plan.md` still says platform-admin create is the fallback. The gate copy no longer says that.
- Handler tests skip 401 and a slug that fails `^[a-z0-9]+(?:-[a-z0-9]+)*$`. The mutual-exclusion check in `coworkers.ts` runs only after the workspace GET. Move it next to the confirm check.

Web Developer create form and OpenAPI client regen can stay a follow-up. The CLI posts JSON itself. Do not hand-edit the generated client.

**Verdict:** no. Ship `vendors create` and `POST /v1/vendors` only after the directory lookup cannot be used by every new account, and after create is capped. Leave `register --create-vendor` refused until vendor admins can create coworkers.

Sarthib7:
@mrosberghaus thanks for the review. Pushed `2ee31a4ff..a73b927d5` addressing both blockers, all should-fix, and the later items. Three commits:

- `66f15b982` feat(cli): drop vendor-create from `coworkers register`
- `6177801d0` feat(core): replace vendor member-add oracle with a pending-invite model
- `a73b927d5` feat(core): harden developer vendor create against abuse

**Blocker 1: `register --create-vendor` orphan.** Removed the chain entirely. `coworkers register` no longer creates a vendor; it requires an administered `--vendor-id`. Developer vendor create is only the standalone `sokosumi vendors create`. Gate copy in `registration-authority.ts` points there and states coworker create is still platform-admin only. `--create-vendor`/`--confirm-create-vendor`/`--vendor-name`/`--vendor-slug` are gone from the parser.

**Blocker 2: member-add directory oracle.** Removed `POST /v1/vendors/{id}/members` (add-by-email/userId). Replaced with an email-only pending-invite model:
- `POST /v1/vendors/{id}/invites` always creates a `PENDING` invite whether or not the email is registered, returns only the invite (id/email/role/status/expiresAt) with no invitee user id or name, and is rate-limited (live-pending-per-vendor plus creates-per-inviter/hour). Already a member of that vendor returns 409 (no new signal; the admin sees their own member list).
- A `VendorMember` row is created only when the invitee accepts under their own auth: `GET /v1/vendors/invites` (mine), `POST /v1/vendors/invites/{id}/accept`, `.../decline`. Admin management: `GET /v1/vendors/{id}/invites`, `DELETE .../{inviteId}` (revoke).
- New `VendorMemberInvite` model plus migration; a partial unique index keeps one live pending invite per (vendor, lowercased email). Mirrors the chat-room invitation subsystem.

**Should-fix: vendor create.** Require an organization workspace (server-side match to the CLI gate); cap self-service vendors per user at 1 (`createdByUserId`); keep self-service vendors off `GET /v1/vendors` until a platform admin lists them (`listed` flag, picker filters `listed = true`); return the caller's existing vendor on a same-slug retry instead of 409; cap logo string length.

**Later.** `COMMAND_USAGE` names the flags via the create error (kept the help list free of a bare double-dash to satisfy the existing help-format test); removed the dead `vendor.role ?? "admin"`; updated the plan doc and SPEC V85/T37. Handler test note: a 401 case is not meaningfully testable in the stubbed-auth unit harness, so I left it; the invalid-slug regex is covered by the schema.

**Verification.** CLI `208 passed`; core vendor suites `80 passed` (incl. new `vendor-invites.test.ts`); full-monorepo typecheck plus biome clean via the pre-commit hook.

**Not yet verified:** I could not apply the migration or run the live manual flows here (no local Postgres). The branch is green on typecheck/tests; `prisma migrate deploy` plus the manual test-plan items still need a real DB. OpenAPI client regen and the Web developer create form stay a follow-up as agreed (the CLI posts JSON itself; no hand-edited generated client).

Sarthib7:
**Follow-up push `390ca22b0..4f732638d` (3 commits): remaining should-fix + secondary hardening.**

- `534941c96` fix(core): remove user-directory oracle from vendor member and assignment routes. The invite rewrite closed `POST /v1/vendors/{id}/members`, but member `PATCH`/`DELETE` and coworker assignment `PUT`/`DELETE` still resolved a path segment or body field as user id **or** email, giving the same registered-vs-unknown 404 signal. Dropped `resolveUserIdFromIdentity`/`resolveUserIdFromUserIdOrEmail`; those routes now take a user id only. `assignCoworkerRequestSchema` loses its email branch. Tests assert an unknown email and a registered non-member return the identical 404 and never hit `user.findFirst`.
- `8defaafeb` feat(core): cap self-service vendors per user and gate listing. Unique index on `vendor.createdByUserId` (Postgres allows many NULLs, so platform vendors are uncapped) caps self-service creation at 1, enforced by a pre-check count and the constraint race mapped to 409. Self-service vendors are created `listed: false` and stay off `GET /v1/vendors` until a platform admin publishes them via the new `listed` flag on `PATCH /v1/admin/vendors`. A same-slug retry by the owner returns the vendor they already administer instead of 409. Logo strings gain a max-length bound. New migration `20260922183000_vendor_self_service_cap`.
- `4f732638d` fix(core): upgrade developer to admin on accepting an admin vendor invite. Accepting as an existing member previously kept the old role. A developer accepting an admin invite now updates to admin; an admin accepting a developer invite keeps admin (no self-demote of the last admin). Beyond the review scope, flagged for visibility.

**Verification.** `apps/core` vitest across the vendor suites: 79 passed (8 files). Turbo typecheck: 17/17 successful. Biome clean via pre-commit.

**Still not verified against a live DB.** The cap has a pre-check plus a unique-constraint catch, but `prisma migrate deploy` and the manual test-plan items still need a real Postgres.

mrosberghaus (changes requested):
## Follow-up review

Not ready yet. Both blockers from my last review are fixed, and the should-fix list is done. Replacing add-by-email with invites was the right call. The invite model has one new hole that needs closing before merge, plus a few Important items below.

Checked at `4f732638d35b24b1e0271df4dc9becf8a6c6f367`.

### What's fixed

- **`register --create-vendor` orphan.** `coworkers register` no longer creates a vendor. It requires an administered `--vendor-id`. The gate copy points at `sokosumi vendors create` and says coworker create is still platform-admin only.
- **Directory oracle.** `POST /v1/vendors/{id}/members` is gone. Invite create returns the same response whether or not the email is registered, returns no invitee PII, and is rate limited inside a serializable transaction. The only 409 is for someone already in this vendor, and the admin can already see that in the member list. Member `PATCH`/`DELETE` and assignment `PUT`/`DELETE` take a user id only. Accept and decline return 404 for "not yours", "expired" and "already resolved" alike.
- **Abuse guards.** Create requires an organization workspace from the auth context, is capped at one per user (count plus unique index), creates the vendor `listed: false`, and only platform admins can set `listed`. The same-slug retry returns the caller's vendor. The migration defaults `listed` to true, so existing vendors stay visible, and existing `createdByUserId` values are NULL, so the index cannot fail.
- **Dead `vendor.role ?? "admin"`** is removed.

### Still open from last time

- `COMMAND_USAGE["vendors create"]` is still `"[options]"` (`apps/cli/src/cli/index.ts:158`), so `--help` does not show `--name`/`--slug`. The error message now names the flags, which helps. Help should too.
- `apps/cli/docs/developer-cli-implementation-plan.md:159` (item 1b) still says "platform admin create is the fallback".
- `apps/core/src/routes/v1/vendors/post.test.ts` still has no 401 case and no invalid-slug case. 401 can be tested in the current harness: about 19 Core route tests already assert `toBe(401)`, for example `chat-room-invite-links/[token]/accept/post.test.ts` and `agents/index.auth.test.ts`. The slug regex sits in the schema, but no test exercises it.
- `apps/cli/src/cli/commands/vendors.ts:17-23` still re-declares `optionString`. Import it from `command-helpers.ts`.

### Must fix

**1. Anyone who signs up with an invited email can claim the invite, including an admin invite.**

`POST /v1/vendors/invites/{inviteId}/accept`, `.../decline` and `GET /v1/vendors/invites` match the invite on `user.email` only. None of them read `emailVerified`. Core signup sets `requireEmailVerification: false` and `autoSignIn: true` (`apps/core/src/lib/auth.ts`, `emailAndPassword`). Invites are meant to reach emails that have no account yet.

Here is the sequence:
1. An admin invites `partner@acme.com` as `admin`.
2. Anyone can sign up as `partner@acme.com` and gets a session straight away, without verifying.
3. `GET /v1/vendors/invites` lists the invite for them.
4. Accept makes them vendor admin.
5. From there they can demote or remove the real admins, down to the last-admin guard.

The codebase already handles the same case for Soko Bot. `routes/v1/soko-bots/helpers.ts` selects `emailVerified` and refuses when it is false, and its comment explains why: "Verification is what the whitelist rests on."

Select `emailVerified` in all three routes:
- Accept and decline: return the same 404 when it is false, so an unverified user learns nothing.
- `GET /v1/vendors/invites`: return an empty list.

Add tests for an unverified user on accept and on list.

### Should fix

**2. The logo cap is on the response schema, so one long stored logo takes down vendor reads.**

`vendorLogoSchema` now has `.max(LIMITS.VENDOR_LOGO_MAX_LENGTH)`. `vendorLogosInputSchema` shares it, and so does the response `vendorLogosSchema` → `vendorSchema` that `mapVendor` parses every row against (`apps/core/src/schemas/vendor.schema.ts:5-19`). Before this PR, admin create and patch took logos of any length. If one existing vendor has a logo longer than the cap, `GET /v1/vendors`, `/v1/vendors/me` and the admin list throw a ZodError for every caller. Put the `.max()` on the input schema only.

**3. Regenerate the Web Core client.**

Nothing under `apps/web` changed. `generated/core` still has `addVendorMember`/`AddVendorMemberRequest` and `AssignCoworkerRequest.email`. It has no `listed`, no `AdminVendor`, and no invite operations. Nothing in Web calls the removed operations, so nothing breaks at runtime. AGENTS.md still makes regeneration part of any Core endpoint change: run `pnpm --filter web generate:core:snapshot`, then Web typecheck. In the PR description you said `generate:core:snapshot` fails on main because of an unrelated utils export. If that's still true, link the fix or the issue. Otherwise regenerate here.

**4. Reuse the invitation helpers that already exist.**

`apps/core/src/helpers/vendor-invite.ts` re-implements code from `helpers/chat-room-invitation.ts`:
- the email normalizer (`normalizeVendorInviteEmail` vs `normalizeInvitationEmail`, already imported by `helpers/invitation.ts`)
- the 7-day TTL and the expires-at helper

`[id]/invites/post.ts` also inlines the `findUserIdByEmail` lookup. Import the existing helpers, or move them into a shared invitation module if the chat-room name gets in the way.

**5. An invitee has no way to see or accept an invite.**

Invite create sends no email. The CLI has no `vendors invites` or `vendors accept`, and Web has no UI. The gate copy in `registration-authority.ts` still tells blocked developers to get a "member invite", but the only way to find one is to call `GET /v1/vendors/invites` by hand. Either add the two CLI commands in this PR or take the invite wording out of the gate copy until a client exists. Separately, the comment on `VENDOR_MEMBER_INVITE_PENDING_LIMIT` mentions email-blast abuse, but no email is sent.

### Minor

- **Admin upgrade on accept almost never runs.** Invite create returns 409 when the email is already a member of the vendor (`[id]/invites/post.ts:89-106`). So an admin cannot invite an existing developer to become admin, and the accept upgrade only happens in a race. Either allow an invite that raises the role, or drop the upgrade branch.
- **Re-inviting with a different role is silently ignored.** Re-inviting a pending email with a different role returns the old invite with the old role. Return 409 or update the role.
- **`VendorMemberInvite.invitedById` uses `onDelete: Cascade`.** When an admin deletes their account, that vendor's pending invites go with it. The invite belongs to the vendor, so `SetNull` fits better. The per-inviter hourly cap also filters on `(invitedById, createdAt)` with no index; chat invites have `@@index([inviterId, createdAt])`.
- **Squash the two migrations.** The first creates `vendor_createdByUserId_idx`, and `20260922183000_vendor_self_service_cap` drops it again for the unique index. This PR hasn't merged, so squash them into one.
- **Drop `lower(email)` from the pending-invite unique index.** The chat-room precedent (`chat_room_guest_invitation_room_email_pending_uidx`) indexes the column directly, because the app stores emails lowercased. This PR already normalizes on write, and a plain column index keeps Prisma diffing simpler.
- **A self-service vendor can lose its only admin.** Account deletion never checks vendors. When the only admin of a self-service vendor deletes their account, the vendor has no admin and its slug stays taken. That gap existed before this PR, but now every self-service vendor starts with one admin, so it will come up.
- **A fast double submit of `POST /v1/vendors` gets 409 instead of the idempotent 200.** The second request loses to P2002. The CLI also prints "Created vendor" on the idempotent 200.
- **Missing tests:** accept of an expired or already-accepted invite, the hourly 429, and the serializable-conflict 409.

### Tests

I ran these at the head commit in a scratch worktree:

| Suite | Result |
| --- | --- |
| Core vendor, admin-vendor and helpers vitest | 1764 passed |
| CLI `pnpm test` | 208 passed |
| `packages/database` vitest, including the migration-prefix check | 363 passed |
| Core and CLI `tsc` | clean |

I did not run the Web typecheck, because the client isn't regenerated. I also did not run a live Postgres run of the migrations or of the manual test plan, and those still need doing.

**Verdict:** request changes. Item 1 blocks merge, and the fix is small. Items 2–5 should land in this PR. The minors can go in this PR or a follow-up.

Sarthib7:
Follow-up to the reviews: I pushed `e49dd160a` with the final invite cursor fix.

**VERIFIED at `e49dd160a`:** Both Vendor invite list routes reject malformed UUID cursors with HTTP 400 before querying invites. UUIDv7 cursors remain valid. Focused tests report `Test Files 3 passed (3)` and `Tests 39 passed (39)`.

Earlier fixes are in `0086c9544` and `01287305c`. `coworkers register` no longer creates Vendors. `vendors create` is standalone. Vendor creation handles slug and owner-cap unique conflicts. Invite flows hide whether an email has an account, require a verified email for acceptance and decline, update roles on re-invite, and preserve existing member roles on acceptance. The last-admin guard, inviter `SetNull`, normalized-email uniqueness, input-only logo limit, paginated invite lists, datetime schemas, and generated Web client are also in place.

GitHub reports `Build`, `Biome`, `Test Core`, `Test Web`, `Test CLI`, `Test Packages`, and `Typecheck` passing at `e49dd160a`. Please review the fixes again.

mrosberghaus (changes requested):
Please fix the merge conflicts.

Sarthib7:
fixed the merge conflicts

mrosberghaus:
Pushed 3 fixes from a review pass: `6a8bfe239` declares the pending-invite partial index in `schema.prisma`, so the next `migrate dev` won't drop it. `d3f278590` makes account deletion write the Vendor row, so a racing demotion retries instead of leaving 0 admins (reproduced and fixed on scratch Postgres), and narrows `USER_IS_LAST_VENDOR_ADMIN` to Vendors with another member or an unarchived coworker. `89e4b13ed` revokes a leaving sole admin's pending invites. See SPEC V87/V89. Core 6516/6516 locally.

mrosberghaus (approved):

PR #4986, opened 2026-09-21
feat(chat): retain history across all chat rooms

mrosberghaus:
Switching A → B → A now immediately restores retained messages and the reading position for both channels and direct messages. First visits load history in the background after server-side access validation; returning visits reuse the confirmed transcript without waiting for navigation or history requests.

History uses the existing QueryClient, scoped to user and workspace, with 30-minute inactive retention. Background reconciliation preserves history outside authoritative coverage and rejects stale responses. Pending sends remain local; logout, workspace changes, revoked access, and rejoining fence cached data. Message-link jumps and browser Back/Forward remain supported. The Chat tab title stays stable during room navigation.

Implements [SOK-1146](https://linear.app/masumi/issue/SOK-1146/switch-between-web-chat-channels-instantly-without-reloading-messages), extended to direct messages for consistent room navigation.

### Validation

- Latest Web suite: 832 files passed; 7,083 tests passed and one skipped.
- Channel and direct-message regression tests cover cached navigation before server navigation completes and independent reading-position restoration. Route tests cover direct-message history loading through the retained client cache.
- Biome and root typecheck passed in commit hooks.
- Earlier full root tests and Web build passed. Earlier production browser checks covered channel history restoration during delayed requests, desktop/mobile positions, Back/Forward, live-edge restoration, and message links using controlled local Core fixtures. The direct-message extension was verified with automated tests; those browser checks were not repeated for direct messages.


mrosberghaus:
## Grok review

### Fixed
- Important: jump/older-page reads treated any transcript write as a stale snapshot, then either looped without a cap or failed the older-page load on the first realtime/refresh. Snapshot identity now follows the transcript object (so a completed refresh is visible), retries are capped at 5, and a stale older page is retried instead of applied (`fc80415801e4ad4419c07822f7dd9ffc564ac8d1` — `apps/web/src/app/(app)/chat/hooks/use-room-message-jumps.ts`, `apps/web/src/app/(app)/chat/components/rooms-client.tsx`).

### Remaining
- Minor: `docs/superpowers/plans/2026-09-21-thread-unread-surfacing.md` — drive-by generate-command correction in an unrelated plan.
- Minor: `persistent-channel-view.tsx` — a revoked room still in view can render a blank pane until the `room-unavailable` redirect commits, because `ChannelRouteBootstrap` returns null when register exists even if it skipped seeding.
- Minor: `channel-transcript-cache.ts` — historical refresh uses a hardcoded `limit: 30` instead of `ROOM_HISTORY_WINDOW_LIMIT`.
- Minor: `transcript-viewport-stub.tsx` — RoomsClient tests do not exercise `initialPosition` / `onPositionChange`; restoration is covered by the stated browser checks.
- Skipped / disagreed: `AuthenticatedChannelCache` returning null when the client session is empty — first pass is pending (`AuthSessionHydrator`), so this is a logout/expiry fence rather than a hydration hole.

### Assessment
Cached A→B→A history, position restore, and access fencing match SOK-1146; snapshot retries are now bounded, leftover items are polish.

mrosberghaus:
Addressed the remaining notes in 8287db413:

- Revoked rooms now show the existing disabled room-loading shell while the access-loss redirect completes, without exposing cached messages or a live composer.
- Historical refresh uses `ROOM_HISTORY_WINDOW_LIMIT` for both around and older-page requests.
- The viewport stub now exposes the position bindings. Added an A → B → A regression test proving that A's saved position is restored and does not leak into B.

Keeping the generation-command correction: after the main merges, those three nonexistent script references caused the `Test CI config` job to fail. Reverting them would restore that failure.

Verified the bounded snapshot retries from fc8041580 and retained the session-loss fence: pending hydration still renders, while a confirmed empty session hides cached content.

Validation: all 831 Web test files passed (7,020 tests, one skipped), including the focused cache, navigation, and jump tests. Biome and typecheck passed.

Replying to https://github.com/masumi-network/sokosumi/pull/4986#issuecomment-5762475413.

mrosberghaus:
## Grok review

Picked up a review request from @mrosberghaus.

### Fixed
- Important: switching to a room that is not cached yet unmounted the open transcript onto a page that renders nothing, so the chat pane went blank until the server response (`5ac40c4ba` — `apps/web/src/app/(app)/chat/components/persistent-room-view.tsx`)
- Important: a cross-room message link spent `?message=` with `router.replace` before the link push committed, and Next discards the earlier navigation, so Back skipped the room just left (`5ac40c4ba` — `apps/web/src/app/(app)/chat/components/rooms-client.tsx`)

### Remaining
- Minor: `apps/web/src/app/(app)/chat/utils/room-transcript-cache.ts`: `setQueryDefaults` for the room prefix is left on the shared QueryClient after `clear()`
- Minor: `apps/web/src/app/(app)/chat/components/rooms-client.tsx`: `setMessagesState` and `refreshFocusedRoomMessages` close over the first retained binding. Safe while that room's lifetime object stays stable
- Skipped / disagreed: `removedRoomId` also calls `router.replace("/?notice=room-unavailable")`. Leave and membership revoke already navigate in the same turn, and Next discards the earlier pending navigation, so the caller's destination wins

### Assessment
Cached rooms still paint immediately, and an uncached target keeps the open room until its bootstrap exists; message-link jumps wait for the committed route so Back still returns to the previous room.

mrosberghaus:
Addressed both remaining notes in c35bca394:

- Query defaults now register once per room query type (`["room-transcript"]`) rather than per user/workspace. Subsequent cache instances replace that same registration; clearing a session still removes its scoped query data. Added a regression test covering separate user/workspace scopes.
- `setMessagesState` and `refreshFocusedRoomMessages` now include their retained binding dependencies, so they no longer capture only the first binding.

Verified the navigation fixes in 5ac40c4ba with the focused regression suite. Left the existing leave/revoke navigation behavior unchanged.

Validation: 52 focused tests passed; the full Web suite passed 837 files (7,207 tests, one skipped). Biome and root typecheck passed in commit hooks.

Replying to https://github.com/masumi-network/sokosumi/pull/4986#issuecomment-5764984618.

PR #4988, opened 2026-09-21
perf(apple): isolate transcript projection from scroll state

mrosberghaus:
Scrolling an unchanged transcript still rebuilt its full live-message dictionary/array once per view update after #4970. Keep preparation and projection in the outer room/thread views and put scroll/composer state in private content views, so local scroll updates reuse the supplied rows. Skip `ThreadSession.clearJump()` publication when the target is already nil.

The native `ScrollView`/`LazyVStack`, row identities and scroll behavior remain in place. No persistent cache, dependencies, Xcode project changes or test-target package products are introduced. All changes are under `apps/apple`.

## Measurement

The account-free Release harness drives the real room and thread views at 50 / 500 / 2,000 messages with plain and mixed content (markdown, code, images, unfurls and reactions). A counter inside the actual `PreparedTranscript.overlaying` construction fails before and passes after: **one rebuild per scroll-content evaluation → zero**, in all 12 matched cases. Longer 2,000-message sweeps also rebuild zero times across 538 room / 561 thread evaluations.

The audit retains numbers, method, source hashes and inlined rerunnable probes. Separate live-update checks pass for both revisions, including edits, optimistic reactions/rollback, arrivals, removals, pending sends and scope replacement. The failed-send check observes outbox state.

## Validation

- Full workspace tests: baseline 892 / updated **893 passed**, zero failures or skips (1,183 updated parameterized executions). New jump-publication regression observed red before its guard and green afterward.
- Both Release measurement builds and the shared Workspace iOS 17 build pass.
- Pinned uncached SwiftFormat and strict uncached SwiftLint pass; inlined harness recreates the measured probes byte for byte.
- Existing content-growth/media-scrolling/jump tests pass. All four light/dark room/thread jump captures were inspected, with their rendered-target OCR assertions passing.

This establishes removal of scroll-driven projection work, **not an end-to-end scrolling speedup or presented-frame FPS gain**. M6 remains Partial for live-room/thread acceptance and its existing pagination, historic-jump, gap-loading, follow-latest and image-growth gates. The separate geometry warning remains open.


mrosberghaus:
## Grok review

Ad-hoc review only — reviewer subagent spawn was unavailable. Skills `requesting-code-review` and `receiving-code-review` were loaded and applied in-process against `c984353766b4924978da9e20b6256c6ce419f5bd..75ebf4129c79a81a0d076931fca33b1d45feb0c8`.

### Fixed

### Remaining
- Minor: `RoomTimelineView.swift`: live non-empty / projection-empty now shows `ProgressView` instead of mounting an empty `ScrollView`. First paint waits on prepare; jump/scroll tests still pass.
- Minor: `RoomTimelineView.swift`: `highlightedId = nil` still assigned on every interacting/tracking phase with no nil guard. Local `@State` only, so it does not rebuild the projection.
- Minor: `ReplyThreadView.swift`: outer overlay still runs on any forwarded `WorkspaceState` / `ThreadSession` publication; only the nil `clearJump()` path is gated. Audit already states this.
- Skipped / disagreed: none.

### Assessment
Ready to merge: Yes

Preparation and live overlay stay on the outer room/thread views; scroll/composer `@State` lives in private content views, and `ThreadSession.clearJump()` no longer publishes a no-op. That matches the measured zero `PreparedTranscript.overlaying` rebuilds on unchanged-data scrolling without changing `ScrollView`/`LazyVStack` or adding a cache.

mrosberghaus:
Checked the three remaining observations against the reviewed base (`c98435376`) and head (`75ebf4129`):

- **Loading state:** the base already shows `ProgressView` when live messages exist but the prepared projection is empty. This PR preserves that branch; it does not introduce a first-paint behavior change. See [the existing branch on the base](https://github.com/masumi-network/sokosumi/blob/c984353766b4924978da9e20b6256c6ce419f5bd/apps/apple/Sokosumi/Chat/Timeline/RoomTimelineView.swift#L148-L152).
- **Highlight clearing:** the nil assignment is pre-existing and affects local `@State`. It does not rebuild the outer projection in the measured scrolling cases. There is no measured benefit supporting an additional guard in this change.
- **Forwarded publications:** correct. Workspace/thread publications can still rebuild the outer projection; the audit explicitly records that limitation. The measured guarantee is zero projection rebuilds during unchanged-data scrolling, not reuse across every workspace publication.

No code changes are warranted by these observations. The existing measurements and live-update checks support the scoped CPU-work reduction; end-to-end scrolling smoothness still requires manual acceptance.

PR #4992, opened 2026-09-21
docs(chat): Room unread counts top-level messages only

mrosberghaus:
Stack layer 1 of 5 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread). Documentation only — no application code.

```
(main) <- 1-vocabulary <- 2-reply-bar <- 3-debold-channel <- 4-inset-rows <- 5-rail-flyout
```

In this repo the vocabulary leads the code, so the glossary moves first and the four code layers sit on top of it.

## What changed

**ADR-0037 is now Accepted.** It was merged as `Proposed` in #4984. Its argument is untouched; only the status line moved.

**Three `CONTEXT.md` entries rewritten** to match it:

- **Room unread** — now counts unseen non-self *top-level* messages only. Thread replies are Thread unread and surface on the Thread. A User mention inside a Thread is the one exception that still marks the room.
- **Unread thread** — gains the three surfaces it renders on, and is explicitly the room's second number, never folded into the first.
- **Thread list** — becomes the overflow target for the sidebar's capped inset rows; still per-room.

## Two contradictions the plan didn't predict

The rewrite made two other entries wrong, and the planned grep wouldn't have caught either:

- **Room last-read** said it clears "top-level Room unread". After the split there is no other kind, so the qualifier implies a half that no longer exists.
- **Muted thread** said it stops counting toward **Room unread**. It now stops counting toward **Thread unread** — Room unread never sees thread replies at all.

Both fixed here. The second is the one worth a second look: it's the sentence that carries ADR-0030's meaning, and leaving it would have made the mute rule read as doing something it no longer does.

## Also

Corrected two stale lines in the implementation plan: one telling later layers to wait for an approval that already happened, and a global constraint telling them to assert on SQL strings — which SOK-1147's agreed test seams explicitly reject. That constraint would have steered layer 3 wrong.

## Review focus

This is a decision about words. The question worth your time is whether the **Room unread** entry now says something you agree with, since every later layer is built to match it.


mrosberghaus:

PR #4977, opened 2026-09-21
fix(apple): show the quote of a pinned quote-only message

mrosberghaus:
## What changed

Apple parity row **21a**: a pinned message that is only a quote now shows the quote. Before, the pin card rendered the body unconditionally, so such a pin showed sender and time and nothing else.

- `Packages/SokosumiChat/.../PinnedMessagePreview.swift` (new, UI-free): decides what a pin row prints — `.body` or `.quote(author:snippet:)` — plus the Markdown source, the quoted author and the spoken label of the quote block.
- `Sokosumi/Chat/Rendering/MessageQuoteView.swift`: the transcript quote card's chrome (opening-quote glyph, inset, quaternary rounded background) moved into `MessageQuoteBlock`. The card itself is unchanged; there is still one way a quote is drawn.
- `Sokosumi/Chat/Pins/PinnedMessagesView.swift`: `PinnedMessageCard` asks the preview. A body renders exactly as before; a quote renders in `MessageQuoteBlock` with the author semibold on one truncating line and the snippet through the same `MessageMarkdownView`, six-line clamp and height cap as a body, in the secondary style.
- `PARITY.md`: row 21a in review with evidence, row 13a Done (#4965 merged as `bf8afac93`; its unverified notes kept), work order advanced to 23a.

Pin/unpin, jump to message, loading/empty/error states and realtime updates are untouched. No `apps/web`, `apps/core`, Core contract, OpenAPI snapshot, dependency or project change.

## Web sources audited (at `bf8afac93`)

- `apps/web/src/app/(app)/chat/components/pinned-messages-panel.tsx`:234-235 — `quoteOnly = message.content.trim().length === 0 ? message.quote : null`.
- same file :266-271 — author line `quoteOnly.authorName`, `text-xs font-semibold`, `truncate`.
- same file :272-281 — snippet through the same `ChannelMessageText` a body uses, muted, `border-l-2`, `line-clamp-6`.
- `apps/web/.../room-message-row.tsx`:376 — the transcript card's own empty-snippet check, for comparison.
- Read only, for where the data comes from: `apps/core/src/routes/v1/chats/rooms/helpers.ts`:629 (quote nulled only when the pinned message itself is deleted), :740-745 (author falls back to "Someone" in the stored snapshot), `apps/core/src/schemas/chat-room.schema.ts`:653 (empty content is accepted with a quote, in any room), `packages/utils/src/chat-room-quote-snippet.ts` (snippet is plain text, untruncated).

## Behaviour decisions

[...]

PR #4995, opened 2026-09-21
fix(apple): keep search results visible while refining the query

mrosberghaus:
## What changed

Apple parity row **23a**: room search keeps the previous hits on screen while the query is refined, as web does. Before, `RoomSearch.search` reset the model before its 250 ms debounce and the panel drew an empty list while a query was pending, so every keystroke blanked the list to "Searching…".

- `Packages/SokosumiChat/.../RoomSearch.swift` — `search` supersedes the earlier request (generation bumped before anything suspends) without clearing. Success replaces the hits and selects the first; an error, an emptied query or a different room empties them. New `presentation(for:)` takes the text in the field and answers the one status row (`idle` / `loading` / `empty` / `failed`) or none, the hits to draw, and `isRefining`. The model now owns the selection (`selectedId`, `select`, `moveSelection(by:)`, `selectedResult`). The debounce is an initializer argument (default 250 ms) so tests do not sleep.
- `Sokosumi/Chat/Search/RoomSearchResultsView.swift` — switches over the presentation; rows are no longer disabled while loading.
- `Sokosumi/Chat/Search/RoomToolsModifier.swift` — lost its `selectedId` state and the per-keystroke `selectedId = nil`; `select` opens any hit the presentation shows instead of refusing while loading or pending.
- `apps/apple/PARITY.md` — 23a in review with evidence, **21a Done** (#4977 merged), work order advanced to 12d.

No `apps/web` or `apps/core` change, no contract, dependency, snapshot, project or entitlement change.

## Web audited

`apps/web/src/app/(app)/chat/components/room-search-panel.tsx` at `f8adbeec0`:

- `searchMessages` :74-107 — a starting request leaves `results` alone; success replaces and sets `activeIndex` 0 (:94-95); failure empties `results` and sets the error (:100-101); stale `requestId` answers are dropped (:91, :97).
- Status rows :270-282 — loading row only with `results.length === 0`; the error row yields while loading or pending; the empty row needs a settled, error-free empty answer. Hits render beneath (:470).
- There is **no** secondary progress indicator while stale hits show; the only spinner is the loading row. Apple adds none. (The "trailing column" in PARITY's slice 23 notes is the per-row "Opening message" spinner, not a search-field indicator.)
- Emptying the field clears everything in the same keystroke (:308-327); closing clears the hits, keeps the query, and reopens from the loading row (:177-192).
- Arrows and Enter act on `results` whenever non-empty, pending or not (:243-265).

Web matched the row; nothing was ambiguous.

## Behaviour decisions

Mirrored web: keep until answered; loading row only with no hits; no secondary indicator; replace on success; empty on error with the error row; stale error yields to loading; "No messages found" only for a settled empty answer; immediate clear on an emptied field; clear on close, reopen from loading; selection to the first hit on replace and otherwise untouched; Return and click on a stale hit open it.

Kept Apple: selection by message id rather than index; a jump failure shows above the hits with Retry; the field is disabled during a jump; room or account change closes search and resets (the model now also clears by room on its own).

One deliberate deviation: Apple's debounce lives inside `search`, so a query typed back to the already-answered one is searched again after 250 ms, where web — inside its debounce — sends nothing. The hits stay either way; only after an *empty* answer does Apple show "Searching…" briefly where web shows the empty row at once. Skipping the request would have turned "Retry search" after a jump failure into a no-op.

## Verification (from `apps/apple`, on `origin/main` `f8adbeec0`)

- `swift test --package-path Packages/SokosumiChat` — 588 passed. `SokosumiWorkspace` 125, `CoreAPI` 1, `SokosumiAuth` 35, `SokosumiRealtime` 48 — all passed.
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO` — TEST SUCCEEDED, 916 tests, 0 failed.
- Same invocation with `build` — BUILD SUCCEEDED.
- `mint run swiftformat --lint .` — 0/385; `mint run swiftlint lint --strict` — 0 violations.
- Tests first: `RoomSearchRefiningTests` (12) was red against the old reset with the new API stubbed (12 issues), green after. The app render tests were not run against the old view, which does not compile against the new call site.
- **Flaky test, stated plainly:** before main was merged, the first two full runs each failed one case of `NativeWindowTests/TranscriptContentGrowthTests/growingMessagePreservesReadingPosition` (914/915; offsets off by 194 and 160 points) with the machine's load average at 15-17 from other processes. The suite passed alone, and the next four full runs passed, including the one on the merged tree. The test waits fixed intervals for transcript scroll corrections and touches no search code. A run skipping the new render tests also passed, which does not separate load from the extra tests; details in PARITY.
- Fixtures inspected in light and dark: `docs/screenshots/room-search-refining-*.png` (both hits, first selected, no status row — asserted byte-equal to the settled panel) and `room-search-first-loading-*.png` (spinner over "Searching…", no hits).

## Unverified

The app was not launched. Not exercised live: typing against a real room and watching the hits stay, Return/click on a stale hit, arrow keys while refining, a real failed search over stale hits, switching rooms with search open, VoiceOver.

## How to test

Open a room, Cmd-F, type a word with hits, keep typing: the hits stay until replaced — no "Searching…" flash, no "No messages found" flash. Press Down, keep typing, press Return before the new answer: the highlighted message opens. Clear the field: the list empties at once. Search nonsense: "Searching…", then "No messages found".


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/Packages/SokosumiChat/Tests/SokosumiChatTests/RoomSearchRefiningTests.swift`: `anotherRoomNeverShowsTheFirstRoomsResults` answers with `testMessageJSON`, whose `roomId` is always `testRoomId`. The empty list after that response does not show that a real page for the other room is kept. The clear-before-response assertion is the one that matters.
- Minor: `apps/apple/Sokosumi/Chat/Search/RoomToolsModifier.swift`: after Escape empties the field, arrow keys can still walk the previous hits until the search task clears them. Return does not open those hits, because `presentation(for:)` is already idle.
- Skipped / disagreed: a cancelled in-flight `search` that nobody follows with `search` or `reset` would leave `query` on the cancelled text and `isLoading == false`, so leftover hits would look settled. `RoomToolsModifier`'s `.task` always does one or the other. Not a user path.
- Skipped / disagreed: typing back to the already-answered query searches again after the debounce, unlike web. Deliberate, and the hits stay either way.
- Skipped / disagreed: "Retry search" on a jump failure refetches instead of retrying the jump. Pre-existing, and the PR keeps it on purpose.

### Assessment
Ready to merge: Yes

`RoomSearch.search` keeps the previous hits until the next answer, `presentation(for:)` is what the panel draws, and a late or superseded response still cannot replace or clear them. `swift test --package-path apps/apple/Packages/SokosumiChat --filter RoomSearch` passed, 19 tests. App render tests were not re-run.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/4995#issuecomment-5764731927) — one follow-up, in `3ad2e95ee` (tests and one PARITY sentence; nothing under `Sources/` or the app).

- **`anotherRoomNeverShowsTheFirstRoomsResults` never proved room B's page is kept** — fixed. The gap was real: a model that dropped every answer after a room switch passed. `testMessageJSON` and the gated transport's `respond(to:rows:)` now take a room id (default `testRoomId`). The test keeps its clear-before-response and foreign-row steps, then answers room B with its own rows and asserts `results == ["b1", "b2"]`, `selectedId == "b1"` and a presentation with no placeholder; it then switches back to room A and asserts B's rows and selection are cleared, with the loading row showing, before A answers. Checked that the new lines bite: with the post-response filter deliberately broken to accept only the first room, the old assertions still passed and only the three new ones failed (`RoomSearchRefiningTests.swift:160` first). The break was reverted; `RoomSearch.swift` is untouched in the commit.
- **Arrow keys can walk the previous hits after Escape empties the field** — not changed. `RoomSearch.search` clears `results` and `selectedId` synchronously for an empty query, before any suspension (`RoomSearch.swift`, the `query.isEmpty || roomId != resultsRoomId` branch ahead of the debounce), and `RoomToolsModifier`'s `.task(id: request)` calls it on the query change. The window is the turn between the field emptying and that task starting. In it `moveSelection` can only change a `selectedId` that is about to be set to nil, nothing is drawn because `presentation(for:)` is already idle, and Return opens nothing, as you noted. A guard would need the model to know the field's live text, which is the split this slice removed.
- **Skipped items** (cancelled search nobody follows up, re-searching an answered query, "Retry search" refetching) — agreed, unchanged.

Verification: `swift test --package-path Packages/SokosumiChat` 588 passed; strict SwiftFormat and SwiftLint clean. The app and full-workspace suites were not re-run for a package-test-only edit; CI on `3ad2e95ee` covers them.

PR #5000, opened 2026-09-21
feat(chat): unread reply bar on thread parents (SOK-1147, 2/5)

mrosberghaus:
Layer 2 of 5 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): the unread reply bar. Stacked on #4992.

It sits below the de-bolding layer on purpose: the destination for Thread unread has to exist before the channel mark is removed. This layer changes no count and no sidebar behaviour.

## What changes

**Core**
- The message list (`GET /chats/rooms/{id}/messages`) carries `threadUnreadReplyCount` on each thread parent.
- It comes from the existing Participant- and mute-gated thread aggregate (`getChatRoomThreadAggregates`, `unreadOnly`), read once per request, not per message.
- The read is skipped when the page holds no thread parent. Search hits report `0`.
- `room-unread.ts` is untouched: no gating predicate and no SQL changes in this layer.

**Web**
- A parent with unread replies shows a tinted bar reading "2 new replies". Otherwise it shows the plain reply count, as today. A lurker on a busy thread sees the plain bar.
- New string `App.Channels.Thread.newReplyCount` in `en`, `de` and `es`.

## Where this differs from the plan

1. **The field is optional, not defaulted to 0.** Edit, reaction, pin and realtime payloads do not compute the count. A defaulted `0` there would let the web overwrite a real count and untint the bar. The edit-route test asserts the field is absent.
2. **The web keeps the known count** when a fresher copy of a message states none (`keepKnownThreadUnreadReplyCount`, used by `mergeRoomMessages` and `mergeUpdatedMessage`).
3. **A thread Look and Mark all clear the bar at once** (stories 20 and 21). The plan had no step for this. `onThreadLooked` now names the thread it looked at.

## Tests

Seams as agreed in the issue: Core route contract with mocked Prisma, thin component tests. No SQL-string assertions, no DB-backed tests.

- Core: 6077 passed, 39 skipped. Web: 6999 passed, 1 skipped. Typecheck and Biome clean.
- Route contract: count per parent, lurker reads 0, search reads 0 with no aggregate read, `around` window carries the count, thread-less page makes no aggregate read.
- Component: plain bar, unread bar, singular copy, lurker, absent field, click opens the thread.
- `rooms-client`: opening a thread clears the parent's count. Checked by mutation: the test fails without the wiring.

Not verified in a running app. The tint was checked in component tests only.

## Known gaps

- **Live replies lag by up to 60 s.** A reply arriving while the room is open tints the bar on the next automatic refresh, because the client cannot evaluate Participant or mute gating. Tracked as [SOK-1151](https://linear.app/masumi/issue/SOK-1151/thread-reply-bar-lags-live-replies-by-up-to-60s-drive-it-from-the), which also closes the race below.
- **A message-list refresh in flight while a thread is opened can re-tint the bar** with a stale count. Self-corrects on the next refresh.
- **The lurker case is covered only as "no aggregate row comes back"**, which is the limit of mocked Prisma. The gating itself is existing, unchanged code.

## Before merge

ADR-0037 still reads `Proposed` on `main`. The issue asks for agreement before anything a reader sees ships; #4992 is the PR that flips it to `Accepted`.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/rooms-client.tsx`: Mark all zeros every loaded reply bar. Core `markAllChatRoomThreadsRead` does not mark muted threads, even when a mention is still unread (SOK-1087). That bar stays plain until the next message-list read (up to 60s while Ably is healthy). The thread list and the header count still show it.
- Skipped / disagreed: defaulting `threadUnreadReplyCount` to 0 — edit, reaction, pin, and realtime payloads would then untint a real count. Leaving the field off, and keeping the known count in `keepKnownThreadUnreadReplyCount`, is the right contract.
- Skipped: a live reply tints the bar only on the next list read (SOK-1151), and a list read already in flight can re-tint after Look. Both are named in the PR and correct themselves on the next read.

### Assessment
The bar matches this layer: one gated count per list read, lurkers stay plain, Look clears that parent, and payloads that do not compute the count do not wipe it.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5000#issuecomment-5765696221)

**Minor — Mark all zeroes every loaded reply bar: fixed in 02bd7b694.**
Verified against Core: `markAllChatRoomThreadsRead` filters on `thread_read."mutedAt" IS NULL`, so a muted thread with an unread mention is left unread (SOK-1087) while the transcript cleared its bar anyway.

`ThreadListPanel` now passes `onAllThreadsLooked` the loaded threads that are muted and still unread, which are exactly the ones Core skips. `rooms-client` clears every bar except those. Covered in `thread-list-panel.test.tsx` ("names the muted threads Mark all leaves unread").

One limit: only threads loaded in the panel are named. Unread threads sort first, so a muted unread thread is on the first page in practice. If one were not, its bar would go plain until the next message-list read, as before.

**Optional field instead of a default of 0:** no change, agreed.

**Live reply lag and the Look-versus-in-flight-read race:** no change in this PR. Both are tracked in SOK-1151, which drives the bar and the header count from one read.

mrosberghaus:

PR #5007, opened 2026-09-21
feat(chat): thread replies no longer bold the channel (SOK-1147, 3/5)

mrosberghaus:
Layer 3 of 5 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): the channel stops bolding on Thread replies. Stacked on #5000.

It sits above the reply bar on purpose. Without #5000 this would clear the channel mark and leave nothing pointing at the replies.

## What a reader notices

- A channel whose only unread is Thread replies is no longer bold. It shows no opt-in number, no collapsed-rail pill, and it does not count in the tab title.
- A User mention inside a Thread still bolds and badges the channel.
- Reading a channel clears that channel's mark.

## What changes

**Core**
- `getChatRoomUnreadCounts` already computed two legs and summed them. It now groups by leg and returns `{ channel, thread, total }` per room.
- All 8 room summary routes carry `channelUnreadCount` and `threadUnreadCount` beside `unreadCount`, through one helper (`unreadCountFields`). `unreadCount` is still the sum.
- `mapChatRoom` treats a caller that states only the sum as Room unread, so a future route cannot go silently quiet.

**Web**
- `resolveRoomAttention`: bold = channel half > 0, or an unread mention, or a hand-set unread mark. The opt-in number counts the channel half only. The badge is unchanged.
- The tab title asks the resolver instead of restating the rule.
- One `RoomAttentionCounts` interface replaces five hand-written copies of the same fields.

Mocked Prisma supplies its own rows, so no test here can prove the split. In `room-unread.ts` both `WHERE` bodies are byte-identical to before. The only changes are a `'channel'::text AS source` / `'thread'::text AS source` literal on each leg, `combined.source` in the outer `SELECT`, and `GROUP BY combined."roomId", combined.source`. `sqlViewerIsThreadParticipant` and `sqlThreadReplyPagesViewer` are untouched.

## Two things the plan missed

1. **Three optimistic "room was read" writers set only `unreadCount: 0`.** With the split, the channel half would have stayed stale and the row bold after reading. They now share one rule, `roomAttentionAfterRead`, and the read overlay stores and applies both halves.
2. **`mapChatRoomWithSidebarFlags` forwarded a fixed subset of fields** and dropped the two new ones. The route-level test caught it.

## Tests

Seams as agreed in the issue: Core route contract with mocked Prisma, the pure attention resolver, thin component tests. One narrow SQL tripwire for the mute and Participant gates, marked as such.

- Core: 6250 passed, 39 skipped. Web: 7215 passed, 1 skipped. Typecheck and Biome clean. After the last review edits only the affected tests were re-run (636 room-route, 1708 chat).
- Route contract: single room, room list and `read` each report both halves and the sum; a Thread-only room reports a clean channel; a room with nothing unread reports zeros.
- Resolver: Thread-only room is not bold; a Thread mention is bold and badged; the badge counts mentions only; the opt-in number counts the channel half; muted room is silent; hand-set unread still bolds; a count is never shown without bold.
- Row: no rail pill for a Thread-only room, mention pill for a Thread mention. Checked by mutation: the test fails without the row wiring.

Not verified in a running app.

## Left alone on purpose

- **Reading a channel also clears a mention that came from a Thread.** Existing behaviour, and the glossary says Room last-read clears the room's mention badge.
- **In a 1:1 Direct a Thread reply still bolds the row.** Core writes a direct-message notification for it, and ADR-0037 says the badge keeps counting directs.
- **The older broad SQL-string test in `room-unread.test.ts` is still there.** Only its fixture changed.
- **The Apple client bolds on the total**, so it keeps the old behaviour until its own work lands.


mrosberghaus:
## Grok review

### Fixed
- Important: the opt-in Postgres mute suite treated the new unread breakdown as a number, so every `roomUnread()` assertion failed (`toBe` is identity, and `lurkerUnread` compared two fresh objects). Both helpers now assert `total`, which is the old summed count (`e87f5b43a` — `apps/core/src/routes/v1/chats/rooms/room-unread.mute.integration.test.ts`)

### Remaining
- Minor: `apps/core/src/routes/v1/chats/rooms/room-unread.test.ts` never feeds one map both legs plus a second room. Route tests cover a single room. The opt-in suite still checks only `total`, so a wrong `source` alias would stay green.
- Minor: `apps/core/src/routes/v1/chats/rooms/room-unread.ts` stores any non-`channel` source as thread. Harmless while the query only emits the two literals.
- Skipped / disagreed: Apple still bolds on `unreadCount` — the PR leaves that until Apple’s own work, and the sum is unchanged.
- Skipped / disagreed: create, patch, restore, and join still return zero unread without a recount — same as before this split.
- Skipped / disagreed: `.default(0)` keeps the halves optional in OpenAPI — that is the documented fallback for snapshots that only have the sum. `??` does not treat `0` as missing.

### Assessment
Channel bold, the opt-in number, the rail, and the tab title follow the channel half; a thread mention still badges; `unreadCount` stays the sum. The two `WHERE` bodies match the base aside from the `source` column and `GROUP BY`.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5007#issuecomment-5766399691)

**Important — opt-in Postgres mute suite (e87f5b43a): kept as pushed.** The fix is correct. I missed that file because the suite is skipped without `CHAT_MUTE_INTEGRATION_DATABASE_URL`, and `roomUnread()` has an inferred return type, so typecheck passed with the breakdown object flowing into `toBe(1)`.

**Minor — no test with both legs across two rooms: fixed in 1914c4dc4.** Added at the room list route: two rooms, legs interleaved, one room Thread-only. Checked by mutation: a fold that shares one breakdown between rooms fails it.

**Minor — any non-`channel` source stored as thread: fixed in 1914c4dc4.** The fold now does `breakdown[row.source] = …`. The source literals are the breakdown's own keys, so the catch-all `else` is gone.

**Not done — asserting `thread` rather than `total` in the opt-in suite.** It would let that suite catch a wrong `source` alias, which no mocked test can. I have no database URL here, so I cannot run it, and I did not want to edit a suite blind. Worth doing by whoever next runs it against Postgres.

**Apple, the zero-unread routes, `.default(0)`:** no change, agreed with the reasons given.

mrosberghaus:

PR #5010, opened 2026-09-21
feat(chat): inset unread thread rows under the room (SOK-1147, 4/5)

mrosberghaus:
Layer 4 of 5 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): inset unread thread rows under the room. Stacked on #5007.

## What a reader notices

- A room lists up to three unread Threads under its sidebar row. Each is named by its parent message and shows its unread replies.
- A row is a link to the Thread's first unread reply, the URL a chat notification uses. Following it opens the Thread and clears it.
- Past three, an overflow row says "4 more unread threads" and opens the room's thread list.
- Muted rooms, reorder mode and the collapsed rail list nothing. The rail is layer 5.

## What changes

**Core**
- The room summary carries `unreadThreads` (at most 3, newest unread reply first) and `unreadThreadCount` (the true number of unread Threads).
- Each entry has `parentMessageId`, `firstUnreadReplyId`, `parentContent` (raw, cut to 1000 characters) and `unreadReplyCount`.
- The read runs only for rooms whose counts show Thread unread. The sidebar polls the room list, and on most polls that is no room.

**Web**
- `ChatRoomThreadRows`, built on the sidebar's sub-menu primitives, mounted inside the room's sidebar item.
- The room reads a new URL ask, `?threads=1`, and opens its thread list. `useEditChannelParam` is renamed `useRoomUrlAsk`, since it now reads two asks.
- The thread list and the inset rows name members through one builder, `roomMentionNames`, so `@all` and member mentions read the same in both.
- The read overlay carries the list, so a Thread just read does not come back on remount.
- New copy under `App.Channels.ThreadRows` in `en`, `de` and `es`.

Mocked Prisma supplies its own rows, so no test here can prove either point.

1. **`sqlUnreadThreadReplies` is extracted** from the count's thread leg, so the Thread unread number and this list share one `FROM … WHERE`. A Thread cannot be listed while contributing nothing to the number. I captured the count query's SQL text before and after the extraction and compared the bytes: identical. Neither gating predicate is touched.
2. **The new query groups by `parent.id` alone** and selects `parent."roomId"` and `LEFT(parent.content, …)`. That relies on Postgres letting a table's other columns ride along with its primary key. It ranks with `ROW_NUMBER() OVER (PARTITION BY "roomId" ORDER BY "lastUnreadAt" DESC, "parentMessageId" DESC)` and counts Threads with `COUNT(*) OVER (PARTITION BY "roomId")`.

## Where this differs from the issue and the plan

1. **The single-room routes carry the list too.** The issue says only the room list does, "since the sidebar is the only consumer". That premise does not hold here: a room read, a Thread Look, a star, a mute and mark-unread each answer with the room, and the sidebar redraws the row wholesale from that answer (`handleRoomRead`, `replaceRoom`, `upsertRoomToTop`). An answer without the list would empty the row's inset rows until the next poll, which breaks stories 7 and 20. The extra read runs only when that room has Thread unread.
2. **`unreadThreadCount` is new.** The plan subtracted a Thread count from a reply count for the overflow row. Story 9 asks for the overflow "stated honestly".
3. **Rows are links, not an `onOpenThread` callback.** A reply id on `?message=` already opens its Thread and Looks it, from any page, so no second route shape was needed.
4. **Core sends raw parent content and the web builds the label.** Naming mentioned members in Core would cost up to three lookups per Thread per room on a polled route; the sidebar already holds each room's members.

## Found in review and fixed

- **Layout bug jsdom cannot show.** The row's trailing cluster (mention badge, mute glyph, menu) is centred on its positioned ancestor. That was the whole sidebar item, which now grows with the inset rows, so the badge and menu would have landed on them. The row proper now sits in its own positioned box. A test pins that the cluster is inside that box and the inset rows are outside it.
- `@all` read differently in the inset rows than in the thread list.
- The schema hard-coded `3` and `280` beside the constants. Both are now named once.
- The read overlay could blank the list from a snapshot stored without one.

## Tests

Seams as agreed in the issue: Core route contract with mocked Prisma, thin component tests. No new SQL-string assertions.

- Core: 6256 passed, 39 skipped. Web: 7270 passed, 1 skipped. Typecheck and Biome clean.
- Route contract: the list and the true count on the room list and the single room; a capped list beside a larger count; no list without Thread unread; no second read when no room has Thread unread.
- Component: one row per Thread with its name and unread replies; the link target; nothing when nothing is unread; no overflow at exactly three; the true remainder, singular and plural, and its link; a member mention and `@all` by name; a fallback label for a parent with no text.
- Row: lists under the row; nothing when muted or reordering; trailing cluster stays on the row.
- Room: `?threads=1` opens the thread list, and nothing opens without it.

The ranking and the cap live in SQL, so the route tests cannot exercise them; they assert what the route does with the rows it is given.

**Not verified in a running app.** That matters more here than in the lower layers: this is the first layer with real layout.

## Known gaps

- A hand-typed URL carrying both `?edit=1` and `?threads=1` leaves `threads=1` on the URL. No in-app link produces that pair.
- Layer 5 cannot reuse these rows as they are: the sub-menu primitive hides itself on the collapsed rail, so the flyout needs its own wrapper.
- Live replies reach the rows on the next sidebar poll, as they reach the reply bar on the next message read ([SOK-1151](https://linear.app/masumi/issue/SOK-1151/thread-reply-bar-lags-live-replies-by-up-to-60s-drive-it-from-the)).


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `serializeDirectRoomForViewer` in `apps/core/src/routes/v1/chats/rooms/helpers.ts` still returns a room with empty `unreadThreads` and zero counts. Web upserts that payload on direct, coworker, and personal-assistant open, so inset rows disappear until the next sidebar poll (about 15s). An idempotent join does the same: `POST /chats/rooms/{id}/members/me` maps the room with no unread fields. Read, star, mute, mark-unread, and GET already carry the list.
- Minor: overflow and thread-row links always push. When the reader is already in that room, the URL ask then replaces the param away, so Back can land on the same view. The edit item avoids this with `router.replace` while the room is active.
- Minor: the unread-thread list and the count are two queries with no shared snapshot. A reply that lands between them can leave `threadUnreadCount > 0` with an empty list for one poll. The rows render nothing when the list is empty; the next poll corrects it.
- Skipped: a hand-typed URL with both edit and thread-list params leaves the thread-list param in place. No in-app link builds that pair, and the PR already records it.
- Skipped: grouping the new query by `parent.id` while selecting `parent."roomId"` and `LEFT(parent.content, …)`. `chat_room_message.id` is the primary key, so Postgres accepts the other columns of that row. The count query's `FROM … WHERE` is the extracted fragment, unchanged.

### Assessment
The inset rows, the shared unread predicate, and the sidebar redraw on read, star, mute, and mark-unread match this PR. The leftover notes are staleness outside those routes, not a break in the new list.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5010#issuecomment-5767169807)

**Minor — open, join, create and restore answer with zeroed unread: fixed in a9cac37cb, on the web.**
Verified: `serializeDirectRoomForViewer` and `POST …/members/me` map the room with no unread fields. This is older than the stack. On `main` the same answers already carry `unreadCount: 0`, so re-opening a Direct already unbolds its row until the next poll; this layer only extended that to the inset rows.

Every caller of `upsertRoomToTop` is one of those four answers, and none of them counts what is unread, so their zeros mean "not counted". An upsert now keeps what is unread on a room it already holds (`keepRoomUnreadState`): the counts, both halves, the mention count, the hand-set unread mark and the unread Threads. A room it did not hold is taken as it comes. The tab-title hook had the same upsert and the same problem, so re-opening a Direct also dropped it from the `(N)` title; fixed with the same helper. Both are covered by tests.

Not fixed in Core on purpose: recounting on those routes adds reads to open and join for a value the client already has.

**Minor — thread row and overflow links always push: fixed in a9cac37cb.** `ChatRoomThreadRows` takes `isActive` and passes `replace` to both links while the reader is in that room, the way the row's Edit item does. Covered by tests.

**Minor — the list and the count are two reads with no shared snapshot: no change.** Sharing a snapshot means an interactive transaction, which this route avoids on purpose (SOKOSUMI-Q9, #2559). The rows render from the list, so the mismatch shows as no rows for one poll, never as a wrong row.

**Skipped items (the `edit` + `threads` URL pair, `GROUP BY parent.id`):** no change, agreed.

mrosberghaus:
**Visual fix, 2fad45738: the inset rows did not match the design mock.**

The rows showed a bare 16px outline icon in the text colour, 4px apart. The mock (section 2b) has the unread language the notification rows already use: an 18px tinted circle with an 11px glyph in the accent colour, rows 1px apart, an 11px count, and the overflow row starting on the labels' column. The plan's code had the circle; I dropped it when I moved the rows onto the `SidebarMenuSub*` primitives.

Now matched, and checked in a real browser against the mock's computed styles. The circle is `bg-primary-quaternary` rather than `quinary`: in dark mode quinary is a shade off the sidebar's own background and does not show. #5012 is rebased onto this; the one conflict was the list's `className`, resolved keeping both.

mrosberghaus:
**Visual fix: the inset rows now reach the room row's right edge.** The sidebar's sub-menu primitive insets its list on both sides, so the rows stopped about 24px short and a Thread's name truncated early. They are indented from the left only now, as in the mock. Measured in a browser: every thread row ends at the same x as the room row. #5012 and #5013 are rebased onto it.

mrosberghaus:
**Alignment fix: the inset rows now sit on the room row's axes.** Measured in a browser against the room row: the rule is on the leading mark's axis (19-20px, mark centre 20px; it was at 15px), the Thread icons start on the label column (40px; they were at 33px), and each Thread count is centred in the mention badge's 28px column (236px for both; they were right-aligned 7px outside it). #5012 and #5013 are rebased onto it.

mrosberghaus:

PR #5012, opened 2026-09-21
feat(chat): unread threads in a flyout on the collapsed rail (SOK-1147, 5/5)

mrosberghaus:
Layer 5 of 5 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): unread threads on the collapsed rail. Stacked on #5010. This is the top of the stack.

## What a reader notices

- On the collapsed rail, hovering a room with unread Threads opens a flyout beside its mark: the room's name, the same thread rows the expanded sidebar insets under it, and the "N more unread threads" row.
- The rail's attention pill means what it meant before. A room whose only unread is in Threads shows no pill; a Thread mention shows the mention pill.
- Muted rooms, the expanded sidebar and the phone sheet get no flyout.

## What changes

All in `apps/web/src/components/chat/`.

- `ChatRoomSidebarRow` asks `useSidebar` whether the rail is collapsed. A room with unread Threads there gets a `HoverCard` around its button, and its name tooltip is dropped for as long as the card stands in.
- `ChatRoomThreadRows` gains `variant="flyout"`, which drops the rule and the indent that tie the inset rows to a row above them. Same component, same `room.unreadThreads`. No second data source.

## Where this differs from the issue

The issue says the rail "reuses the expanded rows as tooltip content". The rows are reused; the container is a hover card, not the tooltip.

- The rows are links, and a pointer has to be able to move onto them. Tooltip content is a description of its trigger, not something to move into.
- The tooltip surface is `bg-primary-solid`, on which sidebar-styled rows do not read.
- Rows inside the tooltip plus the name tooltip itself would be two floating layers on one hover. The card replaces the tooltip and carries the room's name as its heading. The name also stays in the link's accessible name.

## Limit, stated plainly

**The flyout is a pointer surface only.** Radix takes a hover card's content out of the tab order and does not open it for touch. From the keyboard, or on a touch device wide enough to show the rail, the rows are reached by expanding the sidebar or by opening the room, where the reply bars and the thread list hold the same Threads. Story 17 ("unread Threads still reachable") is met for pointer users and only indirectly for the others. A click-opened popover would conflict with the mark's own click, which opens the room.

## Verified in a real browser

On a throwaway no-login page, since deleted along with the env files copied for it. Dark mode only: the app sets its theme by class, so the pane's light emulation did not switch it.

- The flyout opens on hover with the room name, three rows and the overflow row. The pointer travels into it and rows highlight.
- The rail's inset lists compute to `display: none` and the flyout's list to `display: flex`, so nothing is announced twice. The `group-data-[collapsible=icon]:hidden` rule does not reach into the portal.
- Pills: none on a thread-only room, the mention pill on a room with a Thread mention, the unread pill on a room with channel unread.
- The same page showed layers 2 and 4 for the first time: the tinted "2 new replies" bar, the inset rows with their overflow, and the row's mention badge sitting on the row rather than on the inset rows.

The check was done with a 150 ms close delay. It is now 120 ms, to be gone before the next room's name tooltip is up, and that value was not re-checked.

## Vocabulary

Found in review, fixed in #4992 because that layer owns vocabulary, and the upper layers rebased onto it. `CONTEXT.md`'s **Sidebar row** entry named exactly three things that may differ between the expanded sidebar and the rail. A room's inset unread Thread rows are a fourth, with the flyout as their form on the rail. ADR-0037 gains a sentence on the rail and on the pill keeping its vocabulary.

## Tests

Thin component tests, as agreed in the issue.

- Web: 7280 passed, 1 skipped, run before the review edits. After them and the stack rewrite, only the chat suites (1769) and the Core room routes (619 passed, 23 skipped) were re-run. Typecheck and Biome clean.
- Collapsed with unread Threads: the flyout holds the room name and a link to the first unread reply, and the name tooltip is gone. Collapsed without: the plain tooltip stays. Expanded: no flyout. Muted: no flyout.

The tests render the card's content inline, so they cover whether there is a card and what it holds. When it opens, and that a pointer can enter it, rest on the browser check above.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/chat-room-sidebar-row.tsx`: `RAIL_FLYOUT_CLOSE_DELAY_MS` is documented as clearing the card before the next room's name tooltip. `TooltipProvider` uses `delayDuration={0}`, so that tooltip is immediate and the two layers overlap for 120ms. The path into the card is the 12px `sideOffset`, which 120ms still covers.
- Minor: `apps/web/src/components/chat/chat-room-sidebar-row.test.tsx`: no case for `isMobile: true` (the phone sheet must not mount the card). Open/close timing and the portal escaping `group-data-[collapsible=icon]:hidden` stay outside the inline hover-card mock, as the PR describes.
- Minor: `apps/web/src/components/chat/chat-room-sidebar-row.tsx`: reorder mode suppresses the inset rows but not the rail flyout. That state is only reachable if reorder starts while expanded and the sidebar is then collapsed; the drag handle is already hidden on the rail.
- Skipped / disagreed: keyboard and touch not reaching the flyout rows — the PR and ADR-0037 require a pointer surface, with the same threads available by expanding the sidebar or opening the room. Radix sets the card's links to `tabindex="-1"` and does not open on touch. On React 19 the trigger's `touchstart` `preventDefault` is a passive-listener no-op, so the room link's click still fires.
- Skipped / disagreed: hover card instead of tooltip content — the rows are links a pointer has to enter, and the tooltip surface is `bg-primary-solid`.
- Skipped / disagreed: not stripping `group-data-[collapsible=icon]:hidden` on the flyout rows — that utility only matches descendants of `.group[data-collapsible=icon]`. The card portals to `document.body`, outside that group, which matches the browser check (inset list `display: none`, flyout list `display: flex`, rows visible).

### Assessment
The collapsed rail reuses `ChatRoomThreadRows` from the same `unreadThreads` list, drops the name tooltip while that card is mounted, and leaves the attention pill on `resolveRoomAttention`.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5012#issuecomment-5767556702)

All three minors verified and fixed in fb01d8624. #5013, #5078 and #5080 are rebased onto it.

**Minor — the close-delay comment: fixed.** Confirmed: `TooltipProvider` runs at `delayDuration={0}`, so a neighbour's name tooltip is up before any close delay could clear the card. The comment now says what the 120ms is for, which is crossing the 12px `sideOffset` onto the card without it closing under the pointer, and that the two layers overlap while the card is on its way out. No behaviour change.

**Minor — no `isMobile: true` case: fixed.** The guard was there (`!isMobileSidebar`); it had no test. One now pins that the phone sheet mounts no card and keeps the plain name tooltip.

**Minor — reorder mode suppressed the inset rows but not the flyout: fixed.** `showRailThreadFlyout` now also requires `reorderHandle == null`, with a test. Reorder starts expanded, and collapsing the sidebar with it still on reached the flyout.

**Skipped items:** agreed on all three. The pointer-only limit stands as the PR states it, and the note that on React 19 the trigger's `touchstart` `preventDefault` is a passive-listener no-op is a useful confirmation that a tap on the mark still opens the room.

mrosberghaus:

PR #5011, opened 2026-09-21
fix(apple): keep unread attention on the open room like web

mrosberghaus:
Apple parity row **05a**: open-room attention follows web. One vertical slice; no `apps/web`, `apps/core`, `docs/adr`, Core contract, dependency, project or entitlement change.

## What changed

- `SokosumiChat/SidebarRooms.swift`: `resolveRoomAttention` lost the `isActive` parameter and its half of the early return. Only muted rooms suppress the chrome now, as on web.
- `ConversationSidebarView.roomRow`: the `isActive:` argument is gone. This was the only call site (closed-section attention never passed it; there is no Dock badge or notification code on this resolver). The row view itself is unchanged, so the selected row's accessibility text (badge spoken form, "N unread messages") is what any other row has.
- Before: an open room holding residual thread unread, or one whose history was hidden or failed to load, showed no bold, badge or count until another room was selected. Now the selected row keeps them until the room is read, marked read or muted.
- Unchanged: muted suppression, the `99+` cap, closed-section attention, Mark unread disabled on the open room and on muted rooms (`ConversationSidebar.canPerform`), native `List` selection, the Settings caption.
- `PARITY.md`: row 05a in review with evidence, row 12d Done (#5006 merged 2026-09-21; its unverified notes, the marked-text exception and the review follow-up are kept), work order advanced so the next row is 12c.

## Web sources audited (read at `612029ada`)

- `apps/web/src/components/chat/room-attention.ts` :24-42 — the resolver takes `unreadCount`, `unreadMentionCount`, `markedUnread`, `isMuted`, `showUnreadCount` and nothing else; muted is one early return (:33-35); the header (:13-19) states that having the room open is not a reason to drop the chrome. `resolveSectionAttention` :53-75 uses the same resolver.
- `apps/web/src/components/chat/chat-room-sidebar-row.tsx` :324-330 — the row passes exactly those five values; `isActive` only drives `aria-current`, the selection bar and the muted dimming; Mark unread is `disabled={isActive || isPending || isMuted}` (:601).
- `apps/web/src/components/chat/chat-unread-document-title.ts` :8-14 — the tab title counts the open room like any other.
- `room-read-overlay.ts`, `use-room-read-attention.ts` and `room-attention.test.ts` — no open-room rule.
[...]

PR #5014, opened 2026-09-21
fix(apple): let the code block control toggle a block back off

mrosberghaus:
Apple parity row **12c**: the composer's Code block control is a toggle on web and was wrap-only on Apple. One vertical slice; built on `origin/main` `9aca29a8e`.

## What changed

- `Packages/SokosumiChat/.../ComposerBlockFormat.swift` — new pure `edit(in:selection:) -> Edit` (range, replacement, caret) over the **whole** attributed text. For Code block it first looks for the enclosing block at the selection's start and unwraps all of it into plain lines; otherwise it wraps the selection or opens an empty block. Quote and the lists go through the same entry point with their behaviour unchanged.
- `Sokosumi/Chat/Composer/MacComposerTextInput.swift` — `applyBlockFormat` now hands the model the whole text and the selection and only applies the edit (one registered undo, caret collapsed). For Code block it also sets the typing attributes from the block the caret was left in, then runs the 12d input rule once, as web's `handleInput` does after the toggle.
- Tests in `ComposerBlockFormatTests.swift` (new suite `ComposerCodeBlockToggleTests`) and `MacComposerTextInputTests.swift` (incl. a light/dark fixture); four fixture PNGs under `apps/apple/docs/screenshots/`; one sentence in `docs/rich-composing.md`.
- `PARITY.md`: row 12c rewritten with evidence, **row 05a marked Done** (#5011 merged 2026-09-21; its unverified notes, the ADR-0037 note and review follow-up are kept), resume checkpoint and work order advanced so the next row is **22a**.
- iOS: there is no iOS composer (every input file under `Chat/Composer/` is `#if os(macOS)`), so nothing was built; the decision lives in the portable package and is tested there.

Defects before this change, each shown red first through the real text view: with the caret inside a block the control opened a second, empty fence in front of it; `hello` + control + typing sent `hellolet x` and an empty fence. Found while auditing the wrap direction and fixed in the same model: a wrapped mention chip was sent as U+FFFC, a block made inside a quote left the quote, and a new block merged into a block beside it.

## Web sources audited (at `9aca29a8e`)

- `apps/web/src/lib/utils/composer-wysiwyg-code-block.ts` — `closestCodeBlock` :13-25, `createCodeBlock` :27-45, `unwrapCodeBlock` :68-102, `toggleComposerCodeBlock` :108-146; unchanged since #4794.
- `apps/web/src/lib/utils/composer-wysiwyg-code-block.test.ts` — all ten tests, mirrored one for one.
- `apps/web/src/lib/utils/composer-code-text.ts` — `readComposerCodeText`.
- `apps/web/src/components/chat/composer-wysiwyg-editor.tsx` — toolbar case :914-922, `handleInput` :496-503, shortcuts :1209-1233.
- `apps/web/src/lib/utils/composer-active-formats.ts` :123, :145; `composer-wysiwyg-input-rules.ts` :151-171.

Where the source left a question open I ran web's own functions in a throwaway vitest file **outside the repository** (nothing under `apps/web` was touched): a selection across two blocks unwraps only the anchor's; a selection anchored in plain text wraps; ` ```swift ` unwraps to bare lines; wrapping flattens bold/inline code and writes a chip as `@alice`; a block stays inside its `blockquote` both ways; inside a list item web nests `pre` in `li` and serializes markdown its own `markdownToHtml` cannot read back.

## What in the row (and the brief) was stale

- "keep … the existing shortcut": there is none. Web binds only ⌘B/I/U/K, Apple the same. None was added.
- "unwrapping should not retro-format" (brief): wrong, in the same way the 12d brief was. The toolbar case calls `handleInput()` after the toggle, which runs `tryApplyComposerInputRuleAtCaret`; after an unwrap the caret is at the end of the last text node, outside any `pre`. Run on web: `a` / `**x**` unwrapped became `a<br><strong>x</strong>` (markdown `a\n**x**` before and after); `**x**` / `tail` stayed literal. Current web wins, so Apple runs the rule once after the toggle. The sent markdown is byte-identical either way.
- "the whole enclosing block, as Quote does": Apple's Quote only ever unwrapped the selected lines; the code block now really is the whole enclosing block, as on web.

## Behaviour decisions

Mirrored web: the anchor's block and only it; plain lines with newlines and blank lines kept; caret at the end of the unwrapped text, and the line break an untouched empty block leaves behind; language dropped with the fence; wrap flattening incl. the chip label; caret at the end of the code; an empty block that holds the caret and sends what is typed into it; quotes kept both ways; the input rule at the caret after the toggle; the highlight.

Kept Apple: (1) AppKit does not expose a selection's anchor, so the selection's **start** stands for it — a selection dragged backwards out of a block behaves like a forward one; (2) wrapping inside a list item still takes the text out of the list (unchanged from 12b), because web defines nothing usable there — unwrapping a block that a restored draft has inside a list item does keep the item and marker; (3) undo is native (web has none).

Observed, not part of this row: a selection wrapped from the middle of a line shares that line in the editor (it is still sent as a fence on its own lines, and unwrapping restores the line exactly); web's `code` and `quote` toolbar cases also end in `handleInput`, which Apple does not mirror for those two controls.

## Verification (from `apps/apple`, on `9aca29a8e`)

- `swift test --package-path Packages/SokosumiChat` — 654 tests passed; `SokosumiWorkspace` 125, `CoreAPI` 1, `SokosumiAuth` 35, `SokosumiRealtime` 48 — all passed.
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi -configuration Debug -destination 'platform=macOS,arch=arm64' -skipPackagePluginValidation DEVELOPMENT_TEAM= CODE_SIGN_IDENTITY=- -enableCodeCoverage NO` — TEST SUCCEEDED, twice; xcresult summary 1019 tests, 0 failed, 0 skipped. No flaky failure in either run.
- The same invocation with `build` — BUILD SUCCEEDED.
- `mint run swiftformat --lint .` — 0/391 files require formatting; `mint run swiftlint lint --strict` — 0 violations in 391 files.
- `origin/main` was fetched again right before committing and had not moved.

Red first: package — 27 of the 31 new tests failed (65 issues) against a port of the old behaviour behind the new signature; app — 18 of 20 new cases failed against the unchanged editor.

Fixtures (recorded as test attachments, exported with `xcresulttool export attachments`, inspected): block monospaced on its tint with the control highlighted — `composer-code-block-on-light.png`, `-on-dark.png`; after the toggle plain lines, no tint, control idle — `-off-light.png`, `-off-dark.png`. The fixture asserts what it shows (`activeBlocks`, tint range, fonts, bound draft, caret), not only records it.

## Not verified

- The app was not launched. Not exercised live: pressing the control with the mouse in a room and a thread, ⌘Z / ⇧⌘Z right after a toggle in the app, the highlight while clicking between a block and plain text (in particular the caret at the very start of a block that follows a paragraph: the highlight reads AppKit's typing attributes, the toggle reads the block under the caret), a selection dragged backwards out of a block, an input method, sending a toggled draft and reading it on web, VoiceOver.
- SwiftUI builds no accessibility tree in the test host, so the toolbar button was not clicked; tests call `MacComposerCommands.apply`, the closure the button runs.
- ⌘Z grouping cannot be shown in the XCTest host (it never closes `NSUndoManager`'s event group). The in-host test asserts the isolated step. Grouping was checked in a throwaway AppKit program with a real run loop using the same `breakUndoCoalescing`/`registerUndo` sequence (unwrap and rule in one event, after typed text): first undo restored the block, second removed the typing, redo walked forward the same way. That program is not in the repository.
- The fixture turns the format-toolbar preference on for its own duration and restores it, because the test host shares the app's defaults.


mrosberghaus:
## Grok review

### Fixed
- Important: toggling a code block off kept a newline the selection never had, so a mid-line selection or the first word of a later paragraph came back split (`3a0f810b9` — `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/ComposerBlockFormat.swift`)
- Important: a mention chip inside a fence unwrapped to U+FFFC instead of the label it shows (`3a0f810b9` — same file)

### Remaining
- Minor: `MacComposerCommands.refresh` still reads typing attributes for a collapsed caret, while the toggle reads the character under the caret. At the boundary between a paragraph and a block those can disagree; the PR already records that click as unverified.
- Minor: a mention inserted from the toolbar while the caret is already inside a fence is still sent as U+FFFC until the block is toggled off. `ComposerBlockText` serializes a code run with `body.string`.
- Skipped: quote and inline-code controls still do not run the input rule after the toggle — the PR leaves that outside this row, matching the current web note only for the code-block control.
- Skipped: a backwards selection uses the range start, and wrapping inside a list item still lifts the text out of the list. Both are stated kept behavior, not regressions.

### Assessment
The code-block control now restores a line it split and writes a chip's label on the way out; package tests for that model pass.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5014#issuecomment-5768036480) — follow-up is in `2f60d537b`, on top of your `3a0f810b9`.

- **`3a0f810b9` (mid-line unwrap rejoins the line; a chip unwraps to its label)** — read and kept. One change to it: `replaceReferenceLabels` moved into `ComposerReferenceText.replaceChipsWithLabels(in:)` so the serializer, unwrap and wrap share one path, and it now repeats the label when two identical chips sit side by side. They share one attribute run, and replacing the run with a single label collapsed two chips into one; a test pins it.
- **A mention inserted while the caret is inside a fence is sent as U+FFFC** — fixed, in the serializer rather than at the button. The path: typed `@` and emoji are guarded by `caretIsInCode`, the toolbar mention button is not, and `ComposerBlockText` serialized a code run with `body.string`. Web's `openMentions` has no code guard either (`composer-wysiwyg-editor.tsx` ~:949), so the picker opening inside a fence is web behavior and stays. What web sends was checked by running its `htmlToMarkdown`: the `pre` branch serializes the block's text content (`composer-markdown-dom.ts`:499-504), so a chip in a fence goes out as the label it shows (`@Alice Example`, `#general`) and notifies nobody. The code branch of `ComposerBlockText` now does the same. Red first: `aChipInAFenceIsSentAsTheLabelItShows` and the quote-nested case failed on U+FFFC against the unchanged serializer; `aMentionPickedInsideAFenceIsSentAsItsLabel` drives the real text view through `openMentionPicker` and `acceptMention` and failed with the serializer line disabled. Also pinned: a chip outside code is still sent as its token, the serialize → parse → serialize round trip is stable, and a channel chip in a fence has the same fix.
- **Inline code** — different root cause, no U+FFFC there, not changed: Apple reads the reference token before the code attribute, so the chip ends the code run and stays a live mention, where web keeps the token literal inside the backticks. The current behavior is pinned in a test and the difference is recorded in PARITY.
- **`refresh` reads typing attributes while the toggle reads the character under the caret** — tested, no disagreement found, no source change. `theHighlightAgreesWithTheToggleAtABlocksEdges` puts a block directly between two paragraphs and checks four carets (start and end of the block, start of the paragraph after, end of the paragraph before): in each the highlight and the control's action agree. This was shown with `setSelectedRange` on the real text view, not with clicks or arrow keys in the running app.
- **Skipped items** (quote and inline-code controls not running the input rule, backwards selection using the range start, wrapping inside a list item) — agreed, unchanged.

Verification on `2f60d537b`: SokosumiChat 663 passed; full workspace `xcodebuild test` 1030 tests, 0 failed; strict SwiftLint/SwiftFormat clean. A whitespace-only SwiftFormat fix to the new test file came after the full run; its suite was re-run (6 of 6), the full run was not. Not exercised in a running app: picking a mention inside a fence with the mouse, and reading the sent message on web.

PR #5016, opened 2026-09-22
fix(core): stop POST /tasks dying on Error 479 trailing bytes

mrosberghaus:
## Root cause

Sentry **SOKOSUMI-CORE-3T** (`Error 479 trailing bytes` on `POST /v1/tasks`) is not first-party text. It is Ably’s msgpack decoder:

```ts
throw new Error(buffer.byteLength - decoder.offset + " trailing bytes");
```

(`ably` / `@ably/msgpack-js`.) Node REST defaults to msgpack. A JSON or padded HTTP body decodes as a tiny msgpack value, then throws with the leftover length (here `479`).

OpenRouter `generateTaskName` already `catch → null` around `generateText`. That path does not produce this string. The uncaught create-path gap was `resolveTaskName` awaiting `generateTaskName` with no try/catch, so any throw (including this parser error) 500’d the request after a legitimate body.

## Fix

- Ably Rest clients: `useBinaryProtocol: false` so REST uses JSON, not msgpack.
- `resolveTaskName`: catch name-generation throws and keep the existing non-LLM fallback so create still returns 201.

Malformed JSON bodies remain 422 from the OpenAPI validator (unchanged). This error is not a request-body parse failure.

## Proof

- `ably rest client > disables msgpack so JSON REST bodies cannot throw trailing-bytes`
- `openrouter.client > returns null when generateText throws trailing bytes`
- `resolveTaskName > falls back when generation throws trailing bytes`
- `POST /tasks > creates the task when name generation throws trailing bytes` (201, not 500)

Tip SHA: `18fc2ead235c8dbc98bf367d8deeee926482a950` (fix commit `a90b6ec224704238eec2f06c21b90ec9e0e78a57`; later `main` merges by @mrosberghaus).

<div><a href="https://cursor.com/agents/bc-5c1acb4f-22a2-5b07-80a4-879c675d52cb?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5c1acb4f-22a2-5b07-80a4-879c675d52cb&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


mrosberghaus:
### Grok human call — Ably-publish throw on POST /tasks

**Decision: dismiss (no extra test).**

Rationale:
1. **Root cause is already guarded at the Rest constructor.** `useBinaryProtocol: false` stops the msgpack decoder path; `ably/client.test.ts` asserts that on both Rest clients.
2. **Create cannot 500 from an Ably publish throw today.** `notifyTaskHumanAssignee` (and the other task notify helpers) already `try/catch` → Sentry and swallow. Un-mocking it in `POST /tasks` to inject `479 trailing bytes` would only re-prove that catch-all, not the msgpack boundary Grok named.
3. **Cost vs signal.** Wiring a thinner publish-boundary test for this string specifically either (a) fights the protocol switch (decoder no longer runs) or (b) stubs `channel.publish` to throw — which is any-error coverage we already get from the notify catch. Not worth expanding the layered trailing-bytes suite Grok already called heavy.

Leaving the intentional skip of trailing-bytes → 400 as-is (would mis-classify Ably failures; OpenAPI still 422s bad JSON).

Still waiting on Andreas before merge.

mrosberghaus:
## Grok review

Reviewed `c980da996..18fc2ead2` (merge-base of `main`).

### Fixed

### Remaining
- Minor: `apps/core/src/helpers/task-name.ts`: the new catch does not run for this error in production. `generateOpenRouterText` already catches `generateText` and returns null, so `POST /v1/tasks` does not 500 on `479 trailing bytes`.
- Minor: `apps/core/src/lib/ably/client.ts` comment, plus the three name-generation tests, describe task create. Sentry [SOKOSUMI-CORE-3T](https://masumi.sentry.io/issues/SOKOSUMI-CORE-3T) is `hasAppInFront` → `getSubscribeRestClient().channels.get(name).status()` while sending a notification email, on `POST /v1/tasks/:id/events`. `appInFront` already catches that, reports it, and returns false, so the email sends now instead of after the delay. The HTTP request is not a 500.
- Skipped / disagreed: an extra test that injects an Ably publish throw into `POST /tasks`. Same call as the earlier human decision to dismiss that test. The production frame is `channel.status()` on the subscribe client, not publish. `ably/client.test.ts` already asserts `useBinaryProtocol: false` on that client. A stubbed `publish` throw would not lock the decoder.

### Assessment
`useBinaryProtocol: false` on the shared Rest clients is what stops the msgpack decode inside `channel.status()`. The task-name catch does not change that path.

PR #5074, opened 2026-09-22
fix(chat): stop the section clip cutting the last room's rail ring

mrosberghaus:
## What

The last room in every sidebar section rendered open-bottomed on hover in the collapsed rail — the ring's bottom stroke was missing.

## Why

`ChatSidebarSectionContent` needs `overflow-hidden` for its height animation, and the rows fill that box top to bottom. The collapsed rail's hover `ring-1` and active `ring-2` paint *outside* the button box, so the clip landed exactly on the first and last row's own edge and took the outer stroke off.

Measured against a running app, the clip left **0px** of vertical headroom at both ends:

```
{"headroomTop":0,"headroomBottom":0,"needed":2,"verdict":"FAIL"}
```

## Fix

`-my-0.5 py-0.5`, mirroring the `-mx-2 px-2` pair already there for the rail attention pill: the negative margin pushes the clip edge out by the 2px the ring needs, the padding puts the rows back.

```
{"headroomTop":2,"headroomBottom":2,"needed":2,"verdict":"PASS"}
```

Outer height is unchanged (−2px margin, +2px padding), so the collapse animation and the gap to the next section stay put. Verified in the browser: the rows render at the same offsets (`105 / 137 / 169`) before and after, and the hovered last row now draws a complete ring.

## Tests

New regression test at the class seam, alongside the existing `-mx-2 px-2` one. Proven red-capable — dropping the two classes fails it:

```
× heightens the clip box by the rail ring the first and last row wear
```

`chat-sidebar-section-header.test.tsx` + `chat-room-sidebar-row.test.tsx`: 75 passed. `biome check` and `turbo typecheck` clean.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/chat-sidebar-section-header.tsx`: `-my-0.5 py-0.5` is `0.125rem`, while the rail ring is a fixed `2px`. Below a 16px root (Apple Dynamic Type xSmall), the pressed `ring-2` and `focus-visible:ring-2` can still lose a fraction of the outer stroke. Hover `ring-1` still fits.
- Minor: that pair overlaps the section header by 2px, so the bottom 2px of the header button does not receive hits. Settled row offsets and the gap to the next section stay the same.
- Skipped / disagreed: collapse jump — not reproduced. At `height: 0` the padding still occupies 4px and the negative margins cancel it, so the next section stays put, same as before this change.

### Assessment
Clip headroom is 2px and the row positions match the layout this change was meant to keep.

mrosberghaus:
Both minor findings were real. Fixed in b25db03d0.

**`0.125rem` vs a fixed 2px ring** — reproduced. Measured against a running app, the outset tracks the root font size while the ring does not:

| root | outset | ring needs |
|---|---|---|
| 12px | 1.5px | 2px |
| 16px | 2px | 2px |

Now `-my-[2px] py-[2px]`. Re-measured at a 12px, 16px and 20px root: headroom is 2px at each, top and bottom.

**Heading loses its bottom 2px** — reproduced. `elementFromPoint` down the heading's centre line, before the fix:

```
-1px from heading bottom → collapsible content  (isHeader: false)
-2px                     → collapsible content  (isHeader: false)
-3px                     → header-1             (isHeader: true)
```

The content box now takes no pointer events and hands them back on its rows (`pointer-events-none [&>*]:pointer-events-auto`). The rows start inside the 2px padding, so the strip falls through to the heading and the rows keep their own edges. After:

```
headerBottomEdgeHitsHeader: true
rowKeepsCenter: true,  rowKeepsTopEdge: true,  rowKeepsBottomEdge: true
```

Rows stay flush with the heading above and the next heading below at every root size, so the offsets this change was meant to keep are unchanged.

**On the mechanism** — `overflow-clip-margin: 2px` would say this directly, with no negative margin and so no hit-test question at all. I tested it and it works in Chromium, but MDN's compat data has `safari: false`, where it would silently fall back to clipping at the row's own edge — the original bug. So the margin/padding pair stays, with the pointer-events pass-through covering what it costs.

**On the collapse jump** — agreed, not a jump. Worth recording why the 4px is invisible rather than merely cancelled: Radix unmounts `CollapsibleContent` when closed, so the 4px padding box only exists while the section is open or animating.

Two regression tests cover the new classes; dropping either fails. Full chat suite: 1712 passed.

PR #5076, opened 2026-09-22
test(apple): wait for native view readiness in GUI fixtures

mrosberghaus:
The composer GUI fixtures can assert before SwiftUI mounts the editor. The failed CI result bundle for SOK-1153 reports `textView(in: host) == nil` after the 200 ms startup sleep, before any keystroke is sent.

Extract the existing transcript polling pattern into a shared readiness helper with a 10-second deadline. Composer fixtures now wait for the editor and final text/binding state; the mention fixture waits for its editor and suggestion options; the scrolling fixture waits for its initial bottom position. Production code and test parallelism are unchanged.

Add a regression that delays mounting a real composer beyond the former 200 ms sleep. It fails with the same nil-editor expectation before the fix and passes afterward.

## Validation

- Three fresh full-target `SokosumiTests` runs: 139 tests / 244 executions each, zero failures.
- Delayed-mount regression: red before the fix, green afterward.
- SwiftFormat and strict SwiftLint: clean across 387 files.
- Normal commit hooks: Biome and all 17 Turbo typecheck/dependency tasks passed.
- `git diff --check`: clean.

The natural composer flake did not reproduce in four unchanged local runs; the deterministic regression covers its captured failure condition. CI validation remains pending. An exploratory Xcode repetition run also exposed a separate sidebar-emphasis failure (`drawnEmphasized == 0`); it did not recur in fresh runs and is outside this focused fix. Other fixed observation periods remain unchanged.

Refs [SOK-1153](https://linear.app/masumi/issue/SOK-1153/apple-gui-tests-flake-under-full-target-parallelism-replace-fixed).


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/SokosumiTests/Chat/Timeline/TranscriptTestReadiness.swift`: `import Testing` is unused after `#require` moved into `waitForView`.
- Minor: `apps/apple/SokosumiTests/NativeViewTestReadiness.swift`: a wrong composer string, mention list, or scroll offset waits the full 10s, then fails with only "Expected view did not become ready within 10 seconds". The old transcript helper distinguished "did not appear" from "did not finish initial layout".
- Skipped / disagreed: other fixed sleeps (message-link 600ms loop, pagination, sidebar) stay on purpose. This PR does not claim to replace them.
- Skipped / disagreed: the delayed-mount test does not pin the 10s deadline. The requirement is a real composer mounted after the old 200ms sleep, and `loadedInput` already calls the shared helper. Hit-testing the mention button as soon as the text view exists is also fine: the successful poll already ran `layoutSubtreeIfNeeded`, and the post-click wait is on mention option ids.

### Assessment
The shared 10s poll matches the old transcript helper, the composer and scroll fixtures wait on the real end state, and the delayed-mount regression covers the captured nil-editor failure.

mrosberghaus:
Addressed the timeout-diagnostic note in eda9e06b1.

- Each wait now supplies its expected condition and observed editor/draft text, mention IDs, or scroll geometry. The message is evaluated at timeout, so it reports the final observed state.
- Transcript readiness again distinguishes an absent transcript from one whose initial layout is incomplete.
- Failures retain the calling source location through `#_sourceLocation`.

The `Testing` import was unused in the reviewed version. With the new caller-location default, it is required at the call site: removing it produces `no macro named '_sourceLocation'`. I retained it with a comment documenting that compiler requirement.

Validation: a temporary forced-timeout probe failed after 10.023 seconds with the updated observed state and the calling test's line number; the probe was then removed. The final full app target passed all 139 tests / 244 executions. SwiftFormat, strict SwiftLint, `git diff --check`, and normal Biome/typecheck commit hooks passed. The 10-second deadline and other fixed observation periods are unchanged.

PR #5013, opened 2026-09-21
feat(chat): two sidebar marks, a muted number and a primary @ pill (SOK-1147, 6)

mrosberghaus:
Layer 6 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): the sidebar's two marks. Stacked on #5012, the top of the stack.

Not one of the issue's five layers. It started from the design the issue was built from ([mock](https://thread-unread-mock.vercel.app/)), which marks mentions with an `@` on rooms and on thread rows, and ended somewhere simpler after several rounds with Andreas. The branch name and the earlier commits still say "amber"; the last commit is what ships.

## What a reader notices

The sidebar had grown four treatments for a count. It now has two.

- **A muted number says how much is unread.** The same number, in the same column, for a channel, a Direct and a Thread's inset row. It follows the reader's count option, as before.
- **A tinted primary pill with an `@` says you were named.** On a channel, a group Direct, and a Thread row. It matches the collapsed rail's mention pill, which was already primary.
- **One number per row.** A row that names you draws the pill and not the count; bold still says there is more.
- **A Thread's icon is the extra signal.** A neutral circle for an unread Thread, a primary `@` circle for one that names you.
- A Direct of two looks like a channel: bold and a muted number.

## What changes

**Core**
- Each entry of `unreadThreads` carries `unreadMentionCount`: the unread replies in that Thread that name the viewer. One `COUNT(*) FILTER (WHERE EXISTS …)` on `chat_room_user_mention` in the query layer 4 added. No gating predicate is touched.

**Web**
- `resolveRoomAttention` takes `badgeCountsMentions` and reports `mentionCount`, what the row draws as its pill. `badgeCount` still drives bold and the rail. `unreadTextCount` is 0 whenever a pill is drawn.
- `MentionCountPill`: one pill, used by the room row and the thread row.
- The reader's count moves from "· 3" after the room's name into the pill's own column, as muted text, and holds the same hole open that a pill does.
- `ChatRoomThreadRows` draws the icon and either the muted number or the pill. New string `App.Channels.ThreadRows.mentions` in `en`, `de` and `es`, for screen readers.
- ADR-0027 carries a correction: it said the count and the badge can sit on one row.

## Rules worth a reviewer's eye

[...]

PR #5078, opened 2026-09-22
feat(chat): thread unread leaves room unread, surfaced on the thread (SOK-1147, layers 1-7)

mrosberghaus:
[SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): a Thread reply stops marking its channel, and Thread unread gets its own surfaces. Layers 1 to 7 of the stack, collapsed into one PR against `main`. Layer 8, the column drop, stays separate as #5080 because it must deploy after this one (see **Deploy order**).

The seven layer PRs are closed and keep their review threads: #4992 (vocabulary), #5000 (reply bar), #5007 (de-bold), #5010 (inset rows), #5012 (rail flyout), #5013 (two marks), and this one (count on by default). Their descriptions carry the per-layer detail; this body is the whole.

## What a reader notices

- **Reading a channel clears that channel's mark.** A channel whose only unread is Thread replies is no longer bold and shows no number. A User mention inside a Thread still bolds and badges it.
- **The parent of an unread Thread says so in the transcript.** A tinted bar reading "2 new replies" replaces the plain count while any are unread. A lurker on a busy thread sees the plain count.
- **Unread Threads sit under their channel in the sidebar.** Up to three inset rows, each named by its parent and carrying its unread count, linking to the first unread reply. Past three, a "4 more unread threads" row opens the room's thread list. Rows vanish as they are read.
- **The collapsed rail keeps them reachable.** Hovering a room with unread Threads opens a flyout with the same rows. The rail's attention pill means what it meant before.
- **Two marks, one number per row.** A muted number says how much is unread, the same way for a channel, a Direct and a Thread row. A tinted primary `@` pill says you were named, and a row that names you draws the pill and not the count. A Thread's icon is the extra signal: a neutral circle, or a primary `@` circle when it names you.
- **The unread count is on for everyone.** The Settings switch "Show unread message counts" still exists and now starts on ([ADR-0038](../blob/sok-1147-07-unread-count-default/docs/adr/0038-room-unread-count-is-on-by-default.md)).

## What changes

**Vocabulary** (layer 1). ADR-0037 is Accepted. `CONTEXT.md`'s Room unread, Unread thread, Thread list, Room last-read, Muted thread and Sidebar row entries match it.

**Core**
- `GET /chats/rooms/{id}/messages` carries `threadUnreadReplyCount` per thread parent, from the existing Participant- and mute-gated thread aggregate, read once per page and skipped when the page holds no thread parent. Optional on the wire: edit, reaction, pin and realtime payloads do not compute it, and a defaulted `0` would let the web untint a bar.
- `getChatRoomUnreadCounts` groups its two legs instead of summing them. All 8 room summary routes carry `channelUnreadCount` and `threadUnreadCount` beside `unreadCount`, which is still the sum, so the Apple client is unaffected.
- The room summary carries `unreadThreads` (at most 3, newest unread reply first, each with `parentMessageId`, `firstUnreadReplyId`, `parentContent`, `unreadReplyCount`, `unreadMentionCount`) and `unreadThreadCount`. The read runs only for rooms whose counts show Thread unread, on the room list and the single-room routes alike, since the sidebar redraws a row from either.
- The preferences route still speaks `showRoomUnreadCount`, derived from the new `hideRoomUnreadCount` column on read and stored as its inverse on write. No client moves.

**Database.** `20260921233000_user_hide_room_unread_count`: `ADD COLUMN IF NOT EXISTS "hideRoomUnreadCount" BOOLEAN NOT NULL DEFAULT false`. Expand only. Its `false`, which every reader starts with, means shown, so every reader lands on the new default with no backfill. `showRoomUnreadCount` stays until #5080, neither read nor written.

**Web**
- `resolveRoomAttention`: bold = channel half > 0, or an unread mention, or a hand-set unread mark. It reports `mentionCount` for the pill and `unreadTextCount` for the muted number, never both. The tab title asks the resolver.
- The reply bar on a thread parent, tinted `bg-primary-quaternary` with `text-primary-variant` and an inset rule, the sidebar mention pill's pair. A thread Look and Mark all clear it at once.
- `ChatRoomThreadRows` under the room's sidebar row, and as `variant="flyout"` inside a `HoverCard` on the collapsed rail. Rows are links to `?message=<firstUnreadReplyId>`; the overflow row opens the thread list through a new `?threads=1` URL ask.
- `MentionCountPill`, `@ 9+` past nine so the glyph never drops.
- `useShowRoomUnreadCount` reads `hideRoomUnreadCount` off the session: shown unless `true`. The Settings switch writes the inverse.
- Copy under `App.Channels.Thread.newReplyCount`, `App.Channels.ThreadRows.*` and the switch's description, in `en`, `de` and `es`. Apple's Settings caption gets the same one line; that is the only Apple change.

Mocked Prisma supplies its own rows, so no test proves the SQL. Three places, all in `apps/core/src/routes/v1/chats/rooms/room-unread.ts`:

1. **The count split.** Both `WHERE` bodies are byte-identical to before. The only changes are a `'channel'::text AS source` / `'thread'::text AS source` literal on each leg and `GROUP BY combined."roomId", combined.source`.
2. **`sqlUnreadThreadReplies`** is extracted from the count's thread leg so the number and the inset list share one `FROM … WHERE`; a Thread cannot be listed while contributing nothing to the number. The count query's SQL text was compared byte for byte before and after the extraction.
3. **The inset list** groups by `parent.id` alone, ranks with `ROW_NUMBER() OVER (PARTITION BY "roomId" …)`, counts with `COUNT(*) OVER`, and adds one `COUNT(*) FILTER (WHERE EXISTS …)` on `chat_room_user_mention` for the per-Thread mention count.

`sqlViewerIsThreadParticipant` and `sqlThreadReplyPagesViewer` are untouched. A diff that modifies either gate has gone wrong.

## Rules worth a reviewer's eye

- **A Direct of two is written to, not named.** Core notifies every message there, so that row draws no pill, shows the muted count, and announces "5 unread messages". Everywhere else the badge counts mentions alone.
- **The two mention counts clear differently.** A Thread row's count comes from the replies, so a Look clears it. The room's comes from notifications, which Room last-read clears.
- **In a 1:1 Direct a Thread reply still bolds the row**, because Core writes a direct-message notification for it. Left alone on purpose.
- **The flyout is a pointer surface only.** Radix takes a hover card's content out of the tab order and does not open it for touch; keyboard and touch readers reach the same Threads by expanding the sidebar or opening the room.
- **A reader who switched the count on and then off again under ADR-0027 gets it on again.** The old column cannot tell "never chose" from "chose off".

## Deploy order

This PR is safe on its own: the migration is additive and the old Core ignores the new column. **#5080 drops `showRoomUnreadCount` and must not run while a Core older than this one is still serving**, since that Core selects the column on every session read. Core runs `prisma migrate deploy` inside its build, before the new build is live, so the two PRs cannot share a deploy. Merge #5080 once this is fully rolled out.

## Rollback

Revert the deploy. The old column still holds each reader's pre-ADR value, the old Core reads it, and the new column sits ignored.

## Verified in a real browser

On throwaway no-login pages, since deleted, dark mode only:

- The tinted reply bar; the inset rows with their overflow; the row's mention badge on the row rather than on the inset rows.
- The rail flyout opens on hover with the room name, three rows and the overflow row, and the pointer travels into it. Pills: none on a thread-only room, the mention pill on a Thread mention, the unread pill on channel unread.
- Every mark centred on one column: room numbers, room pills, thread numbers and thread pills all at 236px. "@ 1" is 21.5px and "@ 9+" 27.9px in the 28px hole.

Not verified: the migration against a database (none in the worktree; `prisma validate` and `generate` pass), and the app end to end with real data.

## Tests

Seams as agreed in the issue: Core route contract with mocked Prisma, the pure attention resolver, thin component tests. One narrow SQL tripwire for the mute and Participant gates. No broad SQL-string assertions, no DB-backed tests.

Core: 6260 passed, 39 skipped. Web: 7300 passed, 1 skipped. All 10 package suites. Typecheck and Biome clean, on the layer 7 head before the final rebase; CI runs on the current head.

Two-axis code reviews and external reviews ran on every layer; their findings are fixed or answered in the closed PRs' threads.

## Follow-ups

- [SOK-1151](https://linear.app/masumi/issue/SOK-1151/thread-reply-bar-lags-live-replies-by-up-to-60s-drive-it-from-the): live replies reach the reply bar and the inset rows on the next refresh, up to 60 s, because the client cannot evaluate Participant or mute gating.
- [SOK-1156](https://linear.app/masumi/issue/SOK-1156/thread-reply-bar-shows-the-repliers-faces-and-the-last-replys-age): the reply bar's repliers' faces and last-reply age from the mock.
- The Apple client bolds on the total and keeps the old behaviour until its own work lands.


mrosberghaus:
## Grok review

Picked up the review request on this PR from @mrosberghaus.

Reviewed `5a46d664d5d6874670247287863bf6417d196197..02885ae511408541766ee7b5b87e01104bb80af9` (base `sok-1147-06-mention-badge`).

### Fixed

### Remaining
- Minor: `apps/core/src/schemas/user.schema.ts`: the `showRoomUnreadCount` example is still `false`, while the description and the PATCH body example now say the count is on unless the reader switches it off.
- Minor: `apps/apple/Packages/CoreAPI/Sources/CoreAPI/openapi.json`: preferences descriptions and examples still describe the old opt-in. The wire field did not change, so Apple does not need a client regen to pick up the new default.
- Minor: `apps/apple/PARITY.md`: row 39 still says the count defaults off and that web writes `showRoomUnreadCount`. The slice 05a note still quotes the old caption.
- Minor: `apps/core/src/routes/v1/users/[id]/coworker-user-route-allowlist.test.ts`: the preferences fixture still has `showRoomUnreadCount` and not `hideRoomUnreadCount`. GET stays 200 because a missing hide flag is treated as shown, and the test never asserts the count.
- Skipped / disagreed: readers who had switched the count off under ADR-0027 see it on again — ADR-0038 and this PR call that unavoidable, because the old column cannot tell "never chose" from "chose off". The sidebar and the Settings switch both read `hideRoomUnreadCount`, so they agree with each other.
- Skipped / disagreed: Apple starts the count off until preferences load — same gate as web's unloaded session. `ChatDisplayPreferences` then applies `showRoomUnreadCount` from GET, which Core now answers as `!hideRoomUnreadCount`.

### Assessment
The new column, the inverted preferences read and write, and the session hook match ADR-0038; what is left is docs and an example, not behavior.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5078#issuecomment-5774850010)

Three of the four minors verified and fixed in 00ddde4c7; #5080 is rebased onto it.

**Minor — `showRoomUnreadCount` example still `false`: fixed.** Now `true`, matching the description and the PATCH body example.

**Minor — the allowlist fixture still named the old column: fixed.** It names `hideRoomUnreadCount` now. This was already fixed on #5080, where the old column goes; it belonged here.

**Minor — `PARITY.md` row 39 and the 05a note: fixed.** Row 39 now says the count is on by default since ADR-0038, stored as `hideRoomUnreadCount`, that web writes it through Better Auth and Apple reads and writes the unchanged wire field through Core, and that the number sits in the badge's column and never beside a badge. The 05a note now records that the caption quoted there was corrected in this layer rather than claiming it "stays so".

**Minor — Apple's `openapi.json` descriptions: not changed here, on purpose.** `apps/apple/AGENTS.md` says not to hand-edit `CoreAPI`'s generated sources and to refresh them with `scripts/update-core-api.py` from a generated Core spec. The stale text is descriptions and examples only; the field's name and type are unchanged, so Apple needs no regen to pick up the new default, as you note. That refresh belongs to whoever next regenerates the Apple client, with the two Core snapshots this stack changed.

**Skipped items:** agreed on both.

PR #4629, opened 2026-09-15
fix(apple): match web thread preview text

mrosberghaus:
## Feature
Thread overview previews now use the same cleanup and mention-resolution rules as web. This closes the remaining preview gap from #4626: HTML/comment text, unnamed mention IDs and addresses assembled around display names no longer produce different thread labels.

The portable formatter replaces the earlier Markdown-flattening path. It runs in the existing detached page preparation, with the room-name lookup built once per page. No view layout, web/Core contract, generated client or dependency changes.

## Web references
- [Thread-row wrapper](https://github.com/masumi-network/sokosumi/blob/05b622d79cbe3957d2ad9a9119caf2886a030f3a/apps/web/src/app/(app)/chat/utils/unread-threads-preview.ts)
- [Shared formatter](https://github.com/masumi-network/sokosumi/blob/05b622d79cbe3957d2ad9a9119caf2886a030f3a/packages/utils/src/chat-message-preview.ts)
- [Existing web tests](https://github.com/masumi-network/sokosumi/blob/05b622d79cbe3957d2ad9a9119caf2886a030f3a/packages/utils/src/chat-message-preview.test.ts)

## Verification
- 177 representable calls from the existing web test suite match exactly. Their original assertions also ran during capture. One lone-surrogate input cannot exist in Swift String; separate native coverage checks other invisible characters and emoji preservation.
- All five package suites pass: Chat 389, Workspace 73, Auth 35, Realtime 27, CoreAPI 1.
- Full macOS app suite passes with coverage disabled. Final formatter/overview rerun: 11 tests pass.
- iOS 17 Workspace compilation, pinned SwiftLint/SwiftFormat and repository pre-commit checks pass.
- Separate macOS build passes.

How to test: run `swift test --package-path apps/apple/Packages/SokosumiChat --filter ChatMessagePreviewTests`; open Threads for messages containing markup or mentions, and compare their previews with web. There is no layout change.

Out of scope: OS notifications, deep links, changing timeline Markdown rendering, new APIs or dependencies. PARITY.md records this as follow-up 24a before slice 25.

PR #5079, opened 2026-09-22
fix(chat): keep intraword underscores in quotes and previews

mrosberghaus:
Quoting `I see it seems to be the COMPOSIO_X_AUTH_CONFIG_ID, correct?` produced a quote block reading `COMPOSIOXAUTHCONFIGID` — every underscore gone, while the message body right beside it printed the name whole.

## Cause

`cleanChatMessageText` (`packages/utils/src/chat-room-quote-snippet.ts`) stripped the whole class `[*_~>#]+`. The body renders through markdown, and CommonMark opens no emphasis on an underscore flanked by word characters, so `SNAKE_CASE` names are literal text there. The plain-text clean was the only place reading those underscores as markers — and it feeds both the stored quote snippet and every sidebar/notification preview.

## Fix

- Strip `_` only where it is **not** flanked by word characters on both sides. `*~>#` behave exactly as before, and `_very_` still cleans to `very`.
- Same rule in the Swift port (`ChatMessagePreview.swift`), so the apple previews stay in parity.
- The shared corpus fixture gains the env var case (177 → 178 examples).

One expectation changed on purpose: a mention slug now keeps its underscore when the roster names nobody (`@ada_lovelace`, not `@adalovelace`). The hyphen case already read that way.

## Not covered

Core stores the snippet at send time (`apps/core/src/routes/v1/chats/rooms/helpers.ts`), so quotes already in the database keep their mangled text. Backfilling those would be a separate migration. Previews are computed live, so they are all fixed.

## Verification

- `packages/utils`: 693 passed (red before the fix: `expected 'COMPOSIOXAUTHCONFIGID' to be 'COMPOSIO_X_AUTH_CONFIG_ID'`)
- Swift `ChatMessagePreviewTests`: 2 passed
- `apps/core` `rooms/helpers.test.ts` 89 passed, `ably/publish.test.ts` 29 passed
- `apps/web` quote/preview suites (unread threads preview, pending quote, composer quote preview): 12 passed


mrosberghaus:
## Grok review

### Fixed
- Important: an underscore run between words was still deleted (`foo__bar` became `foobar`) because `+` backtracked into the run, and a non-ASCII letter was not treated as a word character (`变量_名` became `变量名`). (`05244ecc7` — `packages/utils/src/chat-room-quote-snippet.ts`, `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/ChatMessagePreview.swift`)

### Remaining
- Minor: `packages/utils/src/chat-message-preview.ts`: comments still say the clean turns `R_D` into `RD` and that `@<uuid-with-underscores>` collapses into a mention. Neither happens now.
- Skipped: `apps/web/src/lib/utils/strip-markdown.ts` still strips every `_`. That path is task descriptions and project briefings, not chat quotes or previews.
- Skipped: unmatched edge underscores (`_PRIVATE`, trailing `foo_`) still drop. That is the stated flank rule, the same approximation as `*` and `#`, not full CommonMark.
- Skipped: quotes already stored in the database stay mangled. The PR defers a backfill.
- Skipped: `apps/apple/PARITY.md` still says 177 in the historical #4629 note. That is the old snapshot, not the live corpus (181).

### Assessment
Single intraword underscores were already fixed; runs of them and non-ASCII letters now survive the same cleaner, and `__init__` still strips as strong emphasis.

mrosberghaus:
Checked the two fixed items and the regex in `05244ecc7` holds: `foo__bar` matched the second alternative one underscore at a time before, and `\p{L}\p{N}` is the right class for `变量_名`. `__init__` still cleans to `init`, which is what the room renders.

**Stale comments — fixed** in `a65fe57f4` (`packages/utils/src/chat-message-preview.ts`). Verified both claims against the built formatter before editing:

- `R_D` now survives the clean, so it is not the name that would be mangled. The comment's point stands — names go in after the clean — with `R*D` as the example, which does still read as `RD` when handed to it.
- An underscore inside a uuid no longer spells a key: `@019fc7e4-e4bd-7005-900c-66e44d33f_5e4` stays literal text now. A `*` in the same place still cleans into a mention (`@Ada Lovelace`), so the comment keeps its warning with `*`.

**On the skipped items:**

- `apps/web/src/lib/utils/strip-markdown.ts` — agreed, out of scope here: task descriptions and project briefings, a different surface with its own tests. It is the same bug class though, so I have queued it separately rather than widen this PR.
- Unmatched edge underscores (`_PRIVATE`, `foo_`) — agreed, that is the flank rule as stated. Matching those needs delimiter pairing, not a regex, and it is the same approximation `*` and `#` already run on.
- Stored snippets — agreed, the PR says so; a backfill is its own migration.
- `PARITY.md` 177 — agreed, leaving it. Line 946 records what #4629 captured at `05b622d79`, so it is history, not a claim about the live corpus.

PR #5080, opened 2026-09-22
feat(database): drop showRoomUnreadCount (SOK-1147, 8)

mrosberghaus:
Layer 8 for [SOK-1147](https://linear.app/masumi/issue/SOK-1147/thread-unread-leaves-room-unread-surface-unread-replies-on-the-thread): the contract step of ADR-0038. Follows #5078, layers 1 to 7, merged to `main` as 226cf36f4. This PR is the two commits on top of it, rebased onto that merge.

## What it does

Drops the `showRoomUnreadCount` column, superseded by `hideRoomUnreadCount` in #5078, and with it the Better Auth field and the session type line declared for it.

- **Migration** `20260922090000_user_drop_show_room_unread_count`: `ALTER TABLE "user" DROP COLUMN IF EXISTS "showRoomUnreadCount"`. Idempotent, so a retry after a partial failure is safe.
- Prisma field, `betterAuthUserAdditionalFields` entry and `SessionUser` type line removed. Three test fixtures that still named the old field now name the new one.
- A test pins that the field is gone from the Better Auth schema: declaring a field the database cannot store would break sign-in.
- ADR-0038 records this as its contract step rather than "a later migration".

Nothing reads or writes this column since #5078. The wire field `showRoomUnreadCount` on the preferences route is unaffected; it is derived from the new column and keeps its name, so no client, the Apple app included, sees a change.

## Deploy order

**Do not run this migration while a Core older than #5078's is still serving.** That Core selects `showRoomUnreadCount` on every preferences read and would 500 the moment the column is gone. #5078 has to be fully rolled out first. The migration's header says the same.

## No rollback after it runs

The value it drops was each reader's ADR-0027 opt-in, which #5078's default made moot for every reader. Once the column is gone there is nothing to restore. Until this merges, #5078 stays reversible: revert its deploy and the old column still holds each reader's old value.

## Not verified

The migration has not run against a database; there is none in this worktree. It has passed `prisma validate` and `prisma generate`, and it is one `DROP COLUMN IF EXISTS`.

## Tests

Web: 7297 passed, 1 skipped. Core: 6295 passed, 39 skipped. All 10 package suites. Typecheck and Biome clean.

The web's generated Core client was not regenerated: the wire field is unchanged, and regeneration produced only unrelated drift from `main`, which was reverted.


mrosberghaus:
## Grok review

Reviewed `02885ae511408541766ee7b5b87e01104bb80af9..df1d506ae58b96fa5f15020fe5edace4c8f00122` (base `sok-1147-07-unread-count-default`).

### Fixed

### Remaining
- Minor: `packages/utils/src/better-auth-client-schema.test.ts`: the new "field is gone" test sits inside `describe("betterAuthOrganizationAdditionalFields")` instead of `describe("betterAuthUserAdditionalFields")`. The assertion still runs against the user schema.
- Minor: `apps/core/src/routes/v1/users/[id]/preferences/patch.ts`: the comment above the hide-column write still says `showRoomUnreadCount` stays so a rollback can find each reader's old value. This migration drops that column, and ADR-0038 now says there is nothing to restore.
- Skipped / disagreed: deploy-while-old-Core-serves — the migration header and ADR already require the #5078 Core (which no longer selects the column) to be the serving release first. That matches this stack's previous layer; it is not a code defect. Historical plan docs, ADR-0027, and the already-applied expand migration still name the old column on purpose. The wire field stays, which is the requirement.

### Assessment
The column, Prisma field, Better Auth additional field, and `SessionUser` line drop together, and nothing in this range still reads or writes the column.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5080#issuecomment-5774816304)

Both minors verified and fixed in 7dd798cea.

**Minor — test in the wrong `describe`: fixed.** It was appended to the file's last block, which is the organization one. It now sits under `betterAuthUserAdditionalFields`, the schema it asserts against.

**Minor — the rollback comment in `patch.ts`: fixed.** It said the superseded column stays for a rollback to find, which this layer makes untrue. It now says only what the write does: the wire says "show", the column stores "hide", so a reader who never chose is shown the count. `grep` finds no other rollback note in Core or utils source; the ones that remain are in the expand migration's header and ADR-0027, both of which describe that earlier step on purpose.

**Skipped item:** agreed. The deploy-order requirement is in the migration header, ADR-0038 and the PR description; whoever merges has to sequence #5078's deploy before this migration runs, and `gh stack merge` will not do that on its own.

PR #5082, opened 2026-09-22
fix(chat): stop Seen by faces shortening every line on mobile

mrosberghaus:
Reported by Alexa: "the chat view text padding looks squished" on mobile — one message wrapping normally and the one under it wrapping at about two thirds the width.

## Cause

The Seen by faces are pinned to the row's bottom-right corner (`absolute end-2 bottom-1`), and the message column reserved `pe-20` for them whenever they were present. That is 80px off **every** line to protect the one line that can actually collide with them.

On a phone the body column is ~312px, so the newest message in every room lost a quarter of each line. Since the faces ride `newestMessageId` only, the narrow row is always the one people are reading, and it moves down the transcript as new messages arrive — which is why it read as a random squished message rather than a layout rule.

Measured at iPhone 14 width with the real row markup:

| Row | Body width | `padding-inline-end` |
| --- | --- | --- |
| previous message | 312px | 0px |
| newest message (Seen by) | **232px** | **80px** |

## Change

1. `pe-20` is gone from the content column. No row reserves anything block-level any more.
2. `SEEN_BY_INLINE_RESERVE_CLASS` — a 66px inline-block spacer riding the body's existing `trailing` slot (the same mechanism that sits the "edited" label after the final paragraph), so it shortens the last line only.
3. `bodyEndsTheRow` — the spacer renders only when nothing follows the body. Reactions, a thread link, an unfurl or the failed-outbound row all push the faces beside *them*, so those rows reserve nothing at all. Alexa's message is in that bucket.

66px is what the corner occupies: 52px of faces, plus the 14px the touch target reaches further left below `md`.

After: both rows 312px.

## Verification

- New `room-message-seen-by-reserve.test.tsx`, 5 cases. Confirmed **red** on the old code (restored `pe-20`, watched the column assertion fail) and green on the new — not a test that only ever passes.
- Full chat suite: 137 files, 1314 tests, all green.
- `tsc --noEmit` and `biome check` clean.
- Rendered the real `ChatMessageRow` markup in **real iOS WebKit** (iPhone 18 Pro simulator), not just jsdom: the reported message goes from 8 ragged lines to 5 full-width ones, faces clear below.

Chromium was no use for this one — it now blockifies `-webkit-line-clamp` (computed `display: flow-root`), so the phone rendering only shows up on WebKit.

## Tradeoff

When the last line genuinely runs long, the spacer wraps and adds one ~24px line. That is never worse than before: `pe-20` would have pushed the same word down anyway, and charged every other line for it.

Residual, left alone: a row with many reaction pills could still have its last wrapped pill row reach under the faces. That is cosmetic overlap of bordered pills, not text.


mrosberghaus:
## Grok review

### Fixed
- Important: the inline reserve set the last paragraph to `display: inline`, so `[&_p+p]:mt-3` (the blank line) stopped applying. The previous block paragraph now keeps that gap with `mb-3` (`7785730de` — `apps/web/src/app/(app)/chat/components/room-message-row.tsx`)
- Important: the reserve was `w-[66px]` while the faces and the touch inset are rem and scale with Dynamic Type. It is now `w-[4.125rem]` (66px at the default root) (`7785730de` — `apps/web/src/app/(app)/chat/components/room-message-row.tsx`)
- Important: a Soko Bot reply always renders a footer under the body, but the text still reserved space. The reserve is skipped when that footer is what ends the row (`7785730de` — `apps/web/src/app/(app)/chat/components/soko-bot-message-footer.tsx`, `apps/web/src/app/(app)/chat/components/room-message-row.tsx`)
- Important: deleted rows and quote-only rows never mounted the reserve, so that text still ran under the faces (`7785730de` — `apps/web/src/app/(app)/chat/components/room-message-row.tsx`)

### Remaining
- Minor: `room-message-row.tsx` — reaction pills can still wrap under the faces. The PR left that as cosmetic overlap.
- Minor: `soko-bot-message-footer.tsx` — enough task chips to wrap can still meet the faces. The message text no longer reserves for them.
- Minor: `room-message-row.tsx` — a live thinking row or a failed-mention shell does not mount the reserve.
- Minor: `room-message-row.tsx` — on a 16-line clamped body the spacer can be clipped. Show more then sits in the corner instead of the text.
- Skipped: title-only unfurls — Core already drops those cards, so a raw `unfurls.length` check matches what the row renders.

### Assessment
The newest message keeps full line width, and the reserve no longer eats a blank line or misses a deleted, quote-only, or bot-footer row.

mrosberghaus:
Checked all four fixed items against the codebase and in a browser. All correct, nothing reverted.

**Blank-line gap** — this was a real regression I introduced, and the worst of the four: vertical margins do not apply to non-replaced inline boxes, so `[&_p+p]:mt-3` silently stopped producing the blank line as soon as the reserve made the last paragraph inline. It also pre-existed for edited continuation rows, which take the same `trailing` path.

The added test asserts the class string is present, which happy-dom can confirm but cannot prove renders. Measured it instead, at iPhone 14 width with the real row markup:

| | body height | first-line delta (p1 → p2) | p2 display |
| --- | --- | --- | --- |
| no reserve | 60px | 36px | block |
| with reserve | 60px | 36px | inline |

Pixel-identical, so `[&_p:has(+_p:last-of-type)]:mb-3` restores the gap exactly. (A naive `getBoundingClientRect` diff reads 14px rather than 12px here — that is the inline box reporting its content area rather than its line box, not a real shift.)

**rem instead of px** — confirmed at the source rather than by inspection: `room-seen-by-line.tsx:204` carries `after:-inset-3.5 … md:after:hidden`, so the touch inset is 0.875rem and drops above `md`, exactly as the comment claims. Tailwind's spacing scale is rem throughout, so the corner does scale with the root size. The reserve measures 66px at a 16px root, as intended.

**Soko Bot footer** — `SokoBotMessageFooter` returns `null` only when `readSokoBotMetadata` returns nothing, and otherwise always renders `FeedbackButtons`. So `hasSokoBotMessageFooter` is an exact predicate for "the footer ends the row", not an approximation.

**Deleted and quote-only rows** — rendered both at 390px and confirmed no overlap. The deleted row takes the cheap path (reserve inline inside the `<p>`). The quote-only row pays a full extra line (120px vs 96px) because the spacer lands below the card rather than on a text line — acceptable for a send-to-yourself row, and the card is otherwise full-width under the faces.

**On the remaining notes** — agreed, leaving all four. Worth recording why the clamp one is benign: when a 16-line body overflows, the spacer is clipped, but `Show more` then renders below the clamp and becomes the row's tail. It is short and left-aligned, so the faces sit clear of it and no text ends up underneath. The reaction-pill and task-chip cases stay cosmetic overlap of bordered chips, never text.

PR #5083, opened 2026-09-22
feat(projects): remove project deletion

mrosberghaus:
## Summary

- The project menu no longer offers Delete. Close stays on that menu.
- `DELETE /v1/projects/{id}` returns `409` with `kind: project_deletion_removed` and does not delete the project. Coworker keys still get `403`.
- Removing one task or job from a project is unchanged. Account and workspace erasure can still delete the row.

This is the user-facing cut discussed on #4569. It does not change Close (the project list, project files, or `CLOSE_FAILED` recovery). #4569 should drop its ordinary-delete protection, or rebase after this, so the trigger and tombstone are not added for a route that no longer deletes.

## Verification

- `pnpm check`
- `pnpm typecheck`
- Core `project-id.routes.test.ts`: 20 passed
- Web project detail actions, project actions, project service, and project page tests: passed
- Browser: not opened. The detail-actions test opens the menu and asserts Delete is absent.

## Tracking

- Follow-up to the recommendation on #4569


mrosberghaus:
## Grok review

### Fixed
- Important: brand-context remove dialog still used `App.Projects.Detail.deleteDialog.cancel` after that project-delete copy was removed, so Cancel rendered as a missing key (`9aed5ef6f` — `apps/web/src/app/(app)/projects/components/project-brand-card.tsx`)

### Remaining
- Minor: `apps/core/src/routes/v1/projects/[id]/delete.ts`: `c.req.valid("param")` discards its result. The OpenAPI hook already rejects a bad id with 422.
- Minor: `apps/core/src/routes/v1/projects/[id]/project-id.routes.test.ts`: DELETE asserts `project_deletion_removed` but not the conflict message.
- Skipped / disagreed: generated client also dropped reviews `offset` and added `500` on `listAdminTaskPaymentClaimsRequiringReview`. Both match current Core routes, so they stay.
- Skipped / disagreed: `PROJECT_HAS_CALENDAR_HISTORY` is unused by this route. The error-kind contract says never remove an existing value.

### Assessment
Delete is gone from the project menu, `DELETE /v1/projects/{id}` returns 409 `project_deletion_removed` without deleting the row, and the brand-remove cancel label is restored.

mrosberghaus:
Follow-up on https://github.com/masumi-network/sokosumi/pull/5083#issuecomment-5774910952

The brand-remove Cancel label stays as fixed in `9aed5ef6f`.

`c.req.valid("param")` is removed in `6a990a0d6`. `defaultValidationHook` already returns 422 for a bad id, and the handler never used the parsed id. The DELETE suite now covers `DELETE /projects/not-a-uuid` → 422. Core route tests: 21 passed.

The conflict message stays unasserted. `CORE_API_ERROR_KINDS` says clients match `kind`, and the message may be reworded. The test already checks `project_deletion_removed`.

`PROJECT_HAS_CALENDAR_HISTORY` stays. The generated `offset` drop and the payment-claim `500` match the current Core routes, so those stay too.

PR #5108, opened 2026-09-23
feat(chat): badge the room header's Threads icon with its unread count (SOK-1160)

mrosberghaus:
Closes [SOK-1160](https://linear.app/masumi/issue/SOK-1160).

## What a reader sees

The room header's Threads icon now carries a primary badge holding the unread thread count on its top-right corner, as section 3 of the [thread-unread mock](https://thread-unread-mock.vercel.app/) asks. Every reader sees it, including readers who switched numeric counts off. Past 99 it reads `99+`, and the spoken name still says "More than 99 unread threads".

This replaces both old states:
- the 8px dot, plus a heavier glyph stroke, for readers with counts off
- the bare number beside the glyph, which widened the button, for readers with counts on

The button is now `size-8` in every state. The glyph keeps its normal stroke, as the mock's does.

## Decisions

- **One badge for the whole header.** The badge moved into a new `CornerCountBadge` component, taken from the notification bell. The bell and the Threads icon both use it now, so they match in size, position, weight and colour (`bg-primary-solid` / `text-primary-solid-foreground`). Its geometry is the sidebar `MentionCountPill`'s: 16px tall, with 10px semibold tabular digits. It deliberately does not copy the mock's own badge, which is 11px tall and bold: in Inter that looked cramped and matched no other count in the product. No new colour tokens.
- **The count switch now governs the sidebar only.** A new paragraph in ADR-0038 records this. `showUnreadCount` no longer gated anything here, so it is removed from `UnreadThreadsPanel`, `RoomHeaderChrome` and `rooms-client`, together with the ten `useShowRoomUnreadCount` test mocks that only existed for it. The sidebar row still reads the switch.
- **The mention pill keeps its tint.** It says *you were named*, with an `@`, capped at 9. The Threads badge says *how much*, as a solid fill capped at 99+. The two share geometry but not colour.
- **Correction to the issue:** it calls counts-off "the reader's default", but ADR-0038 made counts on by default. The rewritten comment describes the switch as something readers turn off.

## Computed styles (Inter, real component markup, compiled `globals.css`)

| | Threads badge | Bell badge | Sidebar `@` pill |
| --- | --- | --- | --- |
| Height | 16px | 16px | 16px |
| Width at `2`/`12` / `99+` | 18 / 23.4px | 18px | — |
| Font | 10px semibold, tabular | same | same |
| Space above / below digits | 4.6 / 4.0px | 4.6 / 4.0px | 4.6 / 3.9px |
| Offset from the button (top / right) | −2 / +2px | −2 / +2px | — |
| Fill (light / dark) | `rgb(43,92,120)` / `rgb(51,109,143)` | same | tint |

The ghost `Button` has a 1px border that the bell's plain `div` doesn't, so the Threads badge needs a 1px extra offset (`-top-0.75 -right-0.75`) to land where the bell's badge does.

## Verification

- `pnpm exec vitest run` on the chat components, the header and the `color-tokens` / `whole-pixels` guards: 92 files, 898 tests pass
- `unread-threads-panel.test.tsx` checks:
  - the badge carries the count
  - no badge at zero
  - `99+` with "More than 99" past the cap
  - exact `99` at the cap
  - the badge stays while the panel is open
  - no `unread-threads-dot` in any state
- The bell's existing badge tests pass unchanged
- Pre-commit hook passes: `pnpm check` and `pnpm typecheck`


mrosberghaus:
## Grok review

Picked up a review request from @mrosberghaus.

### Fixed
- Important: the room header row's `overflow-hidden` clipped the threads badge. The badge hangs 4px past the icon button (1px border, 3px offset, 2px ring), so the top of the count and its ring were cut off. The row now has `py-1.5` and the actions group `pe-1.5`, which also keeps the last control's focus ring inside the clip. (`a60fe31e9` — `apps/web/src/app/(app)/chat/components/room-header-chrome.tsx`)

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/unread-threads-panel.tsx`: the component comment splits the sentence "looking at a list" across two lines.
- Minor: tests never mount the badge inside the clipping header, so they still would not catch this clip.
- Skipped: Apple `RoomToolsModifier` still shows a dot when counts are off and an inline number when they are on. This PR's requirements are the web mock and the ADR-0038 note.
- Skipped: copying the mock's 11px bold pill. The written decision is the bell's 16px semibold badge, and that is what shipped.

### Assessment
Web now badges the threads icon for every reader, capped at 99+, and the header no longer clips that badge. Apple still uses the old mark.

mrosberghaus:
Follow-up to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5108#issuecomment-5790674483).

**Header clip (a60fe31e9):** checked in a real browser and kept as is. I rendered the Threads trigger inside the row's exact classes, before and after the change:
- **Before:** the row was 32px tall, and the `99+` badge with its ring was clipped by 4px at the top and at the right.
- **After:** the row is 44px tall, and the badge clears the clip by 2px on both sides.

Both hosts are `h-16`, the desktop `ROOM_SHELL_DESKTOP_HEADER_CLASSNAME` and the mobile app-header row that holds `data-app-header-room-slot`, so the extra height fits inside them and the header's height does not change. The action group moves 4px left (`pe-0.5` → `pe-1.5`), which is what the badge's overhang and the focus ring need.

**Split comment sentence:** fixed in dd0aa9485.

**No test for the clip:** I left this out on purpose. jsdom does no layout, so a unit test cannot tell whether an element is clipped. The only thing it could check is that `py-1.5` is present, and that just repeats the source. The browser check above is the evidence instead.

**Apple parity:** out of scope here, as you say. SOK-1160 covers the web control only. I'll raise the Apple `RoomToolsModifier` separately.

**11px mock pill:** the geometry shared with the bell is deliberate. The mock's 11px bold pill looked cramped in Inter and matched no other count in the product.

PR #5116, opened 2026-09-23
feat(chat): show repliers' faces and last reply age on the thread reply bar (SOK-1156)

mrosberghaus:
Closes [SOK-1156](https://linear.app/masumi/issue/SOK-1156/thread-reply-bar-shows-the-repliers-faces-and-the-last-replys-age). Follow-up to #5000.

## What changes for users

The reply bar under a thread parent now shows who replied and how recently, as in the approved [mock](https://thread-unread-mock.vercel.app/) (section 1):

- **Plain:** faces, `3 replies · 2d ago`
- **Unread:** faces, `2 new replies · 4m ago`, on the tinted bar from #5000

Up to three overlapping faces of the people who replied, in the order they joined. The first face is whoever started the thread. The parent author appears only where they replied, because the row already shows them. This order was changed in review; the issue originally asked for newest replier first. The age is muted and updates every minute.

## How

**Core**
- `ChatRoomMessage.threadRepliers` (optional): up to three distinct reply senders, in the order they first replied, in the existing `ChatRoomMessageSender` shape.
- `chatRoomMessageInclude.replies` now takes the newest 12 replies with their sender relations, so every route that returns a message carries the faces, including single-message and realtime.
- The sender closure in `mapChatRoomMessage` is lifted into `mapChatRoomMessageSender`, so a replier is mapped exactly like a sender.
- The 12-reply scan is an approximation, documented on the field. In a thread longer than 12 replies, someone whose only replies are older than the scan is left out, and order is by first reply within the scan.

**Web**
- New `ThreadReplyBar` component. The button's accessible name is still the count; the age is its `aria-describedby`; the faces are `aria-hidden`.
- `SenderFace` is extracted from the message row and shared by the row avatar and the bar faces.
- `senderProfile(sender)` in `room-helpers.ts` now backs `messageSender`.
- The Ably full-event schema and `hydrateChatRoomMessageFromRealtime` now carry `threadRepliers`. Without this, a live update to the parent (an edit, or a reply deleted) cleared the faces.
- Web Core client regenerated.

## Faces

- At 16px each face shows one initial, grey (`bg-muted text-muted-foreground`), like the read-receipt faces. With two initials, the overlap hid the second letter.
- The ring is a 1px hairline in the page colour. It keeps overlapping faces apart without reading as a halo on the tinted bar. The spec asks for the bar's own colour, but the tint token is alpha-baked, so a ring in it wouldn't separate the faces.
- After you reply, the local preview appends your face after the earlier repliers, unless you are already shown (`applyReplyToParentThreadPreview`).

## Also fixed (already on main)

- The unread bar's left rule and rounded corner were cut off. The bar kept the plain state's `-mx-1`, which pushed its edge 4px into the message content column's `overflow-x-clip`. The unread state now uses `mx-0`; measured, the bar sits at 0px from the column edge (−4px on main).

## Out of scope (per the issue)

- The sidebar's inset thread rows, the thread list panel, and Apple.

## Verification

- `pnpm check` and `pnpm typecheck` pass (commit hook).
- Core: all tests pass. `messages/get.test.ts` covers three repliers in the order they first replied, the cap at three, a repeat sender counted once, and no replies giving an empty list.
- Web:
  - `room-message-row-thread-bar.test.tsx`: one face per replier, accessible name unchanged, age rendered as `4m ago`.
  - `hydrate-chat-room-message.test.ts`: a regression test for the dropped field.
- Full `pnpm test`: one web failure outside this change, a 5 s timeout in `admin-soko-bot-actions.test.tsx` that passes on its own.
- Checked visually with a static mock: real `ChatMessageRow` markup from vitest plus the compiled `globals.css`, before vs after, light and dark. Screenshot to follow.


mrosberghaus:
## Grok review

### Fixed
- Important: after you reply, the bar updated the count and age but kept the previous faces, so "just now" could point at the wrong people. The local preview now puts that sender first, deduped, capped at three (`144175bc1` — `apps/web/src/app/(app)/chat/utils/parent-thread-preview.ts`).

### Remaining
- Minor: `apps/core/src/routes/v1/chats/rooms/helpers.ts`: faces come from the newest 12 replies, so one recent sender can hide earlier people. Documented approximation.
- Minor: `apps/web/src/app/(app)/chat/components/room-message-row.tsx`: the face ring uses the page background, not the bar colour. Documented; the tint token is alpha-baked.
- Minor: other open clients still keep the old faces, count, and age until a refetch when someone else replies. The reply count already worked that way. Core does not republish the parent on create.
- Skipped / disagreed: Apple OpenAPI snapshot was not refreshed. The thread-bar UI is out of scope, and the extra field does not break decode.

### Assessment
Loaded transcripts show up to three faces and the last reply's age, and the sender's own bar now includes their face without waiting for a refetch.

mrosberghaus:
Checked each remaining item against the branch (head `ec17509`):

- **Faces from the newest 12 replies:** no change. This is the approximation the spec chose (SOK-1156: "The scan is an approximation and says so"). It's documented on `THREAD_REPLIER_SCAN` and on the OpenAPI field.
- **Face ring in the page background:** no change to the colour. Since this review, `9b7ee0d` made the ring a 1px hairline (it was a 2px halo on the tint) and made the initials a single grey letter. Mock in the PR description.
- **Other open clients keep stale faces, count and age:** confirmed. `messages/post.ts` publishes only the reply (`publishChatRoomMessageRealtime(message, "create")`), and the web has no handler that bumps a parent for someone else's reply. The count already behaved this way before this PR, so this isn't a regression. The fix is in Core (republish the parent on reply create) and is out of scope here.
- **Apple OpenAPI snapshot not refreshed:** agreed. `threadUnreadReplyCount` from #5000 isn't in `apps/apple/.../openapi.json` either, so additive web-only fields are already handled this way. The field is optional, and the snapshot doesn't set `additionalProperties: false`, so Apple decoding ignores it.

Since this review, the branch also has `ec17509`, which fixes the unread bar's clipped left edge. That bug was already on main: the bar's `-mx-1` pushed it 4px into the content column's `overflow-x-clip`.

mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/core/src/routes/v1/chats/rooms/helpers.ts`: faces come from the newest 12 replies, so a thread longer than that can omit whoever actually replied first. Documented approximation.
- Minor: `apps/web/src/app/(app)/chat/components/room-message-row.tsx`: the face ring is the page background, not the bar colour. Documented; the tint token is alpha-baked.
- Minor: `apps/web/src/app/(app)/chat/utils/parent-thread-preview.ts`: a local reply delete drops the count but leaves the faces until the parent update (or the next refetch) arrives. Delete already republishes the parent.
- Skipped / disagreed: do not put the latest replier first again. `144175bc1` did that; `5111cd825` changed the product to join order, and `applyReplyToParentThreadPreview` now appends. That matches the PR body, and the tests lock it.
- Skipped / disagreed: other open clients still keep the old faces, count, and age until a refetch when someone else replies. Core publishes the reply, not the parent. The count already worked that way; republishing the parent is out of scope.
- Skipped / disagreed: Apple OpenAPI snapshot was not refreshed. The field is optional, the snapshot does not set `additionalProperties: false`, and the thread-bar UI is out of scope.

### Assessment
Loaded transcripts show up to three repliers in the order they joined, plus the last reply's age, and your own reply appends your face without waiting for a refetch.

mrosberghaus:
Checked the new item (head `17686f1`):

- **Local reply delete leaves the faces until the parent update:** no change. Removing the face locally would be wrong whenever the deleter has other replies in the thread, and the client can't tell. The server update corrects it right away:
  - `delete.ts:216` republishes the parent (`publishChatRoomMessageRealtimeById(parentMessageId, "update")`).
  - That payload goes through `mapChatRoomMessage`, which carries the recomputed `threadRepliers`.
  - `hydrateChatRoomMessageFromRealtime` passes them through (fixed earlier in this PR, with a regression test).

The other items are unchanged from my previous reply.

Since this review, `17686f1` narrowed the faces to repliers only. The parent author appears only where they replied, since the row already shows them above the bar.

PR #5118, opened 2026-09-23
fix(apple): replace the edit Save and Cancel buttons with compact controls

mrosberghaus:
Parity row **18b**: the Apple edit composer loses its Save and Cancel button row. Compact ✕/✓ icon controls at the trailing edge of the field replace it. This PR also marks 18a Done.

The user made two decisions on 2026-09-23:
1. The button row goes, as on web. Web's inline `MessageEditComposer` has had no buttons since #3664.
2. On review of this PR: a native Mac app keeps a pointer path. Compact always-visible controls replace the row, like editing a message in Messages. This is a **recorded deliberate deviation from web**, like the kept toolbar New chat.

The PR has two commits: `c08c3ad63` removes the row and aligns the count, hint and dimming with web. `fa0fbfdff` adds the controls. A merge of the branch, which the user had updated with main, sits on top; it touches nothing in `apps/apple`.

## User-facing impact
- **Controls:** editing a message in a room, thread parent or reply shows no button row under the field. At the right of the first line sit a grey ✕ (`xmark.circle.fill`, "Cancel (Escape)") and an accent ✓ (`checkmark.circle.fill`, "Save (Return)"). VoiceOver labels are "Cancel" and "Save".
- **What they do:** ✓ runs the same commit as Return. An unchanged or empty draft cancels, and an over-limit draft keeps editing. ✓ is disabled and grey over the limit. Both controls are disabled and dimmed while a save is in flight. ✕ cancels. Return, Escape, Shift-Return and blur behave as in 18a.
- **Count and hint:** from 9,500 characters (trimmed, UTF-16), one line under the field shows the count on the right. Over 10,000 the count turns red and **"Too long to send as text"** appears on the left. This follows web's copy and layout, and replaces "Message exceeds the 10,000-character limit."
- **Plain digits:** the count prints `10001/10000`, as web's ICU argument does. Apple used the locale's grouping (`10.001/10.000`). The message composer shares the count view; nothing else in it changes.
- **Dimming:** while a save is in flight the field dims to 50 % and stays locked, as web's `opacity-50` + `disabled`.

## Web audit (`9ebafac40`, `room-message-row.tsx`)
1. **Controls:** web's edit field has no buttons, icons or visible hint (:1843-1876). `App.Channels.Edit` has no save or cancel string.
2. **Count:** shown from `trimmed.length >= 9500`, reading `{count}/{max}`. It is muted, red over the limit, and the hint sits on the same line (:1838-1841, :1878-1903).
3. **Return over the limit:** web toasts and keeps editing (:1817-1836).
4. **Saving:** `opacity-50` and `disabled`, with no spinner (:1848, :1862). A failed save toasts and keeps the draft (`rooms-client.tsx` :2411).
5. **Message composer:** unchanged. The thread panel passes the same props.
6. **Pointer path:** web has none for a changed edit. Apple keeps one through the ✕/✓ controls, per the user's decision. This replaces the accessibility question the first commit raised.

Kept Apple (recorded in PARITY):
- The ✕/✓ controls.
- No toast on Return over the limit.
- The formatting bar, Emoji and Mention controls in the edit composer.
- No live region on the hint or count.

## Changes
- `MessageEditControls.swift` (new): two native borderless icon `Button`s, framed to the field's first-line height so they centre on it.
  - `.help` and `.accessibilityLabel` on each.
  - Save goes `.tertiary` when disabled, because an explicit foreground style does not dim with `.disabled`.
- `ComposerTextInput.swift`:
  - Edit mode puts the controls beside the text view, through the existing `cancelEdit` seam.
  - `canSend` enables Send or, in edit mode, Save.
  - The row is gone.
- `MacComposerTextInput.swift`: `textInset` and `firstLineHeight`, the values the field's own height already uses.
- `MessageEditComposer.swift`:
  - `canSend: !content.isTooLong`.
  - The hint and count line under the field.
  - `.opacity(0.5)` while saving.
- `ComposerCharacterCount.swift` (new, first commit): the shared count view, with plain digits.
- Tests:
  - `MessageEditComposerFixture.swift` (new): the shared hosted fixture. It binds the workspace's own `MessageEditing`, as `MessageRowView` does, and uses a recording, never-answering `URLProtocol`. It also carries the pixel and click helpers.
  - `MessageEditComposerControlsTests` (new, 5 tests, 7 cases).
  - `MessageEditComposerChromeTests` (4 tests, 6 cases).
  - `MessageEditComposerReturnTests` now uses the shared fixture.
- `docs/images/message-editing.png`: the new render, light over dark.
- `PARITY.md`:
  - Row 18b is In review, with the refined decision recorded as a deviation.
  - 18a is Done.
  - The section 18 paragraph and the 18a deviation note are updated.
  - Checkpoint and work order: 24b is next.
  - The slice 18b section is rewritten.

- **Red first (controls):** against `c08c3ad63` all 5 control tests (7 cases) failed, with no control found beside the field. The refactored chrome and Return suites passed.
- **What the control tests check:**
  - **Clicks:** a click on ✓ over a changed draft sends exactly one `PATCH /v1/chats/rooms/room_1/messages/message_1`, and the edit survives any blur. A click on ✕ over a changed or unchanged draft cancels and sends nothing.
  - **Disabled states:** ✓ is weaker and inert over the limit. Both controls are weaker and inert while a save is held in flight.
  - **Layout:** both controls are centred within 2 pt of the first line, fit one line, and sit at least 8 pt from the border.
- **Composer suites:** `xcodebuild test … -only-testing:` the controls, chrome and Return suites plus `MacComposerTextInputTests` passed 73 tests, 101 cases.
- **Full run:** `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO` passed **1102/1102**: SokosumiTests 184, Chat 705, Auth 35, Realtime 48, Workspace 129, CoreAPI 1.
  - An earlier full run failed only the known load-flaky `TranscriptContentGrowthTests/growingMessagePreservesReadingPosition`, which passed when rerun alone.
- **Builds:** `xcodebuild … build` (macOS) succeeded, and `swift build --package-path Packages/SokosumiWorkspace --triple arm64-apple-ios17.0` completed.
- **Lint:** `mint run swiftformat --lint .` reports 0/401 files; `mint run swiftlint lint --strict` reports 0 violations.
- **Render, light and dark:** the controls are centred on the first text line and nothing is clipped. ✓ keeps 12 pt from the border, the same inset as the formatting toggle. Wrapped text never runs under the controls. Over the limit, ✓ is grey next to a normal ✕.
  - In the light render the accent is desaturated, because the test window is not key.
- **OCR:** the count and hint checks are OCR-only, so CI skips their text assertions.

## Not exercised in the running app
- Saving or cancelling a live edit with the keys and with ✕/✓ against Core.
- The controls in a key window with the real accent colour and Increase Contrast.
- Whether a real click moves focus (the tests prove only the outcome).
- VoiceOver and Switch Control on the controls.
- A real failed save or a live over-limit edit.
- The dimmed field during a slow save.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/SokosumiTests/Chat/Composer/MessageEditComposerControlsTests.swift`: the over-limit and in-flight click expects do not prove the buttons ignore the click. `commit()` already returns on `isSaving` or an over-limit draft, and `cancel()` returns while `isSaving`. A button that still fires would pass. The strength checks only show the pixels got weaker.
- Minor: `apps/apple/Sokosumi/Chat/Composer/MessageEditControls.swift`: ✓ calls `submit()` and ✕ calls `cancel()` directly, so they skip the suggestion list. Return accepts the highlighted row; Escape dismisses the list. Help says "Save (Return)" and "Cancel (Escape)", which is wrong while a list is open.
- Skipped / disagreed: an enabled ✓ on an unchanged or empty draft is the commit path (it cancels), same as Return. The old row disabled Save via `editing.canSave`; this matches the written decision.
- Skipped / disagreed: Save's `.disabled(false)` does not re-enable it while the composer is `.disabled` during a save. SwiftUI ANDs `isEnabled`, so both controls stay disabled. Checked, not a bug.
- Skipped / disagreed: no over-limit toast, the kept formatting bar, and no live region on the hint or count are recorded in `PARITY.md`. Not defects.

### Assessment
Ready to merge: Yes
Compact ✕/✓ controls use the same commit and cancel path as Return and Escape, the count and hint follow web, and nothing here should block merge.

mrosberghaus:
**1. Disabled-state tests don't prove the controls ignore the click:** fixed in daaf99aa4.

Three tests in `MessageEditComposerControlsTests` host `MessageEditControls` alone, with counting closures, and click by drawn pixels:
- Enabled: ✓ and ✕ each fire once.
- `canSave: false`: ✓ fires 0 times, ✕ still fires once.
- Inside `.disabled(true)` (the composer's saving state): both fire 0 times.

Checked against a regression: with `.disabled(!canSave)` removed from the view, the second test fails (`saves == 1`). The composer-level `saveIsDisabledOverTheLimit` still passed in that run, which confirms your point. The composer-level tests stay, because they prove the outcome.

Composer suites: 76 tests, 104 cases. Strict SwiftLint and SwiftFormat are clean.

**2. ✓/✕ skip the suggestion list:** not changed.
- **Consistent with the existing buttons.** The message composer's Send (`ComposerTextInput.swift`, `Button("Send", …) { _ = submit() }`) calls `submit()` directly while a list is open. So did the removed Save (`Button("Save") { _ = submit() }`). ✓ behaves the same way.
- **A click is an explicit save.** A pointer click on ✓ is not a Return keystroke, so it does not go through the list's key handling.
- **What happens to the list.** The list lives in the edit composer's own `MacComposerCommands`.
  - A successful save or a cancel removes `MessageEditComposer` (`MessageRowView.swift`:260), and the list goes with it.
  - A failed save, or ✓ on an over-limit draft, keeps the editor open. The list then stays open too, still attached and still usable by keyboard. It is not orphaned, but it is not closed either.
  - While a save is in flight, the list stays drawn inside the disabled, dimmed composer.
- **The tooltip.** `.help("Save (Return)")` names the key equivalent, as Mac tooltips do. It does not promise that Return saves while a completion list has the key; there Return accepts the row, as in the message composer.

**Skipped items:** no action. Your own checks agree:
- An enabled ✓ on an unchanged or empty draft follows the commit path.
- SwiftUI ANDs `isEnabled` while saving.
- The kept differences are recorded in PARITY.

PR #5119, opened 2026-09-23
feat(chat): add Threads and All unreads views across rooms (SOK-1159)

mrosberghaus:
Implements [SOK-1159](https://linear.app/masumi/issue/SOK-1159/global-threads-entry-above-the-channel-list-with-an-all): a global **Threads** entry above the channel list, and an **All unreads** filter beside it. Follows [SOK-1147](https://linear.app/masumi/issue/SOK-1147). Design: section 4 and 5 of the [mock](https://thread-unread-mock.vercel.app/).

## What a reader gets

- **Sidebar**: a Threads row above Pinned, counting unread **Threads** (not replies) across every room. It shows the `@` pill when a Thread names the reader, a muted number otherwise, and nothing at zero. Below it sits the All unreads toggle.
- **Collapsed rail**: both are permanent items, and Threads carries the existing edge pill (unread / mention).
- **Threads on the desktop gets the Projects flyout**: the row stays a link to `/chat/threads`, and the pointer (after a 200ms beat) or →/↓ opens the unread Threads beside it. Each names its room, newest unread reply first, in the Thread list's row style, with an **All threads** footer. The list is prefetched, so the panel never opens onto a spinner. Following a row opens the Thread in its room via `?message=<firstUnreadReplyId>`. The behaviour is shared with Projects through `useSidebarFlyout`, so the two rows cannot drift.
- **`/chat/threads`** lists every Thread the reader is part of, grouped like a room's Thread list: **Unread** (the flyout's list) then **Earlier**, newest reply first, each naming its room, reply count and age, opening at the newest reply. Where the row goes on click, and the only way in on the phone. The breadcrumb reads "Threads".
- **All unreads is a filter, not a page**: toggling it (the row takes the primary tint while on, since a mode is not the page you are on) keeps the sidebar's own sections (Pinned, Channels, External, Directs) but only the rooms a read would still change, bold or holding an unread Thread, with their Threads inset. The open room stays visible so it cannot vanish while being read. Empty sections and Archived drop out and pinned reordering pauses. Rooms read while the filter is on move to a dimmed **Just read** section, newest first, instead of vanishing; the section is tracked in the browser for that pass only and cleared when the filter goes off. At zero the caught-up state is a row on the sidebar's axes: a tinted ✓✓ in the icon slot, "All caught up", **Show all** where a row keeps its action, and "Nothing unread in N chats" beneath. While the filter is on, **Mark all as read** sits on its row and runs each room's existing reads: room read, plus Mark all threads where the room has Thread unread.
- **Phone**: both entries sit above the room list in the Chats tab, where the filter works the same way. The Chats tab stays active on `/chat/threads`.

## How

- **Core**: `GET /v1/chats/threads/unread` returns unread Threads across the rooms the sidebar lists, minus rooms the reader muted, cursor-paged. It is built on the same `sqlUnreadThreadReplies` fragment as the room counts; the per-room list and the cross-room list now share one grouping (`sqlUnreadThreadsByParent`).
- **Core**: `GET /v1/chats/threads/earlier` is the Earlier group: the reader's Participant Threads (ADR-0013) less every Thread the unread fragment finds, so a Thread is in exactly one group; same room scope and paging as the unread read.
- **Core**: the room summary gains `unreadThreadMentionCount`, which counts mentions across **all** of a room's unread Threads, including those past the three-row cap. A mention in a room's 4th Thread still lights the Threads pill.
- **Web**: the pill and the All unreads filter read the room list the sidebar already holds; no new read. `roomUnreadReads` decides both what the filter keeps and what Mark all as read runs. The Threads view reads through a background route handler (`/api/chat/threads/unread`), not a server action, and reads again when a sidebar poll moves the rooms' Thread counts.
- **Web**: `useLiveChatRooms` is extracted from `useChatTabUnreadPresence`, so the Threads page gets live rooms without the sidebar mounted (the phone). `UnreadThreadsList` and `useUnreadThreadsQuery` are shared by the flyout and the page.

## Deviations from the ticket

- **Route shape**: `GET /chats/threads/unread`, not `?unread=true`; the endpoint serves only unread.
- **Room mute** is gated in the route and client (as everywhere else in the sidebar), not in the SQL fragment, which only knows Thread mute.
- **Pill colour**: the ticket says "primary pill" but also "`MentionCountPill` for mentions and the muted number otherwise". This follows the latter, the one-number rule.
- The rail mention pill is 8px, not the mock's 9px. That is the existing `RailAttentionPill`, unchanged here.
- **All unreads** was specified as a page. It is a sidebar filter instead, at the author's request.
- **Phone**: while the Threads view is open with no sidebar mounted, it runs its own room-list read beside the Chats tab dot's.

## Verification

- `pnpm test`: Core 563 files pass, web 846 files pass; `pnpm typecheck` and `pnpm check` are clean.
- Core route contract with mocked Prisma, with one narrow gating tripwire. Thin tests cover the pill, the flyout (keyboard open, prefetch, All threads footer), the tinted filter row, the list rows, the empty state, the filter (kept rooms, dropped sections, caught-up, Just read across a pass, off again), Mark all as read and the mobile slot.
- The new SQL (cross-room ranking, cursor paging, total, muted-Thread exclusion, per-room mention sum) was run once against a throwaway local Postgres with migrations applied. That check is not committed, per the no-DB-tests rule.
- The real component markup (sidebar entry, the Threads list at flyout width), rendered with the app's compiled CSS in light and dark, matched the mock. The filter's visual state has not been rendered yet.
- **Not yet checked** in a running app with a signed-in account. Please check on the preview: hover the Threads row to open its flyout, read a Thread, and confirm it drains; then toggle All unreads and use Mark all as read.


mrosberghaus:
## Grok review

### Fixed
- Important: the Threads view stored the server page as fresh query data under the live room fingerprint. The app query client keeps that data fresh for 60s, so a sidebar snapshot that had already moved kept the stale list until some later count change (`9c2464974` — `apps/web/src/app/(app)/chat/threads/components/unread-threads-view.tsx`)

### Remaining
- Minor: `apps/core/src/routes/v1/chats/rooms/room-unread.ts`: an empty cursor page reports `total: 0`, because the total is read off the first row. A cursor thread that was just read returns no rows.
- Minor: `apps/web/src/components/chat/chat-unread-nav-rows.tsx`: the Mark all control is absolute on the All unreads row and the row does not reserve its width, so a longer label can run under it.
- Minor: `apps/web/src/components/chat/organization-chat-list.client.tsx`: pending invitations still open the external section while the unread filter is on.
- Minor: `apps/web/src/app/(app)/chat/threads/components/unread-threads-view.tsx`: an empty live room list is treated as not loaded, so a settled empty read keeps the server rooms.
- Skipped / disagreed: Mark all does not Look a muted Thread or a non-participant mention — `markAllChatRoomThreadsRead` already excludes those (ADR-0030), and this PR calls that read on purpose.

### Assessment
Threads and the All unreads sidebar filter match the ticket; the stale first page was the bug that would stick.

mrosberghaus:
Replying to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5119#issuecomment-5792681374). Three of the four remaining items are fixed in a32460a4b. The fourth stays as intended; the reason is below.

**The stale-seed fix (9c2464974)**
Kept. The Threads list has since moved into the shared `UnreadThreadsList` (the desktop popover and the `/chat/threads` page). The fix is carried there: the list takes the server page together with the rooms it was rendered for, and only that exact pair seeds `initialData`. Your regression test passes against the moved code.

**1. An empty cursor page reported `total: 0`.** Fixed.
- The count is now one row, joined to the page with `LEFT JOIN page ON true`. A stale cursor returns no threads but the real total.
- Checked against a local Postgres with migrations applied: normal page, next page, stale cursor (total kept), and no unread (total 0).
- New unit test: `states the total even when the page after the cursor is empty`.

**2. The Mark all control could overlap the label.** Fixed. While the button shows, the All unreads row reserves its 28px with a spacer. Covered in `chat-unread-nav-rows.test.tsx`.

**3. Pending invitations stay under the unread filter.** Intentional, so left as is.
- An invitation is addressed to the reader and waits on them. `resolveSectionAttention` already marks a closed section with one as a mention.
- Hiding it under "unread only" would hide the one thing that needs the reader.
- The intent is now a comment in `organization-chat-list.client.tsx`.

**4. An empty live room list was treated as not loaded.** Fixed.
- `useLiveChatRooms` now returns `null` until its first read or a session snapshot. An empty list is taken as an answer, so the cached server rooms no longer stand in for it.
- The Chats tab dot uses `?? []`.
- New test: `says the reader is caught up when the live read finds no room`.

**Skipped: Mark all leaves muted Threads unlooked.** Agreed. `markAllChatRoomThreadsRead` leaves those out on purpose (ADR-0030).

**Checks:** `pnpm test` passes (web 846 files, Core 565); `pnpm check` and `pnpm typecheck` are clean.

mrosberghaus:
## Grok review

### Fixed
- Important: a failed older page in the room Thread panel shared the list error, so a retry that succeeded left the row failed and scroll-loading never resumed. Mark all used that same error and looked like a failed page (`778f931fd` — `apps/web/src/app/(app)/chat/components/thread-list-panel.tsx`)
- Important: after a live room read, a cleared roster fell back to the Threads page's server rooms, so a workspace switch kept the previous workspace's threads on screen (`778f931fd` — `apps/web/src/app/(app)/chat/threads/components/unread-threads-view.tsx`)

### Remaining
- Minor: `apps/web/src/components/chat/unread-threads-list.tsx`: the refetch key is room id plus `threadUnreadCount`. A net-zero change (one thread read, another reply in the same room) does not refetch.
- Minor: `apps/web/src/components/chat/earlier-threads-list.tsx`: Earlier uses that same key, so a reply that does not change unread counts leaves its order stale.
- Minor: `apps/web/src/components/chat/unread-threads-list.tsx`: a page whose threads are all filtered out is treated as caught up even when a next page exists. Core already drops muted rooms, so this is the placeholder window.
- Skipped / disagreed: pending invitations stay under All unreads. An invitation is addressed to the reader, and a closed section already marks it as a mention. Left as the author stated.
- Skipped / disagreed: the earlier stale-seed fix stays. `useUnreadThreadsQuery` still seeds `initialData` only when the live fingerprint matches the server page.

### Assessment
Threads, the flyout, and the All unreads filter match the ticket; paging recovers after a failed page, and a workspace switch no longer keeps the previous roster on the Threads page.

mrosberghaus:
Replying to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5119#issuecomment-5794766995). The three remaining minors are fixed in b6a171e96.

**Fixes already on the branch (778f931fd)**
Kept as pushed: the room panel's own failed-page state, and the workspace-switch roster.

**1. A net-zero change did not refetch the unread list.** Fixed.
- `unreadThreadsFingerprint` now also carries each room's `unreadThreadCount` and the Threads it lists (`parentMessageId/unreadReplyCount`).
- So a read in one Thread plus a new reply in another moves the key, even when the room's reply total stays level.
- Room activity with no unread change still does not refetch, which keeps the always-mounted flyout prefetch quiet.
- Test: `unread-threads-fingerprint.test.ts`.

**2. Earlier kept a stale order.** Fixed.
- Earlier now keys on `earlierThreadsFingerprint`: the unread key plus each room's `updatedAt`, which every new message moves.
- A reply that changes no count, such as the reader's own, now reorders the group.
- Only the Threads page mounts Earlier, so the extra reads stay there.
- Tested in the same file.

**3. A page whose rows all dropped out read as caught up.** Fixed.
- "Caught up" from Core's answer now also requires `!hasNextPage`.
- A page whose rows were all filtered out keeps its load-more boundary, which pages on.
- Test: `keeps paging rather than saying caught up while another page follows`.

**Skipped items:** agreed on both, the invitations under the filter and the seeding.

**Checks:** `pnpm test` passes (web 851 files, Core 566); `pnpm check` and `pnpm typecheck` are clean.

PR #5122, opened 2026-09-23
fix(apple): draw the composer emoji icon as an outline in dark mode

mrosberghaus:
## Summary

In dark mode, the composer's **Emoji & Symbols** button drew as a filled light-grey disc with a dark face. Its neighbours (B, I, U, link, lists, "Aa") are outlines. It now draws as an outline in the same grey. Light mode is unchanged: the light render is byte-identical before and after.

This is outside the parity rows, so `PARITY.md` is not touched.

## Root cause

The cause is the symbol. On macOS 27, `face.smiling` picks its glyph by colour scheme:

- In light it draws a ring with eyes and a mouth.
- In dark it draws a filled disc with cut-out features, which is what `face.smiling.inverse` draws in light. `face.smiling.inverse` in dark draws the outline.

Rendering the symbol on its own in scratch windows shows this:

- The disc appears in dark with every styling option tried: `.secondary`, `.primary`, a plain `Color.gray`, `.monochrome`, `.hierarchical`, `.symbolVariant(.none)` and `Label`. It also appears with an AppKit `NSImageView` and `NSImage(systemSymbolName:)`.
- Palette rendering with red, green and blue shows the swap. The light glyph is features plus a ring. The dark glyph is features plus a filled disc on the tertiary layer.
- SwiftUI's `colorScheme` decides which glyph is used, not the window's `NSAppearance`. A dark environment in an aqua window draws the disc. A light environment in a darkAqua window draws the outline.
- The disc appears in non-key windows. No window became key, either in the scratch program or in the test host, so key-window rendering is unverified.
- Other symbols (`circle`, `person.circle`, `heart`, `at`, `textformat`, `list.number`, and so on) do not change with the scheme. `smiley` swaps the same way `face.smiling` does.

`ComposerToolbarButton` itself applies no variant, rendering mode or fill, and no container sets any either.

## Fix

The fix is in `ComposerToolbarButton` only:

- The button resolves the glyph in the light scheme (`.environment(\.colorScheme, .light)` on the `Image`).
- It colours the glyph with `Color.primary` or `Color.secondary`, resolved in the real environment.

Compared with the old hierarchical `.primary` / `.secondary` style, the resolved colours are pixel-identical in light and dark, and when disabled (checked on `textformat`). Every other toolbar symbol therefore draws exactly as before. `ComposerTextInput.swift`, `MessageEditComposer.swift` and `ComposerFormatToolbar.swift` are not touched, so this PR does not conflict with #5118 or `claude/apple-composer-code-block-icon`.

**Not fixed here:** the message row's **React** action (`MessageRowView`, `face.smiling`, callout, `.secondary`) has the same cause and also draws as a disc in dark. It is fixed in stacked #5124.

## Test

New `NativeWindowTests/ComposerToolbarButtonTests/emojiIconIsAnOutlineInTheNeighboursGrey(dark:)`, run in light and dark. It renders the emoji button, the "Aa" toggle, and a reference icon in one window. The reference is `textformat` drawn the old way: plain `.foregroundStyle(.secondary)`, with no scheme override. The test checks two things:

- **Shape:** the face's ink covers less than half of its bounding box. The disc covers 78 %, the outline 35–38 %.
- **Colour:** the strongest ink contrast of both toolbar buttons matches the reference's within 0.08. The reference does not go through the new colour path, so a colour miss shared by both buttons still fails. To confirm this, I temporarily resolved the button's grey to `.primary`: both cases failed, with a difference of 0.37 in light and 0.22 in dark.

The test uses a non-key window only. No test window becomes key in the test host, so key-window rendering stays unverified.

Before the fix both dark cases fail (78 %) and both light cases pass. After the fix both pass. The renders are recorded as `composer-toolbar-{light,dark}.png`.

In the recorded renders the dark emoji goes from a filled disc to an outline, and the light render does not change.

## Verification

- Composer suites plus the new test: 111/111 passed. After the independent-reference change and merging `main`: 123/123 passed.
- Full workspace run (`xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`): 1093/1094 passed. The one failure was `HistoryGapRowTests.aFailedGapStaysOnItsRowWithoutAnAlertAndTryAgainReloads`, a known load-flaky OCR test; rerun alone it passed 3/3.
- `mint run swiftlint lint --strict`: 0 violations. `mint run swiftformat --lint .`: 0/397 files need formatting (0/402 after merging `main`).
- macOS Debug build: succeeded.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/SokosumiTests/Chat/Composer/ComposerToolbarButtonTests.swift`: the contrast check compares the emoji button with the `textformat` button, and both use the new resolved-color path, so a shared color miss still passes. The shape check is the real lock (disc ~78% vs outline ~37%). A separate render of these modifiers matched the old `.secondary` path for `textformat` (same ink count and luminance) and drew the dark emoji as an outline at the same max luminance.
- Minor: `key: true` never asserts `isKeyWindow`, so the key and non-key cases can be the same render. The PR already treats key state as irrelevant.
- Skipped / disagreed: React `face.smiling` in `MessageRowView` still fills in dark mode. This change does not touch that control, and the PR leaves it for a separate change.
- Skipped / disagreed: `@Environment(\.self)` on `ComposerToolbarButton` — `Color.resolve(in:)` needs the full environment so the light glyph keeps the real scheme’s grey. Not a behavior bug.

### Assessment
Ready to merge: Yes
Dark mode draws the composer emoji icon as an outline in the same grey as the other toolbar symbols, without shifting those symbols off the previous `.secondary` pixels.

mrosberghaus:
Addressed in 1cef8bca9 (merged `main` first, head was 6e9947070).

1. **Colour check has no independent reference.** Fixed in 1cef8bca9. The window now also draws `Image(systemName: "textformat").font(.body).imageScale(.medium).foregroundStyle(.secondary)` in a 28 pt cell. This is the old path, with no `resolve(in:)` and no scheme override. The emoji and "Aa" buttons must each match its ink contrast within 0.08. To check that it catches a shared miss, I temporarily resolved the button to `.primary`: both cases failed, emoji and "Aa" alike (Δ 0.37 in light, 0.22 in dark). Then I restored it.
2. **`key: true` never asserts `isKeyWindow`.** Fixed in 1cef8bca9. The `key` argument is removed and the test runs a non-key window only. No test window becomes key in the test host, so a key case would render the same window. Key-window rendering stays unverified; the test comment and the PR body now say so.

Skipped: the React `face.smiling` in `MessageRowView` is fixed in stacked #5124. `@Environment(\.self)` needs no action.

Composer suites 123/123. `swiftlint --strict` and `swiftformat --lint` are clean.

PR #5123, opened 2026-09-23
fix(web): clip sidebar nav labels on collapse instead of re-ellipsizing

mrosberghaus:
## What changes for users

When the sidebar collapses or expands, the **New Task** pill, the nav rows (Search, Agents, Task Manager, …) and the **Soko Bots** row no longer re-truncate their name on every frame (`New Task` → `New T…` → `Ne…`). The name keeps its full layout and the narrowing row clips it at its edge, so the row slides over the text instead of the text squeezing into the row.

The New Task pill also stops sliding: it used to sit flush with 8px padding when expanded and take a 4px margin with 4px padding on the rail, so its left edge jumped 4px on every toggle. It now keeps the rail's 4px inset and padding at every width, so collapsing only narrows it.

## How

- New `SIDEBAR_ROW_FIXED_LABEL_CLASS` (`whitespace-nowrap text-clip!`) in `sidebar-classes.ts`, used by `menu-items.tsx` and `personal-assistant-nav.client.tsx` instead of `truncate`.
- `!` because `SidebarMenuButton`'s own `[&>span:last-child]:truncate` would put the ellipsis back.
- New Task pill: `ml-1 w-[calc(100%-0.5rem)] pl-1` at every width, matching the rail square's `ml-1` / `pl-1!`, so only `width` animates.
- Only for rows whose names always fit at full width. Chat rooms, projects and other variable-length rows keep their ellipsis.

## Verification

- Rendered the real `MenuItems` markup (before and after) with the app's compiled `globals.css`, slowed the collapse 20× in headless Chromium. Mid-collapse, before showed `New …` and after showed `New Ta` clipped by the pill.
- Measured the pill in headless Chromium (real markup, compiled CSS) expanded, mid-collapse and collapsed: after the change its left edge stays at 12px and its padding at 4px in every frame; before, the edge moved 8 → 10 → 12px and the padding 8 → 5.9 → 4px.
- `pnpm exec vitest run "src/app/(app)/components/sidebar"`: 124 passed, including new assertions for the Soko Bots label clip and the New Task pill's constant inset.
- Pre-commit hook: `pnpm check && pnpm typecheck` green.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/components/sidebar/components/projects-menu-item.tsx` — the Projects row label is a fixed short string (`Projects` / `Projekte` / `Proyectos`) and still uses `truncate`, so that one nav row still re-ellipsizes on collapse. Flyout project names should keep the ellipsis. Left as-is because this PR explicitly keeps ellipsis on project rows.
- Skipped / disagreed: expanded New Task pill keeps `padding-inline` from `px-2` on the right while `pl-1` sets the left to 4px. Outer inset is still 4px each side (`ml-1` + `w-[calc(100%-0.5rem)]`), and these labels never reach the extra right padding. Icon axis stays put. Not a behavior bug.
- Skipped / disagreed: the new tests assert class tokens, not computed `text-overflow`. Checked Tailwind 4.3.3: `.text-clip\!` is `text-overflow: clip !important` and beats `[&>span:last-child]:truncate`. `twMerge` drops `w-full` for `w-[calc(100%-0.5rem)]`. `.pl-1` is later than `.px-2`, so `padding-left` wins.

### Assessment
Ready to merge: Yes
Clip-instead-of-ellipsis and the constant 4px New Task inset match the collapse behavior this PR describes. `menu-items.test.tsx` and `personal-assistant-nav.client.test.tsx` passed (27 tests).

mrosberghaus:
Re the [Grok review](https://github.com/masumi-network/sokosumi/pull/5123#issuecomment-5793229943):

- **Projects row still re-ellipsizes:** fixed in 5f053d77c. The row's own `Projects` label now uses `SIDEBAR_ROW_FIXED_LABEL_CLASS` like the other nav rows. Project names in the flyout keep `truncate`. `projects-menu-item.test.tsx` now asserts `text-clip!` and no `truncate` on that label.
- **Right padding of the expanded pill:** agreed, no change. The 4px outer inset is symmetric, and the label never reaches the extra right padding.
- **Class-token tests instead of computed style:** agreed, no change. The computed geometry was measured in headless Chromium before pushing (see PR description), and the tokens are what can regress.

Sidebar tests: 124 passed. Pre-commit `check` + `typecheck` green.

PR #5121, opened 2026-09-23
feat(chat): step through a message's images in the image viewer (SOK-1163)

mrosberghaus:
Resolves [SOK-1163](https://linear.app/masumi/issue/SOK-1163/step-through-a-chat-messages-images-in-the-image-viewer). Drive follows in [SOK-1162](https://linear.app/masumi/issue/SOK-1162/step-through-a-drive-folders-images-in-the-image-viewer).

## What changes for users

Opening one of several images in a chat message (room, thread or pinned panel) now lets you step through that message's images without closing the viewer:

- `2 / 3` before the file name in the toolbar
- previous / next buttons on the image edges; the one at the first or last image fades and does nothing (no wrap)
- ← / → keys, and a swipe on touch while the image is at 100% zoom
- zoom resets to 100% on every image; the previous and next image preload
- a message with a single image looks exactly as before

The gallery is the **Message image gallery** (new glossary term in `CONTEXT.md`): the message's own image attachments, in body order across all its attachment rows, a file linked twice counted once. Other file kinds, quote and unfurl images are not in it. If an edit removes the open image, the viewer closes and does not reopen when the file comes back.

## How

- `ImageViewer` takes `images`, `activeSrc` (`null` = closed) and `onActiveSrcChange` instead of one `src`. It is open only while `activeSrc` is in `images`, so Drive can pass a folder's images later.
- `ChannelMessageText` builds the gallery and owns one viewer per message. `FileChipMiniPreviewFrame` now requires `onOpenImage`, so a chat tile never opens a viewer of its own.
- The composer chip, the file chip and the Drive card pass a one-image list.
- Disabled arrows use `aria-disabled`, not `disabled`, so focus stays in the viewer and the arrow keys keep working after clicking to an end.
- Apple parity: new row 15b (Todo) in `apps/apple/PARITY.md`; row 15a's web line references updated.

## Verification

- `pnpm check`, `pnpm typecheck` (17/17), `pnpm --filter web test` (7409 passed, 1 skipped), full `pnpm test` (17/17)
- New tests: `image-viewer.test.tsx` (position, one image, buttons, keys, focus after the end, swipe, zoomed and pinch-zoomed swipe, zoom reset, live close, preload) and `room-message-row.test.tsx` (gallery across rows skipping a PDF, duplicate link, edit removes and restores the open image)
- Visual: the real viewer markup rendered with the compiled app CSS at desktop and 375 px. Not yet clicked through in a running app: keys, swipe and preloading are covered by tests only.

[...]

PR #5125, opened 2026-09-23
fix(chat): stop nesting attachments inside the pinned row's jump button

mrosberghaus:
## What changes for users

In the Pinned Messages panel, a row's attachments now work in place. Opening a document, following a file link, playing a video or audio file, or hovering a mention no longer jumps to the message and closes the panel. Clicking anywhere else on the row still jumps, and keyboard users still reach the jump with Tab and activate it with Enter or Space.

## Why

Each pinned row was one `<button>` wrapping `ChannelMessageText`. That body renders `FileChipMiniPreviewFrame` tiles (image and document buttons, file links), media players, mention hover-card triggers and markdown links. Two problems followed:

1. The HTML was invalid: interactive content was nested inside a `<button>`.
2. Using any of those controls bubbled to the row and jumped. #5121 patched only the image tile, with `stopPropagation` in the image trigger and on the ImageViewer's `DialogContent`.

## How

- The header (sender and time) is now the jump `<button>`. Its `::after` stretches over the row (`after:absolute after:inset-0`), so a click anywhere on the row still jumps.
- The body is a sibling of the button, not a child. It sits above the stretched target (`relative z-[1] pointer-events-none`) and passes clicks through to it. Its own controls (`a, button, audio, video, [role=button], [data-slot=hover-card-trigger]`) opt back in with `pointer-events-auto`. The unpin button sits above the target too.
- Because no control in the body has the jump button as an ancestor, React's portal bubbling (image viewer, document viewer) no longer reaches a jump handler.
- The body is the button's `aria-describedby`, so screen readers hear the message and not only "Ada, 1m ago". The focus ring is drawn on the `::after`, so it wraps the whole row.
- This reuses the layering from `DriveItemCard` (`drive-item-card.tsx`): a full-card button underneath a `relative z-[1] pointer-events-none` body.

## Removes the #5121 workaround

#5121 kept image tiles from jumping by stopping propagation: in the tile's `handleOpenImageClick` and in an `onClick` on the ImageViewer's `DialogContent`. With no jump handler above the body any more, both go, along with their two "ancestor click" tests. The panel test now covers that case end to end: it opens a real image tile into a real gallery viewer, zooms, and closes by clicking the stage, all without jumping.

## Verification

- `pinned-messages-panel.test.tsx`: 5 new tests. No control sits inside the jump button, and the body describes it. Opening an attachment or following a link doesn't jump. The layering classes are pinned. Tab + Enter jumps. The image viewer opens and is used without a jump.
- The attachment test and the image-viewer test both fail against the old row (with the workaround removed): `onJump` is called twice in each.
- Hit test in headless Chromium, using the rendered markup and compiled `globals.css`. `elementFromPoint` on body text, row padding and header → jump button. On document tile, file link, mention and unpin → those controls.
- Tab focus lands on the jump button, and the ring wraps the full row at `rounded-lg`.
- `vitest run src/app/(app)/chat src/components src/app/(app)/drive`: 287 files, 2587 tests pass.
- Pre-commit `pnpm check && pnpm typecheck` passes.
- Not verified in the running app, which needs sign-in against the worktree DB.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/pinned-messages-panel.tsx`: body text cannot be selected. It is `pointer-events: none`, so a drag never starts. Click-to-jump is the stated behavior.
- Minor: same file: `data-testid="pinned-message-body"` is repeated on every row.
- Minor: same file: only `a`, `button`, `audio`, `video`, `[role=button]`, and `[data-slot=hover-card-trigger]` take hits again. A later control outside that list will jump. Disabled task-list checkboxes already fall through.
- Skipped / disagreed: restoring `stopPropagation` on `ImageViewer` — the drive activate control is a sibling button, not an ancestor, and the message article has no `onClick`. The panel test covers the portaled viewer.
- Skipped / disagreed: replacing `aria-describedby` — the body as the description is the requirement, and the test asserts it.

### Assessment
The header button's stretched target matches the drive-card hit model. Compiled Tailwind and a headless hit test send body text, padding, and the header to the jump button, and a document button, link, mention, video, and unpin control to those controls.

mrosberghaus:
Reply to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5125#issuecomment-5794025361):

**Hard-coded control list: fixed in 1ff101f73.** The body now names its controls by kind instead of by component: anything natively interactive or focusable (`a, button, input:enabled, select, textarea, summary, audio, video, [role=button], [tabindex]`), hover-card triggers, and horizontal scrollers (`pre, .overflow-x-auto`).

The old list also had a gap the review didn't mention. Code blocks and tables sit in `overflow-x-auto` wrappers, and under `pointer-events: none` they couldn't be scrolled. Inside the old button they could.

I re-checked hits in headless Chromium against the rendered markup and compiled CSS:
- Body text and disabled task-list checkboxes still go to the jump button. A disabled checkbox has no action of its own, so jumping from it is fine.
- An enabled checkbox, the code block, the table scroller and a focusable mention take their own pointer events. The code block scrolls.

**Body text can't be selected: leaving as is.** The old row was a `<button>`, and you couldn't drag-select its text either, so nothing regressed. Click-anywhere-to-jump and drag-to-select also fight each other, and the jump is the row's job.

**Repeated `data-testid="pinned-message-body"`: leaving as is.** A test id names a role, not an instance. `room-message-attachment-row` already repeats per row the same way. The unique handle is the body's `id`, which the button's `aria-describedby` points to.

PR #5127, opened 2026-09-23
feat(tasks): move recurring rules into Task Schedule

mrosberghaus:
Spec: [SOK-1164](https://linear.app/masumi/issue/SOK-1164/feattasks-move-recurring-rules-into-task-schedule) · Decision record: `docs/adr/0041-recurring-rules-move-to-task-schedule.md`

A repeating rule becomes its own **Task Schedule** instead of a template Task. Each Occurrence creates an ordinary Ready Task that points back to its schedule, and a single Task only ever has one **Run at**.

**Do not mark ready or merge before ticket 10 lands.** From ticket 02 on, each SOK-1164 ticket is reviewed in its own draft PR, based on this branch. Merging that PR adds the ticket here, and the whole change ships in one deploy (ADR 0041).

## Progress

- [x] 01/10 [SOK-1165](https://linear.app/masumi/issue/SOK-1165): Task Schedule API (tracer bullet), `0bb7cf8`
- [x] 02/10 [SOK-1166](https://linear.app/masumi/issue/SOK-1166): Occurrences create Tasks, #5129 (`8fa8543`)
- [x] 03/10 [SOK-1167](https://linear.app/masumi/issue/SOK-1167): Change single Runs via the API (Occurrence renamed to Run; ADR now 0041), #5138 (`15980a5`)
- [x] 04/10 [SOK-1168](https://linear.app/masumi/issue/SOK-1168): Run at, #5155 (`0f16734`)
- [x] 05/10 [SOK-1169](https://linear.app/masumi/issue/SOK-1169): Schedules view in the web app, #5163 (`36b9c6a`)
- [ ] 06–10: see SOK-1164

## 01/10: Task Schedule API

Additive expand step. People and Coworker vendors can create, read, list, update, pause, resume, end, and delete Task Schedules at `/v1/tasks/schedules`. Today's per-Task series behaviour is unchanged.

- **Schema:** migration `20260923130000_task_schedule` adds the `task_schedule` table. It stores the rule as typed columns, the blueprint, the state, and a revision, with DB checks for exactly one creator, at most one assignee, and a consistent end rule. It also adds nullable `task.scheduleId` (ON DELETE SET NULL) and `task.runAt`, and nullable `task_schedule_occurrence.scheduleId`. No column is dropped.
- **Access:**
  - Every caller passes the organization Seat gate.
  - People read through the Task visibility rule and change only schedules they own.
  - Coworkers also need a GRANTED vendor workspace grant. They see private schedules only while acting for the owner, and only on their vendor family.
- **Revisions and state:** PATCH is revision-checked and returns 409 `schedule_revision_conflict`. Illegal state moves return 409 `schedule_state_conflict`.
- **Delete:** deleting a schedule keeps the Tasks it created. Their `scheduleId` goes to null.
- **Deletions:** user deletion and Soko Bot deletion now handle the new RESTRICT creator FKs. The Soko Bot deletion result gains `retained.taskSchedules`.
- **Docs:** adds the `CONTEXT.md` terms, ADR 0041, and a status note on ADR 0029.

## Verification

- Ran `pnpm check`, `pnpm typecheck`, and `pnpm test`: all green. Core ran 6,440 tests and Web 7,405.
- Added 67 mocked-Prisma route tests beside the new routes. They cover rule, timezone, and end-rule validation; each assignee kind; private visibility; an unseated member; a Coworker with and without a grant; the revision conflict; and illegal state moves.
- Ran `prisma migrate deploy` of all migrations on an empty Postgres. `migrate diff` shows no drift on the new objects.
- Regenerated the Web Core client with `pnpm --filter web generate:core:snapshot`.


mrosberghaus:
## Grok review

### Fixed
- Important: every-N-days next run skipped a same-day slot whose local time was still ahead, so create, PATCH, and resume stored the following occurrence (`a1df222ef` — `apps/core/src/helpers/task-schedule.ts`)

### Remaining
- Minor: `apps/core/src/schemas/task-schedule.schema.ts`: a PATCH `rule` that omits `timezone` or `endsMode` resets them to UTC / NEVER, because those fields default on the shared rule schema.
- Minor: `apps/core/src/test-fixtures/task-schedule.ts`: the Prisma double ignores coworker capability and whitelist, and treats null as matching `{ not }`.
- Minor: pause, resume, and end compare state only, not `expectedRevision`.
- Skipped / disagreed: project close or delete does not End schedules — this slice is the additive API, nothing releases occurrences yet, and the project FK is SetNull on purpose.
- Skipped / disagreed: a Coworker cannot read a private schedule outside its vendor family, including one it created with no coworker assignee — that matches the access rules and the GET tests.
- Skipped / disagreed: a Coworker cannot assign a Soko Bot — `requireTaskReferences` uses the same `taskAssigner` (`kind: "other"`) as task create.

### Assessment
SOK-1165 matches the access, revision, and state rules; keep this draft until tickets 02–10 land.

mrosberghaus:
Reply to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5127#issuecomment-5793990177):

**Fixed interval slot (`a1df222ef`):** checked and kept. The fix is in the shared `computeIntervalNextRun`, so existing every-N-days series pick it up too. That is a bug fix, not a behaviour redesign.

**Remaining**
- **PATCH `rule` resets `timezone` / `endsMode`:** fixed in `f42859478`. PATCH now takes a separate `TaskScheduleRuleReplacement` schema with no defaults, so a replacement rule must send both fields and an omitted one returns 422 instead of silently resetting. Create keeps its UTC / NEVER defaults. Tests cover both fields.
- **Prisma double treats NULL as matching `{ not }`:** fixed in `f42859478`. It now follows SQL semantics. I did not add Coworker capability or whitelist modelling: no route test depends on it, and those lookups are covered by `access-control` tests.
- **Pause / resume / end ignore `expectedRevision`:** keeping as is. The ticket asks for revision checks on PATCH only. These actions are intents guarded by the current state (`updateMany … where state`), so a stale caller gets 409 `schedule_state_conflict` rather than overwriting anything.
- **Skipped / disagreed:** agreed with all three. Project close and delete ending schedules is ticket 7.

Checks: `pnpm check`, `pnpm typecheck` and `pnpm test` are green (Core 6,443, Web 7,424). Keeping this PR as a draft until ticket 10 lands.

enjojoy:
It's definitely not ok right now
image

enjojoy:
I can open the calendar of each project but they are empty

enjojoy:
Calendar still not loading

enjojoy:
Right now when I want to schedule a task for a coworker I can only pick one date, no recurrence or custom rule, is it intended? 

image

PR #5126, opened 2026-09-23
feat(apple): mute a thread from the thread view

mrosberghaus:
Apple chat parity row **24b** — mute a thread (ADR 0030, web `ThreadMuteButton`).

## What users see
- In a thread, a bell appears in the window toolbar once Core confirms the parent is a thread ("Mute thread"). Click it: it stays pressed with a struck-through bell ("Unmute thread"); the thread's replies stop counting toward the room's unread and the Threads count and stop notifying, unless they mention you. Click again to unmute.
- A parent with no replies shows no bell until the first reply arrives.
- The Threads overview marks a muted thread with a struck-through bell ("Muted") beside its time.
- If Core refuses, the toggle springs back and "Could not mute this thread." / "Could not unmute this thread." shows at the top of the thread (dismissible). Web reverts silently; this is a recorded deviation.

## How
- Snapshot: `POST`/`DELETE /chats/rooms/{id}/threads/{parentMessageId}/mute` selected with `scripts/update-core-api.py` from the deployed `openapi.json` (ops 68 → 70; additive deployed drift on `ChatRoom`, `ChatRoomMessage`, notification enums came along; details in PARITY).
- `SokosumiChat`: `ThreadMuteState` (unknown → read → optimistic toggle → settle/revert), `ChatService.muteThread`/`unmuteThread`, and `ThreadSession.readMute`/`toggleMute` with stale-response guards.
- `SokosumiWorkspace`: `readThreadMuteIfNeeded`/`toggleThreadMute`; after a change the Threads count re-counts and the room read runs (shared with the thread Look, as web's `bumpThreadUnread` + `syncRoomAttentionAfterThreadLook`).
- App: `ThreadMuteToggle` (native button-style `Toggle` in the thread view's toolbar), the overview marker, and `ThreadMuteFailureRow`.
- PARITY: row 24b In review with the web/Core audit, row 18b Done ([#5118](https://github.com/masumi-network/sokosumi/pull/5118) merged), 24c next.

Thread mute, light over dark

## Verification (from `apps/apple`)
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO` — 1138 tests, 0 failed (SokosumiTests 199, Chat 724, Auth 35, Realtime 48, Workspace 131, CoreAPI 1). A first run failed only the known load-flaky `HistoryGapRowTests` OCR case, which passed on rerun; the second full run passed 1138/1138.
[...]

PR #5128, opened 2026-09-23
fix(apple): page the thread overview and settle reads like web

mrosberghaus:
Apple parity row **24c**: thread overview paging and read effects. Current web was re-audited at `ca25794db`, after 24b. All three rules in the row were still open. For (c), only the automatic Look was still open: 24b had already made the explicit Look re-count.

## What changes for the user
- **Paging:** the Threads overview now loads 50 threads per page, first and older, like web's `THREAD_LIST_PAGE_LIMIT`. Before, Apple got Core's default of 20.
- **Mark all as read:** once Core accepts it, the Threads count is re-counted and the room read is posted, and then the list reloads. The room's sidebar row now loses its thread attention, as on web. The count waits for Core's answer instead of dropping to zero, because Mark all skips a muted thread that has an unread mention. If Mark all fails, none of these follow-ups run.
- **Automatic Look:** returning to the window with a thread open, or a new reply arriving there, now re-counts the Threads count after the automatic Look. Web does this with `onThreadLooked`.

## How
- `ChatService.listThreads` sends `limit: 50`.
- `RoomThreadOverview.markAllRead(…, looked:)` awaits `looked` between Core's answer and the reload. The overview no longer zeroes or re-counts the count itself.
- `RoomReadAttention.readIfNeeded(…, threadLooked:)` calls `threadLooked` for every Look that reaches Core.
- The workspace passes the shared `syncRoomAttentionAfterThreadChange` from 24b to Mark all, and bumps `threadAttentionRevision` from `syncReadAttention`. There is no second path.

## Found beside it, not implemented
Web's overview puts its rows under **Unread** and **Earlier** headings, with "All caught up." when nothing is unread, and its row lines also differ. Apple has no headings. This is recorded as new row **24d** (Todo, depends on 24), and it now heads the work order ahead of 09c.

## PARITY
- Row 24b → Done (#5126, merged 2026-09-23), noting Grok's follow-up `2f3cbd9e4`.
- Row 24c → In review, with corrected text.
- New row 24d.
- New "Slice 24c" section with the web audit answers.
- Resume checkpoint and work order updated.

Tests were written first. Red was run against the new signatures with the old behaviour, so the failures were behavioural, not compile errors.
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`: TEST SUCCEEDED on the first run, 1149 tests, 0 failed, 0 skipped.
- macOS `xcodebuild … build`: BUILD SUCCEEDED.
- iOS 17 `swift build` of `SokosumiWorkspace`: Build complete.
- `mint run swiftformat --lint .`: 0/410 files need formatting.
- `mint run swiftlint lint --strict`: 0 violations.

**Not verified:** the app was not launched. Mark all against Core, Load older on a room with more than 50 threads, and the Threads count after returning to a window were not exercised live.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `RoomThreadOverview.markAllRead` does not fetch the unread-thread count. It only invalidates an in-flight count, then the caller bumps `threadAttentionRevision`. The toolbar number still comes from the existing 150ms task in `RoomToolsModifier`. That matches web. The new workspace test checks the revision bump and the room read, not `GET …/threads/unread-count`.
- Skipped / disagreed: posting the room read while a history jump is open — `syncRoomAttentionAfterThreadChange` already refuses that, because `POST …/read` would mark channel messages the reader has not seen. On the live tail the room read still runs.
- Skipped / disagreed: web also clears `threadUnreadReplyCount` on transcript rows after Mark all and Look. Apple rows do not render that field, so there is nothing stale on screen. Row 24d is the headings gap, not this.

### Assessment
Ready to merge: Yes
Every page asks for 50 threads, Mark all waits for Core before the room read and the reload, and an automatic Look bumps the Threads count only after Core accepts it.

mrosberghaus:
Re the Grok review above (no change pushed):

- **Mark all doesn't fetch the unread-thread count itself:** intended, and it matches web. Web's Mark all calls `onAllThreadsLooked`, which bumps the trigger, and `use-unread-thread-count.ts` refetches 150 ms later. Apple's `markAllRead` does the same by bumping `threadAttentionRevision` once Core accepts. `RoomToolsModifier`'s `.task(id: … threadAttentionRevision …)` (`RoomToolsModifier.swift:54`) then runs `GET …/threads/unread-count`. The workspace tests pin the part this PR changes: the bump, the room read and the order. The revision → GET task is slice-24 code this PR doesn't touch. You're right that no test pins it, so that gap is noted for a follow-up; it isn't added here.
- **No room read while a history jump is open:** agreed, no change. `POST …/read` would mark channel messages the reader hasn't seen. On the live tail the read runs.
- **Web clears `threadUnreadReplyCount` on transcript rows:** agreed, no change. Apple rows don't render that field, so nothing on screen goes stale.

PR #5129, opened 2026-09-23
feat(tasks): create a Task at every Task Schedule Occurrence (SOK-1166)

mrosberghaus:
Ticket 02/10 of [SOK-1164](https://linear.app/masumi/issue/SOK-1164) · [SOK-1166](https://linear.app/masumi/issue/SOK-1166) · ADR 0040

This PR is stacked on #5127: its base is the SOK-1164 branch, not `main`. Merging it adds this ticket to #5127, which ships the whole change in one deploy after ticket 10.

An Active Task Schedule now creates a new Ready Task at every Occurrence. The existing `/sync/task-schedules` cron releases Task Schedules after the old series path, which keeps running until ticket 09.

## What changes

- **Schema:** migration `20260923160000_task_schedule_occurrences` is additive.
  - It adds `task_schedule.epochId`, the rule epoch. A rule edit starts a new epoch, so the new rule's Occurrences never collide with the old one's.
  - `task_schedule_occurrence.seriesTaskId` becomes nullable. A CHECK requires exactly one parent (series Task or Task Schedule).
  - A unique key on `(scheduleId, epochId, originalScheduledAt)` stops the same Occurrence being planned twice.
- **Planning Occurrences** (new module `services/task-schedule-occurrences.service.ts`):
  - It plans Occurrences over the 90-day calendar horizon, always including the next one, capped at the index limit.
  - `nextOccurrenceAt` is the earliest planned Occurrence.
  - Planning happens on create, rule edit, and resume.
  - A rule edit drops the future plan but keeps Occurrences that are already due and not yet released.
  - Pause and end drop the plan.
  - When the blueprint's project changes, the planned Occurrences move with it.
- **Release:**
  - Creates a Ready Task from the blueprint: name, description, project, visibility, any assignee kind, the schedule's creator, `scheduleId`, and a Ready event.
  - Creates a new Task even if the previous one is still open.
  - Advances `releasedCount` and `nextOccurrenceAt`, and moves the schedule to Ended when its end rule is met.
  - Skips Paused and Ended schedules, and schedules whose project is closing.
  - Is idempotent: each Occurrence is claimed from planned to released, and the schedule by its revision, so a retried run or a concurrent edit never duplicates a Task.
  - Runs at most 50 releases per transaction and catches errors per schedule, so one broken schedule does not block the others.
  - Notifies a person assignee, as Task create does.
  - Does no Seat check (ADR 0020).
- **Old series code:** the calendar read (until SOK-1170), project close, and reconciliation now ignore Task Schedule rows.

## Review notes

- **Why not reuse existing Task creation:** `createTaskForActor` re-checks references and vendor grants, which would stop org work at release (ADR 0020). `cloneRecurringTaskScheduleOccurrence` depends on the template Task's metadata and is removed in ticket 09. Only `getOccurrenceSource` is reused.
- **Deploy window:** while the new Core rolls out, the old Core could briefly meet a Task Schedule row. This is harmless: such rows only exist once the new Core has created them.

## Verification

- Ran `pnpm check`, `pnpm typecheck`, and `pnpm test`: all green (Core 6,478 passed). The old series release tests are unchanged and pass.
- Added 23 release-service tests (mocked Prisma). They cover:
  - a Task from the blueprint for each assignee kind, and the Coworker creator;
  - overlap, overdue Occurrences, and the rolling horizon;
  - end after N and end on date;
  - Paused and Ended schedules never firing, and closing or closed projects being skipped;
  - retries, a lost Occurrence claim, and an edit during release;
  - an Occurrence owed across a rule edit;
  - a stopped backlog that does not overrun end-after-N;
  - a batched backlog, and one failing schedule not blocking the others.
- Extended the route tests:
  - create plans Occurrences (horizon, end after N, beyond the horizon);
  - edit re-plans the rule, keeps released Occurrences and their Tasks, keeps owed ones, and counts them against end after N;
  - a project change moves the planned Occurrences;
  - pause and end drop the planned Occurrences;
  - resume plans again from now.
- Ran `prisma migrate deploy` on an empty Postgres. `migrate diff` shows no drift on the new objects.


mrosberghaus:
## Grok review

### Fixed
- Important: a full page of schedules that stay due (create throws, or the claim is lost) was read again on every pass and starved every schedule behind them (`803669566` — `apps/core/src/services/task-schedule-occurrences.service.ts`)
- Important: a PATCH that changed the rule and the project returned before moving owed Occurrences, so those rows kept the old project (`803669566` — `apps/core/src/services/task-schedule.service.ts`)

### Remaining
- Minor: `apps/core/src/services/task-schedule-occurrences.service.ts`: each release re-inserts the whole horizon (up to 2000 rows) in the same interactive transaction as the Task inserts. Fine for a weekly rule; a minute rule can get slow.
- Minor: release and route tests use the in-memory Prisma double, so the parent CHECK, the new unique index, and the calendar invalidation trigger are not exercised here.
- Skipped / disagreed: do not drop owed Occurrences when a new end-after-N is below released+owed. The ticket says a rule edit keeps Occurrences already due, and those rows can then release past the new cap.
- Skipped / disagreed: release does not bump `revision`. An edit that commits first still rolls the release back, because the schedule write requires the revision it read. Bumping on every release would 409 an in-flight user edit.
- Skipped / disagreed: closing projects are left planned for ticket 7, calendar hides these rows until SOK-1170, and release does no Seat check (ADR 0020). `createTaskForActor` stays unused so release does not re-check vendor grants.

### Assessment
Occurrence planning, the release claim, and the cron handoff match SOK-1166; the two fixes stop a failed page from blocking the queue and keep owed rows on the project from the same PATCH.

mrosberghaus:
Follow-up on the Grok review.

**Fixes pushed by the review (`803669566`):** checked both, and they are correct. A full page of schedules that stay due no longer starves the ones behind it, and a PATCH that changes the rule and the project together now moves the owed Occurrences as well.

**Remaining, minor: horizon re-insert.** Fixed in `d872dc023`, and it was worse than it looked. On Postgres 17, a `BEFORE INSERT` trigger runs even for a row that `ON CONFLICT DO NOTHING` then skips. So each release ran `invalidate_calendar_for_occurrence_change` once per Occurrence in the plan, up to 2,000 times for a dense rule. The projection now continues after the last planned Occurrence of the current epoch, and writes only the missing tail. Every planned Occurrence counts against end-after-N and the index limit. The first Occurrence beyond the horizon is planned only when nothing is planned yet. The new test `writes only the Occurrences the plan is missing` covers this.

**Remaining, minor: DB constraints not exercised by the Prisma double.** Checked by hand against a throwaway Postgres with all migrations applied, inside a rolled-back transaction:
- A Task Schedule Occurrence with no series Task inserts, and the calendar invalidation trigger writes an outbox row.
- A duplicate `(scheduleId, epochId, originalScheduledAt)` is rejected by `task_schedule_occurrence_schedule_epoch_original_key`.
- A row with neither parent is rejected by `task_schedule_occurrence_parent_check`.

These checks are not committed. Per the repo rule, new coverage stays on mocked Prisma; the SQL migration test is ticket 08 (SOK-1172).

**Skipped / disagreed items:** agreed with all three. Not bumping `revision` on release is deliberate, and the same goes for the ticket 07 and SOK-1170 boundaries and for keeping `createTaskForActor` out of release (ADR 0020).

**Branch sync:** `main` was merged into the SOK-1164 branch to fix the conflict on #5127. The conflicts were only in the generated Web Core client, which was regenerated with `pnpm --filter web generate:core:snapshot`, not hand-merged. That merge is carried into this branch too.

**Verification:**
- `pnpm check` and `pnpm typecheck` pass.
- Core: 6,467 tests pass.
- Task Schedule tests: 157 pass.
- Web: two tests failed only in the full parallel run (`tasks/[taskId]/page.test.tsx` timed out at 5s, and `admin-soko-bot-actions.test.tsx`). Both pass when run alone, and neither touches Task Schedules.

PR #5130, opened 2026-09-23
fix(chat): quote a Message link from a public channel for any org member

mrosberghaus:
## What changes for users

Pasting a link to a message from a **public or external channel** now turns it into a quote even when the room you paste into has people who are not on that channel's roster, as long as they belong to the channel's organization (they can join it themselves). Before, the link stayed a plain URL.

Guests, members of other organizations and **private** channels keep the old rule: every reader of the target room must already be a member of the source room.

## Cause

The cross-room quote audience rule (`canQuoteIntoRoom`) counted only source-room members as readers. It ignored that any organization member can self-join a public or external channel.

## Fix

- **Core** (`resolveCrossRoomQuoteSnapshot`): for a public/external source channel, target readers who are members of the source's organization count as readers. This is the authoritative check at send time.
- **Web** and **Mac app** paste check: the same rule, using non-guest members of a room in the same organization as the stand-in for organization membership (clients can't see organization membership).
- `isSelfJoinableChannelDiscoverability` in `@sokosumi/utils` is now the one definition of "public or external", shared with Core's self-join check.
- `apps/apple/PARITY.md` row 17a updated.

No schema or OpenAPI change.

## Verification

- Core `post.test.ts`: new cases for a public/external source (201), a reader outside the organization (400), and a private source staying roster-only (400). The public case failed before the fix.
- `pnpm` vitest: utils 700 passed, Core chat 662 passed, Web chat 1359 passed.
- `swift test --filter MessageLinkQuoteTests`: 9 passed. SwiftLint and SwiftFormat clean.
- `pnpm check` and `pnpm typecheck` (commit hook) pass.
- Not checked in a running app or on a preview deploy yet.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/PARITY.md` row 17a attaches "and never in the coworker 1:1 stream" to the Self Direct sentence. The coworker limit is still on cross-room paste, not on Self Direct.
- Minor: Web and Mac only treat non-guest members of a same-organization room as able to join. A Personal Direct (`organizationId` null) stays a plain link even when `peerInActiveOrganization` is true and Core would accept the quote. That misses closed (no snippet).
- Skipped / disagreed: Private-channel owners and admins can self-join, but quotes stay roster-only. The PR requires that. Not a bug.

### Assessment
Core's send-time check matches who can self-join a public or external channel, and the paste check fails closed when it cannot see organization membership.

mrosberghaus:
Replying to the Grok review (https://github.com/masumi-network/sokosumi/pull/5130#issuecomment-5794649979):

- **PARITY 17a sentence**: Fixed in ae3895af7. The Self Direct clause is its own sentence again, and the coworker 1:1 limit now applies to cross-room links.
- **Personal Direct with a peer in the organization**: Leaving as is. The paste check fails closed, so the link just stays a link, and Core would accept the quote. Supporting this would mean trusting `peerInActiveOrganization` in the paste check on both Web and Mac, plus a parity row, for a narrow case. Can follow up if it comes up in use.
- **Private-channel owners/admins**: Agreed, not a bug. Quotes from private channels stay roster-only on purpose.

PR #5131, opened 2026-09-23
test(apple): pin that a Threads attention bump re-fetches the toolbar count

mrosberghaus:
## Summary

This adds a test. It changes no app behaviour and no `PARITY.md` row.

`RoomToolsModifier` keys the Threads trigger's count task on `WorkspaceState.threadAttentionRevision`. After a 150 ms debounce the task calls `updateThreadOverview(.count)`, which sends `GET /chats/rooms/{id}/threads/unread-count`. The package tests already prove that Mark all, a thread Look and a mute each bump the revision. No test proved that a bump makes the toolbar fetch the count again. The Grok note on #5128 raised this gap: https://github.com/masumi-network/sokosumi/pull/5128#issuecomment-5794512296

`SokosumiTests/Chat/Search/RoomToolsThreadsCountTests.swift` mounts the real `RoomToolsModifier` in a window with bridged toolbars. The window has an open room and a stubbed Core, set up through a `ClientTransport` that answers by `operationID`. The unread-count request returns 3 and then 1. The test then:

1. Waits for the first count: one unread-count GET, and the trigger state shows 3. It uses the existing `waitForView` readiness helper.
2. Runs Mark all through the public `updateThreadOverview(.markAllRead)` path, the same call the overview's button makes. Core accepts the request and the room bumps the revision.
3. Checks that exactly one more unread-count GET goes out, for a total of 2, so Mark all sent none of its own. It also checks that `threadOverview.unreadCount`, which the trigger draws, is now 1. The bridged `NSToolbarItem` for Threads has an empty label and exposes only its tooltip, so the test reads the state value rather than the drawn item.
4. Renders again without a bump, using `objectWillChange` and a probe view that counts renders. It then checks that no further unread-count GET goes out.

An absent request can't be awaited, so step 4 watches for three debounce windows and stops early if the count changes. Every other wait uses the readiness helper.

`RoomToolsModifier.swift` has one change: the 150 ms sleep now reads a named constant, `RoomToolsModifier.threadsCountDebounce`. The test sizes its quiet window from the same constant, so the window follows the production value. Behaviour is unchanged.

### Red proof

With `threadAttentionRevision` temporarily removed from the task id, the test fails:

```
Expectation failed: readyView(): Timed out after 10 seconds: the re-count after the bump; 1 unread-count requests, trigger shows 3
```

After I restored the id, the test passes. I re-ran this proof after the review fixes and got the same message.

## Verification

- The new test passed 3 times in isolation: `-only-testing:SokosumiTests/NativeWindowTests/RoomToolsThreadsCountTests -test-iterations 3`
- After the review fixes: the new test passed 3 more times in isolation, and the composer and search app suites passed (136 tests, 202 runs, 0 failures).
- Before the review fixes, the full workspace run passed: `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`, 1150 tests (1561 runs), 0 failures
- `mint run swiftlint lint --strict`: 0 violations
- `mint run swiftformat --lint .`: 0 files need formatting


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/SokosumiTests/Chat/Search/RoomToolsThreadsCountTests.swift`: `roomThreadsAccessibilityLabel(unreadCount:)` does not read the mounted toolbar item. A button that stopped using `threadOverview.unreadCount` would still pass. The second unread-count GET and `unreadCount == 1` are the real pin.
- Minor: `RoomToolsThreadsCountTests.swift`: `countDebounce` copies the 150ms sleep in `RoomToolsModifier`. A longer production sleep lets the quiet check return before a mistaken refetch. The positive check still waits up to 10s.
- Minor: `RoomToolsThreadsCountTests.swift`: `#require(transport.countRequests == 1)` runs immediately after `markAllRead`, inside the same 150ms debounce window. A stalled main actor can let the toolbar recount land first and fail a correct run.
- Skipped / disagreed: Look, mute, and burst coalescing — package tests already bump `threadAttentionRevision` for Look and mute, and this task only had to pin one bump to one more `GET …/threads/unread-count`. The only `.count` caller is the modifier task.
- Skipped / disagreed: drawn digit vs dot — `showsRoomUnreadCount` starts false until the preferences GET. The requirement checks `threadOverview.unreadCount`, which is what the trigger's label reads. No pixel or accessibility-tree assertion.
- Skipped / disagreed: Mark all's room `POST /read` — `timeline.reset` leaves `hasLoadedHistory` false, so `syncRoomAttentionAfterThreadChange` bumps the revision and returns before the room read. `WorkspaceStateTests.markAllReadPostsTheRoomReadAndReCountsBeforeReloading` already covers that read. This test still requires the bump.

### Assessment
Ready to merge: Yes
The window test mounts the real `RoomToolsModifier`, and a mark-all bump is what sends the second unread-count GET and stores Core's new count on the trigger.

mrosberghaus:
Grok review on #5131:

- **Accessibility-label check**: fixed in a1622b177. I removed the assertion. I checked whether the mounted item could be read directly: the bridged Threads `NSToolbarItem` has an empty label and only the tooltip "Threads", and its hosted view has no accessibility label. So the pin is still the second unread-count GET plus `threadOverview.unreadCount == 1`.
- **Copied debounce**: fixed in a1622b177. `RoomToolsModifier.threadsCountDebounce` (150 ms) now drives the modifier's sleep, and the test's quiet window is 3 × that constant. Behaviour is unchanged.
- **Racing count check after Mark all**: fixed in a1622b177. I removed the check. The final `countRequests == 2` already shows that Mark all sent no count request of its own.
- Look, mute and burst coalescing: agreed, no action.
- Drawn digit vs dot: agreed, no action.
- Mark all's room `POST /read`: agreed, no action.

Red proof re-run: with `threadAttentionRevision` dropped from the task id, the test still fails with `Timed out after 10 seconds: the re-count after the bump; 1 unread-count requests, trigger shows 3`. The new test passed 3 times in isolation, and the composer and search app suites passed (136 tests, 202 runs, 0 failures). Strict SwiftLint and SwiftFormat lint are clean.

PR #5132, opened 2026-09-23
fix(chat): show retained room sends as pending rows again

mrosberghaus:
Sending a message in a chat room shows it at once as a pending row again. Since #4986 (SOK-1146), a send only appeared after the server confirmed it. When that took a while, the composer cleared and the message seemed lost until it arrived.

### Cause

`mergeRoomMessages(existing, incoming)` keeps local pending shells only from `existing`. It skips shells that arrive in `incoming`. `RetainedRoomsClient` passed the pending shells as `incoming`, both when rendering the transcript and when rebuilding it for the next write, so every pending row was dropped. The fix passes them as `existing`. Server rows still confirm them in place by client message id.

### Validation

- New regression test in `rooms-client-messages-pending.test.tsx`: two sends in a retained room show two pending rows, and confirming the first leaves one. It fails without the fix and passes with it.
- `pnpm vitest run src/app/(app)/chat src/components/chat`: 176 files, 1,830 tests pass.
- Commit hook (`pnpm check && pnpm typecheck`) passed.

Slow room switches (the second symptom reported alongside this one) are a separate issue and get a separate PR.


mrosberghaus:
## Grok review

### Fixed
- Important: a pending shell’s local clock was treated as the end of loaded history, so a later jump during a slow send joined across the gap and hid the missing messages (`5eda9f9ed` — `apps/web/src/app/(app)/chat/utils/room-transcript-ranges.ts`)

### Remaining
- Skipped: slow room switches — the PR body keeps that as a separate change, and it is not in this diff.

### Assessment
Ready to merge: Yes
Pending rows show again on a retained room, and a jump while that send is in flight still keeps the history gap.

mrosberghaus:
Verified `5eda9f9ed` (keep history gaps while a send is pending).

- The issue is real, and this PR caused it. Before, pending shells never reached the `current` transcript passed to updates. Now they do, so `mergeRoomJumpWindow(current, page)` read a shell's local `createdAt` as the newest loaded row and joined a later jump window across the gap.
- The fix is complete. `loadedRanges` has one caller (`mergeContiguousRange`), and that call now filters shells.
- The new test fails on the parent commit and passes with the fix. Chat suites: 183 files, 1,882 tests pass.

Nothing else to change here. Slow room switches stay out of this PR, as noted.

PR #5133, opened 2026-09-23
feat(chat): list All unreads as one flat list, read rooms dimmed in place (SOK-1159)

mrosberghaus:
Follow-up to #5119 (SOK-1159). The All unreads filter looked wrong when you were caught up: the message was drawn like a chat row, and a "Pinned" heading sat over the single read room you had open.

## What changes under the filter
- **One flat list, no section headings.** It is ordered by newest activity. Pinned, Channels and Directs headings over one or two rooms were clutter.
- **Read rooms stay in place, dimmed** (`data-read`, `opacity-60`). Nothing moves out from under the reader. The Just read section is removed.
- **Caught up is a quiet centred message** (`role="status"`), not a row. The Show all link is removed, because the tinted All unreads row above already turns the filter off.
- **"Read just now"** labels the dimmed rooms once everything is read: the open room plus anything read during the pass.
- **Caught up counts unread rooms, not listed rows.** It shows even when only the open room is left.
- Pending invitations stay under the filter in External, and still prevent "caught up".

`unreadFilterPass` is now just the order rooms first appeared in. A room that turns unread later goes on top, and a listed room never moves.

## Verification
- Chat tests: 510 pass. Typecheck and Biome are clean.
- A static page built from the harness's markup with the compiled `globals.css` shows the four states: unread, reading, caught up with the room open, and caught up.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/__tests__/organization-chat-list-unread-filter.test.tsx`: `inboxRows(...).sort()` accepts either order, so the flat list being newest-activity is not locked. Fixtures share one `updatedAt`, so id order is the only activity order.
- Minor: `apps/web/src/components/chat/organization-chat-list.client.tsx`: `id="unread-inbox-read-label"` is fixed. The app sidebar and mobile `/chat` each mount this list. If both are caught up, `aria-labelledby` can name the other list. `useId` would keep them apart. Same label text, so this only matters when both filters are on.
- Skipped / disagreed: sorting with `compareChatRoomsByRecentActivity` — that helper ranks public before private ahead of recency. This list is newest `updatedAt`, then id. Muted rooms never enter it (`roomUnreadReads`).
- Skipped / disagreed: Channel and Direct compose actions disappear while the filter is on — those controls sit on the headings this PR removes. The All unreads toggle brings the headings back.
- Skipped / disagreed: an open room that was never unread is drawn at the bottom and moves to the top the first time it becomes unread. It is not in the pass until then. After that it stays put, which `advanceUnreadFilterPass` already tests.

### Assessment
Caught up follows unread rooms and pending invites, not the open row; rooms read during the pass stay dimmed in place; External invites still block the empty state.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5133#issuecomment-5795728594). Both minors are fixed in 94bad2c47.

- **Order not locked:** fixed. The tests no longer sort the rows before comparing. `#launch` is now newer than `#design`, so id order and activity order differ. The tests assert `["launch", "design"]` exactly. The in-place test also checks that a read room keeps its place after the read moves its `updatedAt` on.
- **Fixed label id:** fixed. The "Read just now" label now takes its id from `useId`, so the app sidebar and the mobile `/chat` list no longer share an `aria-labelledby` target.

I agree with the three skipped items as reasoned:
- `compareChatRoomsByRecentActivity` isn't used here, because its public-before-private buckets don't suit an inbox.
- The compose actions come back with the headings when the filter is turned off.
- An open room that was never unread joins the pass the first time it turns unread, and then stays put.

Chat tests: 510 pass. Typecheck and Biome are clean.

PR #5134, opened 2026-09-23
feat(core): name group Directs (SOK-1175)

mrosberghaus:
Closes [SOK-1175](https://linear.app/masumi/issue/SOK-1175/name-group-directs).

Any member of a group Direct can now give it a **Group name** that every member sees in place of the member list. Clearing it brings the member list back. Each change leaves a quiet status row ("Ada named the group Launch crew" / "Ada removed the group name"). The group stays a Direct: private, listed under Direct Messages, same members, same participant key.

## What changed

**Core**
- New nullable `chat_room.groupName`, separate from the stored room name ([ADR-0040](docs/adr/0040-group-name-is-separate-from-stored-room-name.md)); a CHECK keeps it null for channels.
- `PATCH /v1/chats/rooms/{id}` accepts `groupName` on group Directs only (started for 3+ humans, read from the participant key, so a shrunk group keeps its name). Any member may set it, no org role needed. It is trimmed and capped at 80 characters, and `""` or `null` clears it. Every other field on a Direct is still rejected. Saving the same value again writes nothing and publishes nothing.
- One sender-less status row per change (`metadata.groupNameChange`, promoted on the message DTO as `groupNameChange`). It cannot be reacted to, edited, pinned or quoted, and it is left out of the unread count (see below).
- The room DTO gains `groupName` and `isGroupDirect` (the latter works like `isSelfDirect`, so clients never parse the key).
- Per-reader notification titles and Soko Bot's chat listing and room lookup prefer the Group name.

**Web**
- The group Direct header title, and a "Rename" item in the sidebar row menu (where channel rows show "Edit channel"), open a one-field "Name this group" dialog. It reuses `updateRoomAction` and the room's existing `?edit=1` request.
- The sidebar, header, breadcrumb and composer read the name through `getRoomDisplayName`.
- `MembershipStatusRow` is now `RoomStatusRow` and also renders rename rows; the shared `isRoomStatusMessage` check replaces five `membership != null` checks.
- New strings in en/de/es.

**Apple (macOS; shared packages build for iOS 17)**
- "Name Group…" in the sidebar context menu and in the Members inspector.
- Rename rows render like joined/left rows, and the open room's title updates as a row arrives.
- `GroupNameDraft` / `GroupNaming` are modelled on `ChannelEditing`. New catalog: `ChatGroupName.xcstrings`.
- Renamed so the names still fit: `updateChannel` → `updateRoom`, `updatingChannel` → `updatingRoom`, `channelMutationInFlight` → `roomMutationInFlight`, `channelErrorMessage` → `chatErrorMessage`.

## Worth a look
- **Unread:** the spec says rename rows are excluded from unread "exactly like joined/left rows", but joined/left rows do count today. Rename rows are excluded, because with no sender the row would otherwise mark the room unread for the person who renamed it. Recorded in ADR-0040.
- **Deploy order:** Apple requires `isGroupDirect`, so Core must deploy before an Apple build that includes this.
- **Web, other members:** the sidebar picks up another member's rename live, but an already-open room's header updates on the next navigation. Channel renames behave the same way.
- Apple's regenerated `openapi.json` also carries earlier, unrelated Core schema additions. The app does not read them yet.

## Schema changes
- `packages/database/prisma/migrations/20260923120000_chat_room_group_name` adds a nullable column and a CHECK. It is additive; no data migration.

## Verification
- Core: `pnpm --filter @sokosumi/core test`, 6,396 passed. New mocked-Prisma tests beside `PATCH /chats/rooms/{id}`, room GET list and detail, `loadDirectRoomNamesByReader`, the Soko Bot runtime and the unread SQL.
- Web: `pnpm --filter web test`, 7,425 passed. 4 drive and tasks tests failed under the full run and pass on rerun; they are unrelated to this change.
- Apple: `xcodebuild test` on the `Sokosumi` workspace scheme (macOS), `swiftformat --lint`, `swiftlint --strict`.
- `pnpm check`, `pnpm typecheck` and `messages:parity` pass.
- Web dialog and rename rows: I rendered the real components with the app's compiled CSS and checked them in a browser.
- **Not verified:** the rename flow in a running app against a live Core, on Web or Apple. The Apple sheet and context menu were not inspected by eye.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/lib/ably/schema.ts`: the full message schema does not declare `groupNameChange`. Hydrate still reads it because the object is `.passthrough()`. A later strip would render the row as a normal bubble.
- Minor: `apps/web/src/app/(app)/chat/components/room-header-chrome.tsx`: the group title button's accessible name is the display name. "Name this group" is only the `title` tooltip, same as Edit channel.
- Minor: `apps/web/src/app/(app)/chat/components/name-group-dialog.tsx`: no Cancel control. Dismiss is the dialog close button. Apple's sheet has Cancel.
- Minor: `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/MessagePresentation.swift`: a `named` row whose `name` is null or empty renders as cleared. Core does not write that shape.
- Skipped / disagreed: open Web header stays on the member list until the next navigation — called out in the PR and the same as channel renames. Sidebar refetch and the status row are live; Apple retitles the open room from the row.
- Skipped / disagreed: `POST` can still thread a status row. Join/leave rows have the same hole, and neither client offers reply on a status row.
- Skipped / disagreed: the CHECK only forces `kind = 'direct'`, not "three or more humans". The group rule is the participant key, and PATCH is the only writer.
- Skipped / disagreed: no rename rate limit, and the stored `content` line is English. Clients localize from `groupNameChange`. Same pattern as membership rows. Rename does not fan out a notification.

### Assessment
Ready to merge: Yes
Group name is its own nullable column, PATCH accepts only that field on group Directs, the status row is left out of unread, and Web, Apple, notifications, and Soko Bot prefer it over the member list.

mrosberghaus:
Replying to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5134#issuecomment-5795808787).

**Fixed in 0dc350161**
- `apps/web/src/lib/ably/schema.ts` now declares `groupNameChange` beside `membership` instead of relying on `.passthrough()`. It is optional because Web and Core deploy separately, so an event from a Core without Group names still parses. Hydrate turns a missing value into `null`. Tests cover a rename row surviving hydrate and a payload without the field.

**Left as is**
- **Header button's accessible name is the group name.** This matches the channel title button, and it's what story 35 asks for: the title announces the name, and "Name this group" is the hover title.
- **No Cancel in the Web dialog.** The Edit channel dialog has only Save in its footer too. Esc and the close button dismiss both. I kept Web consistent with itself rather than with Apple's sheet.
- **Apple renders a `named` row with an empty name as cleared.** Core never writes that shape (PATCH stores null for an empty name and records `cleared`), so this is only a fallback for bad data.

PR #5135, opened 2026-09-23
feat(apple): group the thread overview under Unread and Earlier

mrosberghaus:
Apple chat parity row **24d**: the thread overview groups its rows under **Unread** and **Earlier**, like web's thread panel.

## What the user sees
- Opening Threads shows "Unread" with a count pill above the threads that have new replies, then "Earlier" above the rest. When nothing is unread, "All caught up." stands under Unread. A room with no threads shows only "No threads yet." and no headings.
- An unread row sits on a light accent fill and leads with an accent-tinted bubble. Its time is in the accent, and its second line reads "N new · starter". A read row has a grey bubble, a secondary preview and "Started by X · N replies". The 24b muted bell stays beside the time.
- Rows move only when the overview reloads from Core: after Back from a thread you looked at, or after Mark all as read.

## Web audit (current main, re-audited after #5119 landed mid-slice)
- The split uses `unreadReplyCount > 0` and nothing else (`threadNeedsOverviewUnread`). Mute and mention rules apply only through Core's count.
- Unread shows whenever at least one thread is loaded, carrying its count as a capped pill hidden from speech. Earlier shows only once a read thread is loaded. There are no headings for an empty room, before the first page loads, or after a failed first page.
- #5119 moved the row look into `thread-list-row.tsx`. It added the primary-tinted unread row, the tinted time and the heading count, and this PR follows all three.
- Recorded in PARITY but not changed, because they belong to a follow-up row: web's self-loading "Load older threads" row with its inline failure and retry, first-page failure clearing the list, and the loading and error copy.

## Apple changes
- `SokosumiChat`: new portable `RoomThreadOverviewGroups` (the split plus heading visibility), exposed as `RoomThreadOverview.groups`.
- `RoomThreadOverviewView`: native `Section`s with header-trait headings, the count capsule (reusing `roomCountLabel`), "All caught up.", and web's row lines and emphasis. The existing row layout is kept, and a private `RoomThreadOverviewMark` replaces the accent dot. Apple had no existing tinted thread circle to reuse.
- `ThreadMuteToggleTests`: the muted fixture row is read, so it now sits under Earlier.
- `PARITY.md`: row 24d is In review with evidence and corrected text, 24c is Done (#5128), there is a new "Slice 24d" section, and the work order now starts at 09c.

Thread overview groups, light over dark

## Verification (from `apps/apple`, on `59e173dd9` plus this change)
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`: 1160 tests, 0 failed (SokosumiTests 205, Chat 737, Auth 35, Realtime 48, Workspace 134, CoreAPI 1), passing on the first run.
- New tests: 6 UI-free grouping tests (9 cases) in `RoomThreadOverviewGroupsTests`, and 2 hosted render tests (7 cases, light and dark) in `RoomThreadOverviewGroupsViewTests`. Both were red before the change.
- macOS build and the iOS 17 `SokosumiWorkspace` build succeeded.
- `mint run swiftformat --lint .` reports 0/414 files, and `mint run swiftlint lint --strict` reports 0 violations.

Not exercised: the app was not launched. The groups against Core, a row moving live, the tint with a real key-window accent colour, and VoiceOver on the headings all remain unverified.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/Sokosumi/Chat/Threads/RoomThreadOverviewView.swift:108` — `onHover` sets `hoveredId` to nil on exit. Moving from one row onto another can drop the hover fill when the new row's enter runs first. Entering from outside the list still works. Same shape as the previous row.
- Minor: `apps/apple/Sokosumi/Chat/Threads/RoomThreadOverviewView.swift:97` — the read line is a `LocalizedStringKey` (`Text("Started by \(starter) · …")`). The unread line passes the name as `Text(verbatim:)`. Both show the name. Only the read line would pick up a catalog string later.
- Minor: `apps/apple/SokosumiTests/Chat/Threads/RoomThreadOverviewGroupsViewTests.swift:30` — a Vision failure returns and skips the copy checks. CI still runs the pixel checks. `RoomThreadOverviewGroupsTests` covers the split.
- Skipped / disagreed: A failed first page leaving loaded rows, the self-loading older-page control, and the error and loading copy are recorded in the PR and in PARITY as a later row. Native section headers instead of web's uppercase caption and rule are the recorded Mac deviation. The parent `.foregroundStyle(.secondary)` does not wipe the "N new" tint; the committed render shows the accent.

### Assessment
The Unread and Earlier split follows web's `unreadReplyCount > 0` partition, the row lines match the audited panel, and Xcode test plus Swift lint are green on this head.

mrosberghaus:
Re the Grok review above:

- **Hover dropped when moving between rows:** fixed in `886856dcd`. `onHover` now sets `hoveredId` on enter and clears it only if the leaving row still owns it (`else if hoveredId == item.parentMessage.id`), so an enter that arrives before the previous row's exit survives. The same shape existed before this PR; fixed here anyway. There's no hosted test: the order of enter and exit events isn't something the test host can drive reliably. The macOS build passes and SwiftLint/SwiftFormat are clean on the file.
- **Read line vs unread line localization:** no change. Both are `LocalizedStringKey` format strings. The unread line is `Text("\(Text(...)) · \(Text(verbatim: starter))")`, and its outer literal is a key just like the read line's `Text("Started by \(starter) · …")`. `Text(verbatim:)` only keeps the bare name from being looked up as a key itself; a `String` interpolated into a key is already a `%@` argument. Either line picks up a catalog entry the same way.
- **Vision failure skips the copy checks:** no change. This is the known CI limitation recorded in PARITY (the GitHub macOS runner has no Vision text recognition; `recognizedLines` returns nil). The pixel and tint checks still run on CI, `RoomThreadOverviewGroupsTests` covers the split and heading rules, and the copy checks pass locally.
- **Skipped items:** agreed, no action. They're recorded as deviations or the later row in PARITY.

PR #5136, opened 2026-09-23
feat(notifications): show account and push hints as rows on Needs you

mrosberghaus:

- The **account notice** (verify your email, credits running low, credits used up) and the **push hint** now show only on the **Needs you** tab. This applies to the bell panel and the `/notifications` page, which phones use too. They no longer sit above every tab.
- They are **rows of the list**, not cards above it. They use the same rail, icon circle and text column as a notification row, so Needs you reads as one list.
- **Fill shows urgency. Colour stays the notice's tone.**

  | Hint | Row | Colour |
  | --- | --- | --- |
  | Email unverified | Filled | Warning |
  | Credits used up | Filled | Destructive |
  | Credits running low | Marked (rail + medium title) | Warning |
  | Push (enable, blocked, install, stopped) | Quiet | Neutral |

- An open account notice **counts +1 on the Needs you tab**, so a reader on All still sees something is waiting. The push hint does not count. It's an offer, and a blocked or iPhone browser would carry the count for good.
- The **bell dot is unchanged**: it still lights for an open account notice in the notice's tone.
- All and Unread now show their normal empty state while a notice is open, because the notice isn't on those tabs.
- A row's action sits beside the text when there is room. In the narrow panel, a long label drops under the text instead of squeezing it. No copy changed.

## How

- New `NotificationRequestRow` (`components/notifications/notification-request-row.tsx`): the shared row with the four emphasis styles.
- `AccountNoticeRow` and `NotificationBrowserPermissionPrimer` render through it. The old card styling and the unused `className` props are gone.
- New `NotificationNeedsYouRequests` gates both rows on the Needs you view. It is the one place both frames use.
- The view filter adds the notice to the Needs you count. The list keeps its own `needsActionCount` for paging.

## Verification

- `pnpm --filter web test`: 851 files, 7483 passed
- Island test (`notification-center.island.test.tsx`) covers both frames: hints only on Needs you, the notice counted on the tab, no "Nothing needs you" under a notice, and All empty under a notice.
- Visual check: the real panel and page components were rendered for all three notices in dark and light, with the app's compiled CSS.


mrosberghaus:
## Grok review

### Fixed
- Important: Needs you rows clipped on a narrow phone. The text column could not shrink below 12rem, and sm buttons kept long labels on one line, so German and Spanish actions (`Bestätigungs-E-Mail erneut senden`, `Reenviar correo de verificación`) were cut off by the card. The 12rem floor now yields, and those actions wrap. (`dbe32c718` — `apps/web/src/components/notifications/notification-request-row.tsx`)

### Remaining
- Minor: `apps/web/src/app/(app)/components/account-notice-action.ts`: the comment still says the notifications page notice card always hosts the resend button. That button is only on Needs you.
- Minor: `apps/web/src/components/notifications/__tests__/notification-center.island.test.tsx`: the row and primer are mocked, so filled versus marked emphasis is not asserted.
- Skipped / disagreed: toast and panel "View" still open `/notifications` on the current view, usually All — the spec keeps the notice off All and uses the Needs you count as the signal.
- Skipped / disagreed: the push hint does not add to the count and stays off All and Unread — the spec says a blocked or iPhone browser would carry that count for good.
- Skipped / disagreed: low-credits title stays default color with a warning rail and circle — that matches "marked (rail + medium title)".

### Assessment
Needs you is the only place the account notice and push hint render, the notice adds one to that tab, and the phone clip is fixed.

mrosberghaus:
Follow-up to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5136#issuecomment-5796079406) in `76227caed`:

- **Phone clip fix (`dbe32c718`):** kept as is.
- **Stale comment in `account-notice-action.ts`:** fixed. It now says the resend button lives on the Needs you row, and names who puts the page on Needs you.
- **Filled vs marked not asserted:** covered. `getAccountNoticeEmphasis` is exported and tested in `account-notice-row.test.tsx`: email → `filled-warning`, no credits → `filled-destructive`, low credits → `marked`.
- **Toast "View" opens `/notifications` on All:** fixed. I disagree with skipping this one. The spec keeps the notice off All, so landing on All shows neither the notice nor the resend button, which is the whole point of that link. `useAccountNoticeAction` now switches the view to Needs you before navigating for email verification. Credit notices go to billing and leave the view alone. Test: `use-account-notice-action.test.tsx`. The panel's "View" needed no change: the row only renders on Needs you, and the page shares the panel's view.
- **Push not counted / low-credits title colour:** agreed, no change.

Verification: `pnpm check`, `pnpm typecheck`, and the notification and account-notice tests (62 files, 475 tests).

PR #5137, opened 2026-09-23
feat(chat): keep pinned rooms one click away under All unreads (SOK-1159)

mrosberghaus:
Follow-up to #5133 (SOK-1159), which is merged. `main` is merged in, so this diff holds only the Pinned change.

## Why
With All unreads on, reaching a read pinned room meant switching the filter off and back on.

## What
- Under the filter, **Pinned sits on top** and holds every pinned room in your own order (`data-slot="unread-inbox-pinned"`, named by its label). This mirrors the full sidebar.
- **Pinned rooms appear only there.** The flat list never takes one, whether it's unread, open, or read during the pass. Nothing is listed twice, and opening a pinned room doesn't pull it out of Pinned.
- An unread pinned room shows bold with its count; a read one is dimmed.
- "Caught up" still waits for unread pinned rooms. Once they're read, Pinned stays above "All caught up".
- There's no setting.

## Verification
- Chat tests: 513 pass. The new tests cover:
  - Pinned holds every pinned room and sits above the flat list.
  - An opened pinned room stays in Pinned.
  - "Caught up" shows when only pinned rooms are left.
- Typecheck and Biome are clean.


mrosberghaus:
## Grok review

### Fixed
- Important: a room pinned during the All unreads pass stayed in the flat list and also appeared under Pinned, because the pass only grows. The flat list now drops pinned ids and keeps that pass place for unpin (`e46652a02` — `apps/web/src/components/chat/organization-chat-list.client.tsx`)

### Remaining
- Minor: `apps/web/src/components/chat/organization-chat-list.client.tsx`: the `canReorderPinned` comment says a filtered reorder would move rooms relative to hidden ones. This filter shows every pin; reorder stays off because that control lives on the full sidebar section.
- Minor: `apps/web/src/components/chat/__tests__/organization-chat-list-unread-filter.test.tsx`: "keeps an opened pinned room" sets the path before the filter is turned on, so it does not cover a later navigation.
- Minor: no test that an unread pinned room alone blocks "Caught up". `unreadRooms` still includes pinned rooms, so the banner waits.
- Skipped / disagreed: reorder under the filter — the spec says the list is fixed in the reader's order and there is no setting.
- Skipped / disagreed: thread-only unread does not bold the row — same rule as every other row (thread unread does not bold).
- Skipped / disagreed: a collapsed Pinned section still shows every pin under the filter — the spec says every pinned room and no setting.

### Assessment
Pinned stays on top, in pin order, and a room pinned mid-pass is no longer listed twice. Leftover notes are a stale comment and two test gaps.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5137#issuecomment-5796158769)

- **Pinned mid-pass fix (`e46652a02`):** checked. The flat list filters out pinned ids after assembling the pass, so a room pinned during the pass leaves the list and returns to its old place if it's unpinned.

The three minors are fixed in 6b9eee4f8:
- **Stale `canReorderPinned` comment:** fixed. It now says that reordering lives on the full Pinned section's header, which the filter replaces.
- **Opened pinned room test:** it now turns the filter on first and then navigates to the pinned room, as a click from Pinned would. Without the pinned-id filter, the room would appear in the flat list and the test would fail.
- **A lone unread pinned room:** new test. With a single unread pinned room, it shows in Pinned, the flat list is empty and there is no "All caught up".

I agree with the three skipped items; each follows the spec as written: no reorder under the filter, the thread-unread rule for bold, and every pin shown.

Chat tests: 515 pass. Typecheck and Biome are clean.

PR #5138, opened 2026-09-23
feat(tasks): skip, move, and restore single Task Schedule Runs (SOK-1167)

mrosberghaus:
Ticket 03/10 of [SOK-1164](https://linear.app/masumi/issue/SOK-1164) (ADR 0041): [SOK-1167](https://linear.app/masumi/issue/SOK-1167/task-schedule-0310-change-single-occurrences-via-the-api). Stacked on the SOK-1164 branch (umbrella #5127); do not merge the umbrella before ticket 10.

## What changes

People and Coworker vendors can list a Task Schedule's Runs and skip, move, or restore one upcoming Run without changing the rule.

- `GET /v1/tasks/schedules/{id}/runs`: Runs in time order, paged, optional `from`/`to`. Released, skipped, moved, planned, and canceled rows (the schedule's history).
- `PATCH /v1/tasks/schedules/{id}/runs/{runId}` with `skip`, `move` (`scheduledAt`), `restore`. Only future, unreleased Runs of an Active schedule inside the 90-day horizon; checked against `expectedRevision`. `restore` puts a skipped or moved Run back at the rule's time.
- Exceptions live on the Run row (state, effective vs original time, actor), not as Task events. The row gains `actorCoworkerId` so a vendor's change keeps its actor.
- End-after-N: a skip frees its place (one more Run is planned), a restore takes it back.
- Rule edit and end cancel upcoming skips and moves into the history instead of dropping them; pause keeps them for the resume; resume cancels a move whose time passed while paused.
- The projection now runs from now and passes over every rule time that already has a Run (inserts stay limited to missing rows).
- A Run change claims the schedule by revision and release count, so a release committed mid-request is a 409, never an extra Run.

Not carried over from the per-series route (deliberately): `operationId` replay (a retry gets 409), the open-project guard, immediate calendar invalidation delivery and the member notification. The DB trigger still enqueues calendar invalidation; calendar on Task Schedules is ticket 6.

## Naming: Occurrence is now Run

A Task Schedule's points in time are **Runs** (`CONTEXT.md`, ADR 0041). The rename covers all Task Schedule code from tickets 01–03: routes, schemas (`TaskScheduleRun`, `UpdateTaskScheduleRunRequest`, `TaskScheduleRunUpdate`), `TaskSchedule.nextRunAt` and `rule.targetRunCount`, error kinds `schedule_run_state_conflict` / `schedule_run_target_invalid`, services, functions, variables, and tests. The SOK-1164 migrations are edited in place (`20260923130000_task_schedule`, `20260923160000_task_schedule_runs`), since none has been merged. The shared ledger (`TaskScheduleOccurrence` model, `task_schedule_occurrence` table) and the old per-Task schedule code keep their names until the SOK-1172 cutover rewrites them.

## Schema change

- `packages/database/prisma/migrations/20260923180000_task_schedule_run_actor_coworker/migration.sql`: nullable `task_schedule_occurrence.actorCoworkerId` (FK `coworker`, SetNull, indexed) and an at-most-one-actor check. Applied cleanly on a throwaway Postgres 17; no drift for the new column.

## Verification

- Mocked-Prisma route tests beside both routes (skip/move/restore, refusals for released, past, outside horizon, stale revision, non-Active schedule, canceled row, release race, access: private schedule, non-owner, unseated member, Coworker grant, Soko Bot), plus pause/resume/end/rule-edit cases.
- Release service: a moved Run releases at its new time, a skipped one never releases, a schedule whose last Run was skipped Ends.
- `pnpm --filter web generate:core:snapshot`, `pnpm typecheck`, `pnpm check`, `pnpm test` green.


mrosberghaus:
## Grok review

Reviewed `7cda744a68a510986eb32f23627d10ed41c37229..44b86d3cd9cc799c8350b297ff00640c19c49d43` (merge-base with `sok-1164-feattasks-move-recurring-rules-into-task-schedule`).

### Fixed

### Remaining
- Minor: `apps/core/src/schemas/task-schedule.schema.ts` — `TaskScheduleRun.state` says CANCELED is "dropped by a rule edit". End, and resume of a move whose time passed while paused, cancel runs too.
- Minor: `apps/core/src/services/task-schedule.service.ts` — if the run row loses the update before the schedule claim, the 409 is `concurrency_conflict` ("Task Schedule changed since it was read"), not `schedule_run_state_conflict`. A stale `expectedRevision` caught on the first read is still `schedule_revision_conflict`.
- Minor: migration directory renamed `20260923160000_task_schedule_occurrences` → `20260923160000_task_schedule_runs`. A database that already applied the old folder name from the base branch will try to apply it again.
- Skipped / disagreed: no `operationId` replay, no open-project guard, and no synchronous calendar fan-out — the PR states those are deliberate; the occurrence trigger still enqueues invalidation.
- Skipped / disagreed: a restored run keeps `actorUserId` / `actorCoworkerId`. The run schema defines those as who last restored it. ADR "leaves no trace" is the skip itself not staying in history (state and time match the rule again).
- Skipped / disagreed: after the last in-horizon run of an end-on-date schedule is skipped, `nextRunAt` stays on that skipped time. That is the wake-up so release ends the schedule, and tests lock it in. Clients must not treat `nextRunAt` alone as "a Task will be created".

### Assessment
Skip, move, and restore match the ticket: horizon and revision checks, end-after-N give-back, pause keeps exceptions, rule edit and end cancel them, and a mid-request release conflicts instead of planning an extra run.

mrosberghaus:
Addressed in `ec8d28025`:

- **CANCELED description** — fixed. `TaskScheduleRun.state` now names all three paths: a rule edit, ending the schedule, and a move whose time passed while the schedule was Paused.
- **Run lost before the schedule claim** — fixed. When the Run row fails its conditional update, the response is now `409 schedule_run_state_conflict` ("This Run changed since it was read"). `concurrency_conflict` is kept for a schedule claim lost to a release. Covered by the new test "a Run a release took between the read and the write".
- **Renamed migration folder** — keeping it. Editing the SOK-1164 migrations in place is deliberate: nothing from SOK-1164 has been merged, and Andreas asked for the Occurrence → Run rename to replace those files. A database that applied the base branch's migrations has to be reset either way, because `20260923120000_task_schedule` was also edited in place (`nextRunAt`, `targetRunCount`). Prisma would not re-apply it, so that database would keep the old column names whether or not the folder was renamed. The full chain applies cleanly to an empty Postgres 17, with no drift on the Task Schedule tables.
- **Skipped items** — agreed on all three, and the rationale is in the PR description. A restored Run keeps its last actor. `nextRunAt` stays on a skipped last Run so that the release Ends the schedule on time.

PR #5139, opened 2026-09-23
test(web): keep the better-auth client out of the composer test

mrosberghaus:
## Why

[Test Web failed](https://github.com/masumi-network/sokosumi/actions/runs/35866902456) with every test passing but one unhandled error from `room-session-composer.test.tsx`:

```
ReferenceError: window is not defined
  at cleanupBroadcastSetup (better-auth/dist/client/session-refresh…)
  at better-auth/dist/client/session-atom.mjs:168
  at nanostores/lifecycle/index.js:139 (_onTimeout)
```

Import chain: `room-session-composer.test.tsx` → `room-session-composer.tsx` → `room-composer.tsx` → `components/drive/drive-file-picker.tsx` → `useSession` from `@/lib/auth/auth.client`.

Rendering the picker subscribes better-auth's session atom. On unmount, nanostores waits `STORE_UNMOUNT_DELAY` (1s) before running the atom's cleanup, which reads `window`. If happy-dom has torn down the environment by then, the timer throws. It depends on CI timing and does not reproduce locally.

## What

A colocated `vi.mock("@/lib/auth/auth.client")` in the composer test returns `useSession: () => ({ data: null })`, the value the unauthenticated real client gave. No session atom, no timer.

After review, the same mock was added to the sibling tests that also render `RoomComposer`: `room-session-composer.typing`, `room-composer-focus` and `room-composer-quote-preview`. `room-composer-overflow` already mocks the picker.

A shared setup mock was ruled out. With `importOriginal` it still loads the real client. A full stub would have to cover `signIn`, `authClient.organization` and every other export that other tests rely on.

## Verification

- `pnpm --filter web test`: 851 files, 7486 passed, 1 skipped
- Real client proven unloaded: with a top-level `throw` added to `auth.client.ts`, all 5 composer test files still pass (37 tests). Change reverted.
- `biome check` on the file and `pnpm --filter web typecheck` are clean

## Follow-up (not in this PR)

The same rig shows 13 other web test files still load the real client, e.g. `markdown-editor`, `task-activity`, `workspace-calendar` and the chat sidebar tests. The ones that render `useSession` could hit the same timer.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/app/(app)/chat/components/room-session-composer.typing.test.tsx`, `apps/web/src/app/(app)/chat/components/__tests__/room-composer-focus.test.tsx`, `apps/web/src/app/(app)/chat/components/__tests__/room-composer-quote-preview.test.tsx`, and `apps/web/src/app/(app)/chat/components/__tests__/room-composer-overflow.test.tsx` still render the composer. `DriveFilePicker` still calls the real `useSession`, so the same nanostores unmount timer can fire after happy-dom drops `window`. This change leaves that as follow-up.
- Skipped / disagreed: a shared setup mock of `@/lib/auth/auth.client` — `importOriginal` would still construct the session atom, and a full stub would have to cover every export other tests use. The colocated mock is the right scope.

### Assessment
The mock stops this file from loading the better-auth session atom, which is the failure in the cited Test Web run.

mrosberghaus:
Re: the Grok review above.

**Minor (sibling composer tests):** fixed for 3 of the 4 files in 708c4a193.

- `room-session-composer.typing.test.tsx`, `__tests__/room-composer-focus.test.tsx` and `__tests__/room-composer-quote-preview.test.tsx` render `RoomComposer`. It always mounts `DriveFilePicker` (`room-composer.tsx:888`), so the real `useSession` subscribed there. Each now has the same colocated `useSession: () => ({ data: null })` mock as the composer test.
- `__tests__/room-composer-overflow.test.tsx` needed no change. It already mocks `@/components/drive/drive-file-picker` to `() => null` (line 98), so `useSession` never runs there.

**Verification**
- With a top-level `throw` added to `auth.client.ts`, all 5 composer test files pass (37 tests), so none of them loads the real client. The `throw` was reverted.
- `pnpm --filter web test`: 851 files, 7486 passed, 1 skipped.
- Biome is clean and the commit hook passed.

**Shared setup mock:** agreed, and none was added.

The other web tests that still load the real client (calendar, sidebar, task activity, markdown editor and others) stay as a separate follow-up.

PR #5141, opened 2026-09-23
test(web): mock the auth client in tests that render useSession

mrosberghaus:
## Summary

Follow-up to #5139, which mocked the auth client in the four composer tests. Five more web test files render a component that calls the real better-auth `useSession`. Its session atom schedules a nanostores unmount timer (1s) that can fire after happy-dom tears down `window`, which fails CI with `ReferenceError: window is not defined at cleanupBroadcastSetup`.

Each file now has a colocated `vi.mock("@/lib/auth/auth.client", () => ({ useSession: () => ({ data: null }) }))`. `{ data: null }` is the unauthenticated state the real client already produced in these tests. Every consumer in these graphs gates on `data`; `ProjectsMenuItem` also reads `isPending`, `isRefetching` and `error`, but its scope guard checks `session` first, so a null session bails before those flags matter.

| Test file | Component calling `useSession` |
| --- | --- |
| `markdown-editor`, `task-activity` | `DriveFilePicker` |
| `app-mobile-chrome` | `AppMobileBottomChrome`, `ChatMobileBottomNav`, `useLiveChatRooms` |
| `sidebar-row-slot-contract` | `ProjectsMenuItem`, `useShowRoomUnreadCount` |
| `app-shell-loading-frame` | mobile chrome, `ProjectsMenuItem` |

No global mock in `src/test/setup.ts`: an `importOriginal` spread still loads the real client, and a full stub breaks tests that use `signIn` or `authClient.organization`.

## Not changed

These 9 files import the client but never render `useSession`, so no timer starts: `workspace-calendar`, `workspace-calendar.week-layout`, `resolve-credit-usage`, `tasks/[taskId]/page`, `chat-unread-nav-rows-filter-hover`, `chat-sidebar-section-header`, `chat-unread-nav-rows`, `pending-invitation-rail-button`, `pinned-rooms-dnd`.

## Verification (on `main` with #5139 merged)

- A temporary wrapper that logged each `useSession` render per test file: no test file renders the real hook any more.
- Throw rig (`throw` at the top of `auth.client.ts`): 10 files fail, the 9 import-only files above plus `auth.client.test.ts`.
- `pnpm --filter web test`: 851 files, 7491 tests passed.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `projects-menu-item.tsx` reads `isPending`, `isRefetching`, and `error` as well as `data`, so the PR note that every consumer reads only `data` is slightly off. `{ data: null }` still matches logged-out: the scope guard bails before those flags. Same stub as the composer tests. No code change.
- Skipped / disagreed: none. A shared or global auth mock would still load the real client (`importOriginal`) or break `signIn` / `authClient.organization` tests. Colocated stubs are the right cut.

### Assessment
Ready to merge: Yes
Five colocated `useSession` stubs stop the nanostores unmount timer in the trees that actually render the hook, and Test Web is green on `7bc3e187`.

mrosberghaus:
Minor finding: fixed in the PR description. It now says `ProjectsMenuItem` also reads `isPending`, `isRefetching` and `error`, and that `session && …` in its scope guard short-circuits on a null session, so `{ data: null }` still yields `scope = null`, the same result as the real logged-out client. No code change.

PR #5142, opened 2026-09-23
fix(chat): put Pinned below the All unreads list and caught-up message (SOK-1159)

mrosberghaus:
Follow-up to #5137 (SOK-1159).

## Why
- **Pinned on top:** in #5137, Pinned sat on top under the filter. Once you were caught up, you saw a column of dimmed pinned rooms first, with "All caught up" pushed down below them.
- **Dimmed open room:** the room you're viewing was dimmed together with its highlight, so it read as disabled.
- **Invitation position:** a pending invitation sat under Pinned's read rooms, even though it waits on you like an unread room.

## What
Under the filter, the order is now:
- **While unread:** the flat unread list, then External (pending invitations), then Pinned.
- **When caught up:** "All caught up", then "Read just now", then Pinned.

The room you're viewing stays at full strength even once it's read, whether it's in the flat list or in Pinned. It dims in place when you move to another room.

Pinned's contents are unchanged: every pinned room, fixed, listed only there.

## Verification
- Chat tests: 516 pass. The tests cover:
  - Order: Pinned comes after the flat list and after the caught-up message, and a pending invitation comes before Pinned.
  - The open room stays undimmed, then dims after you leave it.
- Typecheck and Biome are clean.
- Before and after mocks were rendered from the list's real markup with the compiled CSS.


mrosberghaus:
## Grok review

### Fixed
- Important: a read room stayed faded under the selection bar until the route committed. The highlight follows optimistic selection; dim followed `usePathname`. Dim now uses the same room id (`96135d484` — `apps/web/src/components/chat/organization-chat-list.client.tsx`, `active-room-id.ts`).

### Remaining
- Minor: `organization-chat-list-unread-filter.test.tsx`: the invitation test does not include an unread flat-list room, so "flat list, then External, then Pinned" is not locked together.
- Minor: no test that "Read just now" sits between the caught-up message and Pinned when both exist. The render order does that.
- Skipped / disagreed: Apple sidebar has no All unreads filter, so it is out of scope. The render-time `setFilterPass` update is pre-existing and unchanged.

### Assessment
Pinned stays last (flat list or caught-up message, then External, then Pinned), and the open read room undims with the highlight instead of after the route commits.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5142#issuecomment-5796809404)

- **Dim from the selection (`96135d484`):** checked. `activeRoomId` now comes from the optimistic room selection (`useRoomSelection`), the same source as the row highlight. It falls back to `usePathname`, so the highlight and the dimming can't disagree while a route commits.

The two minors are fixed in 8d60fe92e:
- **Invitation order:** the test now includes an unread room, so it checks the flat list, then External, then Pinned together, in document order.
- **Caught-up order:** new test. After the last unread room is read, it checks "All caught up", then "Read just now", then the read room, then Pinned.

I agree with the skipped items:
- The Apple sidebar has no All unreads filter.
- The render-time `setFilterPass` update is unchanged.

Chat tests: 523 pass. Typecheck and Biome are clean.

PR #5144, opened 2026-09-23
feat(apple): load older threads on scroll and show paging failures like web

mrosberghaus:
Apple parity row **24e**: the thread overview pages and fails the way web's `ThreadListPanel` does. Also marks **24d** Done ([#5135](https://github.com/masumi-network/sokosumi/pull/5135)) and makes **24f** next in the work order.

## What users see
- The row at the end of the thread list loads the next page by itself once it scrolls into view, showing a spinner with "Loading threads…". It is still a button you can click.
- A failed older page stays on that row: "Could not load threads. Try again." in red over "Load older threads", which retries. The loaded rows, headings and "All caught up." stay, and nothing reloads by itself until you click.
- A failed first page (on open, after Back, or in the reload after Mark all) replaces the list with Core's message or "Could not load threads. Try again." and a Retry button. No rows, headings, Mark all or paging row are left.
- A failed Mark all keeps the rows under Core's message or "Could not mark unread threads as read. Try again.".
- While Mark all runs, its button reads "Loading threads…", as on web.

## Web audit (at `1d2c25c37`)
`thread-list-load-more.tsx` watches itself with `useLoadWhenVisible`: an `IntersectionObserver` with default options, armed only while idle, re-armed when the last thread changes, firing once per arming. `thread-list-panel.tsx` keeps `olderFailed` apart from the list's `error`. A failed first page clears `items` and `nextCursor`; a failed Mark all keeps them. Web itself is consistent, and no API, Ably, project, entitlement or dependency change was needed. File and line references for each answer are in PARITY's "Slice 24e" section.

## Changes
- `SokosumiChat/RoomThreadOverview`: `isLoading` now covers the first page only. The paging row has its own `olderPageStatus` and `loadsOlderAutomatically`. `failure: Error?` becomes `failureMessage: String?` with web's clearing rules and copy.
- The transcript's `TranscriptBoundaryRow` becomes the shared `Shared/PageBoundaryRow`, which takes its copy per list. `TranscriptBoundaryStatus` becomes `PageBoundaryStatus`. The transcript's strings and identifiers are unchanged.
- `RoomThreadOverviewView` uses the row: `onScrollVisibilityChange` feeds an `onChange` arming, with a main-actor hop before `older`.
- Kept native (recorded in PARITY):
  - a Retry under a failed first page;
  - the tinted action label;
  - Mark all waits for a page in flight;
  - the old rows stay visible while Back reloads.

## Verification
- Tests were written first. 5 Chat tests (7 cases) were red against the old behaviour, then green. 6 hosted render cases and both automatic-load tests were red on the old view, then green.
- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`: on the base, 1170 tests passed on the first run; after merging main (`6ed3a0baa`), 1181 of 1182 passed, and the one failure was the known `HistoryGapRowTests` flake, which passed on rerun (2 of 3 iterations)
- `swift test --package-path Packages/SokosumiChat`: 743 passed.
- The macOS build and the iOS 17 `SokosumiWorkspace` build both succeed.
- `mint run swiftformat --lint .` and `mint run swiftlint lint --strict` report 0 issues.

Renders in light and dark: loading older, failed older page, failed first page, Mark all running.

thread overview paging states

Unverified: the app was not launched. Not exercised:
- paging against Core by scrolling;
- a real failed page;
- VoiceOver on the row and on the failure announcement;
- a click on the failed row in a real window.


mrosberghaus:
## Grok review

Read-only review of `100eadea68788a82ba73eab91a2f70f9aae4b510..7e448025f3014d8885753a278016044539ece113` against the PR body and web `ThreadListPanel` / `useLoadWhenVisible`. No Critical or Important findings. `swift test --package-path apps/apple/Packages/SokosumiChat --filter RoomThreadOverviewTests`: 16 tests passed.

### Fixed

### Remaining
- Minor: `apps/apple/Sokosumi/Chat/Threads/RoomThreadOverviewView.swift`: the paging row disables only while `olderPageStatus == .loading`. During a first-page load or Mark all it still reads "Load older threads" and accepts a click; `RoomThreadOverview.load` then returns without a request. The previous button was disabled for `isLoading` and `isMarkingRead`. The idle label during Mark all matches the renders and web's copy.
- Minor: `RoomThreadOverviewView.swift`: `OlderPageArming` includes visibility, so an idle row that scrolls out and back asks again without a new last thread or a new idle transition. Web's observer fires once per `armed` / `boundaryKey` arming. A second call is dropped once `.loading` is set.
- Minor: `apps/apple/Packages/SokosumiChat/Sources/SokosumiChat/RoomThreadOverview.swift`: `message(for:)` keeps Core's text only for `.unprocessable`. A 401 shows the fallback; `updateThreadOverview` still signs out.
- Skipped / disagreed: Retry under a failed first page, rows staying while Back reloads, and Mark all waiting for an in-flight page — kept on purpose in the PR body and PARITY, not defects. An older-page failure using the fixed "Could not load threads. Try again." string, not Core's message, matches web. Threshold `0.01` matches the transcript row.

### Assessment
First page, older page, and Mark all failures are split the way web splits them, and a failed older page stays on the row until a click.

mrosberghaus:
Re the Grok review above:

- **Paging row clickable during a first-page load or Mark all:** fixed in `f381fe1c9`. The overview now passes `.disabled(overview.isLoading || overview.isMarkingRead)` to its `PageBoundaryRow`, as the old button did. `RoomThreadOverview.load` already dropped the request, so nothing changes in behaviour; the row just stops offering a click that did nothing. Automatic loading doesn't go through the button, so it's unaffected. `RoomThreadOverviewPagingViewTests` and `RoomThreadOverviewGroupsViewTests` pass (18 cases), and lint is clean. There's no new test: in both states the observable outcome, no request, is the same as before.
- **Visibility re-arms the automatic load:** no change. It re-fires only while the row is idle, visible, still has a cursor and the last thread hasn't changed. That happens only when an earlier trigger was dropped (a first-page load or Mark all was running). Asking again then is what we want. It avoids the stuck state this slice found on web, where an older page started during Mark all's reload can leave `isLoading` set. `.loading` rules out a duplicate request.
- **`message(for:)` keeps only `.unprocessable` text:** no change. `.unprocessable` is the case that carries Core's message. `.unauthorized` signs the user out (`updateThreadOverview` → `signOutIfUnauthorized`), so no list copy is ever seen. `.unexpectedResponse` strings are the client's own diagnostics, not Core copy, so web's fallback wording is right there.
- **Skipped items:** agreed, no action. They're recorded as deviations in the PR and PARITY.

PR #5145, opened 2026-09-23
feat(chat): shrink All caught up to one animated row (SOK-1159)

mrosberghaus:
## What changes

Follow-up to #5142 for SOK-1159.

**All caught up is one row now.** The centered block (icon circle, heading, "Nothing unread in 30 chats") took about 130px above Pinned. It is now a single sidebar row: tinted double check and "All caught up". The chat count is gone: it was every room the reader is in, not something to act on, and next to the read rooms it looked like an unread count.

**"All unreads" is now "Unreads"** (de "Ungelesene", es "No leídos"). Stacked above "All caught up" it said All twice, and the filter only ever covers the reader's own rooms. It now pairs with the Threads row above. Not "Unread": that is already the group's screen-reader label.

**The caught-up copy is aligned across the app.** The phrase stays; it comes in two forms. "You're all caught up" heads a full empty state (Notifications, Threads page) and "All caught up" stands in a row or list (this row, the thread panel's Unread group), with no full stop either way. German used three phrasings for the one state and now has "Du hast alles gelesen" / "Alles gelesen". The Threads page's two curly apostrophes become straight like the rest of `en.json`, and the Mac thread overview drops its full stop with web (string, test fragment and the PARITY row).

**Everything around it animates** (200ms, one easing, off under reduced motion):

- A new unread room slides into the list, and the caught-up row and its "Read just now" label collapse at the same time. With only the caught-up row showing, the room takes its place and nothing below moves. With the label showing too, what is below moves up by the label's height, and slides rather than jumps.
- Reading the last unread brings the caught-up row and label back in.
- A room read during the pass eases into its dimmed state. Coming back to full is instant, because that means new activity.
- Switching All unreads on or off fades the new list in. It does not fade on page load.

Before, an arriving unread removed the ~130px block in one frame and everything below jumped up about 100px, often under the pointer.

## How

- `ChatRoomSidebarRow` takes an optional `itemMotion` and renders its `<li>` as a motion component only when given one. Every other sidebar row stays a plain `<li>`.
- The filter's list stays mounted for the whole pass, even when empty, so `AnimatePresence initial={false}` animates rooms arriving in an empty list but not the rows already there when the filter comes on.
- Rooms animate height only. A read room's dimming is a class on the same `<li>`, and a finished opacity animation would pin it at full.
- `UnreadNav.nothingUnread` is removed in en, de and es.

## Verification

- `vitest run src/components/chat src/components/notifications "src/app/(app)/chat"`: 188 files, 1997 tests pass; `messages:parity` ok for de and es
- Apple: only the overview's string and its Vision test fragment changed; not run locally, left to CI (`Xcode test`, where Vision does not run)
- Web typecheck and Biome pass (pre-commit hook)
- Real `OrganizationChatList` in a throwaway `/share` page (not committed), with every element's position and opacity recorded frame by frame in headless Chromium:

| Transition | Result |
| --- | --- |
| Unread arrives while caught up | Room grows into the caught-up row's slot; Pinned holds at the same pixel every frame |
| Last unread read | Caught-up row and label grow back, the room eases to 0.6, the list below slides down 72px |
| Unread in a pinned room | Caught-up row and label collapse, rows below slide up 52px |
| Second arrival | Rows below slide down 44px |
| Room pinned during the pass | Its row collapses out of the list |
| Filter on or off | New list fades in over ~175ms; rows already listed don't grow in |
| Reduced motion (from page load) | Every transition is instant, dim included |

The review found one bug, fixed in the second commit: a bare `duration-200` next to `motion-safe:transition-opacity` left the transition property at its default of `all`. Read rooms still eased under reduced motion, and every property on those rows transitioned.

Known limit: motion's `useReducedMotion` reads the setting at mount, so switching the OS setting while the page is open takes effect on the next load.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/web/src/components/chat/organization-chat-list.client.tsx` — `aria-labelledby` is set whenever caught up, but the "Read just now" id is mounted only when the flat list has rows. Caught up with an empty list (including pinned-only) points at a missing id.
- Minor: same file — on the collapsed rail, caught up with no inbox rows, the list wrapper stays in the tree because `:has([role=status])` matches a `display: none` status. It is 0px tall and not `display: none`, so `space-y-2` adds about 8px before the next rail icon. Expanded layout is fine: an empty list with no status is `display: none`, and Tailwind 4's spacing margin is on that box.
- Minor: height uses `cubic-bezier(0.2, 0, 0, 1)` via `MotionConfig`. The dim and the filter fade use Tailwind's default ease. Duration is 200ms on all three.
- Minor: a new unread while "Read just now" is showing still shifts what is below by the label height. The pixel-stable case is an empty caught-up row replaced by one room. That matches the measured table, not the sentence that includes the label.
- Skipped / disagreed: `useReducedMotion` in motion 13.4.1 stores the media query in `useState` and does not subscribe. The PR's "next load" limit is accurate. Not a defect.

### Assessment
The caught-up row, presence height, reduced-motion classes, and removed count match the requirement. Nothing here blocks merge.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5145#issuecomment-5797643110). All four minor notes were valid. Fixed in f5c3947d0:

- **`aria-labelledby` pointing at a missing id:** the list and the "Read just now" label now share one condition (`showsReadLabel = caughtUp && inboxRooms.length > 0`). A new assertion in `organization-chat-list-unread-filter.test.tsx` covers caught up with an empty list. It fails without the fix.
- **8px gap on the collapsed rail:** the wrapper now also hides on the rail when it holds no room (`group-data-[collapsible=icon]:[&:not(:has(li))]:hidden`). I compiled the class with the app's Tailwind; it emits `:is(:where(.group)[data-collapsible="icon"] *):not(:has(li)) { display: none }`.
- **Easing mismatch:** the dim and the filter fade now use `motion-safe:ease-[cubic-bezier(0.2,0,0,1)]` with `motion-safe:duration-200`, the same curve and length as the presence animations. The curve was already in use in `room-search-panel.tsx`.
- **Description overstated the stable case:** corrected. The description now says nothing moves only when the caught-up row is replaced on its own. With "Read just now" showing, the list below slides up by the label's height.

`useReducedMotion`: agreed, no change.

Also merged `main` (1f0bbba53). The only conflict was the 24d row in `apps/apple/PARITY.md`: I kept main's Done status and new 24e row, with "All caught up" (no full stop) to match the aligned copy. The 1997 web tests in the chat and notifications folders pass on the merged tree.

PR #5147, opened 2026-09-23
test(web): run non-DOM unit tests in node, not happy-dom

mrosberghaus:
## Why

Web Vitest currently boots happy-dom for every file, including hundreds of pure unit and source-scan tests that never touch `window`. That is wasted setup cost.

Vitest 5 removed `environmentMatchGlobs`. The equivalent is two projects in `apps/web/vitest.config.ts`:

- **node** — `src/**/*.test.ts`, excluding files that need a DOM
- **happy-dom** — every `.test.tsx`, plus `.test.ts` files that import Testing Library or otherwise use `window` / `document` / `localStorage`

`src/test/setup.ts` still runs for both; `cleanup()` is skipped when `document` is absent.

## Clarity preserved

- No production code.
- No assertion changes.
- Testing Library / hook `.test.ts` files stay on happy-dom.
- Test Web is not sharded.

## Verification

- `vitest run --project node`: 406 files, 3330 passed
- `vitest run --project happy-dom`: 447 files, 4182 passed (1 existing empty skip)

<div><a href="https://cursor.com/agents/bc-2b3c84e4-5e3e-5fc2-abae-6ebbf441cf74?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2b3c84e4-5e3e-5fc2-abae-6ebbf441cf74&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #5154, opened 2026-09-23
test(web): replace wall-clock sleeps with fake timers

mrosberghaus:
## Why

Web unit tests were paying real 30–300ms `setTimeout` waits for production timers (composer blur, edit-cancel blur, notification leave duration, transcript touch-end / virtualizer settle). Those sleeps add wall-clock to every run and still miss the timer if CI is slow. Fake timers keep the same delays, fire them immediately, and keep the same assertions.

Also deletes the empty skip in `resolver-svg-icon.test.tsx` (`should handle server-side rendering without DOMPurify`) rather than leave a no-op.

## Clarity preserved

- Same `it()` names, same expects. No assertions dropped or loosened.
- Notification row-keep case still unhovers, then advances past leave duration before asserting the row and mark-read control remain. Fake timers start after `userEvent` so pointer setup does not hang.
- Composer blur still waits past the suggestion-dismiss delay before asserting `handleStopTyping`.
- Message-row dirty-blur cases still blur the editor and wait past the deferred `onBlur` timeout before asserting cancel did not fire (`fireEvent.blur` so user-event is not mixed with fake timers).
- Transcript `settle()` still does three rounds of yield + synthetic `scroll`/`scrollend`; the two 200ms waits still cover touch-end grace and virtualizer scroll lapse.
- Empty skip removed; remaining resolver tests unchanged (including the AbortError `setTimeout` in the fetch-timeout case, which is not a wait).

## Verification

`pnpm exec vitest run` on the five files: 5 files, 258 tests passed in 7.90s.

<div><a href="https://cursor.com/agents/bc-2b3c84e4-5e3e-5fc2-abae-6ebbf441cf74?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2b3c84e4-5e3e-5fc2-abae-6ebbf441cf74&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


mrosberghaus:
## Jarvis triage

`/grok review` is posted; no `## Grok review` reply yet. Will pick it up when Grok lands (standing watch).

mrosberghaus:
## Grok review

### Fixed
- Important: the unread row-keep test unhovered before fake timers started, so the 200ms fold stayed on the real clock and advancing timers never ran it. A row that should have left still passed. `pointerout` (what React turns into `onPointerLeave`) now runs after timers are faked, then the advance fires the fold. (`066be4f6a` — `apps/web/src/components/notifications/__tests__/notification-center.island.test.tsx`)

### Remaining
- Minor: `apps/web/src/components/notifications/__tests__/notification-center.island.test.tsx`: the "drops once the pointer leaves" and "keeps paging" cases still wait on the real 200ms timer via `waitFor`. They do fire it. Left alone because `userEvent` hangs once timers are fake.
- Minor: `apps/web/src/app/(app)/chat/components/transcript-viewport.test.tsx`: fake timers wrap the whole file, including `rowHoldAfterRowsChange`. The suite passed (26). Not narrowed.
- Skipped / disagreed: `fireEvent.blur` on the dirty edit rows does reach the editor's deferred `onBlur`. The composer stop-typing spy fires only after that blur, so those waits are not false greens.
- Skipped / disagreed: deleting the empty `it.skip` in `resolver-svg-icon.test.tsx` matches the PR. The skip had no body.

### Assessment
The fold wait now runs the leave timer. The other fake-timer swaps already armed their timers before advancing them.

mrosberghaus:
## Jarvis triage

Verified Grok’s review against `cursor/web-fake-timers-cf74` head (`066be4f6`).

### Fixed (accepted, already on branch)
- **Important — unread row-keep leave timer:** Sound. Unhover-before-fake-timers meant the 200ms fold never ran under the advance, so a “keep” could pass without exercising leave. `pointerOut` after `useFakeTimers`, then `advanceTimersByTimeAsync(300)`, is the right order. No further code change.

### Left alone (agree)
- **Minor — “drops once the pointer leaves” / “keeps paging” still on real `waitFor`:** Agree. Those cases still fire the real leave timer; converting them needs the same `pointerOut` pattern because `userEvent` hangs under fake timers. Out of scope for this draft; not a false green.
- **Minor — transcript suite-wide fake timers including `rowHoldAfterRowsChange`:** Agree leave. Suite already green; narrowing is optional cleanup, not a correctness gap.

### Pushed back / agree with Grok’s skip
- **`fireEvent.blur` dirty-edit / composer stop-typing waits:** Agree these are not false greens. Blur arms the deferred `onBlur`; the spy/assertion only passes after that delay is advanced.
- **Delete empty `it.skip` in `resolver-svg-icon.test.tsx`:** Agree — matches the PR; the skip had no body.

No new cloud-agent work on this PR (attached agent `bc-2b3c84e4…` is currently on #5147). Draft status unchanged.

PR #5157, opened 2026-09-23
feat(apple): add the cross-room Threads view

mrosberghaus:

The Mac app gets web's chat-level **Threads** view (PARITY row 24f1, SOK-1159). A **Threads** row now heads the sidebar, above every section. It is bold while any thread you are part of has unread replies, and shows one number, taken from the rooms alone: the mention badge when an unread reply names you, the muted count of unread threads otherwise, and nothing at zero.

Selecting it replaces the room in the detail column with a reading column titled "Threads":
- **Unread** lists every unread thread across the workspace's rooms, newest unread reply first. Its heading counts them, and "All caught up" stands in when there are none.
- **Earlier** lists the threads you are part of that have nothing unread, newest reply first. It is absent until it holds a row.
- Each row names its room: "2 new · #general" on unread rows, with an `@` mark and pill when a reply names you, and "#design · 4 replies" on read rows.
- Both groups load more as you scroll, using the same paging row as the room's thread overview.

Clicking a row opens that thread in its room at the first unread reply (or at the newest reply for an Earlier row). The thread's normal Look then marks it read. While the Threads view is up, nothing in the room behind it is marked read.

## Why only half of row 24f

Row 24f held two independent features, so this PR splits it into two rows:
- **24f1**, the Threads view (this PR). It needs two new Core reads and nothing else.
- **24f2**, web's Unreads filter (still `Todo`). It filters the room list by each row's own bold, and web takes that bold from ADR 0037 (#5078). Apple has not ported that rule yet, so 24f2 has to bring it over first.

## Core snapshot

I added `get /chats/threads/unread` and `get /chats/threads/earlier` to the snapshot with `scripts/update-core-api.py`, run against a fresh copy of `https://api.sokosumi.com/v1/openapi.json`. Structural diff:
- Operations: those two added, none removed or changed.
- Schemas: `ChatUnreadThread` and `ChatEarlierThread` added.
- `ChatRoom` gained the optional `unreadThreadMentionCount`.

Nothing under `apps/web` or `apps/core` changed.

## Verification

All commands were run from `apps/apple` on `91a005760` plus this change. Every new test failed against a compile-only skeleton before the implementation went in; PARITY lists each one.

- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`: TEST SUCCEEDED on the first run, 1203 tests, 0 failed.
  - Per suite: `SokosumiTests` 216, Chat 765, Auth 35, Realtime 48, Workspace 138, CoreAPI 1.
  - New in Chat: `CrossRoomThreadsTests` (11) and `UnreadThreadsAttentionTests` (2).
  - New in Workspace: 3 `WorkspaceStateTests`.
  - New in the app: `NativeWindowTests.CrossRoomThreadsViewTests` (3 tests, 6 cases, light and dark).
- `xcodebuild … build` (macOS): BUILD SUCCEEDED.
- `swift build --package-path Packages/SokosumiWorkspace --triple arm64-apple-ios17.0 --sdk <iPhoneOS SDK>`: build complete.
- `mint run swiftformat --lint .`: 0 of 427 files need formatting.
- `mint run swiftlint lint --strict`: 0 violations.

Render (light over dark: the view with both groups, the caught-up view, and the sidebar row with a count and with a mention badge):

Threads view

Not exercised in a running app against Core: the view itself, opening a row into another room, the count settling after a Look, and VoiceOver. The test host builds no accessibility tree. The full list is in PARITY under "Slice 24f1".

**Also in this PR (`c25e51a49`, docs only):** at the user's request, new parity rows 24g and 24h for the rest of #5078. 24g covers the sidebar room rows following ADR 0037 (bold on channel unread plus mentions, inset unread thread rows); 24f2 now depends on it. 24h covers the transcript's thread reply bar (#5078, #5116). Work order: 24g, 24f2, 24h, then 09c.


mrosberghaus:
## Grok review

### Fixed
- Important: sign-out, reload, and a successful workspace switch left Threads parent previews in memory, so the next open in that workspace could show the previous reader's rows until Core answered (`91a9fe591` — `apps/apple/Packages/SokosumiWorkspace/Sources/SokosumiWorkspace/WorkspaceState.swift`)

### Remaining
- Minor: `CrossRoomThreads.swift`: row labels are baked when the page is fetched. A mention name that is not on the room yet stays unresolved until the next refetch. Web rebuilds the label on render.
- Skipped / disagreed: restoring the shared hover id on thread rows — per-row hover does not have the old bug, where one row's exit cleared the next row's highlight. Live Core, VoiceOver, and opening a row in the running app were not re-checked here; the PR already lists those as unverified.

### Assessment
The Threads view matches web's unread and earlier lists, the sidebar's one count, and the off-screen read guard; session changes now drop the cached parent text.

mrosberghaus:
Re the Grok review above:

- **Fixed in `91a9fe591` (Threads parent text survived a session change):** checked against the code. `crossRoomThreads.reset()` now runs on sign-out (`reset`), on reload and on a successful workspace switch, next to `clearTranscript()`. A failed switch keeps the lists, and the new `signOutAndWorkspaceSwitchDropCrossRoomThreadText` covers all three paths. This is the same lifecycle rule the transcript already follows.
- **Row labels baked at fetch (minor):** no change in this PR. It follows the room overview's existing pattern: `RoomThreadOverview.previews` is also built when a page lands (`RoomThreadOverview.swift:103-107`), not on render. An unresolved mention clears on the next fetch, which reopening the view or a change in the rooms' thread counts triggers (the fingerprint re-read). Resolving on render would need to happen in both lists at once, so I'm treating it as a follow-up, not a 24f1 change.
- **Shared hover id (skipped):** agreed, no action. `ThreadListRow` keeps its hover per row, so the old bug where one row's exit cleared the next row's highlight can't happen.

PR #5158, opened 2026-09-23
feat(chat): remember the Unreads filter per browser

mrosberghaus:

The chat list's **Unreads** filter is now remembered. Turn it on, reload or open a new tab, and the list opens on Unreads again instead of snapping back to All.

- Stored per browser in a year-long cookie (`chat_unreads_filter`). No Core change, the Apple app is untouched.
- One value for every organisation, shared by the sidebar and the phone's Chats page.
- Never flips by itself: with nothing unread it stays on and reads *All caught up*.
- Sign-out leaves it, like the other display cookies (`sidebar_state`, `tasks_view_mode`).

## No All-then-Unreads flash

The app shell is a static prerender and cannot read cookies, so it always draws All. The existing sidebar boot script now also sets `data-chat-unreads-boot` on `<html>` before paint when the cookie is on. While marked, a `chat-unreads-boot:` variant hides the All list and tints the Unreads toggle, so the first frame already matches what React renders. `useChatUnreadsFilter` drops the mark once React renders the value that agrees with the cookie, so a filter later switched off (in another tab, say) never leaves All hidden.

`CONTEXT.md` gains **Unreads filter**.

## Verification

- `CI=1 pnpm --filter web test` on the branch merged with `main` — 848 files, 7520 tests pass
- `pnpm check`, `pnpm typecheck` (commit hook)
- New tests: cookie parse/serialize and the boot script, alone and shipped after the sidebar's script (`chat-unreads-filter.test.ts`, listed in `DOM_TEST_TS` since it needs `document`); the list opens on the remembered filter, writes it on toggle, and drops the boot mark (`organization-chat-list-unread-filter.test.tsx`)
- Compiled the app's Tailwind CSS and rendered harness markup in a browser: with the mark set, the All list is hidden and the toggle tinted, identical to the post-hydration Unreads toggle; without it, the normal All list


mrosberghaus:
## Grok review

### Fixed
- Important: the boot mark was cleared as soon as the first chat list agreed with the cookie, so a sibling All list that had not hydrated yet (sidebar and the phone Chats page both mount) became visible (`7043a6486` — `apps/web/src/components/chat/use-chat-unreads-filter.ts`)

### Remaining
- Minor: `apps/web/src/components/chat/chat-unread-nav-rows.tsx`: before hydration, collapsed-rail hover and active still use the off-state ring. `group-data-[collapsible=icon]:hover` outranks `chat-unreads-boot:bg-*`. Resting state matches.
- Skipped / disagreed: live cross-tab updates — the requirement is reload and a new tab.
- Skipped / disagreed: clearing the cookie on sign-out, or one value per account — the requirement is one browser cookie, left in place like `sidebar_state`.
- Skipped / disagreed: painting unread rows in the static shell — the requirement hides the All list until React reads the cookie. Setting `aria-pressed` from the boot script would fight hydration.
- Skipped / disagreed: boot-script tests running under the node project — already fixed on the branch in `a172dc36c`.

### Assessment
The filter is one year-long cookie, shared by both lists, and the All list stays hidden until each list has rendered Unreads.

mrosberghaus:
Re the Grok review above:

- **Boot mark cleared by the first list (`7043a6486`)**: verified, and I'm keeping it. The sidebar and the phone's Chats page can each hydrate on their own schedule. While the cookie is on, the mark only hides All lists, and a hydrated list shows Unreads, so keeping it costs nothing. Switching the filter off, in this tab or another, still clears it. The chat and ui-preferences tests pass on the merged branch (569).
- **Collapsed-rail hover and press before hydration**: fixed in `64b42b1e7`. The toggle now has the same two `group-data-[collapsible=icon]` hover and active tints as the on state, under the `chat-unreads-boot:` variant. In the compiled CSS the boot rules are `html[data-chat-unreads-boot] … :is(…[data-collapsible="icon"] *):hover`, one attribute more specific than the rail's own rule, so they win.
- **Skipped items**: agreed on all four. They match what was agreed for this PR: per-browser cookie, kept on sign-out, All hidden rather than painted, and the Node test fix already landed in `a172dc36c`.

PR #5159, opened 2026-09-23
ci: skip test legs a PR does not need instead of starting runners

mrosberghaus:
## What

Tests legs that a PR does not need no longer start a runner. Build, Biome and Typecheck gate the same way.

- `test.yml`: the 7-leg matrix becomes 7 named jobs (`Test Web`, `Test Core`, `Test Packages`, `Test CLI`, `Test Local env`, `Test CI config`, `Test Cloud agent db`). Each has a job-level `if:` on its own path filter. YAML anchors share the checkout/setup steps, the permissions, and the Postgres service.
- `js-paths-filter.yml`: new per-leg filters.
  - `web` = `js` minus `apps/core/**`
  - `core` = `js` minus `apps/web/**`
  - `packages` = `js` minus both apps
  - The advisory legs match the files their `node --test` suites read.
- `build.yml` / `lint.yml`: step-level gates become one job-level `if:`.
- `turbo-remote-cache.test.mjs` asserts the new structure. `delivery.md` documents the rule.

## Why a skip is safe for required checks

- A job skipped by a job-level `if:` gets no runner and still reports a check under its real name.
- The ruleset accepts a skipped check. #4475 auto-merged with the required `Swift lint and format` and `Xcode test` both `skipped`, and auto-merge cannot bypass the ruleset.
- A matrix could not do this, because it expands after the job-level `if:`. That is why the old file started every leg and gated per step.
- Gates are `!= 'false'`. If `Detect JS changes` fails, its outputs are empty and every job runs, so nothing skips silently.

## Branch-protection risk

- **Low.** Required contexts `Build`, `Biome`, `Typecheck`, `Test Web`, `Test Core`, `Test Packages` keep their exact names, and none is renamed or removed.
- The residual risk is a path filter that is too narrow.
  - Web and Core share no workspace dependency.
  - `packages/*` depend on neither app.
  - Root files, `packages/**`, `.github/**` and `pnpm-lock.yaml` still run all three required legs.
  - The filters were replayed with dorny's own matcher (picomatch 2, `dot: true`, `every`) against 24 sample paths.

## Billed minutes per push, if private (from measured runs)

| PR kind | Before | After |
| --- | --- | --- |
| Apple-only / docs-only | Tests 8 + Build 2 + Lint 3 = **13** | Tests 1–2 + Build 1 + Lint 1 = **3–4** |
| Web-only | Tests 8–13 + Build 4 + Lint 3–4 | Tests 2 + web leg (1–6) + same Build/Lint → **−5 to −6** |
| Core-only | Tests 8–15 | Tests 2 + core leg → **−5** |

## Verification

- `pnpm ci:test`: 78/78 passed
- `actionlint` 1.7.12: clean
- Filter replay: 24/24 paths matched the expected legs
- This PR touches `.github/**`, so every required leg runs here. `Test CLI`, `Test Local env` and `Test Cloud agent db` should report **skipped**.

Draft. Do not merge.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `.github/js-paths-filter.yml`: the `docs` filter is unused. Its comment still says Test CI config uses it. The job reads `ci-config`.
- Minor: `.github/workflows/lint.yml`: the `changes` output mentions `workflow_dispatch`, but that workflow only runs on `pull_request`. The CI contract test requires the same expression on every `changes` job, so this was left as-is.
- Minor: `scripts/ci/__tests__/turbo-remote-cache.test.mjs` locks YAML text, not matcher results. A broken anchor or brace would still pass that test.

- Skipped / disagreed: job-level skip does not satisfy the ruleset — disagreed. Ruleset `3855070` still lists `Build`, `Biome`, `Typecheck`, `Test Web`, `Test Core`, and `Test Packages`. This PR is `MERGEABLE` with required `Xcode test` and `Swift lint and format` already `SKIPPED` (`mergeStateStatus` is `BEHIND` only).
- Skipped / disagreed: `web` / `core` / `packages` anchors or braces mis-parse — disagreed. dorny flattens the `*js` alias. Run `35909720995` set `web`, `core`, `packages`, and `ci-config` to true and `cli`, `local-env`, and `cloud-agent-db` to false. A picomatch 2.3.1 replay (`dot: true`, `every`) matches that, and a core-only path does not select Test Web. Web and core do not depend on each other; no package depends on either app.

### Assessment
Required check names are unchanged, skipped checks still report those names, and the per-leg filters match the workspace graph.

mrosberghaus:
Re: [Grok review](https://github.com/masumi-network/sokosumi/pull/5159#issuecomment-5801706672)

- **`docs` filter unused**: already removed in b0eef3068 (`ci: drop the docs path filter nothing reads`). The branch has no `docs:` key.
- **`workflow_dispatch` in `lint.yml`**: fixed in 77663013c. Linting now has a `workflow_dispatch` trigger like Build and Tests, so the `changes` output's dispatch branch is live rather than dead. #5162 folds this file into `ci.yml`, which already has the trigger.
- **Test locks YAML text, not matcher results**: not changed.
  - A matcher-level test needs `picomatch` and `js-yaml`, and neither resolves from the repo root today. Both would become new root devDependencies.
  - The installed `picomatch` is v4; dorny/paths-filter@v3 bundles v2. Their behaviour differs, and root-level brace patterns are the case that differs. A v4 test could pass while dorny evaluates the filter differently.
  - What actually proves the filters is dorny's real outputs on CI, which this PR's runs already show (`web`, `core`, `packages` and `ci-config` true; `cli`, `local-env` and `cloud-agent-db` false), plus the v2 replay in the description.
  - If we want this locked down, the cleaner option is a scheduled job that runs dorny itself against fixture diffs. That's a separate change.

PR #5163, opened 2026-09-23
feat(tasks): manage Task Schedules in the Task Manager (SOK-1169)

mrosberghaus:
Task Schedule 05/10 ([SOK-1169](https://linear.app/masumi/issue/SOK-1169/task-schedule-0510-schedules-view-in-the-web-app)). This PR goes into the SOK-1164 branch (#5127), not `main`.

## What people get

- **Schedules tab** in the Task Manager, next to Tasks and Jobs (`/tasks?tab=schedules`).
  - Each row shows the schedule's rule, next Run, assignee, and state, with a lock on private schedules.
  - The list follows the project switcher. A new state filter offers All, Active, Paused, and Ended.
  - "Load more" pages through long lists.
- **New / Edit schedule dialog** with name, description, assignee (Coworker, Soko Bot, or member), project, private (create only), and the rule.
  - The rule part reuses the existing schedule form, set to repeating only.
  - Editing says "Changes apply to future runs only."
- **Schedule page** `/tasks/schedules/{id}` shows upcoming Runs (skipped and moved ones marked), the Tasks this schedule created, and its properties. The owner can edit, pause, resume, end, and delete it. End and Delete ask for confirmation.
- **Repeat** in a Task's ⋯ menu opens the dialog prefilled from that Task. The Task itself is not changed or linked.
- **"From schedule …"** appears on a Task that a Run created and links to its schedule. If the viewer can't open that schedule, it shows as plain text with no name.

## Core

- The Task DTO now includes a read-only `scheduleId`.
- `GET /v1/tasks?scheduleId=` filters for every actor (people, Coworkers, Soko Bots).
- Both have mocked-Prisma route tests.
- The Web Core client is regenerated.

## Decisions for review

- **Tab, not a third board/list mode.** The ticket says "beside board and list". The board/list switch is a per-user saved preference, and it comes with Task-only filters (status, assignee, scope) that don't apply to schedules. A tab next to Jobs keeps those apart. Moving it into the switch later only means moving the trigger.
- **An edit that keeps the rule doesn't send it.** Core cancels skipped and moved Runs whenever PATCH replaces the rule, so renaming a schedule would have thrown them away.
- **Fixed on the way:** the shared schedule form read an every-N-days rule as the Daily preset and dropped the day step. The Calendar's series editor used the same form and had the same bug.
- **Every-N-days runs at the anchor's time.** Core ignores the cron's time for these rules. The form's time of day now sets the anchor's time, and the preview steps N days from the anchor. Rules saved by the old form, where the cron time and anchor time disagree, open with the anchor's time.

## Known limits

- The schedule page shows the 20 most recently created Tasks of the schedule (`GET /v1/tasks?scheduleId=…&sort=createdAt`), with no way to page further.
- A Paused schedule still lists the skipped and moved Runs it keeps. Pausing keeps them, so they show.

## Verification

- `pnpm check`, `pnpm typecheck`, `pnpm test` all green (Core 6634 tests, Web 7521).
- New component tests:
  - Schedules view: renders; filters by state and project.
  - Dialog: create; edit with the future-only notice; unchanged rule not sent for weekly, custom-cron, and every-N-days rules; changed rule sent; no members offered on a private schedule; stale revision.
  - Pause, resume, end, delete.
  - Repeat prefills from the Task and leaves it unchanged.
  - "From schedule …" shows only with `scheduleId`.
- UI checked by rendering the real components with the real `en.json` and app CSS: light, dark, and 375px (no horizontal scroll). Not checked against a running app with real data.


mrosberghaus:
## Grok review

### Fixed
- Important: an every-N-days rule uses a daily cron, so edit treated it as the daily preset and replaced the stored anchor with the next local slot. A later rule edit shifted the series (`0badab5ec` — `apps/web/src/lib/utils/task-schedule.ts`)
- Important: Repeat on a private Task assigned to a member saved that Task's name and description as a public schedule. The member assignee is dropped so the new schedule stays private (`0badab5ec` — `apps/web/src/app/(app)/tasks/components/task-schedule-dialog.tsx`)
- Important: the schedule page took the first 10 runs and then dropped canceled ones, so planned runs behind canceled skips disappeared (`0badab5ec` — `apps/web/src/app/(app)/tasks/schedules/[scheduleId]/page.tsx`)

### Remaining
- Minor: `apps/web/src/components/task-schedule-section.tsx`: every-N-days opens as custom cron. The time-of-day field shows the current clock and does not change the saved cron while the cron text is filled.
- Minor: `apps/web/src/components/task-schedule-section.tsx`: the preview lists the next daily cron hits, not every N days.
- Minor: `apps/web/src/app/(app)/tasks/components/task-schedule-dialog.tsx`: a stale revision tells the user to reload, and the dialog does not refresh the revision itself.
- Minor: `apps/web/src/app/(app)/tasks/schedules/[scheduleId]/page.tsx`: "Tasks created" is the 20 most recently updated tasks. `GET /tasks` has no `createdAt` sort. Same limit the PR already names.
- Skipped / disagreed: a tab beside Jobs, instead of a third board/list mode, matches the PR decision. Task-only filters do not apply to schedules.
- Skipped / disagreed: `scheduleId` on the Task DTO is response-only. Create and update schemas do not accept it.

### Assessment
The schedules tab, dialog, detail page, and `scheduleId` filter match SOK-1169. The three data bugs above are fixed; what is left is display-only.

mrosberghaus:
Reply to the [Grok review](https://github.com/masumi-network/sokosumi/pull/5163#issuecomment-5802456071).

**Your fixes in `0badab5ec`:** I kept all three. I adjusted how every-N-days rules load, see Remaining 1 below: they now load the cron as `cron` and no longer as `customCronExpr`. `task-schedule.test.ts` is updated to match.

**Remaining**

1. **Every-N-days time of day: fixed in `cb2d69b59`, and it was more than a display bug.** Core runs these rules at the anchor's local time and ignores the cron's time (`computeIntervalNextRun`). The form only put "Time of day" into the cron, so the anchor stayed at "now + 5 min". A new every-N-days schedule ran at the moment the form was opened, whatever time was picked. Changes:
   - The form's time of day now sets the anchor's time.
   - A rule opens in the builder with no cron text, and its time of day comes from the anchor.
   - A rule saved by the old form (cron time different from anchor time) opens with the anchor's time. Editing only the name therefore doesn't resend the rule and cancel skipped or moved Runs.
   - The Calendar's series editor uses the same form, so it gets the same fix.
2. **Preview: fixed in `cb2d69b59`.** It starts at the anchor and keeps every Nth day. Dialog test: `2030-01-07 / 01-10 / 01-13`.
3. **Stale revision: left as is, on purpose.** If the dialog picked up the new revision and saved again, a second click would overwrite whoever changed the schedule, and the revision check exists to stop exactly that. The toast asks for a reload so the person sees the other change first. If you'd rather keep what they typed and rebase it onto the newer revision, that's a small follow-up. Say so and I'll add it.
4. **"Tasks created": fixed in `c66c4b80a`.** `GET /v1/tasks` accepts `sort=createdAt` (newest created first, with a route test), and the schedule page uses it. It still shows 20 with no paging; that's in the PR's known limits.

The last two disagreements (tab placement, `scheduleId` response-only) match the PR.

`pnpm check`, `pnpm typecheck`, `pnpm test` all pass (Core 6635, Web 7526).

PR #5166, opened 2026-09-23
feat(apple): bold sidebar rooms on channel unread and mentions (ADR 0037)

mrosberghaus:
Apple parity row **24g1**: the sidebar's room rows follow [ADR 0037](https://github.com/masumi-network/sokosumi/blob/main/docs/adr/0037-thread-unread-leaves-room-unread.md) as web has since #5078.

## What the reader sees

- A room that only holds replies in threads is no longer bold. Reading the channel empties its row, as on web; the Threads row (24f1) keeps counting the threads.
- A thread reply that names you still bolds the room and shows the mention badge.
- A row shows **one number**: the mention badge when you were named, otherwise the unread count. The count now sits at the right edge, where the badge stands, in the Threads row's muted style, instead of `· N` after the name.
- In a Direct with one other person every message counts toward the badge, so that row shows the count instead of a badge, as web does. Group Directs keep the mention badge.
- Reading a room, looking at a thread, muting a thread and Mark all now settle the row from Core's two numbers, so a read room cannot come back bold from an old channel count.

Sidebar, light and dark

## Split

Row 24g held two independent features, so this PR splits it in `PARITY.md`: **24g1** (bold, number and mention rule, this PR) and **24g2** (inset unread thread rows under a room, Todo). 24f2 needs only 24g1. 24g2 comes next in the work order. Rows 05a (its leftover-thread sentence was written before ADR 0037 shipped) and 24f1 (now Done) are corrected here too.

## Changes

- `SokosumiChat/SidebarRooms.swift`: `resolveRoomAttention` now takes the channel half and web's Direct-of-two rule. Also added: `RoomAttention.mentionCount`, `resolveRoomAttention(_ room:)`, `roomBadgeCountsMentions` and `roomAttentionAfterRead`.
- `SokosumiChat/RoomReadAttention.swift`: the optimistic read keeps the thread half. The overlay now keeps both halves and the room's thread list, as web's `RoomReadOverlay` does.
- `ConversationSidebarView`: rows use the resolver's room overload, and the count moves to the trailing edge.
- No Core, snapshot, project or dependency change.

- `xcodebuild test -workspace Sokosumi.xcworkspace -scheme Sokosumi … -enableCodeCoverage NO`: 1224 tests passed (app 220, Chat 779, Auth 35, Realtime 48, Workspace 141, CoreAPI 1).
- The macOS build and the iOS 17 `SokosumiWorkspace` build pass.
- `mint run swiftformat --lint .` finds 0 files to format; `mint run swiftlint lint --strict` finds 0 violations.
- The tests were written first and seen to fail: 20 issues in the package suites against a compile-only skeleton, 3 in the workspace suite against the old overlay, and 2 of 3 new pixel tests against the unchanged view.

Not checked in the running app: the rows against Core, a live read, Look, thread mute or Mark all, and VoiceOver on the moved count. Details are under "Slice 24g1" in `apps/apple/PARITY.md`.


mrosberghaus:
## Grok review

### Fixed

### Remaining
- Minor: `apps/apple/SokosumiTests/Chat/Sidebar/SidebarRoomAttentionTests.swift`: the pixel checks pass only when the subject pair differs by no more bytes than two control renders of the same sidebar. A control pair that matches exactly fails the test on a one-pixel difference. The suite was green in this change; treat a later red run as flake before changing the resolver.
- Skipped / disagreed: negative-count clamp (`max(0, …)` and `mentionCount <= 0` in `SidebarRooms.swift`) — web leaves the half raw and uses `mentionCount === 0`. Core counts are non-negative, so the row does not change.
- Skipped / disagreed: settled overlays ignore `updatedAt` — pre-existing. List refresh still reconciles by revision, and the snapshot now keeps `channelUnreadCount`, so a stale total cannot re-bold a read room.
- Skipped / disagreed: mention badge stays `99+` rather than web's `@ 9+` — row 26b, called out in `PARITY.md`. Inset thread rows (24g2), Dock badge, and window-title count are out of this slice.

### Assessment
Ready to merge: Yes
The row, the optimistic read, and the overlay follow web's ADR 0037 rules: thread replies alone stay quiet, a thread mention still bolds and badges, one number is drawn, and a Direct of two shows the count.

mrosberghaus:
Minor (pixel checks with an exact-match control pair): fixed in 10f6d8c37.
- The three picture checks in `SidebarRoomAttentionTests` now allow `max(control pair, pixelNoiseFloor)`, with a floor of 50 bytes.
- I measured the floor by regressing each guarded rule. Thread replies bolding the row changed 3294 bytes (light) and 3333 (dark). The count drawn beside a badge changed 540 in both. A badged Direct of two changed 579 and 612. The control pairs differed by 0, which confirms the case you flagged.
- The smallest real difference is 540, so the floor is about 9% of it.
- With the floor in place and the regressions applied, all six red cases still fail (3294, 3333, 540, 540, 921 and 975 bytes against 50). The Direct numbers are higher there because the badge regression was also active.
- `SidebarRoomAttentionTests` passes 24 of 24 with `-test-iterations 3`.

Skipped, no change:
- Negative-count clamp: harmless, because Core counts are non-negative.
- Overlay and `updatedAt`: this behaviour predates this PR, and the overlay now keeps `channelUnreadCount`, so a read room stays quiet.
- 99+ versus web's `@ 9+`: this is row 26b and is recorded in PARITY.

